PR #237, opened 2025-12-11
Token burn mint wallet update

Pravdyvy:
## 🎯 Purpose

To complete token program wallet functionality.

## ⚙️ Approach

Adding mint and burn calls to wallet.

- [x] Burn calls.
- [x] Mint calls.

## 🧪 How to Test

Updated/created new integration tests.

## 🔗 Dependencies

None

## 🔜 Future Work

- `nssa::public_transaction::Message::try_new` is semantically different from other tx helpers because it serializes intruction on itself.
- If definition is private, then you must own it to burn, because definition is guaranteed to be non-default.
- Return an error instead of panicing at CLI if invalid account id provided.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


wallet/src/cli/programs/token.rs
```diff
@@ -49,6 +50,50 @@ pub enum TokenProgramAgnosticSubcommand {
...
         amount: u128,
     },
+    /// Burn tokens on `holder`, modify `definition`.
+    ///
+    /// `holder` is owned
+    ///
+    /// Also if `definition` is private then it is owned, because
+    /// we can not modify foreign accounts.
+    ///
+    /// ToDo: Return and add foreign variant when we could modify foreign accounts
```
Arjentix: IMO this shouldn't be in user docs as clap will render it. Also let's have an issue

wallet/src/cli/programs/token.rs
```diff
@@ -386,6 +669,34 @@ impl WalletSubcommand for TokenProgramSubcommandPublic {
...
                 Ok(SubcommandReturnValue::Empty)
             }
+            TokenProgramSubcommandPublic::BurnToken {
+                definition_account_id,
+                holder_account_id,
+                amount,
+            } => {
+                Token(wallet_core)
+                    .send_burn_transaction(
+                        definition_account_id.parse().unwrap(),
```
Arjentix: So we're panicking if user provided wrong ids... It should be an error then, not a panic. Let's have an issue for that?

moudyellaz (approved):
lgtm, thanks

PR #240, opened 2025-12-12
Allow private authorized uninitialized accounts

Arjentix:
## 🎯 Purpose

Solves issue:
- https://github.com/logos-blockchain/lssa/issues/191

## ⚙️ Approach

- Applied @schouhy commit with unziping auth keys from proof
- Made proofs optional as their number should always be equal to the number of private accounts
- Modified privacy preserving circuit to allow private authorized uninitialized accounts

## 🧪 How to Test

Added unit test:

```bash
RISC0_DEV_MODE=1 cargo test -- test_private_authorized_uninitialized_account
```

Added integration test:

```bash
NSSA_WALLET_HOME_DIR=$(pwd)/integration_tests/configs/debug/wallet/ RUST_LOG=info RISC0_DEV_MODE=1 cargo run --bin integration_tests -- ./integration_tests/configs/debug test_authenticated_transfer_initialize_function_private
```

## 🔗 Dependencies

None

## 🔜 Future Work

None

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests

### DIFF
--- a/nssa/program_methods/guest/src/bin/privacy_preserving_circuit.rs
+++ b/nssa/program_methods/guest/src/bin/privacy_preserving_circuit.rs
@@ -163,29 +163,44 @@ fn main() {
                     // Private account with authentication
                     let nsk = private_nsks_iter.next().expect("Missing nsk");
 
-                    let membership_proof = private_membership_proofs_iter
-                        .next()
-                        .expect("Missing membership proof");
-
                     // Verify the nullifier public key
                     let expected_npk = NullifierPublicKey::from(nsk);
                     if &expected_npk != npk {
                         panic!("Nullifier public key mismatch");
                     }
 
-                    // Compute commitment set digest associated with provided auth path
-                    let commitment_pre = Commitment::new(npk, &pre_states[i].account);
-                    let set_digest = compute_digest_for_path(&commitment_pre, membership_proof);
-
                     // Check pre_state authorization
                     if !pre_states[i].is_authorized {
                         panic!("Pre-state not authorized");
                     }
 
-                    // Compute update nullifier
-                    let nullifier = Nullifier::for_account_update(&commitment_pre, nsk);
+                    let membership_proof_opt = private_membership_proofs_iter
+                        .next()
+                        .expect("Missing membership proof");
+                    let (nullifier, set_digest) = membership_proof_opt
+                        .as_ref()
+                        .map(|membership_proof| {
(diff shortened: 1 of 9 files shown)


nssa/src/state.rs
```diff
@@ -2606,4 +2660,56 @@ pub mod tests {
...
+            vec![(private_keys.npk(), private_keys.ivk(), epk)],
+            output,
+        )
+        .unwrap();
+
+        let witness_set = WitnessSet::for_message(&message, proof, &[]);
+
+        let tx = PrivacyPreservingTransaction::new(message, witness_set);
+        let result = state.transition_from_privacy_preserving_transaction(&tx);
+        assert!(result.is_ok());
```
schouhy: We could add an assert that the initialization nullifier is in the nullifier state after transaction has been processed.

nssa/program_methods/guest/src/bin/privacy_preserving_circuit.rs
```diff
@@ -160,28 +163,42 @@ fn main() {
...
+                                compute_digest_for_path(&commitment_pre, membership_proof);
+
+                            // Compute update nullifier
+                            let nullifier = Nullifier::for_account_update(&commitment_pre, nsk);
+                            (nullifier, set_digest)
+                        })
+                        .unwrap_or_else(|| {
+                            // Compute initialization nullifier
+                            let nullifier = Nullifier::for_account_initialization(npk);
+                            (nullifier, DUMMY_COMMITMENT_HASH)
```
schouhy: When initializing private accounts without authentication we are checking that the pre state account is default. I think we need that check here too. 
Otherwise I think I could deploy a dummy program A that does a noop on a single account and execute it privately with authentication with some crafted account owned by another program B (therefore injecting state into that program). This would pass the well beheaved program rules because it does nothing. And so the circuit needs to check that any new initialized private account comes from an uninitialized pre state.

Maybe a good test to add!

Arjentix: Added, thanks. Took me some time to understand how to trigger this. Actually even without this check malicious transaction wouldn't pass on-chain checks, as it will have the same nullifier which has been already seen.

schouhy: Ah nice thanks

schouhy (approved):
LGTM thanks!

PR #152, opened 2025-11-12
initialize amm program PR

jonesmarvin8:
## 🎯 Purpose

Implements a simple AMM program based on [AMM specifications](https://www.notion.so/AMM-spec-28a8f96fb65c80bf93edde876c5a6932).

## ⚙️ Approach

Write AMM program functions
- [x] new
- [x] swap
- [x] add liquidity
- [x] remove liquidity

Tests
- [x] unit tests for AMM
- [x] state level tests

## 🧪 How to Test

`RISCZERO_DEV=1 cargo test --release` in `nssa/program_methods/guest` for unit tests
`RISCZERO_DEV=1cargo test amm --release` in `nssa` for for state level tests.

## 🔗 Dependencies

Depends on [tail calls PR 151](https://github.com/vacp2p/nescience-testnet/pull/151); necessary parts merged into this PR.

## 🔜 Future Work

PDA needs to be added to AMM in a future PR.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Implement the core functionality
- [x] Add/update tests
- [x] Add/update documentation and inline comments

PR #245, opened 2025-12-15
AMM-wallet integration (Public part)

Pravdyvy:
## 🎯 Purpose

Add ability to call AMM methods from wallet publicly.

## ⚙️ Approach

Public amm definition creation, swap, adding/removing liquidity.

- [x] Public amm definition.
- [x] Public swap.
- [x] Public adding liquidity.
- [x] Public removing liquidity.

## 🧪 How to Test

Added new integration test with all AMM methods.

## 🔗 Dependencies

#152 

## 🔜 Future Work

- Implement private calls for amm (depends on multi-chain calls in private executions).

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


wallet/src/cli/programs/amm.rs
```diff
@@ -0,0 +1,286 @@
...
+    WalletCore,
+    cli::{SubcommandReturnValue, WalletSubcommand},
+    helperfunctions::{AccountPrivacyKind, parse_addr_with_privacy_prefix},
+    program_facades::amm::AMM,
+};
+
+/// Represents generic CLI subcommand for a wallet working with amm program
+#[derive(Subcommand, Debug, Clone)]
+pub enum AmmProgramAgnosticSubcommand {
+    /// Produce a new token
```
schouhy:
```suggestion
    /// Produce a new pool
```

wallet/src/cli/programs/amm.rs
```diff
@@ -0,0 +1,286 @@
...
+        /// user_holding_lp - valid 32 byte base58 string with privacy prefix
+        #[arg(long)]
+        user_holding_lp: String,
+        #[arg(long)]
+        balance_lp: u128,
+        #[arg(long)]
+        max_amount_a: u128,
+        #[arg(long)]
+        max_amount_b: u128,
+    },
```
schouhy: min_amount_*

wallet/src/cli/programs/amm.rs
```diff
@@ -0,0 +1,286 @@
...
+
+                match (user_holding_a_privacy, user_holding_b_privacy) {
+                    (AccountPrivacyKind::Public, AccountPrivacyKind::Public) => {
+                        AMM(wallet_core)
+                            .send_swap(
+                                user_holding_a,
+                                user_holding_b,
+                                amount_in,
+                                min_amount_out,
+                                token_definition.parse()?,
```
schouhy: shouldn't token_definition go first through `parse_addr_with_privacy_prefix`?

wallet/src/program_facades/amm.rs
```diff
@@ -0,0 +1,472 @@
...
+        };
+        let Ok(user_b_acc) = self.0.get_account_public(user_holding_b).await else {
+            return Err(ExecutionFailureKind::SequencerError);
+        };
+
+        let definition_token_a_id = TokenHolding::parse(&user_a_acc.data)
+            .ok_or(ExecutionFailureKind::AccountDataError(user_holding_a))?
+            .definition_id;
+        let definition_token_b_id = TokenHolding::parse(&user_b_acc.data)
+            .ok_or(ExecutionFailureKind::AccountDataError(user_holding_a))?
```
schouhy:
```suggestion
        let definition_token_b_id = TokenHolding::parse(&user_b_acc.data)
            .ok_or(ExecutionFailureKind::AccountDataError(user_holding_b))?
```

wallet/src/cli/programs/amm.rs
Pravdyvy: It is an argument, not an input account. We do not need its privacy, so I added it as just an account. Also added line here, that it is just an account id.
https://github.com/logos-blockchain/lssa/blob/1b0dea11f68e5d268317b2195e986f23251e1f29/wallet/src/cli/programs/amm.rs#L51-L53

wallet/src/cli/programs/amm.rs
Pravdyvy: Added, thank you.

wallet/src/program_facades/mod.rs
```diff
@@ -1,6 +1,142 @@
...
+        let mut seq = serializer.serialize_seq(Some(N))?;
+        for word in self.0 {
+            seq.serialize_element(&word)?;
+        }
+        seq.end()
+    }
+}
+
+type OrphanHack65BytesInput = OrphanHackNBytesInput<65>;
+type OrphanHack49BytesInput = OrphanHackNBytesInput<49>;
```
Arjentix: Let's move all of these into `amm.rs` as all of that are strongly related only to amm

wallet/src/program_facades/amm.rs
```diff
@@ -0,0 +1,451 @@
...
+
+        let witness_set =
+            nssa::public_transaction::WitnessSet::for_message(&message, &[signing_key]);
+
+        let tx = nssa::PublicTransaction::new(message, witness_set);
+
+        Ok(self.0.sequencer_client.send_tx_public(tx).await?)
+    }
+
+    pub async fn send_add_liq(
```
Arjentix:
```suggestion
    pub async fn send_add_liquidity(
```

Let's stick to a full names especially for public api if possible

wallet/src/program_facades/mod.rs
```diff
@@ -1,6 +1,142 @@
...
+    definition_token_a_id: AccountId,
+    definition_token_b_id: AccountId,
+) -> AccountId {
+    AccountId::from((
+        &amm_program_id,
+        &compute_pool_pda_seed(definition_token_a_id, definition_token_b_id),
+    ))
+}
+
+fn compute_pool_pda_seed(
```
Arjentix: I think in future it will be cool to have something like `amm_core` crate library to import into amm guest and wallet code to share this kind of functions.

Not related to this PR actually, just to mention.

wallet/src/program_facades/amm.rs
```diff
@@ -0,0 +1,451 @@
+use common::{error::ExecutionFailureKind, rpc_primitives::requests::SendTxResponse};
+use nssa::{AccountId, program::Program};
+
+use crate::{
+    WalletCore,
+    cli::account::TokenHolding,
```
Arjentix: Hm, I think something's wrong here... Ideally I'd like to have `cli` as a thing used only in `main.rs` stuff. Not `lib.rs` stuff.

Maybe someday even make `cli` module to be included only with `cli` feature, although there are still unresolved UX [issues](https://github.com/rust-lang/cargo/issues/4663) in rustlang.

wallet/src/program_facades/amm.rs
```diff
@@ -0,0 +1,451 @@
...
+            user_holding_lp,
+        ];
+
+        let Ok(nonces) = self
+            .0
+            .get_accounts_nonces(vec![user_holding_a, user_holding_b])
+            .await
+        else {
+            return Err(ExecutionFailureKind::SequencerError);
+        };
```
Arjentix:
```suggestion
        let nonces = self
            .0
            .get_accounts_nonces(vec![user_holding_a, user_holding_b])
            .await
            .map(|_| ExecutionFailureKind::SequencerError)?;
```

More rust way.

Some idea for other places

wallet/src/cli/account.rs
```diff
@@ -25,10 +25,10 @@ struct TokenDefinition {
     total_supply: u128,
 }
 
-struct TokenHolding {
+pub struct TokenHolding {
     #[allow(unused)]
     account_type: u8,
-    definition_id: AccountId,
+    pub definition_id: AccountId,
```
Arjentix: Let's make all pub then? So that we either have a [POD](https://en.wikipedia.org/wiki/Passive_data_structure) with all things public and no incapsulated behavior or object with private fields and methods

wallet/src/lib.rs
```diff
@@ -38,6 +38,11 @@ pub mod poller;
 mod privacy_preserving_tx;
 pub mod program_facades;
 
+pub enum AccDecodeData {
+    Skip,
```
Arjentix: I haven't seen the use of it, can you point out?

PR #248, opened 2025-12-17
Publish sequencer_runner Docker image with CI

Arjentix:
## 🎯 Purpose

This PR implements a proper deployment by publishing a sequencer runner docker image into a registry.

Our hosts will run watchtower which will listen for a new images and run them.

Also fixes:
- https://github.com/logos-blockchain/lssa/issues/179
- https://github.com/logos-blockchain/lssa/issues/180
- https://github.com/logos-blockchain/lssa/issues/181
- https://github.com/logos-blockchain/lssa/issues/222

Also implemented:
- Add `cargo machete` to CI
- Use workspace dependencies as much as possible
- Move program_methods to workspace. This allows to run `cargo test` and `cargo clippy` on them easily. The only problem is feature unification so that when running `cargo clippy` from workspace our program_methods will be checked with `host` feature, which is not very good. That's why I've added another job to check our programs directly by using `cargo clippy -p "*programs"`

## ⚙️ Approach

- Removed `no_docker` feature. Instead all builds are deterministic now. They use program methods binaries from `artifacts` folder. To regenerate artifacts run `just build-artifacts` (you need `just` installed). CI also checks if artifacts are up to date
- `just` is a Rust-written modern analogue to `make`. I used that thing because we need the same artifacts build logic for us -- developers -- and for CI. And it's really nice to use
- Having artifacts really helps in docker build as we don't need docker in docker now
- Improved CI and added docker image build and publish action which will be triggered by a manual action activation from anyone with write access to the repo

## 🧪 How to Test

First terminal:
```bash
cd sequencer_runner
docker compose up
```

Second terminal:
```bash
NSSA_WALLET_HOME_DIR=$(pwd)/wallet/configs/debug/ RUST_LOG=info cargo run --bin wallet -- check-health 
```

You can also run docker in background with `docker compose up -d` and stop it later with `docker compose down`.

## 🔗 Dependencies

None

## 🔜 Future Work

None

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


Arjentix:
@schouhy that's strange, works good for me. It tries to create a rocksdb directory in `/var/lib/sequencer_runner` (defined in `sequencer_runner/configs/docker/sequencer_config.json`) **inside docker container**, but that path is linked to a `./data` in host machine, as defined in `docker-compose.yml`.

So looks like your docker user has no access to your user's directories. This is not the case on my mac...

PR #249, opened 2025-12-17
public key protocol to match specs

jonesmarvin8:
## 🎯 Purpose

Fix public key protocol for children keys based on [HD key specs](https://www.notion.so/Hierarchical-Deterministic-wallets-in-NSSA-2878f96fb65c80659325d84dd66a5630#2878f96fb65c80b78e2dfb6625662139).

## ⚙️ Approach

- [x] Fix child key implementation for harden keys
- [x] Fix child key implementation for nonharden keys
- [x] Replace deterministic tests with outputs from Python implementation

## 🧪 How to Test

In `key_protocol` run `RISC0_DEV_MODE=1 cargo test --release` for public master key and public child key tests.

## 🔗 Dependencies

N/A

## 🔜 Future Work

Eventually, add harden keys.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


key_protocol/src/key_management/key_tree/keys_public.rs
```diff
@@ -29,20 +50,39 @@ impl KeyNode for ChildKeysPublic {
 
     fn nth_child(&self, cci: u32) -> Self {
         let mut hash_input = vec![];
-        hash_input.extend_from_slice(self.csk.value());
+        hash_input.extend_from_slice(self.cpk.value());
```
schouhy: this `hash_input` variable is unused in the rest of the function

key_protocol/src/key_management/key_tree/keys_public.rs
```diff
@@ -11,6 +14,24 @@ pub struct ChildKeysPublic {
...
+        hash_input.extend_from_slice(&cci.to_le_bytes());
+
+        hmac_sha512::HMAC::mac(hash_input, self.ccc)
+    }
+
+    fn nth_child_harden_hash(&self, cci: u32) -> [u8; 64] {
+        let mut hash_input = vec![];
+        hash_input.extend_from_slice(self.csk.value());
+        hash_input.extend_from_slice(&(cci - TWO_POWER_31).to_le_bytes());
```
schouhy: both of these functions depend on cci but they are supposed to be called on different intervals of [0, 2**32]. Maybe best to have one function `nth_child(cci: u32)` that branches internally depending on the input cci, to avoid having one function being called with the incorrect cci by mistake.

jonesmarvin8: Absolutely. In the `nth_child` function, the appropriate hash function is called based on the branching path. The functions `nth_child_harden_hash` and `nth_child_nonharden_hash` probably should be private functions. Though, these functions externally have no real use?

key_protocol/src/key_management/key_tree/keys_public.rs
jonesmarvin8: Thanks, I'll delete `hash_input`; this was prior to branching logic inclusion as mentioned below.

Pravdyvy (changes requested):
Left comment on one of thests. Default Rust behaviour is to panic on overflows, so it would not work. Otherwise looks good.

Pravdyvy (approved):
lgtm

moudyellaz (approved):
Thanks, it loogs good after all the changes (also checked the specs and they match) !

Issue #235, opened 2025-12-10
Redundancy in token program CLI methods handling.


Pravdyvy:
Remove repeating code in token program CLI methods handling.

Issue #234, opened 2025-12-10
Redundancy in transfers parameters in wallet CLI


Pravdyvy:
Remove redundancy in transfers parameters in wallet CLI by introducing shared args structures.

PR #251, opened 2025-12-19
Facade and CLI unification

Pravdyvy:
## 🎯 Purpose

To remove redundancy and boilerplate from cli calls handling.
Tries to solve issues:
- #234 
- #235 

## ⚙️ Approach

Unified privacy-preserving calls. Introduced macro for named account fields in CLI. Refactored token and native token cli and facades.

- [x] Unified privacy-preserving call for native transfer.
- [x] Unified privacy-preserving call for token program.
- [x] Introduced macro for named fields in CLI.

## 🧪 How to Test

Refactor, usual testing must succed.

## 🔗 Dependencies

#237 

## 🔜 Future Work

- The main leftover in unification is public executions. Need to unify it as well.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests

### DIFF
--- a/wallet/Cargo.toml
+++ b/wallet/Cargo.toml
@@ -28,5 +28,5 @@ futures.workspace = true
 risc0-zkvm.workspace = true
 async-stream = "0.3.6"
 indicatif = { version = "0.18.3", features = ["improved_unicode"] }
-paste.workspace = true
+paste = "1.0.15"
 optfield = "0.4.0"
--- a/wallet/src/cli/programs/mod.rs
+++ b/wallet/src/cli/programs/mod.rs
@@ -18,15 +18,15 @@ trait ParsePrivacyPreservingAccount {
 
 #[macro_export]
 macro_rules! owned_account_name {
-    ($classname: ident, $field: ident) => {
+    ($structname: ident, $field: ident) => {
         #[derive(Debug, Args, Clone)]
-        pub struct $classname {
+        pub struct $structname {
             /// $field - valid 32 byte base58 string with privacy prefix
             #[arg(long)]
             pub $field: String,
         }
 
-        impl ParsePrivacyPreservingAccount for $classname {
+        impl ParsePrivacyPreservingAccount for $structname {
             fn parse(&self) -> Result<PrivacyPreservingAccount> {
                 let (account_id, privacy) = parse_addr_with_privacy_prefix(&self.$field)?;
 
@@ -47,10 +47,10 @@ owned_account_name!(ArgsSenderOwned, from);
 
 #[macro_export]
 macro_rules! maybe_unowned_account_name {
-    ($classname: ident, $field: ident) => {
+    ($structname: ident, $field: ident) => {
         paste! {
         #[derive(Debug, Args, Clone)]
-        pub struct $classname {
+        pub struct $structname {
             /// $field - valid 32 byte base58 string with privacy prefix
             #[arg(long)]
             pub $field: Option<String>,
@@ -62,7 +62,7 @@ macro_rules! maybe_unowned_account_name {
             pub [<$field _ipk>]: Option<String>,
         }
 
-        impl ParsePrivacyPreservingAccount for $classname {
+        impl ParsePrivacyPreservingAccount for $structname {
             fn parse(&self) -> Result<PrivacyPreservingAccount> {
                 match (&self.$field, &self.[<$field _npk>], &self.[<$field _ipk>]) {
                     (None, None, None) => {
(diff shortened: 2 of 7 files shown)


moudyellaz (approved):
lgtm, thanks

wallet/src/cli/programs/mod.rs
```diff
@@ -2,3 +2,109 @@ pub mod amm;
...
+                }
+            }
+        }
+    };
+}
+
+owned_account_name!(ArgsSenderOwned, from);
+owned_account_name!(ArgsDefinitionOwned, definition_account_id);
+owned_account_name!(ArgsSupplyOwned, supply_account_id);
+owned_account_name!(ArgsHolderOwned, holder_account_id);
```
schouhy: These are specific to the token program. Can't they go into the token.rs file? Same with the other usages of the macro.

wallet/src/program_facades/native_token_transfer/mod.rs
```diff
@@ -2,32 +2,55 @@ use common::error::ExecutionFailureKind;
...
-        } else {
-            Err(ExecutionFailureKind::InsufficientFundsError)
-        }
-    };
-
-    (instruction_data, program, tx_pre_check)
+#[derive(Debug, Clone, Copy)]
+pub struct NativeBalanceToMove {
+    pub balance_to_move: u128,
+}
```
Arjentix: Hmmmmm, right now I'm pretty skeptical about that... The whole idea with facades was that they reduce the bloat of methods on `WalletCore`  type and at the same time any developer using wallet as library could get easy interactions with our default programs like that:

```rust
NativeTokenTransfer(&wallet).register_account_private()
```

Currently I see no straightforward way to do the same

wallet/src/program_facades/token.rs
```diff
@@ -474,194 +146,104 @@ impl Token<'_> {
...
+    ) -> (
+        InstructionData,
+        Program,
+        impl FnOnce(&[&nssa::Account]) -> Result<(), ExecutionFailureKind>,
+    ) {
+        // Instruction must be: [0x00 || total_supply (little-endian 16 bytes) || name (6 bytes)]
+        let mut instruction = [0; 23];
+        instruction[1..17].copy_from_slice(&self.total_supply.to_le_bytes());
+        instruction[17..].copy_from_slice(&self.name);
+        let instruction_data = Program::serialize_instruction(instruction.to_vec()).unwrap();
```
Arjentix: Why to serialize instruction here and then use `Message::new_preserialized()` if we can just not serialize here and pass instruction into `Message::new()`?

wallet/src/cli/programs/mod.rs
```diff
@@ -2,3 +2,107 @@ pub mod amm;
...
+                    (None, None, None) => {
+                        anyhow::bail!("Provide either account account_id or their public keys");
+                    }
+                    (Some(_), Some(_), Some(_)) => {
+                        anyhow::bail!(
+                            "Provide only one variant: either account account_id or their public keys"
+                        );
+                    }
+                    (_, Some(_), None) | (_, None, Some(_)) => {
+                        anyhow::bail!("List of public keys is uncomplete");
```
Arjentix:
```suggestion
                        anyhow::bail!("List of public keys is incomplete");
```

wallet/src/cli/programs/mod.rs
```diff
@@ -2,3 +2,107 @@ pub mod amm;
...
+                        [<$field _ipk>].copy_from_slice(&[<$field _ipk_res>]);
+                        let [<$field _ipk>] = nssa_core::encryption::shared_key_derivation::Secp256k1Point(
+                            [<$field _ipk>].to_vec(),
+                        );
+
+                        Ok(PrivacyPreservingAccount::PrivateForeign {
+                            npk: [<$field _npk>],
+                            ipk: [<$field _ipk>],
+                        })
+                    }
```
Arjentix: Looks like there is no reason to use `[<$field _npk>]` things here, why not just `npk` and `ipk`?

Arjentix (changes requested):
Overall very good code size reduction, cool!

I can live with macro although I suggest looking at my comment with generics approach.

But I'd like to have a dev-friendly privacy preserving actions with programs like it was before with facades. So that executing private transaction would be as easy as public if you add wallet as a library crate.

This still mean you can use your traits and etc, but I encourage to hide it inside user-friendly facade methods. I understand that all combinations of private accounts in input creates a lot of duplication, but we could go with `PrivacyPreservingAccount` enum as facade method parameter to fix that.

wallet/Cargo.toml
```diff
@@ -28,4 +28,5 @@ futures.workspace = true
 risc0-zkvm.workspace = true
 async-stream = "0.3.6"
 indicatif = { version = "0.18.3", features = ["improved_unicode"] }
+paste.workspace = true
```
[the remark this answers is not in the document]
Pravdyvy: Added specific dependency

wallet/src/cli/programs/mod.rs
```diff
@@ -2,3 +2,107 @@ pub mod amm;
...
+    helperfunctions::{AccountPrivacyKind, parse_addr_with_privacy_prefix},
+};
+
+trait ParsePrivacyPreservingAccount {
+    fn parse(&self) -> Result<PrivacyPreservingAccount>;
+}
+
+#[macro_export]
+macro_rules! owned_account_name {
+    ($classname: ident, $field: ident) => {
```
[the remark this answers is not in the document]
Pravdyvy: Renamed

wallet/src/cli/programs/mod.rs
Pravdyvy: Renamed

wallet/src/cli/programs/token.rs
```diff
@@ -1,27 +1,36 @@
...
     helperfunctions::{AccountPrivacyKind, parse_addr_with_privacy_prefix},
-    program_facades::token::Token,
+    maybe_unowned_account_name, owned_account_name,
+    program_facades::{
+        send_privacy_preserving_transaction_unified,
+        token::{Token, TokenBurnArgs, TokenDefinitionArgs, TokenMintArgs, TokenTransferArgs},
+    },
 };
 
+owned_account_name!(ArgsDefinitionOwned, definition_account_id);
```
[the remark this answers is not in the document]
Arjentix: Honestly I think, that this kind of redundancy is not so bad. It's pretty clear and obvious. There is no issues like with other copy-pasted code where "If I modify here I should not forget to modify there and there" because all commands are independent and the code is pretty simple.

However macros are not so scary on the second look. So I have no strong opinion here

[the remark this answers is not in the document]
Arjentix: Honestly I think, that this kind of redundancy is not so bad. It's pretty clear and obvious. There is no issues like with other copy-pasted code where "If I modify here I should not forget to modify there and there" because all commands are independent and the code is pretty simple.

However macros are not so scary on the second look. So I have no strong opinion here

wallet/src/program_facades/native_token_transfer/mod.rs
Arjentix: I don't get the idea here.

What I was thinking of is to have consistent api call wrappers for each built-in program in our LSSA for both public and private transactions. In this PR we still have consistent methods for public transactions but we lost methods for private transactions.

I suggest to either keep both public and private or to refactor public as well, but I'm not sure how.
I think it's possible to unify our approarches like so:

```rust
struct SomeProgram(&WalletCore);

impl SomeProgram {
    pub fn do_stuff(&self, account_info: AccountInfo) -> Result<...> {
       // Instead of todos apply your approach with args or
       // anything like that to reduce currently existing boilerplate of current main branch
        match account_info {
            Public(account_id) => todo!("public execution"),
            PrivateOwned(account_id) => todo!("private execution"),
            PrivateForeign { ... } => todo!("private execution with foreign account")
        }
    }
}
```

PR #262, opened 2025-12-31
Big integration tests refactor

Arjentix:
## 🎯 Purpose

I was frustrated with our integration tests and decided to rewrite them to allow:
- just `cargo test` or `cargo nextest` usage, no custom framework
- remove `nssa_integration_test` macro
- allow running in parallel (which gives huge speed boost in CI)

Closes:
- https://github.com/logos-blockchain/lssa/issues/175
- https://github.com/logos-blockchain/lssa/issues/158

Partially addresses (gives big boost to CI speed):
- https://github.com/logos-blockchain/lssa/issues/182

## ⚙️ Approach

-  Instead of `nssa_integration_test` macro I've introduced `TestContext` which works really good for what we need
- All tests were rewritten to the new rails. Everything moved to a appropriate modules. Some names were changed, some duplicated asserts were removed
- One of the most challenging part was removal of env and globals dependencies from wallet. So that test could use temporary directories and configs for wallets. With that we can execute multiple tests in parallel

## 🧪 How to Test

I suggest using nextest as it's just really cool (we use it in CI as well):

```bash 
RISC0_DEV_MODE=1 cargo nextest run
```

But you can go with classic way:

```bash
RISC0_DEV_MODE=1 cargo test
```

See CI btw

## 🔗 Dependencies

None

## 🔜 Future Work

- This issue is still unresolved: https://github.com/logos-blockchain/lssa/issues/197

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -101,11 +101,11 @@ jobs:
       - name: Install nextest
         run: cargo install cargo-nextest
 
-      - name: Run unit tests
+      - name: Run tests
         env:
           RISC0_DEV_MODE: "1"
           RUST_LOG: "info"
-        run: cargo nextest run --no-fail-fast
+        run: cargo nextest run --no-fail-fast -- --skip tps_test
 
   valid-proof-test:
     runs-on: ubuntu-latest


Pravdyvy (approved):
lgtm

PR #281, opened 2026-01-08
chore(wallet): use --all instead of fake option name all

vpavlin:
## 🎯 Purpose

`wallet config get` uses a "fake" option key `all` to print full config, which feels wrong - user has to know/guess that it is available and it is kinda abusing the option keys.

THis PR adds `--all` flag which is listed in help and leaves `key` for actual options

TO COMPLETE

## ⚙️ Approach

- [ ] add `--all` flag for `wallet config get`
- [ ] remove `all` key

## 🧪 How to Test

```
cargo run --bin wallet config get -h
cargo run --bin wallet config get --all
```

## 🔗 Dependencies

None

## 🔜 Future Work

None

## 📋 PR Completion Checklist

- [x] Complete PR description
- [ ] Add/update tests

### DIFF
--- a/wallet/src/cli/config.rs
+++ b/wallet/src/cli/config.rs
@@ -10,7 +10,13 @@ use crate::{
 #[derive(Subcommand, Debug, Clone)]
 pub enum ConfigSubcommand {
     /// Getter of config fields
-    Get { key: String },
+    Get {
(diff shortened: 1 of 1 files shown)


schouhy (approved):
Thanks!

PR #282, opened 2026-01-09
Indexer block parsing

Pravdyvy:
## 🎯 Purpose

Implementing parsing of blocks through indexer. 

## ⚙️ Approach

Added indexer lib crate. It now can parse from incoming L1 blocks our L2 blocks. Extended tests infrastructure. Added RPC method.

- [x] Indexer crate.
- [x] Indexer L1 block parsing.
- [x] Extended tests infrastructure. 
- [x] Sending messages to sequencer.

## 🧪 How to Test

Corresponding test added into integration tests.
For now decided to not add this test into CI, because we will need to setup nomos circuits and build the node inside CI step which will effectively make it the longest step in our CI.
Instructions what to do with the test:
- Install nomos circuits and run a local nomos node. See instructions [here](https://github.com/logos-blockchain/logos-blockchain). I ran the node with following command: 
```RUST_LOG=info CONSENSUS_SLOT_TIME=15 POL_PROOF_DEV_MODE=true target/debug/nomos-node nodes/nomos-node/config-one-node.yaml --dev-mode-reset-chain-clock```
- Go to `integration_tests`
- Run the indexer test with following command:
`RUST_LOG=info cargo test indexer_run_local_node -- --ignore`
- After some time you will see messages like this:
```
Sent message L2BlockFinalized {
        l2_block_height: 11,
    } to sequencer; status PostIndexerMessageResponse {
        status: "Success",
    }
```

## 🔗 Dependencies

#286 

## 🔜 Future Work

Indexer state transition and history.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests

### CODE REVIEW

File: sequencer_core/src/lib.rs [hunk 1]
```diff
@@ -24,6 +27,10 @@ pub struct SequencerCore {
     mempool: MemPool<EncodedTransaction>,
     sequencer_config: SequencerConfig,
     chain_height: u64,
+    // No logic here for now
+    #[allow(unused)]
+    receiver: Option<Receiver<IndexerToSequencerMessage>>,
```
schouhy: Is the `receiver` used in this PR?

indexer/src/config.rs
```diff
@@ -0,0 +1,8 @@
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct IndexerConfig {
+    pub resubscribe_interval: u64,
+    pub start_delay: u64,
+    pub limit_retry: usize,
```
Arjentix:
```suggestion
    pub resubscribe_interval: Duration,
    pub start_delay: Duration,
    pub max_retries: usize,
```

if deserializing `Duration` is a problem consider [`Duration`](https://docs.rs/humantime/2.3.0/humantime/struct.Duration.html) from [`humantime`](https://crates.io/crates/humantime) library, it's really nice

indexer/src/lib.rs
```diff
@@ -0,0 +1,139 @@
...
+            info!("Block stream joined");
+
+            while let Some(block_info) = stream_pinned.next().await {
+                let header_id = block_info.header_id;
+
+                info!("Observed L1 block at height {}", block_info.height);
+
+                // Simple retry strategy on requests
+                let strategy =
+                    tokio_retry::strategy::FibonacciBackoff::from_millis(self.config.start_delay)
```
Arjentix: Need to mention fibonacci backoff strategy in indexer config. Because it's not obvious. E.g. first backoff strategy I think of is exponential

indexer/src/lib.rs
```diff
@@ -0,0 +1,139 @@
...
+                        .take(self.config.limit_retry);
+
+                if let Some(l1_block) = Retry::spawn(strategy, || {
+                    self.bedrock_client
+                        .0
+                        .get_block_by_id(self.bedrock_url.clone(), header_id)
+                        .inspect_err(|err| warn!("Block fetching failed with err: {err:#?}"))
```
Arjentix: Should also be encapsulated in `BedrockClient`

indexer/src/lib.rs
```diff
@@ -0,0 +1,139 @@
...
+                    Op::ChannelInscribe(InscriptionOp {
+                        channel_id,
+                        inscription,
+                        ..
+                    }) if channel_id == decoded_channel_id => {
+                        borsh::from_slice::<HashableBlockData>(inscription).ok()
+                    }
+                    _ => None,
+                })
+                .collect::<Vec<_>>()
```
Arjentix: I think this line could be safely removed

Arjentix (commented):
in the latest discussions on Discord we came up with that we need a separate Indexer binary. So I wonder, maybe it makes sense to remove Indexer setup from sequencer at all. And run it with a docker compose? Like, you know, Immich has a lot of sub-services as different binaries but in the end you run them all with docker and it just works.
Probably not for this PR anyway.

sequencer_runner/src/lib.rs
```diff
@@ -79,9 +130,13 @@ pub async fn main_runner() -> Result<()> {
...
 
     // ToDo: Add restart on failures
-    let (_, _, main_loop_handle) = startup_sequencer(app_config).await?;
+    let (_, _, main_loop_handle, indexer_loop_handle) = startup_sequencer(app_config).await?;
 
     main_loop_handle.await??;
 
+    if indexer_loop_handle.is_some() {
+        indexer_loop_handle.unwrap().await??;
+    }
```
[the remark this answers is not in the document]
Pravdyvy: I need to think about it. Sequencer may live some time without indexer, while indexer can not, so this order seems correct. However I think that indexer will usually fail in cases where we lost connection to a bedrock and I am not sure that we should continue producing blocks in this case. So maybe both indexer and sequencer need to have cancellation switches.

indexer/src/lib.rs
```diff
@@ -0,0 +1,139 @@
...
+        })
+    }
+
+    pub async fn subscribe_parse_block_stream(&self) -> Result<()> {
+        loop {
+            let mut stream_pinned = Box::pin(
+                self.bedrock_client
+                    .0
+                    .get_lib_stream(self.bedrock_url.clone())
+                    .await?,
```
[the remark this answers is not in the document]
Pravdyvy: Encapsulated

integration_tests/Cargo.toml
```diff
@@ -11,6 +11,8 @@ sequencer_runner.workspace = true
 wallet.workspace = true
 common.workspace = true
 key_protocol.workspace = true
+indexer_core.workspace = true
+url.workspace = true
```
Arjentix: `url` should be grouped with external dependencies, but currently it's grouped with our dependencies

PR #285, opened 2026-01-13
docs(README): replace wallet `--account-id` with `--addr` option

0x-r4bbit:
This is likely a leftover. There's no `--account-id` option in `wallet.` Need to use `--addr` instead.


0x-r4bbit:
Closing this. Turned out I was using an older version.

PR #292, opened 2026-01-16
add `wallet account label` feature

fryorcraken:
## 🎯 Purpose

LSSA's account based approach is such as a user has to manipulate a good number of accounts.

To make the task easier, we add a label feature to the `wallet` CLI.

## ⚙️ Approach

- can add a label to own account via `wallet` CLI
- labels are displayed with `wallet account get` command
- labels are displayed with `wallet account list` command
- labels are persisted.

## 🧪 How to Test

Try the following commands:

- `wallet account label --label "some string" --account-id <...>`
- `wallet account get --account-id <...>`
- `wallet account list`

## 🔗 Dependencies

None

## 🔜 Future Work

- Set the label on account creation like `wallet account new public --label ...`
- Use the label in other commands like `wallet auth-transfer send --from "my_label"` ...

## 📋 PR Completion Checklist

- [x] Complete PR description
- ~~[ ] Add/update tests~~ No test on CLI output

Note: I used AI to get familiar with the code structure. It works but let me know if code architecture isn't ideal.

### DIFF
--- a/wallet/src/chain_storage.rs
+++ b/wallet/src/chain_storage.rs
@@ -11,17 +11,19 @@ use key_protocol::{
 use log::debug;
 use nssa::program::Program;
 
-use crate::config::{InitialAccountData, PersistentAccountData, WalletConfig};
+use crate::config::{InitialAccountData, Label, PersistentAccountData, WalletConfig};
 
 pub struct WalletChainStore {
     pub user_data: NSSAUserData,
     pub wallet_config: WalletConfig,
+    pub labels: HashMap<String, Label>,
 }
 
 impl WalletChainStore {
     pub fn new(
         config: WalletConfig,
         persistent_accounts: Vec<PersistentAccountData>,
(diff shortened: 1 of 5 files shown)


wallet/src/chain_storage.rs
```diff
@@ -16,12 +16,14 @@ use crate::config::{InitialAccountData, PersistentAccountData, WalletConfig};
 pub struct WalletChainStore {
     pub user_data: NSSAUserData,
     pub wallet_config: WalletConfig,
+    pub labels: HashMap<String, String>,
```
Arjentix:
```suggestion
    pub labels: HashMap<AccountId, String>,
```

Maybe like that? As it's not obvious if it's `AccountId -> Label` or `Label -> AccountId` map

fryorcraken:
Merge because it already got approval - @Arjentix if you have any further comment I am happy to do a follow-up

PR #293, opened 2026-01-16
refactor: split token program into crates

Arjentix:
## 🎯 Purpose

Addresses token program part of the following issues:
- https://github.com/logos-blockchain/lssa/issues/256
- https://github.com/logos-blockchain/lssa/issues/253

## ⚙️ Approach

- Main Token Program logic moved to a library crate `token_program` in `programs/token`. This allows to split it into modules to simplify reading but also gives possibility to use this code in another programs if needed.
- Part of that library is moved to another library called `token_core` and is located in `programs/token/core`. This is the common part for all entities interacting with Token Program: Token Program itself, wallet, tests, AMM Program. Now everybody share the same types and no silent API breaks is possible.
- `progam_methods/guest/token.rs` now contains just `main()` function calling implementation functions from `token_program` lib crate
- `token_program` functions and types were refactored to use more clean and strongly-typed approach reducing the amount of bugs
- Now `Instruction` is a proper enum with readable fields
- `TokenDefinition` and `TokenHolding` are now also enums which helps to distinguish between different token sub-types

## 🧪 How to Test

Tests were reworked to use new types, but they're still the same, so just:

```
RISC0_DEV_MODE=1 cargo nextest run
```

## 🔗 Dependencies

None

## 🔜 Future Work

**Do the same for AMM**

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


programs/token/core/src/lib.rs
```diff
@@ -0,0 +1,221 @@
...
+    /// Required accounts:
+    /// - Token Definition account (initialized),
+    /// - Token Holding account (authorized).
+    Burn { amount_to_burn: u128 },
+
+    /// Mint new tokens to the holder's account.
+    ///
+    /// Required accounts:
+    /// - Token Definition account (authorized),
+    /// - Token Holding account.
```
jonesmarvin8: Should this specify "uninitialized or Token Holding for the provided Token Definition account"?

jonesmarvin8 (changes requested):
Looks mostly good. I left a few comments, and the main changes needed were discussed concerning NFTs on [discord](https://discord.com/channels/973324189794697286/1463289322839998475).

programs/token/core/src/lib.rs
Arjentix: Hm, yeah, that's the only place I didn't mention initialization. The idea was that either one is good. But I can see it's not obvious, will document that

programs/token/src/initialize.rs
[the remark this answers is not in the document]
schouhy: I would keep this function. Not sure if there will be cases of other programs that expect an already initialized token account. If so, not having this function makes it a real pain. Let's keep it and see later if it's not used.

PR #300, opened 2026-01-20
fix(wallet): use cryptographically secure entropy for mnemonic generation

0x-r4bbit:
The mnemonic/wallet generation was using a constant zero-byte array for entropy ([0u8; 32]), making all wallets deterministic based solely on the password. This commit introduces proper random entropy using OsRng and enables users to back up their recovery phrase.

Changes:

- SeedHolder::new_mnemonic() now uses OsRng for 256-bit random entropy and returns the generated mnemonic
- Added SeedHolder::from_mnemonic() to recover a wallet from an existing mnemonic phrase
- WalletChainStore::new_storage() returns the mnemonic for user backup
- Added WalletChainStore::restore_storage() for recovery from a mnemonic
- WalletCore::new_init_storage() now returns the mnemonic
- Renamed reset_storage to restore_storage, which accepts a mnemonic for recovery
- CLI displays the recovery phrase when a new wallet is created
- RestoreKeys command now prompts for the mnemonic phrase via read_mnemonic_from_stdin()

Note: The password parameter is retained for future storage encryption but is no longer used in seed derivation (empty passphrase is used
 instead). This means the mnemonic alone is sufficient to recover accounts.

Usage:

On first wallet initialization, users will see:

```
IMPORTANT: Write down your recovery phrase and store it securely. This is the only way to recover your wallet if you lose access.

Recovery phrase:
  word1 word2 word3 ... word24

To restore keys:
wallet restore-keys --depth 5
Input recovery phrase: <24 words>
Input password: <password>
```

### DIFF
--- a/wallet/src/chain_storage.rs
+++ b/wallet/src/chain_storage.rs
@@ -1,6 +1,7 @@
 use std::collections::{BTreeMap, HashMap, btree_map::Entry};
 
 use anyhow::Result;
+use bip39::Mnemonic;
 use key_protocol::{
     key_management::{
         key_tree::{KeyTreePrivate, KeyTreePublic, chain_index::ChainIndex},
@@ -95,7 +96,7 @@ impl WalletChainStore {
         })
     }
 
-    pub fn new_storage(config: WalletConfig, password: String) -> Result<Self> {
+    pub fn new_storage(config: WalletConfig, password: &str) -> Result<(Self, Mnemonic)> {
         let mut public_init_acc_map = BTreeMap::new();
         let mut private_init_acc_map = BTreeMap::new();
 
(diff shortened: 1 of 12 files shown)


wallet/src/chain_storage.rs
```diff
@@ -109,13 +110,42 @@ impl WalletChainStore {
             }
         }
 
-        let public_tree = KeyTreePublic::new(&SeedHolder::new_mnemonic(password.clone()));
-        let private_tree = KeyTreePrivate::new(&SeedHolder::new_mnemonic(password));
+        // TODO: Use password for storage encryption
+        let _ = password;
+        let (seed_holder, mnemonic) = SeedHolder::new_mnemonic("");
+        let public_tree = KeyTreePublic::new(&seed_holder);
+        let private_tree = KeyTreePrivate::new(&seed_holder);
```
0x-r4bbit: Notice how we no longer use the password for the mnemonic. This is because we no longer use zero bytes entropy. We can now use the password for the keystore encryption (and, if we want can re-add it to the seed, but then recovery requires the passphase as well)

wallet/src/lib.rs
```diff
@@ -220,9 +230,13 @@ impl WalletCore {
         &self.storage
     }
 
-    /// Reset storage
-    pub fn reset_storage(&mut self, password: String) -> Result<()> {
-        self.storage = WalletChainStore::new_storage(self.storage.wallet_config.clone(), password)?;
+    /// Restore storage from an existing mnemonic phrase.
+    pub fn restore_storage(&mut self, mnemonic: &Mnemonic, password: &str) -> Result<()> {
```
0x-r4bbit: Can also keep the `reset_storage` name but I thought `restore` or `recover` would be a better fit. Your call

PR #302, opened 2026-01-22
Add pending block resubmit loop to the sequencer

schouhy:
## 🎯 Purpose

This PR adds a task to the sequencer to retry pending blocks. It also adds the nssa state to the sequencer storage and provides a method to delete already finalized blocks.

## ⚙️ Approach

- Add a task on sequencer startup to retry pending blocks
https://github.com/logos-blockchain/lssa/blob/deca3ffe69bb0d8eb18a281ab16f5515ee7d9198/sequencer_runner/src/lib.rs#L48-L51
- Every time a new block is produced and stored on the sequencer store, a snapshot of the state is also stored. This operation is atomic.
https://github.com/logos-blockchain/lssa/blob/deca3ffe69bb0d8eb18a281ab16f5515ee7d9198/sequencer_core/src/lib.rs#L197
- A method to delete finalized blocks is provided. This will be hooked as a callback in the future when the sequencer subscribes to the indexer and receives notification of the last finalized block id.
https://github.com/logos-blockchain/lssa/blob/22258435ff80fb51460be7555d4ee215b6e49e5a/sequencer_core/src/lib.rs#L234

## 🧪 How to Test

Run a nomos node and the sequencer runner. You should see the pending block submission logs. 

## 🔗 Dependencies

None

## 🔜 Future Work

Integrate with indexer

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests

### DIFF
--- a/integration_tests/src/lib.rs
+++ b/integration_tests/src/lib.rs
@@ -40,6 +40,7 @@ static LOGGER: LazyLock<()> = LazyLock::new(env_logger::init);
 pub struct TestContext {
     sequencer_server_handle: ServerHandle,
     sequencer_loop_handle: JoinHandle<Result<()>>,
+    sequencer_retry_pending_blocks_handle: JoinHandle<Result<()>>,
     indexer_loop_handle: Option<JoinHandle<Result<()>>>,
     sequencer_client: SequencerClient,
     wallet: WalletCore,
@@ -94,10 +95,15 @@ impl TestContext {
 
         debug!("Test context setup");
 
-        let (sequencer_server_handle, sequencer_addr, sequencer_loop_handle, temp_sequencer_dir) =
-            Self::setup_sequencer(sequencer_config)
-                .await
-                .context("Failed to setup sequencer")?;
+        let (
+            sequencer_server_handle,
+            sequencer_addr,
+            sequencer_loop_handle,
+            sequencer_retry_pending_blocks_handle,
+            temp_sequencer_dir,
+        ) = Self::setup_sequencer(sequencer_config)
+            .await
+            .context("Failed to setup sequencer")?;
 
         // Convert 0.0.0.0 to 127.0.0.1 for client connections
         // When binding to port 0, the server binds to 0.0.0.0:<random_port>
@@ -130,6 +136,7 @@ impl TestContext {
             Ok(Self {
                 sequencer_server_handle,
                 sequencer_loop_handle,
+                sequencer_retry_pending_blocks_handle,
                 indexer_loop_handle,
                 sequencer_client,
                 wallet,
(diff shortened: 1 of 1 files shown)


davidrusu (approved):

bedrock_client/src/lib.rs
```diff
@@ -17,6 +17,7 @@ pub struct BackoffConfig {
 
 // Simple wrapper
 // maybe extend in the future for our purposes
+#[derive(Clone)]
```
Arjentix: I had to check `CommonHttpClient` internals to be sure that it's an `Arc` inside and cloning will be cheap. Let's add a doc comment about cheap clonning

common/src/block.rs
```diff
@@ -95,6 +106,22 @@ impl From<Block> for HashableBlockData {
     }
 }
 
+mod borsh_msg_id {
+    use std::io::{Read, Write};
+
+    use logos_blockchain_core::mantle::ops::channel::MsgId;
+
+    pub fn serialize<W: Write>(v: &MsgId, w: &mut W) -> std::io::Result<()> {
```
Arjentix: Hm, why do we need this? Doesn't borsh do the same automatically?

common/Cargo.toml
```diff
@@ -17,5 +17,6 @@ log.workspace = true
 hex.workspace = true
 borsh.workspace = true
 base64.workspace = true
+logos-blockchain-core.workspace = true
```
Arjentix: Looks heavy just for MsgId type, but looks like we have no other option

common/src/block.rs
schouhy: the problem is that MsgId is from another crate and it doesn't implement borsh serialize/deserialize. Let me know if you think there's a better approach

common/src/block.rs
```diff
@@ -49,6 +50,11 @@ pub struct Block {
     pub header: BlockHeader,
     pub body: BlockBody,
     pub bedrock_status: BedrockStatus,
+    #[borsh(
+        serialize_with = "borsh_msg_id::serialize",
+        deserialize_with = "borsh_msg_id::deserialize"
+    )]
```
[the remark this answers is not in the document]
schouhy: that syntax doesn't seem to work with borsh

sequencer_runner/src/lib.rs
```diff
@@ -39,8 +45,41 @@ pub async fn startup_sequencer(
...
+                    .expect("Sequencer should be able to retrieve pending blocks");
+                (pending_blocks, client)
+            };
+
+            let Some(client) = block_settlement_client else {
+                continue;
+            };
+
+            info!("Resubmitting {} pending blocks", pending_blocks.len());
+            for block in pending_blocks.iter() {
```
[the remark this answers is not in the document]
schouhy: updated

Arjentix:
LGTM, thanks for addressing my comments

davidrusu:
⛵

PR #345, opened 2026-02-23
feat: add marketplace program example

leohhhn:
## 🎯 Purpose

*What problem does this PR solve or what feature does it add? Mention issues related to it*

This PR adds a simple Marketplace program to the example programs. It also adds a runner script to allow running the first action available in the marketplace: listing items.

The Marketplace program is not complete yet, and I have ran into issues trying to execute the runner for listing an item. Specifically, when executing the runner, I get the `InvalidProgramBehaviour` error without many more details. This is why there are no runners for buying or withdrawing yet. 

This was my first try at creating something using LEE and the surrounding tooling & environment.

[Here](https://hackmd.io/FoyQT4sXQPagoJjlh13drg) is a document with feedback about my onboarding process.


moudyellaz:
Closing this, @leohhhn. It targets `schouhy/full-bedrock-integration`, which was abandoned in January, so this has no path to merge as-is. There is now an `examples/` directory on `dev` if you want to bring a marketplace example back against current APIs.

PR #362, opened 2026-02-26
feat!(key_protocol): update public account keys to be PQ resilient.

jonesmarvin8:
## 🎯 Purpose

Adds PQ resilience to LEE public account keys. **Disclaimer**: This will cause breaking changes between nodes as key and signature logic are altered.

## ⚙️ Approach

Change public account key logic:
- Rename usual private key from `csk` to `cssk`
- Tweaked private key stored in `csk` (used for signing); logic for computing tweaked private key added to `Signature module`.
- Tweaked public key stored in `cpk` (used for verifying signatures)

**Removed unnecessary tests**: `nonharden_child_keys_generation` and `edge_case_child_keys_generation_2_power_31` as non-harden keys are not supported to maintain PQ resilience.

## 🧪 How to Test

RISC0_DEV_MODE=1 cargo test in key_protocol.

## 🔗 Dependencies

Depends on [PR 356](https://github.com/logos-blockchain/lssa/pull/356); uses its private account key modifications.

## 🔜 Future Work

Public key should be renamed to signing key to remove any confusion with BIP-032 public keys.

## 📋 PR Completion Checklist

- [X] Complete PR description
- [X] Add/update tests


moudyellaz (commented):
Thanks for this draft. Left some comments, things are changing on sequencer and indexer files so i would not touch these files

nssa/src/signature/private_key.rs
```diff
@@ -35,6 +36,23 @@ impl PrivateKey {
     pub fn value(&self) -> &[u8; 32] {
         &self.0
     }
+
```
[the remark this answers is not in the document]
jonesmarvin8: Replaced unwrap with expect.

moudyellaz (approved):
lgtm, thanks. Left some comments.

One general note: `PrivateKey::tweak` has no standalone unit test. It's tested indirectly through key_tree tests, but a direct test would help catch regressions in the tweaking logic itself.

nssa/src/signature/private_key.rs
```diff
@@ -60,6 +62,27 @@ impl PrivateKey {
...
+    pub fn tweak(value: &[u8; 32]) -> Result<Self, NssaError> {
+        if !Self::is_valid_key(*value) {
+            return Err(NssaError::InvalidPrivateKey);
+        }
+
+        let sk = k256::SecretKey::from_slice(value).expect("Expect a valid secret key");
+
+        let hashed: [u8; 32] = Impl::hash_bytes(sk.public_key().to_encoded_point(true).as_bytes())
```
schouhy: I think `sk.public_key().to_encoded_point(true)` encodes the point as 33-byte with last 32-byte being the x-coordinate and first byte being the parity flag. Using this to compute the hash deviates from bip341 because public keys in bip341 are always such that the y-coordinate is even and the parity flag is dropped, meaning public keys are 32-byte only.

wallet/configs/debug/wallet_config.json
```diff
@@ -7,14 +7,80 @@
     "initial_accounts": [
         {
             "Public": {
-                "account_id": "CbgR6tj5kWx5oziiFptM7jMvrQeYY3Mzaao6ciuhSr2r",
```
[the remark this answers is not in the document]
schouhy: accound ids have been base58 for a long time. But I think @moudyellaz means another thing. The `pub_sign_key` was previously expressed as a hex, and this PR is changing it to a byte array in the config. The type of `pub_sign_key` is `nssa::PrivateKey` which derives `SerializeDisplay` and `DeserializeFromStr` ([here](https://github.com/logos-blockchain/logos-execution-zone/blob/e5c00282fc2b8fbab563bac31cf352ea92b28ec0/nssa/src/signature/private_key.rs#L12-L13)). So, I think the byte array will make things crash. It should be expressed as hex

schouhy (commented):
Thanks for the PR! I'm leaving two comments to discuss!

nssa/src/signature/private_key.rs
jonesmarvin8: You are correct. However, `tweak` is used for a very subtle and specific use case. In the context of this PR, public accounts have 3 keys (ssk, sk, pk); probably need to fix these labels.

(sk, pk) are the keys used for Schnorr.
ssk is the PQ resilient key. sk is the hash of ssk.

Given a quantum attacker, pk compromises sk. But sk does not compromise ssk.

So, when the time comes, we can upgrade LEE's signature scheme to a PQ signature. This new signature will use ssk from then on.

Thus, using `sk.public_key().to_encoded_point(true)` on an intermediate value from ssk to (sk) should not be a problem.

wallet/configs/debug/wallet_config.json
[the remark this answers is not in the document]
jonesmarvin8: Due to changes to `wallet_config.json`. This is no longer necessary.

key_protocol/src/key_management/key_tree/keys_public.rs
```diff
@@ -39,28 +43,33 @@ impl ChildKeysPublic {
...
+                (*hash_value
+                    .first_chunk::<32>()
+                    .expect("hash_value is 64 bytes, must be safe to get first 32"))
+                .into(),
+            )
+            .expect("Expect a valid k256 scalar"))
+            .add(
+                &k256::Scalar::from_repr((*self.cssk.value()).into())
+                    .expect("Expect a valid k256 scalar"),
+            )
```
schouhy: this is hard to chew. Maybe refactor it to define variables for each term and use them here `try_new(lhs.add(rhs))` like it was before

key_protocol/src/key_management/key_tree/keys_public.rs
```diff
@@ -75,20 +84,9 @@ impl ChildKeysPublic {
...
-            let sk = k256::SecretKey::from_bytes(self.csk.value().into())
-                .expect("32 bytes, within curve order");
-            let pk = sk.public_key();
-            hash_input.extend_from_slice(pk.to_encoded_point(true).as_bytes());
-        } else {
-            // Harden.
-            hash_input.extend_from_slice(&[0_u8]);
-            hash_input.extend_from_slice(self.csk.value());
-        }
+        // Only support harden keys to maintain PQ resilience.
```
schouhy: Can't the wording about "PQ resilience" be a bit confusing? Doesn't it make it look like the scheme is PQ resistant?

nssa/src/signature/private_key.rs
```diff
@@ -60,6 +62,27 @@ impl PrivateKey {
     pub const fn value(&self) -> &[u8; 32] {
         &self.0
     }
+
+    pub fn tweak(value: &[u8; 32]) -> Result<Self, NssaError> {
+        if !Self::is_valid_key(*value) {
+            return Err(NssaError::InvalidPrivateKey);
+        }
+
+        let sk = k256::SecretKey::from_slice(value).expect("Expect a valid secret key");
```
schouhy: `expect` will panic. I think it should return an Err instead given that the function already returns a `Result` type

nssa/src/signature/private_key.rs
schouhy: Ok I see thanks. Can you add this to the docstring of tweak?

nssa/src/signature/private_key.rs
schouhy: Or better I think you can pass a PrivateKey as input instead of &[u8; 32]. That way you avoid the is_valid_key check and also deserializing it inside the tweak function, and the callers avoid passing `cssk.value()`

key_protocol/src/key_management/key_tree/keys_public.rs
```diff
@@ -18,16 +19,19 @@ impl ChildKeysPublic {
...
         let hash_value = hmac_sha512::HMAC::mac(seed, "LEE_master_pub");
 
-        let csk = nssa::PrivateKey::try_new(
+        let cssk = nssa::PrivateKey::try_new(
             *hash_value
                 .first_chunk::<32>()
                 .expect("hash_value is 64 bytes, must be safe to get first 32"),
         )
         .expect("Expect a valid Private Key");
+        let csk = nssa::PrivateKey::tweak(cssk.value()).unwrap();
```
schouhy: let's use `expect`

schouhy (approved):
thanks for the PR. Left some comments

key_protocol/src/key_management/key_tree/keys_public.rs
jonesmarvin8: I'll rephrase. 

For additional context: this key protocol protects the `cssk` from quantum attacks. But the keys used for signatures (tweaked keys) are still vulnerable. When Schnorr signature is replaced, we can continue to use `cssk` and drop the tweaked keys altogether (okay...we'll keep the tweaked public key for account ids). But `csk` won't be used for anything that required authorization.

nssa/src/signature/private_key.rs
jonesmarvin8: It seems that the alternate approach would pass `cssk` through more function calls? Perhaps this is something we can discuss during a clean up/refactoring during Q3?

PR #375, opened 2026-03-02
fix(wallet): await on-chain confirmation before returning from deploy-program

jimmy-claw:
## 🎯 Purpose

Fixes a bug where `wallet deploy-program` exits 0 immediately after TX submission, even if the sequencer rejected it. Caused silent failures during the LEZ demo — command reported success but program was not deployed.

## ⚙️ Approach

- After submitting the deploy TX, poll the sequencer until the program ID appears in state or ~45s timeout (3 blocks)
- Returns exit 1 with a clear error if the program fails to land on-chain
- No change to TX submission logic

## 🧪 How to Test

```bash
# Should exit 0 and confirm deployment
wallet deploy-program path/to/program.bin

# With an invalid binary — should now exit 1 with error (previously silent exit 0)
```

## 🔗 Dependencies

None.

## 🔜 Future Work

- Dedicated `get_program` RPC method instead of polling nonces as a proxy for confirmation

## 📋 PR Completion Checklist

- [x] Complete PR description
- [ ] Add/update tests

### DIFF
--- a/wallet/src/cli/mod.rs
+++ b/wallet/src/cli/mod.rs
@@ -211,15 +211,15 @@ pub async fn execute_subcommand(
             let confirmed = wait_for_tx_confirmation(
                 &wallet_core.sequencer_client,
                 tx_hash,
-                30,   // max attempts
-                1000, // ms between polls
+                23,   // max attempts (3 blocks × 15s, polling every 2s)
+                2000, // ms between polls (2s)
             )
             .await?;
 
             if !confirmed {
                 anyhow::bail!(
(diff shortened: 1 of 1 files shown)


Arjentix (changes requested):

moudyellaz:
Closing this, @jimmy-claw. The behaviour landed in #587, which makes deploy wait for the tx to be included in a block before returning. On dev the deploy path now goes through `poll_transaction` (`lez/wallet/src/lib.rs:685`). Thanks for flagging it originally, and sorry it sat this long.

PR #386, opened 2026-03-07
fix(examples): make tail call example programs work again

0x-r4bbit:
These programs have been referencing incorrect program IDs, resulting in "Unknown program" errors when trying to run them.

This commit fixes the IDs based on the latest source. Ideally this should be fixed by making the program IDs configurable.


Arjentix (approved):
LGTM

Yes, this issue will happen every time we update something

Pravdyvy (approved):
lgtm

jonesmarvin8:
Obsolete. The changed `id`s in this PR are out of date; as are the hardcoded `id`s in main. [`examples::program_deployment::README.md`](https://github.com/logos-blockchain/logos-execution-zone/blob/87f1db203aed8e02613be1066be575d48a778196/examples/program_deployment/README.md?plain=1#L526) explicitly mentions that these ids may need to be updated (and provide the appropriate corresponding command).

cc: @moudyellaz

PR #389, opened 2026-03-09
feat(sequencer, indexer): Use zone-sdk instead of bedrock client 

pradovic:
## 🎯 Purpose

This PR introduces logos-blockchain zone-sdk to hide some complexity from LEZ node in implementing sequencer and indexer. Mainly the part of working with logos blockchain node and submission and lifecycle of transactions.

## ⚙️ Approach

Zone-sdk has taken over the role of submission/resubmitting strategy, along with the back-fill and support for checkpoints.

## 🧪 How to Test
Integration tests and regular tests should pass as before the change.

## 🔗 Dependencies
 No dependencies for now.
 
## 🔜 Future Work

- address todo comments (remove msg_id field from rocksdb)

TO COMPLETE IF APPLICABLE

## 📋 PR Completion Checklist

- [ ] Complete PR description
- [ ] Implement the core functionality
- [ ] Add/update tests


pradovic:
@davidrusu , @bacv , @schouhy  After some debugging I think I've found  an issue with using the current preferred approach for `zone-sdk`.

 The tests pass now using indexer `follow()` API from `zone-sdk`, but ideally we should use the `next_messages()` API since it also reads historical blocks. The integration tests were failing because the pinned docker image (built Feb 3) predates the LIB initialization fix in `47a10384` (merged Feb 16) on logos-blockchain-node. Is there a reason the docker image is pinned to that specific version?
 
If I can update to a newer image, I believe  `next_messages()` approach should work as expected (back-filling before following new blocks). Then I would improve an API on `zone-sdk` a little bit and open the PR for review.

Pravdyvy (approved):
lgtm

Cargo.toml
```diff
@@ -122,11 +120,12 @@ tokio-retry = "0.3.0"
...
-logos-blockchain-chain-service = { git = "https://github.com/logos-blockchain/logos-blockchain.git", rev = "1da154c74b911318fb853d37261f8a05ffe513b4" }
+logos-blockchain-common-http-client = { git = "https://github.com/logos-blockchain/logos-blockchain.git", rev = "ee281a447d95a951752461ee0a6e88eb4a0f17cf" }
+logos-blockchain-key-management-system-service = { git = "https://github.com/logos-blockchain/logos-blockchain.git", rev = "ee281a447d95a951752461ee0a6e88eb4a0f17cf" }
+logos-blockchain-core = { git = "https://github.com/logos-blockchain/logos-blockchain.git", rev = "ee281a447d95a951752461ee0a6e88eb4a0f17cf" }
+logos-blockchain-chain-broadcast-service = { git = "https://github.com/logos-blockchain/logos-blockchain.git", rev = "ee281a447d95a951752461ee0a6e88eb4a0f17cf" }
+logos-blockchain-chain-service = { git = "https://github.com/logos-blockchain/logos-blockchain.git", rev = "ee281a447d95a951752461ee0a6e88eb4a0f17cf" }
+logos-blockchain-zone-sdk = { git = "https://github.com/logos-blockchain/logos-blockchain.git", rev = "ee281a447d95a951752461ee0a6e88eb4a0f17cf" }
```
Arjentix, line 128: These libraries don't have some stable readable tag yet? E.g. on stage we use [0.1.2 release](https://github.com/logos-blockchain/logos-blockchain/releases/tag/0.1.2), can't we use `0.1.2` tag here?

bedrock/docker-compose.yml
```diff
@@ -1,7 +1,7 @@
 services:
 
   logos-blockchain-node-0:
-    image: ghcr.io/logos-blockchain/logos-blockchain@sha256:c5243681b353278cabb562a176f0a5cfbefc2056f18cebc47fe0e3720c29fb12
+    image: ghcr.io/logos-blockchain/logos-blockchain@sha256:9f1829dea335c56f6ff68ae37ea872ed5313b96b69e8ffe143c02b7217de85fc
```
Arjentix, line 4: Can we also pin 0.1.2 here?

configs/docker-all-in-one/indexer_config.json
```diff
@@ -1,12 +1,8 @@
 {
     "home": "./indexer/service",
     "consensus_info_polling_interval": "1s",
-    "bedrock_client_config": {
-        "addr": "http://logos-blockchain-node-0:18080",
-        "backoff": {
-            "start_delay": "100ms",
-            "max_retries": 5
-        }
```
Arjentix, line 9: Hmm, interesting, why? Shouldn't client be able to configure retry strategy?

sequencer/service/Cargo.toml
```diff
@@ -14,7 +14,6 @@ mempool.workspace = true
 sequencer_core = { workspace = true, features = ["testnet"] }
 sequencer_service_protocol.workspace = true
 sequencer_service_rpc = { workspace = true, features = ["server"] }
-indexer_service_rpc = { workspace = true, features = ["client"] }
```
Arjentix, line 17: Strange... Our initial idea was that Indexer listents for blocks in bedrock and broadcasts it to sequencer.

Maybe current approach is better, but I'm not sure. Needs discussion.

sequencer/service/src/lib.rs
```diff
@@ -29,24 +27,18 @@ pub struct SequencerHandle {
     /// Option because of `Drop` which forbids to simply move out of `self` in `stopped()`.
     server_handle: Option<ServerHandle>,
     main_loop_handle: JoinHandle<Result<Never>>,
-    retry_pending_blocks_loop_handle: JoinHandle<Result<Never>>,
-    listen_for_bedrock_blocks_loop_handle: JoinHandle<Result<Never>>,
```
Arjentix, line 33: So this functionality is included into zone sdk now?

sequencer/core/src/lib.rs
```diff
@@ -20,33 +19,31 @@ pub use storage::error::DbError;
...
 };
 
-pub mod block_settlement_client;
+pub mod block_publisher;
 pub mod block_store;
 pub mod config;
+// Kept as a thin client lib for callers that want to query the indexer
+// directly (e.g. integration tests). The sequencer no longer depends on the
+// indexer at runtime — finalization comes from zone-sdk events.
 pub mod indexer_client;
```
Arjentix: I think you can safely remove it

storage/src/sequencer/mod.rs
```diff
@@ -275,6 +287,22 @@ impl RocksDBIO {
         Ok(())
     }
 
+    /// Mark every pending block with `block_id <= last_finalized` as finalized.
+    /// Idempotent — already-finalized blocks are skipped.
+    pub fn clean_pending_blocks_up_to(&self, last_finalized: u64) -> DbResult<()> {
```
Arjentix, line 292:
```suggestion
    pub fn clean_pending_blocks_up_to(&self, last_finalized: BlockId) -> DbResult<()> {
```

indexer/core/src/lib.rs
```diff
@@ -21,25 +20,11 @@ pub mod config;
 
 #[derive(Clone)]
 pub struct IndexerCore {
-    pub bedrock_client: BedrockClient,
+    pub zone_indexer: Arc<ZoneIndexer<NodeHttpClient>>,
```
Arjentix, line 23: I don't quite get it. Did you implemented Indexer and Sequencer inside sdk? Isn't it too specific? I'd expect sdk to be just some kind of client

Arjentix (approved):
LGTM, thanks

Cargo.toml
pradovic: There was some stable tag, but there were also some improvements after it and zone-sdk and release did not really match so I wanted to use the most recent one.

I wanted to re-export crates through sdk so we can unify everything into a single dep as well. I would like to open a separate PR for this and then let's try to pin some sensible release. I think I can do that in the next week or two.

Does that makes sense?

bedrock/docker-compose.yml
pradovic: Same as the above, I will need to make the docker setup with work with the tag as well.

configs/docker-all-in-one/indexer_config.json
pradovic: Good point, I will add configurable params to zone-sdk and then later open a specific PR for this as well.

sequencer/service/Cargo.toml
pradovic: Answered on discord, the zone-sdk sequencer has the finalized events now. Let me know if we need to revert it.

indexer/core/src/lib.rs
pradovic: Asnwered in the discord. The zone-sdk has sequencer and indexer part now, it handles submitting, re-submittion and utility stuff as reconnects, but still leaves business logic to the user, so it does not replace LEZ sequencer or indexer but aims to offload some of the logic and hide the complexity of following canonical non-finalized chain and handles re-submittion.

storage/src/sequencer/mod.rs
pradovic: Rest of the storage module uses u64 directly so I used it for consistency. I think I would need to add the dep to `nssa_core` in order to use it. Maybe it is better to do it in a separate PR for all calls then?

sequencer/core/src/lib.rs
pradovic: Still used in integration tests. I have simplified it and left it only in tests.

Cargo.toml
Arjentix: Cool, sure

storage/src/sequencer/mod.rs
Arjentix: Ah, I see, we can leave it as is

Arjentix (approved):

PR #415, opened 2026-03-27
fix: allow passthrough signer accounts in rule 7 validation

### LINKED ISSUE #339: Programs cannot include previously-used external signer accounts in pre/post states
## Summary

Rule 7 in `validate_execution` prevents programs from including external signer accounts (regular wallets) in their pre/post states if the account has been used before (nonce > 0). This makes it impossible for programs to verify that a specific wallet signed the transaction using the `is_authorized` flag.

## Problem

When building a multisig program that needs to verify a member signed a transaction, the natural pattern is:

1. Include the signer's wallet account in the program's input accounts
2. Check `is_authorized == true` on that account (proves they signed the transaction)
3. Verify membership against the multisig state
4. Return all accounts (including signer) in post_states unchanged

This works fine when the signer wallet has never been used before (`nonce == 0`, account is `Account::default()`). However, once the wallet has been used in any prior transaction (`nonce >= 1`), it fails validation.

### The failing rule

[`nssa/core/src/program.rs` line 268-272](https://github.com/logos-blockchain/lssa/blob/89ce9f322a1fc4214ec818e094e6efc97be02d9c/nssa/core/src/program.rs#L268-L272):

```rust
// 7. If a post state has default program owner, the pre state must have been a default
//    account
if post.account.program_owner == DEFAULT_PROGRAM_ID && pre.account != Account::default() {
    return false;
}
```

The signer wallet:
- Has `program_owner == DEFAULT_PROGRAM_ID` (it's a regular user account, not owned by any program)
- Has `nonce >= 1` (it's been used before), so `pre.account != Account::default()`
- Is returned **unchanged** in post_states

Result: `validate_execution` returns `false` → `InvalidProgramBehavior`

## Reproduction

This is reproducible in the [lez-multisig](https://github.com/jimmy-claw/lez-multisig) e2e test:

1. Create a 2-of-3 multisig with members m1, m2, m3
2. m1 proposes a transfer (nonce becomes 1)
3. m2 approves (nonce becomes 1) — threshold reached
4. m1 calls Execute with accounts `[multisig_state, m1, proposal, vault, recipient]`
5. **Fails** because m1's nonce is now 1 → not `Account::default()` → rule 7 rejects

Sequencer log with instrumentation:
```
validate_execution failed at line 265
[VALIDATE] FAIL #3: validate_execution failed for program [...]
  acct[1] id=... pre_owner=[0,0,0,0,0,0,0,0] post_owner=[0,0,0,0,0,0,0,0]
         pre_bal=0 post_bal=0 pre_data=0 post_data=0 pre_nonce=1 post_nonce=1
```

Account 1 (the executor/signer) is completely unchanged between pre and post, but fails because `nonce=1` makes it non-default.

## Impact

This affects any program that needs to verify an external signer:
- **Multisig programs** — verifying which member approved/executed
- **Governance programs** — verifying voter identity
- **Any program** that uses `is_authorized` on non-PDA accounts for access control

The treasury program ([`treasury_program/src/send.rs`](https://github.com/logos-blockchain/lssa)) has the same pattern (signer as `accounts[3]`) and would hit this issue if the signer had been used in a prior transaction.

## Comparison with Solana

On Solana, this is handled by the runtime's native `Signer` account type — the program receives the account as read-only, the runtime verifies the signature, and the account remains owned by the System Program. Programs never need to "return" signer accounts in their output.

NSSA currently has no equivalent concept. Every account in `pre_states` must appear in `post_states`, and rule 7 prevents returning non-default accounts with `DEFAULT_PROGRAM_ID` owner.

## Possible Solutions

1. **Read-only / passthrough account type**: Allow programs to declare certain input accounts as read-only signers that don't appear in post_states (or are automatically passed through)
2. **Relax rule 7 for unchanged accounts**: If `pre.account == post.account()` (completely unchanged), skip the default-owner check — the program didn't modify anything
3. **Separate signer verification mechanism**: Provide programs with a list of signing account IDs without requiring them in pre/post states

Option 2 seems like the smallest change that preserves the rule's intent (preventing programs from creating non-default accounts with default owner) while allowing legitimate signer verification.

## Environment

- LSSA commit: `89ce9f32` (main branch)
- Rule added by @schouhy in commit `d677db7f` (2025-12-02)
- Guest program: risc0 zkVM

ygd58:
## Summary

Fixes #339.

## Problem

Rule 7 in `validate_execution` rejected any non-default account with `DEFAULT_PROGRAM_ID` owner in post_states, even when completely unchanged. This made it impossible for programs to verify external signers via `is_authorized` when the signer had prior transactions (`nonce > 0`).

## Fix

Skip rule 7 if `pre.account == post.account` (passthrough signer). This enables multisig/governance programs to include signer accounts as read-only inputs.

## What changes

```rust
// Before:
if post.account.program_owner == DEFAULT_PROGRAM_ID && pre.account != Account::default() {
    return false;
}

// After:
if post.account.program_owner == DEFAULT_PROGRAM_ID
    && pre.account != Account::default()
    && pre.account != post.account  // allow unchanged passthrough
{
    return false;
}
```

## Tests

- All 25 existing tests still pass
- Added: `passthrough_signer_account_allowed_in_rule_7` regression test

Fixes #339

### DIFF
--- a/nssa/core/src/program.rs
+++ b/nssa/core/src/program.rs
@@ -300,8 +300,14 @@ pub fn validate_execution(
         }
 
         // 7. If a post state has default program owner, the pre state must have been a default
-        //    account
-        if post.account.program_owner == DEFAULT_PROGRAM_ID && pre.account != Account::default() {
+        //    account. Exception: if the account is completely unchanged (passthrough signer),
+        //    allow it — this enables programs to verify external signers via is_authorized
+        //    without claiming or modifying the signer's account.
+        //    See: https://github.com/logos-blockchain/logos-execution-zone/issues/339
+        if post.account.program_owner == DEFAULT_PROGRAM_ID
+            && pre.account != Account::default()
+            && pre.account != post.account
+        {
             return false;
         }
     }
@@ -372,6 +378,52 @@ mod tests {
         assert!(!account_post_state.requires_claim());
     }
 
+
+    #[test]
+    fn passthrough_signer_account_allowed_in_rule_7() {
+        // Regression test for https://github.com/logos-blockchain/logos-execution-zone/issues/339
+        // Programs should be able to include external signer accounts (nonce > 0) in pre/post
+        // states unchanged, enabling is_authorized checks for multisig/governance programs.
+        use crate::account::{Account, AccountId, AccountWithMetadata, Data};
+
+        let signer_account = Account {
+            program_owner: DEFAULT_PROGRAM_ID,
+            balance: 0,
+            data: Data::default(),
+            nonce: 1u128.into(), // nonce > 0: previously used account
+        };
+
+        let program_id = [1u32; 8];
+        let account_id: AccountId = "11111111111111111111111111111111".parse().unwrap();
+
+        let pre = AccountWithMetadata {
+            account: signer_account.clone(),
+            is_authorized: true,
+            account_id: account_id.clone(),
+        };
+
+        let post = AccountPostState::new(signer_account.clone()); // unchanged
+
+        // Should pass: account is completely unchanged (passthrough signer)
+        let result = validate_execution(
+            &[pre.clone()],
+            &[post],
+            program_id,
+        );
+        assert!(result, "passthrough signer account should be allowed");
+
+        // Should fail: account is modified while still having DEFAULT_PROGRAM_ID owner
+        let mut modified_account = signer_account.clone();
+        modified_account.balance = 100;
+        let post_modified = AccountPostState::new(modified_account);
+
+        let result_modified = validate_execution(
+            &[pre],
+            &[post_modified],
+            program_id,
+        );
+        assert!(!result_modified, "modified default-owner account should be rejected");
+    }
     #[test]
     fn post_state_account_getter() {
         let mut account = Account {


schouhy:
hey thanks for the PR. But non default accounts are meant to be owned by some program. The programs in the use cases you mention can take ownership of the accounts used for authorization instead.

ygd58:
Thanks for the feedback. I understand the concern — the design intent is that non-default accounts should be program-owned.

However, the issue (#339) describes a legitimate use case: a multisig program needs to verify that a specific wallet account signed the transaction using is_authorized=true. The wallet account has nonce > 0 from prior transactions, making it non-default, but the program does not want to claim/own it — it just needs to read-verify the signature.

Taking ownership of the signer account would change its program_owner, which may not be desired for a general-purpose wallet account that is shared across multiple programs.

Would you be open to a more targeted fix — e.g. allowing passthrough only when is_authorized=true on the account? That would restrict the exception to explicitly signed accounts.

schouhy:
If the "wallet account" used for such thing already has a program owner then the current rules are enough I think. If not, and the program needs such signing verification logic, then it would need to include the necessary init functions for the accounts used for such purpose. Multisig can do that. I still don't see the need of changing the execution validation rules

ygd58:
Thanks for the clarification. You are right — the multisig program can include init functions to claim ownership of the signer accounts before using them for authorization. The current validation rules are sufficient for that pattern.

Closing this PR as the use case can be addressed at the program level without changing the execution validation rules.

PR #416, opened 2026-03-27
feat: add owner_program_id field to AccountWithMetadata

### LINKED ISSUE #347: Programs cannot verify account ownership (no self program ID at runtime)
## Problem

Programs currently have no way to verify that an input account is owned by themselves. The zkVM guest receives `AccountWithMetadata` which includes `account`, `is_authorized`, and `account_id` — but not the **owning program ID** of the account.

This creates a class of vulnerabilities where a malicious program can create accounts with identical data layouts and pass them to a legitimate program. The legitimate program deserializes the data, trusts it, and may execute logic based on attacker-controlled state.

## Example Attack

1. Attacker deploys malicious program P that creates accounts with data matching `MultisigState` (threshold=1, members=[attacker])
2. Attacker calls the real multisig `Execute` instruction, passing the fake account as `multisig_account`
3. Real program deserializes the data, sees threshold met -> executes the proposal

## Proposed Solution

Expose the program's own `ProgramId` to the guest at runtime, so programs can check:

```rust
assert_eq!(accounts[0].owner_program_id, self_program_id, "account not owned by this program");
```

Alternatives:
- Add `owner_program_id` field to `AccountWithMetadata`
- Provide a `self_program_id()` host call in the zkVM environment
- Or both

## Current Workaround

Programs can require that all read-only state accounts are also returned as `AccountPostState::new_claimed()` or modified, which forces the runtime to verify ownership. This is hacky and has performance implications.

## References

- Discovered during review of [lez-multisig](https://github.com/jimmy-claw/lez-multisig)
- Similar to Solana's `owner` field on `AccountInfo`

ygd58:
## Summary

Fixes #347.

Programs currently have no way to verify that an input account is owned by themselves, creating a class of spoofing vulnerabilities.

## Changes

Added optional `owner_program_id` field to `AccountWithMetadata`:

```rust
pub struct AccountWithMetadata {
    pub account: Account,
    pub is_authorized: bool,
    pub account_id: AccountId,
    /// The program that owns this account.
    #[serde(default)]
    pub owner_program_id: Option<ProgramId>,
}
```

Added `with_owner_program_id()` builder method for easy construction.

## Usage in programs

```rust
const SELF_PROGRAM_ID: ProgramId = ...; // from methods/build.rs

if let Some(owner) = account.owner_program_id {
    assert_eq!(owner, SELF_PROGRAM_ID, "account not owned by this program");
}
```

## Backward compatible

- `#[serde(default)]` means existing data deserializes with `None`
- `AccountWithMetadata::new()` sets `owner_program_id = None`
- All 35 existing tests pass

Fixes #347

### DIFF
--- a/nssa/core/src/account.rs
+++ b/nssa/core/src/account.rs
@@ -126,6 +126,11 @@ pub struct AccountWithMetadata {
     pub account: Account,
     pub is_authorized: bool,
     pub account_id: AccountId,
+    /// The program that owns this account. Programs can use this to verify
+    /// that an input account is owned by themselves, preventing spoofing attacks.
+    /// See: https://github.com/logos-blockchain/logos-execution-zone/issues/347
+    #[serde(default)]
+    pub owner_program_id: Option<crate::program::ProgramId>,
 }
 
 #[cfg(feature = "host")]
@@ -135,7 +140,13 @@ impl AccountWithMetadata {
             account,
             is_authorized,
             account_id: account_id.into(),
+            owner_program_id: None,
         }
+    }
+
+    pub fn with_owner_program_id(mut self, program_id: crate::program::ProgramId) -> Self {
+        self.owner_program_id = Some(program_id);
+        self
     }
 }
 
--- a/nssa/src/public_transaction/transaction.rs
+++ b/nssa/src/public_transaction/transaction.rs
@@ -109,11 +109,20 @@ impl PublicTransaction {
             .account_ids
             .iter()
             .map(|account_id| {
+                let account = state.get_account_by_id(*account_id);
+                let owner_program_id = account.program_owner;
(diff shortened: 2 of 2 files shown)


ygd58:
## Update: Full implementation ✅

Added owner_program_id population in sequencer (public_transaction/transaction.rs).

Now when building pre_states, each account's owner_program_id is set from account.program_owner.

Programs can now reliably verify ownership:

    if let Some(owner) = account.owner_program_id {
        assert_eq!(owner, SELF_PROGRAM_ID, "account not owned by this program");
    }

35 tests passing. Ready for review.

schouhy:
`AccountWithMetadata` already has the program owner as it's part of the account that it is wrapping.

Issue #244, opened 2025-12-15
Proper sequencer haelthcheck


Arjentix:
Unfortunately it is not really the health check for sequencer, it is health check for wallet. This coomand just fetch ids of programs.
This ids are stored in server memory, so theoretically seqencer may experience problems in main loop and still respond at this request. 
Probably should add an issue to add real healthcheck method for sequencer. This method should run this checkhealth and, in addition, extract latest block then wait, then extract latest block again to be sure, that number of last block goes up.

_Originally posted by @Pravdyvy in https://github.com/logos-blockchain/lssa/pull/231#discussion_r2613442525_

PR #418, opened 2026-03-28
feat: improve sequencer healthcheck to return HealthStatus

ygd58:
Refs #244.

Returns HealthStatus { chain_height, is_healthy } instead of ().
Clients can poll chain_height to verify sequencer is progressing.

Follow-up: add block progression check inside the endpoint.


schouhy:
Thanks for the PR. The sequencer service already has endpoints to return chain height.

moudyellaz:
Closing this, @ygd58. The chain-height data it surfaces is already available via `getLastBlockId`. A fuller healthcheck report is still worth having, but it should start from the TODO in the RPC crates rather than this branch.

PR #424, opened 2026-04-01
feat: make ValidityWindow generic with Display and datetime conversion

ygd58:
## Summary

Refs #421.

## Changes

Made `ValidityWindow` generic over its bound type `T`:

```rust
// Before
pub struct ValidityWindow(pub (Option<BlockId>, Option<BlockId>));

// After
pub struct ValidityWindow<T = BlockId>(pub (Option<T>, Option<T>>));
```

## Generic Display

```rust
impl<T: Display> Display for ValidityWindow<T> { ... }
```

## DateTime Conversion

```rust
// Convert block-id window to UTC datetime window
let dt_window = window.to_datetime_window(|block_id| get_block_timestamp_ms(block_id));
println!({}, dt_window);
// [2026-03-28 19:49:25 UTC, 2026-03-28 20:49:25 UTC)
```

## Backward Compatible

- Default type parameter is `BlockId` — existing code unchanged
- `#[serde(transparent)]` ensures JSON serialization unchanged

Refs #421


indexer/service/protocol/src/lib.rs
```diff
@@ -316,6 +317,26 @@ impl Display for ValidityWindow {
...
+    {
+        let convert = |id: &BlockId| {
+            block_to_timestamp_ms(*id)
+                .and_then(|ms| chrono::DateTime::from_timestamp_millis(ms as i64))
+        };
+        ValidityWindow((
+            self.0.0.as_ref().and_then(&convert),
+            self.0.1.as_ref().and_then(&convert),
+        ))
+    }
```
schouhy, line 337: How would this be used? Blocks have id and a timestamp already.

schouhy (commented):
Thanks for the PR. Left a comment

ygd58:
Thanks for the question! The use case is displaying validity windows in a human-readable format for tooling (CLIs, explorers, logs).

Currently ValidityWindow<BlockId> can only show raw block IDs like [103311, 103400), which is not meaningful to users. With Display and datetime conversion, a CLI or explorer can show [2026-03-28 19:49 UTC, 2026-03-28 20:49 UTC) instead.

For example, logos-inspector (https://github.com/ygd58/logos-inspector) displays block info to developers. When showing a transaction's validity window, a human-readable time range is much more useful than raw block IDs.

That said, if this is out of scope for the core library, I am happy to implement the conversion in tooling layers instead.

moudyellaz:
Closing this, @ygd58. `ValidityWindow<T>` is already generic on dev and the indexer has its own `Display` impl, so what remains is the datetime helper, which has no caller in the tree.

Issue #209, opened 2025-12-01
Shortcut function to create key tree from list of nodes


Pravdyvy:
Add shortcut function to create key tree from list of nodes

PR #425, opened 2026-04-01
feat: add KeyTree::from_nodes() with consistency check

ygd58:
## Summary

Implements the TODO at `key_tree/mod.rs:62`.

Refs #209.

## Changes

Added `KeyTree::from_nodes(nodes: Vec<(ChainIndex, N)>) -> Result<Self>`:

```rust
// Create a tree from existing nodes
    (ChainIndex::root(), root_node),
    (root.nth_child(0), child_node),
])?;
```

## Consistency Checks

- List must not be empty
- Root node (`ChainIndex::root()`) must be present
- Every non-root node must have its parent present in the list

## Tests

- `from_nodes_empty_fails` ✅
- `from_nodes_missing_root_fails` ✅
- `from_nodes_roundtrip` ✅

All 40 existing tests still pass.

Refs #209

### DIFF
--- a/key_protocol/src/key_management/key_tree/mod.rs
+++ b/key_protocol/src/key_management/key_tree/mod.rs
@@ -61,6 +61,57 @@ impl<N: KeyNode> KeyTree<N> {
 
     // ToDo: Add function to create a tree from list of nodes with consistency check.
 
+    /// Creates a `KeyTree` from a list of `(ChainIndex, N)` pairs.
+    ///
+    /// Performs a consistency check: every non-root node must have its parent present
+    /// in the provided list. Returns an error if the check fails.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if:
+    /// - The list is empty
+    /// - A non-root node's parent is missing from the list
+    /// - The root node (`ChainIndex::root()`) is missing
+    pub fn from_nodes(nodes: Vec<(ChainIndex, N)>) -> Result<Self> {
+        use anyhow::bail;
+
+        if nodes.is_empty() {
+            bail!("Cannot create KeyTree from empty node list");
+        }
+
+        let key_map: BTreeMap<ChainIndex, N> = nodes.into_iter().collect();
+
+        // Consistency check: root must be present
+        if !key_map.contains_key(&ChainIndex::root()) {
+            bail!("Node list must contain the root node");
+        }
+
+        // Consistency check: every non-root node must have its parent present
+        for chain_index in key_map.keys() {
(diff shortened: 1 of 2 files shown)


schouhy (commented):
Thanks for the PR. I wouldn't add functions to our code that are not needed. Would this allow to simplify our code if used? and where?
@Pravdyvy tagging you here because this is related to a TODO added by you some time ago.

ygd58:
Thanks for the review. Looking at the codebase, the most natural use case would be in wallet storage/recovery scenarios  when a KeyTree needs to be reconstructed from persisted flat node data (e.g. from JSON storage or a database).

The consistency checks (non-empty, root present, all parents present) provide safety guarantees that raw serde deserialization does not.

That said, if there is no current code path that would immediately benefit from this, I am happy to close this PR. The TODO at key_tree/mod.rs:62 was the motivating reference @Pravdyvy, could you clarify the intended use case?

Pravdyvy:
@ygd58 @schouhy In wallet we construct trees from file via multiple iterations of following:
https://github.com/logos-blockchain/logos-execution-zone/blob/00d3140490406d869ec994121b7f5e6c55925caa/wallet/src/chain_storage.rs#L70
There is no check for tree consistency.
If you modify your storage, or, more probably, key protocol is changed, tree will not signal about it there.

ygd58:
Thanks @Pravdyvy! That is exactly the use case.

In chain_storage.rs (line 66-84), nodes are inserted one by one into the tree via multiple insert() calls with no consistency validation. If storage is corrupted or the key protocol changes, the tree could be built in an inconsistent state silently.

KeyTree::from_nodes() would allow replacing the iterative insert pattern with a single call that validates:
- Root node is present
- All parent nodes exist before children
- Tree is non-empty

This would make storage loading fail-fast with a clear error instead of silently building a broken tree.

I can update the PR to show a concrete refactor of chain_storage.rs using from_nodes() if that would help demonstrate the value.

ygd58:
Updated the PR to include a concrete refactor of chain_storage.rs using from_nodes().

The iterative insert() pattern is replaced with a single from_nodes() call that validates tree consistency on storage load. This makes the wallet fail-fast with a clear error if storage is corrupted or the key protocol changes, instead of silently building a broken tree.

jonesmarvin8:
@ygd58 Can you rebase and update (as necessary)? Thanks.

moudyellaz:
Closing this, @ygd58. `KeyTree` is still there, but nothing in the tree would call `from_nodes`.

PR #427, opened 2026-04-02
Simple indexer FFI

Pravdyvy:
## 🎯 Purpose

First step of introducing indexer module.

## ⚙️ Approach

Implemented `indexer_ffi` crate in a similar way to bedrock node C bindings. For now it starts RPC on its own and this is an intended way of communicating with indexer service.

- [x] Crate `indexer_ffi` added.
- [x] Add corresponding integration tests.

Also added an item into `deny.toml`. There is issues with `rand` v0.8.5, but it is here transitively for couple of dependencies. It would not even be problematic to fix(just by bumping deps versions), except it is a dependency in a couple `logos-blockchain-*` crates. If I bump them, we may break integration or integrations tests, or both.

## 🧪 How to Test

Added integrations tests to tests indexer ffi into `indexer.rs`.

## 🔗 Dependencies

None

## 🔜 Future Work

- Deal with new deny item.
- Add query API to `indexer_ffi`. This will be necessary in the future for module completeness.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


integration_tests/src/test_context_ffi.rs
```diff
@@ -0,0 +1,480 @@
...
+            initial_data,
+        )
+        .context("Failed to create Sequencer config")?;
+
+        let sequencer_handle = sequencer_service::run(config, 0).await?;
+
+        Ok((sequencer_handle, temp_sequencer_dir))
+    }
+
+    async fn setup_wallet(
```
schouhy: Isn't this the very same function that already exists in `lib.rs`

integration_tests/tests/indexer.rs
schouhy, line 1: I think I would move the indexer ffi tests added here to a new dedicated test file for them `indexer_ffi.rs`

schouhy (approved):
LGTM. Left some minor comments

integration_tests/src/test_context_ffi.rs
Pravdyvy: It is an associated function there. Anyway, moved all associated setup functions from `lib.rs` and from `test_context_ffi.rs` into new module `setup.rs`.

integration_tests/tests/indexer.rs
Pravdyvy: Moved

indexer_ffi/src/indexer.rs
```diff
@@ -0,0 +1,81 @@
...
+                .cast::<IndexerHandle>()
+                .as_ref()
+                .expect("Indexer Handle must be non-null pointer")
+        }
+    }
+}
+
+// Implement Drop to prevent memory leaks
+impl Drop for IndexerServiceFFI {
+    fn drop(&mut self) {
```
Arjentix, line 71: Would be nice to apply destructuring pattern here

Arjentix (approved):
LGTM, thanks

indexer_ffi/src/indexer.rs
Pravdyvy: Do you want something like:
```
let IndexerServiceFFI { indexer_handle, runtime } = self;
```
?
This will create mutable references into fields, which would need to be dereferenced anyway later to produce references into Rust objects. So I do not see the point.

Arjentix: Yes, I mean this pattern. You'd still need to do the same thing but that will ensure that when someone else adds something into the structure they don't forget to update `Drop` impl because compilation would break otherwise

Pravdyvy: Done

PR #431, opened 2026-04-06
Avoid computing program ids to speed up tests

schouhy:
## 🎯 Purpose

Tests were slow because every Program constructor called `compute_image_id()` at runtime. This PR fixes that.

Approx numbers for the crate tests affected by this change:
Before: `nssa` 153s, `amm_program` 92s, `sequencer_core` 4s
After: `nssa` 6s, `amm_program` 0.3s, `sequencer_core` 0.06s

## ⚙️ Approach
- Replace `Program::new(ELF.to_vec())` with direct struct construction using the pre-computed *_ID constants provided by risc0 build, in all program constructors.

## 🧪 How to Test

Run unit tests and check they are faster:
`RISC0_DEV_MODE=1 cargo nextest run --workspace --exclude integration_tests`

## 🔗 Dependencies

None

## 🔜 Future Work

None

## 📋 PR Completion Checklist

- [x] Complete PR description
- [ ] Add/update tests


Arjentix (approved):
Good catch, thanks! Our tests were indeed too slow but I was blaming integration tests, because locally I run everything at once

nssa/src/program.rs
```diff
@@ -98,44 +100,54 @@ impl Program {
 
     #[must_use]
     pub fn authenticated_transfer_program() -> Self {
-        // This unwrap won't panic since the `AUTHENTICATED_TRANSFER_ELF` comes from risc0 build of
-        // `program_methods`
-        Self::new(AUTHENTICATED_TRANSFER_ELF.to_vec()).unwrap()
+        Self {
```
moudyellaz: The old `Program::new` path recomputed the image id from the ELF, so any drift between the cached `*_ID` and `*_ELF` constants would have been caught. After this change a mismatched pair would silently produce a wrong-id Program. Worth a one-shot test that iterates the constructors and asserts `id == compute_image_id(elf)`.

moudyellaz (approved):
nice speedup. one safety nit inline. also worth dropping the before/after numbers in the description so we know what we're trading the runtime check for.

nssa/src/program.rs
schouhy: done

PR #433, opened 2026-04-06
feat: introduce more descriptive error messages for public execution

Arjentix:
## 🎯 Purpose

Programs public execution is hard to debug, because you can get `InvalidProgramBehavior` for various reasons. Some of them even are user mistake. But current error is not descriptive and it's hard to understand what's going wrong.

## ⚙️ Approach

This PR makes `InvalidProgramBehavior` to contain different failure reasons as enum variants. This makes debugging easier and also forces tests which checks for error type to be more robust.

## 🧪 How to Test

See CI.

## 🔗 Dependencies

None

## 🔜 Future Work

None

## 📋 PR Completion Checklist

- [x] Complete PR description
~- [ ] Add/update tests~


moudyellaz (changes requested):
thanks, looks good. left some minor comments

nssa/src/error.rs
```diff
@@ -17,7 +21,7 @@ pub enum NssaError {
     InvalidInput(String),
 
     #[error("Program violated execution rules")]
-    InvalidProgramBehavior,
+    InvalidProgramBehavior(#[from] InvalidProgramBehaviorError),
```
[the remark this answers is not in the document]
Arjentix: In fact your suggestion makes sense for variants containing just strings, but in my case that's another type implementing `Error` and I have `#[from]`. So `thiserror` will understand that internal error is the *source* of this error. In that case the information about underlying error is already there and will be included in [`std::error::Error::source()`](https://doc.rust-lang.org/std/error/trait.Error.html#method.source) implementation.

Even dtonlay (author of `thiserror`, `anyhow`, `serde` and others) [suggests here](https://users.rust-lang.org/t/do-people-not-care-about-printable-error-chains-a-k-a-how-to-nicely-implement-display-for-an-error/35362/2) to use `#[source]` or `#[from]` instead of `{0}`.

Also check out [this playground example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=7e11482be6a2b2d1e54ba427575193d8).

From here you can see that to see a full trace you should wrap error to `anyhow::Error`. See [here](https://docs.rs/anyhow/latest/anyhow/struct.Error.html#display-representations) for docs about displaying `anyhow::Error`. Another option is to implement iteration over error chain manually, but this is too long.

I also find this confusing and not clear in Rust error handling, but this is what we have. So whenever you want to show the full error trace as a string you should better do something like: `format!("{:#}", anyhow!(error))`

program_methods/guest/src/bin/privacy_preserving_circuit.rs
```diff
@@ -116,12 +116,17 @@ impl ExecutionState {
 
             // Check that the program is well behaved.
             // See the # Programs section for the definition of the `validate_execution` method.
-            let execution_valid = validate_execution(
+            let validated_execution = validate_execution(
                 &program_output.pre_states,
                 &program_output.post_states,
                 chained_call.program_id,
```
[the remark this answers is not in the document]
Arjentix: Should be no difference at all

nssa/core/src/program.rs
```diff
@@ -402,6 +402,75 @@ impl WrappedBalanceSum {
...
+    #[error("Unallowed modification of nonce for account {account_id}")]
+    ModifiedNonce { account_id: AccountId },
+
+    #[error("Unallowed modification of program owner for account {account_id}")]
+    ModifiedProgramOwner { account_id: AccountId },
+
+    #[error(
+        "Trying to decrease balance of account {account_id} owned by {owner_program_id:?} in a program {executing_program_id:?} which is not the owner"
+    )]
+    DecreasingBalanceNotOwnedByProgram {
```
schouhy:
```suggestion
    UnauthorizedBalanceDecrease {
```

nssa/src/error.rs
```diff
@@ -77,6 +81,55 @@ pub enum NssaError {
...
+    MismatchedPdaClaim {
+        expected: AccountId,
+        actual: AccountId,
+    },
+
+    #[error("Default account {account_id} was modified without being claimed")]
+    DefaultAccountModifiedWithoutClaim { account_id: AccountId },
+
+    #[error("Called program {program_id:?} which is not listed in dependencies")]
+    CalledProgramWhichIsNotListedInDependencies { program_id: ProgramId },
```
schouhy:
```suggestion
    UndeclaredProgramDependency { program_id: ProgramId },
```

nssa/core/src/program.rs
```diff
@@ -402,6 +402,75 @@ impl WrappedBalanceSum {
...
+    )]
+    UnauthorizedDataModification {
+        account_id: AccountId,
+        executing_program_id: ProgramId,
+    },
+
+    #[error(
+        "Post-state for account {account_id} has default program owner but pre-state was not default"
+    )]
+    AccountWithDefaultOwnerMustBeDefault { account_id: AccountId },
```
schouhy:
```suggestion
    NonDefaultAccountWithDefaultOwner { account_id: AccountId },
```

schouhy (approved):
lgtm. Some of the variant names I found a bit awkward. Left some suggestions in case you want to consider them.

moudyellaz (approved):

PR #439, opened 2026-04-08
chore(ci): enable indexer integration tests with new docker image

pradovic:
## 🎯 Purpose

Use new `logos-blockchain-node` image and enables indexer integration tests. The test is reduced to a single node with modified consensus parameters to make it faster.

Important: The PR does not fix a bug with a tx transaction, it is marked as TODO and needs a follow up PR.

## 🧪 How to Test

*How to verify that this PR works as intended.*

Indexer integration tests should be green.

## 🔜 Future Work

Need to solve the todo, add proper info to the genesis tx.

TO COMPLETE IF APPLICABLE

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


integration_tests/tests/indexer.rs
```diff
@@ -17,27 +17,52 @@ use nssa::AccountId;
 use tokio::test;
 use wallet::cli::{Command, programs::native_token_transfer::AuthTransferSubcommand};
 
-/// Timeout in milliseconds to reliably await for block finalization.
-const L2_TO_L1_TIMEOUT_MILLIS: u64 = 600_000;
+/// Maximum time to wait for the indexer to catch up to the sequencer.
+const L2_TO_L1_TIMEOUT_MILLIS: u64 = 180_000;
```
Arjentix, line 21: This is super cool

bedrock/scripts/run_logos_blockchain_node.sh
Arjentix, line 1: I wish we could just run bedrock without this complex steps. Is it possible to deliver something like *easy startup* into bedrock image? Because this script doesn't look user friendly

Arjentix (approved):
LGTM, thank you! I think we can address the TODO in another PR.

bedrock/scripts/run_logos_blockchain_node.sh
pradovic: It is possible that I have complicated it for no reason, or that we can improve it in the future. @bacv Could you take a look pls, as you know this part the best?

pradovic: @Arjentix We synced, and simplified the script a bit. The config file is from logos-blockchain-node repo, and we just need to modify the chain start time for tests.

schouhy (approved):
LGTM thanks! (deny check is unhappy)

pradovic:
@schouhy Thanks! I believe deny is a present problem on main that just started to appear, not connected to this PR (I tried to run cargo-deny from main). I tried to solve it easily by introducing deny file and ignoring the warning, but it opened a can of worms. I think you need to add deny.toml and make the check work with all current libs, but I might be wrong.

bedrock/scripts/run_logos_blockchain_node.sh
Arjentix: @pradovic , good, if it's a universal thing, maybe it can be included in the image as an entrypoint then?

PR #441, opened 2026-04-11
fix(sequencer): replace panicking .expect() calls with typed error returns

paschal533:
## Summary

The sequencer service was panicking and crashing the whole process when it received a transaction it could not encode to check its size. Two call sites in `send_transaction` used `.expect()` on operations that can legitimately fail with well-formed client input:

- `borsh::to_vec(&tx).expect(...)`, can fail if the transaction graph contains types that cannot be serialized
- `u64::try_from(encoded_tx.len()).expect(...)`, truncation on platforms where `usize > u64` maximum, which is not the current target but is still unreachable-panicking code

Instead of killing the server, these paths now return typed `TransactionMalformationError` variants over RPC so the caller gets a meaningful error code (`InvalidParams`) and message.

### Changes

- **`common/src/transaction.rs`**, add `From<TransactionMalformationError> for ErrorObjectOwned` mapping every variant to `InvalidParams`. Three unit tests verify the mapping for each variant.
- **`common/Cargo.toml`**, add `jsonrpsee` workspace dependency (needed by the `From` impl).
- **`sequencer/service/src/service.rs`**, replace both `.expect()` sites with `map_err(ErrorObjectOwned::from)?`, using the new `From` impl. The transaction-too-large check now uses `saturating_sub` to keep the header overhead arithmetic panic-free as well.

### Why this matters

A single malformed or oversized transaction submitted by any client could previously take down the entire sequencer, requiring a manual restart. With this change the server stays alive and returns the error to the sender.

## Test plan

- [ ] `cargo test -p common`, three new malformation error mapping tests
- [ ] `cargo test -p sequencer_service send_transaction`, existing tests still pass; the new `send_transaction_too_large_returns_invalid_params` test verifies the size limit path returns `InvalidParams` (not a panic)

### DIFF
--- a/common/src/transaction.rs
+++ b/common/src/transaction.rs
@@ -1,4 +1,5 @@
 use borsh::{BorshDeserialize, BorshSerialize};
+use jsonrpsee::types::{ErrorCode, ErrorObjectOwned};
 use log::warn;
 use nssa::{AccountId, V03State, ValidatedStateDiff};
 use nssa_core::{BlockId, Timestamp};
@@ -154,6 +155,12 @@ pub enum TransactionMalformationError {
     TransactionTooLarge { size: usize, max: usize },
 }
 
+impl From<TransactionMalformationError> for ErrorObjectOwned {
+    fn from(err: TransactionMalformationError) -> Self {
+        ErrorObjectOwned::owned(ErrorCode::InvalidParams.code(), err.to_string(), None::<()>)
+    }
+}
+
 /// Returns the canonical Clock Program invocation transaction for the given block timestamp.
 /// Every valid block must end with exactly one occurrence of this transaction.
 #[must_use]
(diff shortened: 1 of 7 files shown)


schouhy (approved):
Thanks for the PR. Good catch!

Pravdyvy (approved):
lgtm

common/src/transaction.rs
```diff
@@ -154,6 +155,12 @@ pub enum TransactionMalformationError {
     TransactionTooLarge { size: usize, max: usize },
 }
 
+impl From<TransactionMalformationError> for ErrorObjectOwned {
+    fn from(err: TransactionMalformationError) -> Self {
+        ErrorObjectOwned::owned(ErrorCode::InvalidParams.code(), err.to_string(), None::<()>)
+    }
+}
```
Arjentix: Honestly I think It's better to not to have this in that crate.

1. This introduces heavy dependency on `jsonrpsee` in a pretty generic `common` crate
2. It's logically strange for `common` crate to care about concrete transport errors like errors of `jsonrpsee`

I understand that you did this because Rust requires to have such conversion in the same crate it was defined. But for this concrete usage I'd better have it in `sequencer_service` crate implemented as a function.

sequencer/service/src/service.rs
```diff
@@ -49,30 +49,26 @@ impl<BC: BlockSettlementClientTrait + Send + 'static, IC: IndexerClientTrait + S
 
         let tx_hash = tx.hash();
 
-        let encoded_tx =
-            borsh::to_vec(&tx).expect("Transaction borsh serialization should not fail");
-        let tx_size = u64::try_from(encoded_tx.len()).expect("Transaction size should fit in u64");
+        let encoded_tx = borsh::to_vec(&tx).map_err(|_| {
+            ErrorObjectOwned::from(TransactionMalformationError::FailedToDecode { tx: tx_hash })
```
Arjentix: It's not decoding, it's encoding, so error variant is wrong here. Probably doesn't make sense to use types from `common` here, instead we could just use [`owned()`](https://docs.rs/jsonrpsee-types/0.26.0/jsonrpsee_types/error/struct.ErrorObject.html#method.owned) constructor

sequencer/service/src/service.rs
```diff
@@ -49,30 +49,26 @@ impl<BC: BlockSettlementClientTrait + Send + 'static, IC: IndexerClientTrait + S
...
+        })?;
+        let tx_size = u64::try_from(encoded_tx.len()).unwrap_or(u64::MAX);
```
Arjentix: IMO expect was good here. We don't expect to work on platforms where `usize` doesn't fit into `u64`, because on must platform it will fit

sequencer/service/src/service.rs
```diff
@@ -49,30 +49,26 @@ impl<BC: BlockSettlementClientTrait + Send + 'static, IC: IndexerClientTrait + S
...
 
         let max_tx_size = self.max_block_size.saturating_sub(BLOCK_HEADER_OVERHEAD);
 
         if tx_size > max_tx_size {
-            return Err(ErrorObjectOwned::owned(
-                ErrorCode::InvalidParams.code(),
-                format!("Transaction too large: size {tx_size}, max {max_tx_size}"),
-                None::<()>,
+            return Err(ErrorObjectOwned::from(
+                TransactionMalformationError::TransactionTooLarge {
```
Arjentix: Looks like this is the only use of `TransactionMalformationError` variant. So I suggest remove it at all and use direct construction of `ErroObjectOwned`

Arjentix (changes requested):

paschal533:
Hi @Arjentix Thank you for the review! I have made the necessary changes.

moudyellaz:
Closing this, @paschal533. The file it patches no longer exists: the RPC server became an actor and the code moved to `lez/sequencer/actors/rpc_server/src/actor/service.rs`. The change is still worth making there, `format!("{err:?}")` at line 76 should be `err.to_string()` so the error renders readably. We will pick that up separately.

PR #442, opened 2026-04-11
feat(sequencer): add persistent transaction receipt system

paschal533:
## Summary

Adds a `get_transaction_receipt` RPC method that answers the question "what happened to the transaction I submitted?" with a structured, typed response rather than the current `null`-returning `get_transaction`.

The four-tier lookup gives callers the exact status of any hash they have ever submitted, even across sequencer restarts:

1. **Rejected**, checked first from a new RocksDB column family that stores rejection records durably. Includes the human-readable reason and a millisecond timestamp.
2. **Included** the block store is searched for the transaction hash; if found, the block ID and the block's timestamp are returned.
3. **Pending**, an in-memory set tracks hashes inserted by `send_transaction` that have not yet reached a terminal state.
4. **Unknown**, hash was never seen, is invalid, or the pending set entry was evicted.

Receipt entries in the pending set are lazily evicted when the receipt is
first queried for a terminal-state transaction, bounding memory growth for
any transaction that callers actually follow up on.

### Changes

- **`common/src/receipt.rs`** (new), `TxStatus`, `TxReceipt`, `RejectedTxRecord` types with Borsh and serde derivations.
- **`common/src/lib.rs`**, expose `pub mod receipt`.
- **`storage/src/sequencer/mod.rs`**, new `cf_rejected_tx` column family, `put_rejected_tx` and `get_rejected_tx` methods.
- **`sequencer/core/src/block_store.rs`**, `store_rejected_tx`, `get_rejected_tx`, and `get_block_id_for_tx` methods.
- **`sequencer/core/src/lib.rs`**, when a transaction fails validation during block production the rejection is now persisted to RocksDB instead of being silently dropped.
- **`sequencer/service/src/service.rs`**, `pending_txs` tracking field,  `get_transaction_receipt` implementation, lazy pending set eviction.
- **`sequencer/service/rpc/src/lib.rs`** and **`sequencer/service/protocol/src/lib.rs`**, RPC trait method and protocol type exports.

### Stacking note

This PR builds on top of #441 (crash fix). The diff against main will include those commits too; the meaningful change in this PR starts at the `feat(common): add TxStatus` commit.

## Test plan

- [ ] `cargo test -p common`, Borsh and serde roundtrip tests for receipt types
- [ ] `cargo test -p sequencer_service get_receipt`, receipt lifecycle tests: pending after submit, unknown for unseen hash
- [ ] Manual: submit a transaction, query receipt (expect Pending), wait for block production, query again (expect Included with block ID and timestamp)
- [ ] Manual: submit an invalid transaction, query receipt (expect Rejected with reason string)
- [ ] Restart sequencer, query receipt for a previously rejected hash (expect Rejected, verifies RocksDB persistence)

### DIFF
--- a/common/Cargo.toml
+++ b/common/Cargo.toml
@@ -23,3 +23,6 @@ hex.workspace = true
 borsh.workspace = true
 logos-blockchain-common-http-client.workspace = true
 jsonrpsee = { workspace = true }
+
+[dev-dependencies]
+serde_json.workspace = true
--- a/common/Cargo.toml
+++ b/common/Cargo.toml
@@ -22,3 +22,4 @@ log.workspace = true
 hex.workspace = true
 borsh.workspace = true
 logos-blockchain-common-http-client.workspace = true
+jsonrpsee = { workspace = true }
--- a/common/src/lib.rs
+++ b/common/src/lib.rs
@@ -7,6 +7,7 @@ pub mod block;
 mod borsh_base64;
 pub mod config;
 pub mod transaction;
+pub mod receipt;
 
 // Module for tests utility functions
 // TODO: Compile only for tests
--- a/common/src/receipt.rs
+++ b/common/src/receipt.rs
@@ -1,0 +1,53 @@
+use borsh::{BorshDeserialize, BorshSerialize};
+use serde::{Deserialize, Serialize};
+
+use crate::HashType;
+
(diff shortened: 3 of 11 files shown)


moudyellaz:
Closing this, @paschal533. Transaction receipts are phase 2 of #379, which is still open and still the plan of record. The branch predates the `lee/`+`lez/` restructure so it no longer applies, but the need is tracked there.

PR #443, opened 2026-04-11
feat(sequencer): add simulate_transaction RPC endpoint

paschal533:
## Summary

Adds a `simulate_transaction` RPC method that executes a transaction against a snapshot of the current state and returns a structured result without ever committing the changes.

Callers can use this to:

- Check whether a transaction will succeed before broadcasting it
- Inspect which accounts will be modified, what nullifiers will be created, and what commitments will be added
- Surface human-readable error messages without wasting a real submission

### Design

The implementation follows a clone-and-discard pattern:

1. Stateless signature/structure checks run without acquiring any lock.
2. The sequencer lock is acquired, `V03State` is cloned, the lock is released immediately.
3. `validate_on_state` runs on the clone, never on the real state, so block production continues concurrently without contention.
4. The resulting diff (or error) is returned without being applied.

The simulation result includes:

| Field | Contents |
|---|---|
| `success` | whether the transaction would succeed |
| `error` | failure reason if `success == false` |
| `accounts_modified` | `Vec<(AccountId, Account)>` post-state for each touched account |
| `nullifiers_created` | nullifiers that would be spent |
| `commitments_created` | commitments that would be added |

### Changes

- **`common/src/simulation.rs`** (new), `SimulationResult` type.
- **`common/src/lib.rs`**, expose `pub mod simulation`.
- **`nssa/src/validated_state_diff.rs`**, expose `new_nullifiers()` and `new_commitments()` accessors on `ValidatedStateDiff` (previously only `public_diff()` was public).
- **`sequencer/service/src/service.rs`**, `simulate_transaction` implementation with lock-release-before-execute pattern.
- **`sequencer/service/rpc/src/lib.rs`** and **`sequencer/service/protocol/src/lib.rs`**, RPC trait method and protocol type exports.

### Stacking note

This PR builds on top of #442 (receipt system) and #441 (crash fix). The
meaningful change in this PR starts at the `feat(common): add SimulationResult`
commit.

## Test plan

- [ ] `cargo test -p sequencer_service simulate`, two tests: `simulate_valid_transaction_returns_success` and `simulate_does_not_modify_state`
- [ ] Manual: simulate a token transfer; confirm `accounts_modified` shows the expected balance delta and the real account state is unchanged after the call
- [ ] Manual: simulate a transaction that should fail (e.g., insufficient balance); confirm `success == false` and `error` contains a readable message
- [ ] Load test: run N concurrent `simulate_transaction` calls while block production is active; confirm no lock contention and no state corruption

### DIFF
--- a/common/Cargo.toml
+++ b/common/Cargo.toml
@@ -23,3 +23,6 @@ hex.workspace = true
 borsh.workspace = true
 logos-blockchain-common-http-client.workspace = true
 jsonrpsee = { workspace = true }
+
+[dev-dependencies]
+serde_json.workspace = true
--- a/common/Cargo.toml
+++ b/common/Cargo.toml
@@ -22,3 +22,4 @@ log.workspace = true
 hex.workspace = true
 borsh.workspace = true
 logos-blockchain-common-http-client.workspace = true
+jsonrpsee = { workspace = true }
--- a/common/src/lib.rs
+++ b/common/src/lib.rs
@@ -7,6 +7,7 @@ pub mod block;
 mod borsh_base64;
 pub mod config;
 pub mod transaction;
+pub mod receipt;
 
 // Module for tests utility functions
 // TODO: Compile only for tests
--- a/common/src/lib.rs
+++ b/common/src/lib.rs
@@ -8,6 +8,7 @@ mod borsh_base64;
 pub mod config;
 pub mod transaction;
 pub mod receipt;
+pub mod simulation;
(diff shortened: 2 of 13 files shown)


moudyellaz:
Closing this, @paschal533. The simulation endpoint is phase 3 of #379, still open and still the plan of record. The branch predates the `lee/`+`lez/` restructure so it no longer applies, but the need is tracked there.

PR #446, opened 2026-04-14
feat: private PDA support in the privacy circuit

moudyellaz:
## 🎯 Purpose

Enable private PDAs via a new visibility mask (3) with a separate `AccountId` derivation
`AccountId::for_private_pda(program_id, seed, npk)`. Programs stay privacy-agnostic: they
emit `Claim::Pda(seed)` and `ChainedCall.pda_seeds` regardless of whether the underlying
account is a public PDA or a private one; the circuit picks the derivation from the
pre_state's mask.

## ⚙️ Approach

- Add `AccountId::for_private_pda(program_id, seed, npk)` derivation with a dedicated
  `PRIVATE_PDA_PREFIX`. Rename the existing public-PDA `From` impl to
  `AccountId::for_public_pda(program_id, seed)` for symmetry. Public PDA formula unchanged.
- Unify the claim/authorization API. `Claim` stays as `Authorized | Pda(PdaSeed)`;
  `ChainedCall.pda_seeds` stays as `Vec<PdaSeed>`. No separate private variants.
- In the privacy circuit:
  - Build `private_pda_npk_by_position` from `visibility_mask` + `private_account_keys`
    as part of `ExecutionState::derive_from_outputs`.
  - Mask-3 `Claim::Pda(seed)` asserts
    `AccountId::for_private_pda(program, seed, npk) == pre.account_id`.
  - Caller-seed authorization matches against both public and private derivations per
    pre_state, extracted into `resolve_authorization_and_record_bindings`.
  - The equality check is the cryptographic binding, no proven-npk attestation channel.
  - Track `private_pda_bound_positions` and assert every private-PDA position is bound
    after the main loop.
- Reject multiple family members for the same `(program_id, seed)` in one tx via a tx-wide
  `pda_family_binding: HashMap<(ProgramId, PdaSeed), AccountId>`. Prevents a single
  `pda_seeds: [S]` entry from authorizing multiple private PDAs under different npks.
- Public-execution path (`validated_state_diff.rs`) only sees public accounts; uses the
  public form.
- Make `NullifierPublicKey` `Copy` (32 bytes of public data) so `Claim` stays `Copy`.

## 🧪 How to Test

`RISC0_DEV_MODE=1 cargo test --release -p nssa_core -p nssa`

nssa_core unit tests:
- `AccountId::for_private_pda` derivation: pinned value, uniqueness across programs / seeds
  / npks, domain separation from public PDAs and standard private accounts.
- `compute_public_authorized_pdas` (with seeds, no-caller cases).

nssa integration tests (exercising the privacy circuit):
- `private_pda_claim_succeeds`: happy path, `Claim::Pda(seed)` on a private-PDA pre_state
  produces the expected `AccountId`, commitment, nullifier, ciphertext.
- `private_pda_without_binding_fails`: private PDA with no claim or caller-seed
  authorization is rejected by the `private_pda_bound_positions` assertion.
- `private_pda_npk_mismatch_fails`: an npk that doesn't match the pre_state's
  `account_id` under `AccountId::for_private_pda(program, seed, npk)` is rejected.
- `caller_pda_seeds_authorize_private_pda_for_callee`: caller claims a private PDA and
  delegates the same seed via `ChainedCall.pda_seeds`; callee (a noop that also asserts
  `is_authorized`) operates on it without a fresh claim.
- `caller_pda_seeds_with_wrong_seed_rejects_private_pda_for_callee`: same setup with a
  different delegated seed; callee authorization check rejects.
- `private_pda_top_level_reuse_rejected_by_binding_check`: pins the current limitation,
  an already-claimed private PDA can't be re-used in a new tx without a fresh claim.
  Follow-up PR in this series will add wallet-side `(seed, owner)` input.
- `two_private_pda_claims_under_same_seed_are_rejected`: the tx-wide family-binding check
  rejects a program claiming two private PDAs (different npks) under the same seed.

Test guests under `test_program_methods/guest/src/bin/`:
- `pda_claimer.rs`: unified public/private PDA claimer via `Claim::Pda(seed)`.
- `private_pda_delegator.rs`: claims a private PDA and delegates the seed to a callee.
- `two_pda_claimer.rs`: claims two pre_states under the same seed (family-binding test).
- `auth_asserting_noop.rs`: noop variant that asserts every pre_state is authorized, used
  as callee in the delegation tests.

## 🔗 Dependencies

None.

## 🔜 Future Work

- Wallet-side support for `(seed, owner)` input so already-claimed private PDAs can be
  re-used across transactions without a fresh claim.
- `GroupKeyHolder` in the wallet for GMS key management.
- Toy example demonstrating the full private PDA pattern end-to-end.
- Refactor the privacy-preserving circuit into a dedicated crate with per-concern modules
  (tracked in #454).
- Strong-type `PrivacyPreservingCircuitInput` with a per-account enum to remove parallel
  vector length assertions (tracked in #455).

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


program_methods/guest/src/bin/privacy_preserving_circuit.rs
```diff
@@ -187,6 +190,7 @@ impl ExecutionState {
         visibility_mask: &[u8],
         program_id: ProgramId,
         authorized_pdas: &HashSet<AccountId>,
+        private_pda_info: &[(ProgramId, PdaSeed, NullifierPublicKey)],
```
schouhy: why is this passed by the user on circuit input? I think program id used for private pda checks should be the same as `program_id` parameter in this function.

program_methods/guest/src/bin/privacy_preserving_circuit.rs
```diff
@@ -494,17 +600,33 @@ fn main() {
         private_account_nsks,
         private_account_membership_proofs,
         program_id,
+        private_pda_info,
```
schouhy: uhm why is private pda info passed as circuit input? I think the program output through the chain call should be used for pda checks. For example, seeds used for pda need to be those given by programs in their output, not manually passed as input no? For example like we are doing with the public pdas already in this circuit.

schouhy (commented):
hey thanks for the PR!
I'm a bit confused about the approach. I see that we are passing private_pda_info as input to the circuit and I think the values used should be those coming from the chained calls.

nssa/core/src/program.rs
```diff
@@ -709,4 +747,136 @@ mod tests {
...
+    /// Same inputs always produce the same `AccountId`.
+    #[test]
+    fn private_pda_account_id_is_deterministic() {
+        let program_id: ProgramId = [1; 8];
+        let seed = PdaSeed::new([2; 32]);
+        let npk = NullifierPublicKey([3; 32]);
+        let id1 = private_pda_account_id(&program_id, &seed, &npk);
+        let id2 = private_pda_account_id(&program_id, &seed, &npk);
+        assert_eq!(id1, id2);
+    }
```
Pravdyvy: This test does not tell, that `pda_account_id` is deterministic. 
This test is useful anyhow, but maybe with different name.
If intent is to pin implementation, it is better to add another test which calculates some pda_account_id for triple `program_id, seed, npk`, puts it into `const`, then checks against this triple.

nssa/core/src/circuit_io.rs
```diff
@@ -26,6 +26,12 @@ pub struct PrivacyPreservingCircuitInput {
     pub private_account_membership_proofs: Vec<Option<MembershipProof>>,
     /// Program ID.
     pub program_id: ProgramId,
+    /// Private PDA info for mask-3 accounts.
+    /// Unlike the other `private_account_*` fields which are parallel arrays indexed by private
+    /// account position, this is a separate lookup table. The circuit matches entries by
+    /// (`program_id`, `seed`) against the chained calls' `pda_seeds` to resolve private PDA
+    /// authorization.
+    pub private_pda_info: Vec<(ProgramId, PdaSeed, NullifierPublicKey)>,
```
Pravdyvy: I don't understand fully intent behind this field in context of chain calls and have the following questions:
- Is PDA id for `seed` and `npk` must be the same for every chained call?(@schouhy comments touch this subject).
- If so, why there is a vector with triplets? Do we want calculate ahead and want every program in chained call have its own PDA? 
- Is this PDA's must be derived from different seeds and npks at every step?

I think it is better to add new struct for this field(something like `PrivatePDAInfo`) with various constructors.

Pravdyvy (commented):
Thank you for PR. Asked some questions for this approach.

moudyellaz:
> hey thanks for the PR! I'm a bit confused about the approach. I see that we are passing private_pda_info as input to the circuit and I think the values used should be those coming from the chained calls.

@schouhy good catch. Removed `private_pda_info` from the circuit input and moved npk into two proven places: `ChainedCall.private_pda_seeds: Vec<(PdaSeed, NullifierPublicKey)>` for caller-side authorization and `Claim::PrivatePda { seed, npk }` for callee-side claims. Mask-3 AccountId binding in `compute_circuit_output` now checks against a `HashMap` built in `derive_from_outputs` from proven data only. Public PDA API unchanged, so AMM, ATA, Token are untouched. Tests pass.

nssa/core/src/program.rs
moudyellaz: removed this test and added `private_pda_account_id_matches_pinned_value` which pins the output for a fixed `(program_id, seed, npk)` against a hardcoded 32-byte AccountId, so any change to the prefix, byte ordering, or hash function breaks the test.

program_methods/guest/src/bin/privacy_preserving_circuit.rs
moudyellaz: fixed

program_methods/guest/src/bin/privacy_preserving_circuit.rs
moudyellaz: fixed

nssa/core/src/circuit_io.rs
moudyellaz: I refactored: `private_pda_info` is removed from the circuit input. The npk now travels inside Risc0-proven data on `ChainedCall.private_pda_seeds: Vec<(PdaSeed, NullifierPublicKey)>` and `Claim::PrivatePda { seed, npk }`.

On your specific questions:
- Each chained call has its own `private_pda_seeds`. Different calls can carry different `(seed, npk)` pairs; uniformity is not required.
- The vector is per-call (one call can authorize multiple PDAs), not a global lookup.
- No constraint on varying seeds or npks across steps; programs are free to reuse or vary.

Each entry is now just `(PdaSeed, NullifierPublicKey)` (no more `ProgramId`, the caller's is implicit). Happy to wrap that pair in a named struct like `PrivatePdaAuthorization { seed, npk }` if you think it's still worth it.

Pravdyvy: Fine as it is

Pravdyvy (approved):
lgtm

nssa/src/state.rs
```diff
@@ -2326,17 +2333,230 @@ pub mod tests {
             true,
             AccountId::new([0; 32]),
         );
-        let public_account_2 =
+        let mask3_account =
```
schouhy: I'd rename all `mask_3` references in test names and variables to a private pda wording. If in the future we change the mask number for the private pda, this will naming will silently get outdated.

nssa/src/state.rs
```diff
@@ -2314,9 +2314,16 @@ pub mod tests {
         assert!(matches!(result, Err(NssaError::CircuitProvingError(_))));
     }
 
+    /// A mask-3 account that no program claims via `Claim::Pda` and no caller authorizes via
+    /// `ChainedCall.pda_seeds` has no binding between its wallet-supplied npk and its
```
schouhy:
```suggestion
    /// `ChainedCall.pda_seeds` has no binding between the supplied npk and its
```

nssa/src/state.rs
```diff
@@ -2326,17 +2333,230 @@ pub mod tests {
...
+        let (output, _proof) = result.expect("mask-3 private PDA claim should succeed");
+        assert_eq!(output.new_nullifiers.len(), 1);
+        assert_eq!(output.new_commitments.len(), 1);
+        assert_eq!(output.ciphertexts.len(), 1);
+        assert!(output.public_pre_states.is_empty());
+        assert!(output.public_post_states.is_empty());
+    }
+
+    /// The wallet supplies an npk that does not match the `pre_state`'s `account_id` under
+    /// `private_pda_account_id(program, claim_seed, npk)`. The claim equality check rejects.
```
schouhy: I'd keep this crate independent of wallet referenes
```suggestion
    /// An npk is supplied that does not match the `pre_state`'s `account_id` under
    /// `private_pda_account_id(program, claim_seed, npk)`. The claim equality check rejects.
```

nssa/src/state.rs
```diff
@@ -2326,17 +2333,230 @@ pub mod tests {
...
+    #[test]
+    fn mask_3_wallet_npk_mismatch_panics() {
+        let program = Program::pda_claimer();
+        let attested_keys = test_private_account_keys_1();
+        let wallet_keys = test_private_account_keys_2();
```
schouhy: same here
```suggestion
        let keys = test_private_account_keys_2();
```

nssa/src/state.rs
```diff
@@ -2326,17 +2333,230 @@ pub mod tests {
...
+        assert_eq!(output.new_commitments.len(), 1);
+        assert_eq!(output.ciphertexts.len(), 1);
+        assert!(output.public_pre_states.is_empty());
+        assert!(output.public_post_states.is_empty());
+    }
+
+    /// The wallet supplies an npk that does not match the `pre_state`'s `account_id` under
+    /// `private_pda_account_id(program, claim_seed, npk)`. The claim equality check rejects.
+    #[test]
+    fn mask_3_wallet_npk_mismatch_panics() {
```
schouhy:
```suggestion
    fn mask_3_wallet_npk_mismatch_fails() {
```

nssa/src/state.rs
```diff
@@ -2326,17 +2333,230 @@ pub mod tests {
...
+            &program.into(),
+        );
+
+        assert!(matches!(result, Err(NssaError::CircuitProvingError(_))));
+    }
+
+    /// Happy path for the caller-seeds authorization of a mask-3 PDA. The delegator claims a
+    /// private PDA via `Claim::Pda(seed)`, then chains to a callee (`noop`) delegating the same
+    /// seed via `ChainedCall.pda_seeds`. In the callee's step, the `pre_state` is already in
+    /// `self.pre_states` (Occupied branch) and its authorization is established via the private
```
schouhy: let's not add implementation details to docs
```suggestion
    /// `self.pre_states` and its authorization is established via the private
```

nssa/src/state.rs
```diff
@@ -2326,17 +2333,230 @@ pub mod tests {
...
+    /// derivation `private_pda_account_id(delegator, seed, npk) == pre.account_id`. This exercises
+    /// the `authorized_via_private` path in `validate_and_sync_states`.
+    #[test]
+    fn caller_pda_seeds_authorize_mask_3_private_pda_for_callee() {
+        let delegator = Program::private_pda_delegator();
+        let noop = Program::noop();
```
schouhy: Shouldn't we use a program that checks authorization in this test as callee? If not, I'm not sure if we are fully testing what the test docs describe (namely, that the callee got the input account with `is_authorized=true`). Maybe add a variant of the noop that checksk the input account is authorized.

nssa/src/state.rs
```diff
@@ -2326,17 +2333,230 @@ pub mod tests {
...
+            vec![None],
+            &program_with_deps,
+        );
+
+        assert!(matches!(result, Err(NssaError::CircuitProvingError(_))));
+    }
+
+    /// Pins the current limitation: a mask-3 PDA that was claimed in a previous transaction
+    /// cannot be re-used in a new transaction as-is, because this PR only binds wallet-supplied
+    /// npks via a fresh `Claim::Pda` or a caller's `ChainedCall.pda_seeds` — neither is present
```
schouhy:
```suggestion
    /// cannot be re-used in a new transaction as-is, because this PR only binds supplied
    /// npks via a fresh `Claim::Pda` or a caller's `ChainedCall.pda_seeds` — neither is present
```

nssa/src/state.rs
```diff
@@ -2326,17 +2333,230 @@ pub mod tests {
...
+    /// can re-verify `private_pda_account_id(owner, seed, npk) == pre.account_id` without a
+    /// claim.
+    /// Exploit-scenario pin. A single `(program_id, seed)` pair can derive a family of
+    /// `AccountId`s (one public PDA and one private PDA per distinct npk). Without the tx-wide
+    /// family-binding check, a program could claim `PDA_alice` (mask-3, `alice_npk`) and
+    /// `PDA_bob` (mask-3, `bob_npk`) under the same seed in one transaction, and — once reuse
+    /// is supported — a later chained call could delegate both to a callee via
+    /// `pda_seeds: [S]` and mix balances across them. The binding check rejects the setup
+    /// here: after the first claim records `(program, seed) → PDA_alice`, the second claim
+    /// tries to record `(program, seed) → PDA_bob` and panics.
```
schouhy: let's dont have this TODO as part of the doc
```suggestion
    // TODO: a follow-up PR in the Private PDAs series needs to let the wallet supply a
    // `(seed, original_owner_program_id)` side input per mask-3 `pre_state` so the circuit
    // can re-verify `private_pda_account_id(owner, seed, npk) == pre.account_id` without a
    // claim.
    // Exploit-scenario pin. A single `(program_id, seed)` pair can derive a family of
    // `AccountId`s (one public PDA and one private PDA per distinct npk). Without the tx-wide
    // family-binding check, a program could claim `PDA_alice` (mask-3, `alice_npk`) and
    // `PDA_bob` (mask-3, `bob_npk`) under the same seed in one transaction, and — once reuse
    // is supported — a later chained call could delegate both to a callee via
    // `pda_seeds: [S]` and mix balances across them. The binding check rejects the setup
    // here: after the first claim records `(program, seed) → PDA_alice`, the second claim
    // tries to record `(program, seed) → PDA_bob` and panics.
```

nssa/src/state.rs
```diff
@@ -2326,17 +2333,230 @@ pub mod tests {
...
+    /// `(seed, original_owner_program_id)` side input per mask-3 `pre_state` so the circuit
+    /// can re-verify `private_pda_account_id(owner, seed, npk) == pre.account_id` without a
+    /// claim.
+    /// Exploit-scenario pin. A single `(program_id, seed)` pair can derive a family of
+    /// `AccountId`s (one public PDA and one private PDA per distinct npk). Without the tx-wide
+    /// family-binding check, a program could claim `PDA_alice` (mask-3, `alice_npk`) and
+    /// `PDA_bob` (mask-3, `bob_npk`) under the same seed in one transaction, and — once reuse
+    /// is supported — a later chained call could delegate both to a callee via
+    /// `pda_seeds: [S]` and mix balances across them. The binding check rejects the setup
+    /// here: after the first claim records `(program, seed) → PDA_alice`, the second claim
```
schouhy: good catch. I think we can protect against this attack by checking in the circuit that every seed is used exactly once for a single account. This way we don't have to change transactions

nssa/src/state.rs
```diff
@@ -2326,17 +2333,230 @@ pub mod tests {
...
+
+        assert!(matches!(result, Err(NssaError::CircuitProvingError(_))));
+    }
+
+    /// Pins the current limitation: a mask-3 PDA that was claimed in a previous transaction
+    /// cannot be re-used in a new transaction as-is, because this PR only binds wallet-supplied
+    /// npks via a fresh `Claim::Pda` or a caller's `ChainedCall.pda_seeds` — neither is present
+    /// when a program operates on an already-owned private PDA at top level. The circuit rejects.
+    ///
+    /// TODO: a follow-up PR in the Private PDAs series needs to let the wallet supply a
```
schouhy: Isn't test `two_mask_3_claims_under_same_seed_are_rejected` already checking that there's a mechanism protecting against this exploit scenario?

nssa/src/state.rs
```diff
@@ -2326,17 +2333,230 @@ pub mod tests {
...
             vec![],
+            vec![None, None],
+            &program.into(),
+        );
+
+        assert!(matches!(result, Err(NssaError::CircuitProvingError(_))));
+    }
+
+    #[test]
+    fn mask_3_reuse_across_txs_currently_unsupported() {
```
schouhy: I don't understand this. I think this should fail (like it does) because the `execute_and_prove` function processes the noop program as the first in the call stack. Since it's the first one all the authorizations need to come from users (no authorization from caller program since there's none). So the input pre_state which is marked with `is_authorized=true` will make things fail in this case.

nssa/src/state.rs
schouhy, line 1: Not sure if this is the best place for these tests since they don't test anything about the `NssaState`

program_methods/guest/src/bin/privacy_preserving_circuit.rs
```diff
@@ -23,12 +23,28 @@ struct ExecutionState {
     post_states: HashMap<AccountId, Account>,
     block_validity_window: BlockValidityWindow,
     timestamp_validity_window: TimestampValidityWindow,
+    /// Positions (in `pre_states`) of mask-3 accounts whose wallet-supplied npk has been bound
+    /// to their `AccountId` via a proven `private_pda_account_id(program_id, seed, npk)` check.
+    /// The binding happens when the circuit validates a `Claim::Pda(seed)` on that `pre_state`,
+    /// or when it authorizes that `pre_state` via a caller's `ChainedCall.pda_seeds`. After the
+    /// main loop, every mask-3 position must appear in this set; otherwise the npk is unbound
+    /// and the circuit rejects.
+    mask3_bound_positions: HashSet<usize>,
```
schouhy: Let's use more descriptive names. `mask3` is not very meaningful.
```suggestion
    private_pda_bound_positions: HashSet<usize>,
```

program_methods/guest/src/bin/privacy_preserving_circuit.rs
```diff
@@ -23,12 +23,28 @@ struct ExecutionState {
     post_states: HashMap<AccountId, Account>,
     block_validity_window: BlockValidityWindow,
     timestamp_validity_window: TimestampValidityWindow,
+    /// Positions (in `pre_states`) of mask-3 accounts whose wallet-supplied npk has been bound
```
schouhy:
```suggestion
    /// Positions (in `pre_states`) of mask-3 accounts whose supplied npk has been bound
```

program_methods/guest/src/bin/privacy_preserving_circuit.rs
```diff
@@ -23,12 +23,28 @@ struct ExecutionState {
...
+    /// to their `AccountId` via a proven `private_pda_account_id(program_id, seed, npk)` check.
+    /// The binding happens when the circuit validates a `Claim::Pda(seed)` on that `pre_state`,
+    /// or when it authorizes that `pre_state` via a caller's `ChainedCall.pda_seeds`. After the
+    /// main loop, every mask-3 position must appear in this set; otherwise the npk is unbound
+    /// and the circuit rejects.
+    mask3_bound_positions: HashSet<usize>,
+    /// Across the whole transaction, each `(program_id, seed)` pair may resolve to at most one
+    /// `AccountId`. A seed under a program can derive a family of accounts (one public PDA and
+    /// one private PDA per distinct npk), and unifying `Claim::PrivatePda` and `ChainedCall`'s
+    /// private seeds into plain `Claim::Pda(seed)` / `pda_seeds` would otherwise let a single
```
schouhy: what does `Claim::Pda(seed)` / `pda_seeds` mean?

program_methods/guest/src/bin/privacy_preserving_circuit.rs
```diff
@@ -439,6 +568,81 @@ fn compute_circuit_output(
...
+                // `Claim::Pda(seed)` match or via a caller `pda_seeds` match, both of which
+                // assert `private_pda_account_id(owner, seed, npk) == account_id`. The post-loop
+                // assertion in `derive_from_outputs` (see the `mask3_bound_positions` check)
+                // guarantees that every mask-3 position has been through at least one such
+                // binding, so this branch can safely use the wallet npk without re-verifying.
+                let Some((npk, shared_secret)) = private_keys_iter.next() else {
+                    panic!("Missing private account key");
+                };
+
+                let (new_nullifier, new_nonce) = if pre_state.is_authorized {
```
schouhy, line 635: This branching logic is only correct because we are not supporting non-authorized private accounts with non-default values. Likely to be changed in the future. I'm sure there's use cases for this. For example the multisig program if ran completely private it would need a private non-default and non-authorized input account.

schouhy, line 635: looks like a lot of repeated code from mask 1 and mask 2 branches. Maybe we could refactor it to avoid code repetition.

schouhy (commented):
Thanks for considering my previous comments. I didn't finish reviewing yet. I'm still going through all the changes the the privacy preserving circuit. I'm leaving some comments to discuss while I finish.

nssa/src/validated_state_diff.rs
```diff
@@ -223,7 +223,8 @@ impl ValidatedStateDiff {
                     }
                     Claim::Pda(seed) => {
                         // The program can only claim accounts that correspond to the PDAs it is
-                        // authorized to claim.
+                        // authorized to claim. The public-execution path only sees mask-0
```
schouhy:
```suggestion
                        // authorized to claim. The public-execution path only sees public
```

program_methods/guest/src/bin/privacy_preserving_circuit.rs
```diff
@@ -304,6 +405,34 @@ impl ExecutionState {
...
 }
 
+/// Record or re-verify the `(program_id, seed) → account_id` family binding for the
+/// transaction. Any claim or caller-seed authorization that resolves a `pre_state` under
+/// `(program_id, seed)` must agree with every prior resolution of the same pair; otherwise a
+/// single `pda_seeds: [seed]` entry could authorize multiple private-PDA family members at
+/// once (different npks under the same seed) and let a callee mix balances across them. Free
+/// function so callers can pass `&mut self.pda_family_binding` without holding a borrow on
+/// the surrounding struct's other fields.
+fn assert_family_binding(
```
schouhy, line 418: I think there is an edge case where a program chains a call to itself (like in flashloan e.g.) more than once and needs to authorize the same pda seed in multiple calls. This will record the binding the first time and error on the next one since the family bindings used are the same across all calls. If so I'll document this behavior to let devs know of this temporary limitation.

nssa/core/src/program.rs
```diff
@@ -477,19 +488,52 @@ pub enum ExecutionValidationError {
     },
 }
 
+/// Derives an [`AccountId`] for a private PDA from the program ID, seed, and nullifier public key.
+///
+/// Unlike public PDAs (`AccountId::from((&ProgramId, &PdaSeed))`), this includes the `npk` in the
+/// derivation, making the address unique per group of controllers sharing viewing keys.
+#[must_use]
+pub fn private_pda_account_id(
```
schouhy: I think this should be a constructor  `AccountId::for_private_pda`. Consider also removing the existing `impl From<(ProgramId, Seed)> for AccountId` for public pdas in favor of a `AccountId::for_public_pda` to have a unified way of constructing pdas

program_methods/guest/src/bin/privacy_preserving_circuit.rs
```diff
@@ -218,19 +253,54 @@ impl ExecutionState {
...
+                    let matched_caller_seed: Option<(PdaSeed, bool)> =
+                        caller_program_id.and_then(|caller| {
+                            caller_pda_seeds.iter().find_map(|seed| {
+                                if AccountId::from((&caller, seed)) == pre_account_id {
+                                    return Some((*seed, false));
+                                }
+                                if let Some(npk) = mask3_npk_by_position.get(&pre_state_position)
+                                    && private_pda_account_id(&caller, seed, npk) == pre_account_id
+                                {
+                                    return Some((*seed, true));
```
schouhy: we can return `Some((*seed, true, caller))` to avoid having to unwrap the `caller_program_id` again in line 290

program_methods/guest/src/bin/privacy_preserving_circuit.rs
```diff
@@ -162,6 +178,19 @@ impl ExecutionState {
...
+        // the wallet-supplied npk and the account_id, and must be rejected.
+        for (pos, &mask) in visibility_mask.iter().enumerate() {
+            if mask == 3 {
+                assert!(
+                    execution_state.mask3_bound_positions.contains(&pos),
+                    "mask-3 pre_state at position {pos} has no proven (seed, npk) binding via Claim::Pda or caller pda_seeds"
+                );
+            }
+        }
+
```
schouhy, line 224: Is it ok to collect all usages of seeds (either for claiming or authorizing) in the same map `mask3_bound_positions`? Since the usages are independent of each other it feels like we shouldn't mix them. Maybe I'm missing something

program_methods/guest/src/bin/privacy_preserving_circuit.rs
```diff
@@ -218,19 +253,54 @@ impl ExecutionState {
...
+                        let caller = caller_program_id
+                            .expect("matched caller seed implies caller_program_id is set");
+                        assert_family_binding(
+                            &mut self.pda_family_binding,
+                            caller,
+                            seed,
+                            pre_account_id,
+                        );
+                        if is_private_form {
+                            self.mask3_bound_positions.insert(pre_state_position);
```
schouhy: It would be nice to have a `assert!(self.mask3_bound_positions.insert(pre_state_position));` instead to check that the item added was not present before. But we can't because we are also using the `mask3_bound_positions` for claiming mechanism.

schouhy (approved):
thanks for the PR. I'm leaving more comments. It looks good to me in general. I didn't see anything off and the tests added seem to cover the functionality.  But the privacy preserving circuit is really hard to chew at this point. I think we need to prioritize refactoring it into a dedicated crate so we can split the code into different files and have something easier to reason about.

Also, this PR is adding more complexity to the circuit. Did you check if performance has regressed in terms of proving time?

Cargo.toml
```diff
@@ -152,6 +152,14 @@ opt-level = 'z'
 lto = true
 codegen-units = 1
 
+# Keep backtraces but drop full DWARF type info to avoid LLD OOM/SIGBUS when
+# linking large integration-test binaries on resource-constrained CI runners.
+[profile.dev]
+debug = "line-tables-only"
+
+[profile.test]
+debug = "line-tables-only"
```
Arjentix, line 161: Hmm, was it really a problem? I've never saw such issues in CI

nssa/core/src/circuit_io.rs
```diff
@@ -17,6 +17,7 @@ pub struct PrivacyPreservingCircuitInput {
     /// - `0` - public account
     /// - `1` - private account with authentication
     /// - `2` - private account without authentication
+    /// - `3` - private PDA account
```
Arjentix, line 20: Not for this PR, but just an idea I have in mind -- need to refactor this input structure to make it more strong-typed. Something like this:

```rust
pub struct PrivacyPreservingCircuitInput {
    pub program_outputs: Vec<ProgramOutput>,
    pub accounts: Vec<PrivacyPreservingCircuitInputAccount>,
    pub program_id: ProgramId,
}

pub enum PrivacyPreservingCircuitInputAccount {
    Public,
    PrivateAuthorized {
        npk: NullifierPublicKey, 
        ssk: SharedSecretKey,
        nsk: NullifierSecretKey,
        membership_proof: Option<MembershipProof>,
    },
    PrivateUnauthorized {
        npk: NullifierPublicKey, 
        ssk: SharedSecretKey,
    },
    PrivatePda { ... } // haven't seen other parts of the PR, not sure what there should be
}
```

Arjentix: This will also remove a lot of vec's length checks

nssa/core/src/program.rs
```diff
@@ -114,7 +120,12 @@ pub enum Claim {
     /// This will give no error if program had authorization in pre state and may be useful
     /// if program decides to give up authorization for a chained call.
     Authorized,
-    /// The program requests ownership of the account through a PDA.
+    /// The program requests ownership of the account through a PDA. The `pre_state`'s
+    /// visibility mask selects the derivation formula: mask 0 uses
+    /// `AccountId::from((&program_id, &seed))`, mask 3 uses
+    /// `private_pda_account_id(&program_id, &seed, &npk)` with the wallet-supplied npk for that
+    /// `pre_state`. Programs stay privacy-agnostic: they emit a seed, the circuit resolves the
+    /// rest from the mask.
```
Arjentix: Same situation here

nssa/core/src/program.rs
Arjentix: Agreed

program_methods/guest/src/bin/privacy_preserving_circuit.rs
moudyellaz: Good to double-check, but the check handles this case correctly. `assert_family_binding` at `privacy_preserving_circuit.rs:421-433` only panics when the stored `AccountId` differs from the new one. For a self-chained program with the same seed, both calls resolve the callee `pre_state` to the same `AccountId` since the derivation is deterministic in `(program_id, seed, npk)` and the npk is pinned by `pre_state` position via `private_pda_npk_by_position`. The `Occupied` arm's `assert_eq!` passes on the re-resolution and no error fires.

`flash_swap_initiator.rs:127-160` never even repeats a `(program, seed)` pair (call 1 uses `[0u8; 32]` on the token program, calls 2 and 3 have empty `pda_seeds`). A private-PDA variant that did repeat one would still resolve identically and pass. The check only blocks the actual exploit, two distinct family members under one seed in one tx.

nssa/src/state.rs
moudyellaz: done

nssa/src/state.rs
moudyellaz: addressed

nssa/src/state.rs
moudyellaz: addressed

program_methods/guest/src/bin/privacy_preserving_circuit.rs
moudyellaz: addressed

program_methods/guest/src/bin/privacy_preserving_circuit.rs
moudyellaz: addressed

moudyellaz: addressed

program_methods/guest/src/bin/privacy_preserving_circuit.rs
moudyellaz: The set answers one question: "is there any proven `private_pda_account_id(program, seed, npk) == account_id` derivation covering this `pre_state`?" Both insertion sites (claim at `privacy_preserving_circuit.rs:377`, caller-seed match at `:305`) establish that same fact via the same equality. Splitting into claim-bound vs auth-bound doesn't strengthen the post-loop check at `:193-198`, the soundness lives in the equality at the insertion, not in which bucket recorded it.

I expanded the field docstring in the doc-cleanup commit to enumerate the two proof paths and note that binding is an idempotent property rather than an event (see `private_pda_bound_positions` in the current HEAD).

program_methods/guest/src/bin/privacy_preserving_circuit.rs
moudyellaz: A position can legitimately be bound via both mechanisms in one tx: a program claims an uninitialized private PDA (insertion at `privacy_preserving_circuit.rs:377`), then delegates the same `pre_state` to a callee via `ChainedCall.pda_seeds` (insertion at `:305`). An insertion-uniqueness assert would reject that valid flow. Bound-ness is an idempotent property rather than an event, which is why the shape is `HashSet` + `contains` instead of `assert!(insert)`.

nssa/core/src/program.rs
moudyellaz: Done in `0183eac5`. The `impl From<(&ProgramId, &PdaSeed)> for AccountId` and the free `private_pda_account_id` function are both gone, replaced by `AccountId::for_public_pda(&program_id, &seed)` and `AccountId::for_private_pda(&program_id, &seed, &npk)` as inherent methods on `AccountId`. All call sites migrated across `nssa_core`, `nssa`, `programs/amm/core`, `programs/associated_token_account/core`, the guest circuit, the example tail-call binary, and the ATA tutorial doc.

moudyellaz:
> But the privacy preserving circuit is really hard to chew at this point. I think we need to prioritize refactoring it into a dedicated crate so we can split the code into different files and have something easier to reason about.

Agreed, will open a follow-up issue to split it into a dedicated crate with modules along those lines.

Edit: tracked in #454.

nssa/core/src/program.rs
```diff
@@ -65,6 +96,12 @@ pub struct ChainedCall {
     pub pre_states: Vec<AccountWithMetadata>,
     /// The instruction data to pass.
     pub instruction_data: InstructionData,
+    /// PDA seeds authorized for the callee. For each callee `pre_state`, the outer circuit
+    /// checks whether its `AccountId` matches a public PDA derivation
+    /// `AccountId::for_public_pda(&caller, seed)` (mask 0) or a private PDA derivation
+    /// `AccountId::for_private_pda(&caller, seed, npk)` (mask 3, where `npk` is the supplied
+    /// npk for that `pre_state`). Programs stay privacy-agnostic: they emit seeds, the circuit
+    /// resolves public vs private based on the `pre_state`'s mask.
```
Arjentix: I think there are too many internal implementation information exposed here. This structure is used by our users -- program devs. And they should not care about distinction between private or public pda or different masks

nssa/core/src/program.rs
```diff
@@ -114,7 +151,12 @@ pub enum Claim {
     /// This will give no error if program had authorization in pre state and may be useful
     /// if program decides to give up authorization for a chained call.
     Authorized,
-    /// The program requests ownership of the account through a PDA.
+    /// The program requests ownership of the account through a PDA. The `pre_state`'s
+    /// visibility mask selects the derivation formula: mask 0 uses
+    /// `AccountId::for_public_pda(&program_id, &seed)`, mask 3 uses
+    /// `AccountId::for_private_pda(&program_id, &seed, &npk)` with the supplied npk for that
+    /// `pre_state`. Programs stay privacy-agnostic: they emit a seed, the circuit resolves the
+    /// rest from the mask.
```
Arjentix: Same here

nssa/core/src/program.rs
```diff
@@ -477,19 +519,24 @@ pub enum ExecutionValidationError {
...
 }
 
+/// Computes the set of public-PDA `AccountId`s the callee is authorized to mutate.
+///
+/// Returns only public-form derivations, suitable for contexts where all accounts are public
+/// (e.g. the public-execution path). The privacy circuit must additionally check each mask-3
+/// `pre_state` against [`AccountId::for_private_pda`] with the supplied npk for that
+/// `pre_state`.
 #[must_use]
 pub fn compute_authorized_pdas(
```
Arjentix:
```suggestion
/// Computes the set of public-PDA `AccountId`s the callee is authorized to mutate.
///
/// Returns only public-form derivations, suitable for contexts where all accounts are public
/// (e.g. the public-execution path). The privacy circuit must additionally check each mask-3
/// `pre_state` against [`AccountId::for_private_pda`] with the supplied npk for that
/// `pre_state`.
#[must_use]
pub fn compute_public_authorized_pdas(
```

Let's reflect the new nuance in the name

program_methods/guest/src/bin/privacy_preserving_circuit.rs
```diff
@@ -501,8 +718,33 @@ fn main() {
...
+                    panic!(
+                        "private_account_keys shorter than visibility_mask demands: no key for masked position {pos} (mask {mask})"
+                    )
+                });
+                if mask == 3 {
+                    private_pda_npk_by_position.insert(pos, *npk);
+                }
+            }
+        }
+    }
```
Arjentix: I think we can move this into `derive_from_outputs()`

program_methods/guest/src/bin/privacy_preserving_circuit.rs
```diff
@@ -218,19 +258,55 @@ impl ExecutionState {
...
+                                    return Some((*seed, false, caller));
+                                }
+                                if let Some(npk) =
+                                    private_pda_npk_by_position.get(&pre_state_position)
+                                    && AccountId::for_private_pda(&caller, seed, npk)
+                                        == pre_account_id
+                                {
+                                    return Some((*seed, true, caller));
+                                }
+                                None
```
Arjentix: This should be never reachable, isn't it? Every pda seed should correspond to either public or private account

program_methods/guest/src/bin/privacy_preserving_circuit.rs
```diff
@@ -218,19 +258,55 @@ impl ExecutionState {
...
+                            caller,
+                            seed,
+                            pre_account_id,
+                        );
+                        if is_private_form {
+                            self.private_pda_bound_positions.insert(pre_state_position);
+                        }
+                    }
+
+                    let is_authorized = previous_is_authorized || matched_caller_seed.is_some();
```
Arjentix: Let's move this whole `is_authorized` computation into a separate function. This became really bulky

Arjentix (commented):
Nice thing to have, thanks for the PR. I left some comments and also have some further questions:

1. So as I understand user first generates `nsk` when it can derive private pda using `program_id`, `seed` and `npk` (which they derived from `nsk`). That means one `nsk` can correspond to two account_id: basic and pda. Is this right? Aren't there any subtle issues with that?
2. Does this implementation mean that ATA will work for private accounts? If so it should mean that private transfers to existing private accounts problem is solved

moudyellaz:
> Also, this PR is adding more complexity to the circuit. Did you check if performance has regressed in terms of proving time?

Measured on my laptop, 3 runs per pair, runs interleaved between branches:

| Test | main mean | HEAD mean |
|------|-----------|-----------|
| `prove_..._fully_private` | 92.0s | 89.0s |
| `prove_..._public_and_private_pre_accounts` | 63.5s | 61.7s |
| `private_pda_claim_succeeds` (new mask-3 path, HEAD only) | n/a | 65.1s |

No measurable regression. Per-branch variance is ±5-10s, cross-branch deltas sit inside that. The new mask-3 claim path proves in the same range as the existing tests, which rules out meaningful overhead from the added `assert_family_binding` hashmap op (it runs on every claim in the new path).

Also, the shared tests don't emit `Claim::Pda`, and the privacy circuit on main has no claim-exercising test to cross-compare against, so the family-binding overhead isn't isolated to a directly comparable delta. It's only bounded by the mask-3 claim test sitting in the same cost range as non-claim tests.

Cargo.toml
moudyellaz: Yes, it showed up in CI run [24568143359](https://github.com/logos-blockchain/logos-execution-zone/actions/runs/24568143359): both `valid-proof-test` and `integration-tests` failed with `collect2: fatal error: ld terminated with signal 7 [Bus error]` during `rust-lld` linking of the test binaries. Classic LLD OOM on `ubuntu-latest` runners. The profile tweak keeps line tables for backtraces but drops full DWARF type info, which was enough to stop the linker from crashing. Next run went green.

nssa/core/src/circuit_io.rs
moudyellaz: Agreed, I'll open a follow-up issue tracking the refactor.

nssa/core/src/program.rs
moudyellaz: Done in `22aa5ef7`. Rewrote to drop the mask/derivation-formula exposition; the doc now just says the seeds authorize the callee to mutate accounts derived from `(caller_program_id, seed)`, regardless of public or private.

nssa/core/src/program.rs
moudyellaz: Done in `22aa5ef7`. Same treatment as the `pda_seeds` doc.

nssa/core/src/program.rs
moudyellaz: Done in `22aa5ef7`. Renamed to `compute_public_authorized_pdas`.

program_methods/guest/src/bin/privacy_preserving_circuit.rs
moudyellaz: Done in `e5b77a27`. Moved the position → npk map construction into `ExecutionState::derive_from_outputs` and stored it as a field (`private_pda_npk_by_position`). `main()` no longer owns the intermediate structure; `derive_from_outputs` takes `private_account_keys` directly and builds the map as part of state initialization.

program_methods/guest/src/bin/privacy_preserving_circuit.rs
moudyellaz: Done in `e5b77a27`. Extracted `resolve_authorization_and_record_bindings` as a free function taking the three field borrows it needs (same shape as `assert_family_binding`). A `&mut self` method would have conflicted with the `&mut self.post_states` borrow held by the `Occupied` match arm, the free function lets rustc split-borrow.

program_methods/guest/src/bin/privacy_preserving_circuit.rs
moudyellaz: The match runs per `pre_state`, not per seed. A caller can emit `pda_seeds: [S1, S2]` and a callee can have multiple `pre_state`s. For a given `pre_state`, most of its caller-side seeds will NOT match (either by public or private derivation), so `find_map` returns `None` on them and only the one matching seed (if any) short-circuits with `Some`. If none of the seeds match, the `find_map` as a whole returns `None`, which is how we know this `pre_state` is not authorized via caller seeds (it may still be authorized from a previous call via `previous_is_authorized`). So `None` is the common case inside the inner loop, and also a valid outcome of the outer `find_map`.

nssa/core/src/circuit_io.rs
schouhy: yes, those ugly masks have lived with us for enough time

schouhy:
> > Also, this PR is adding more complexity to the circuit. Did you check if performance has regressed in terms of proving time?
> 
> Measured on my laptop, 3 runs per pair, runs interleaved between branches:
> 
> Test	main mean	HEAD mean
> `prove_..._fully_private`	92.0s	89.0s
> `prove_..._public_and_private_pre_accounts`	63.5s	61.7s
> `private_pda_claim_succeeds` (new mask-3 path, HEAD only)	n/a	65.1s
> No measurable regression. Per-branch variance is ±5-10s, cross-branch deltas sit inside that. The new mask-3 claim path proves in the same range as the existing tests, which rules out meaningful overhead from the added `assert_family_binding` hashmap op (it runs on every claim in the new path).
> 
> Also, the shared tests don't emit `Claim::Pda`, and the privacy circuit on main has no claim-exercising test to cross-compare against, so the family-binding overhead isn't isolated to a directly comparable delta. It's only bounded by the mask-3 claim test sitting in the same cost range as non-claim tests.

awesome! thanks for running these

Arjentix (approved):
LGTM, thanks

PR #447, opened 2026-04-15
Generalize Npk to multiple accounts

schouhy:
## 🎯 Purpose

Generalizes private account key derivation to support multiple accounts per key pair. Previously, a private account's AccountId was derived from npk alone, limiting each key to a single account. This PR introduces an Identifier field so one (npk, vpk) key pair can back multiple independent private accounts. One for each identifier.

**This PR introduces braking changes in the wallet storage. Old storage files need to be removed.**

## ⚙️  Approach

- Change `AccountId` derivation formula to `SHA256(prefix || npk || identifier_le)`
- Change commitment, nullifier, and nonce init formulas to take `account_id` instead of `npk` directly
- Embed the identifier in the ciphertext (prepended as 16 LE bytes) so recipients can recover their `AccountId` from an incoming transfer
- Refactor private key tree nodes to store `Vec<(Identifier, Account)>` instead of a single `Account`, so one node can track multiple private accounts
- Add `wallet account new private-accounts-key` command that creates a key node and prints npk/vpk without assigning an identifier. Senders supply their own identifiers
- Keep wallet account new private (creates a key node + registers a random identifier immediately, for the single-account use case)
- Update FFI, CLI, and program facades
- Update token transfer tutorial to cover the multi-sender use case

## 🧪 How to Test

For the multi-account flow manually:
1. `wallet account new private-accounts-key`. Note `npk` and `vpk`
2. Send from two different accounts to the same `(npk, vpk)` with different `--to-identifier` values
3. `wallet account sync-private`. Both accounts should appear under the same key path

## 🔗 Dependencies

None.

## 🔜 Future Work

Move to deterministic identifiers derived from the shared secret as source of entropy.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


docs/LEZ testnet v0.1 tutorials/token-transfer.md
```diff
@@ -248,3 +253,65 @@ wallet auth-transfer send \
...
+wallet auth-transfer send \
+    --from Public/CharlieYrP9BmhbFVQyBcbznU8bAXcwrzwRoPTetXdQPB \
+    --to-npk a3f7c21b8e905d4f6a1bc783d0e2f94c1d5a6b7e8f9012345678abcdef012345 \
+    --to-vpk 03b1d2e3f4a5b6c7d8e9f0a1b2c3d4e5f6071819202122232425262728292a2b2c \
+    --to-identifier 2 \
+    --amount 5
+```
+
+> [!Note]
+> Bob and Charlie each chose a different identifier. They do not need to coordinate — any two distinct values work.
```
Arjentix, line 301: Hmm, can't we then make identifier random in wallet?

key_protocol/src/key_protocol_core/mod.rs
```diff
@@ -11,14 +12,15 @@ use crate::key_management::{
 };
 
 pub type PublicKey = AffinePoint;
+pub type DefaultPrivateAccountsMap =
+    BTreeMap<nssa::AccountId, (KeyChain, Vec<(Identifier, nssa_core::account::Account)>)>;
```
Arjentix: Wouldn't it be better if you just make `(KeyChain, Vec<(Identifier, nssa_core::account::Account)>)` a structure? Then you don't need this type def

key_protocol/src/key_management/key_tree/traits.rs
```diff
@@ -1,15 +1,8 @@
...
-    fn chain_code(&self) -> &[u8; 32];
-
-    fn child_index(&self) -> Option<u32>;
-
-    fn account_id(&self) -> nssa::AccountId;
+    fn from_seed(seed: [u8; 64]) -> Self;
+    #[must_use]
+    fn derive_child(&self, cci: u32) -> Self;
+    #[must_use]
+    fn account_ids(&self) -> Vec<nssa::AccountId>;
```
Arjentix: Maybe can be `impl Iterator`?

key_protocol/src/key_management/key_tree/mod.rs
```diff
@@ -171,48 +198,50 @@ impl<N: KeyNode> KeyTree<N> {
...
+        get_account: impl Fn(AccountId) -> F,
+    ) -> Result<()> {
         let depth = usize::try_from(depth).expect("Depth is expected to fit in usize");
         'outer: for i in (1..depth).rev() {
             println!("Cleanup of tree at depth {i}");
             for id in ChainIndex::chain_ids_at_depth(i) {
                 if let Some(node) = self.key_map.get(&id) {
-                    if node.value.1 == nssa::Account::default() {
+                    let address = node.account_id();
+                    let node_acc = get_account(address).await?;
```
Arjentix, line 240: I don't get why we need `get_account()` here now, but haven't needed it before...

key_protocol/src/key_management/key_tree/mod.rs
```diff
@@ -221,41 +250,67 @@ impl KeyTree<ChildKeysPrivate> {
...
     ///
     /// Walks through tree in layers of same depth using `ChainIndex::chain_ids_at_depth()`.
     ///
+    /// Chain must be parsed for accounts beforehand.
+    ///
     /// Slow, maintains tree consistency.
-    pub async fn cleanup_tree_remove_uninit_layered<F: Future<Output = Result<Account>>>(
-        &mut self,
-        depth: u32,
-        get_account: impl Fn(AccountId) -> F,
```
Arjentix, line 238: And the opposite here...

key_protocol/src/key_management/key_tree/keys_private.rs
```diff
@@ -129,12 +119,32 @@ impl<'a> From<&'a ChildKeysPrivate> for &'a (KeyChain, nssa::Account) {
     clippy::single_char_lifetime_names,
     reason = "TODO add meaningful name"
 )]
-impl<'a> From<&'a mut ChildKeysPrivate> for &'a mut (KeyChain, nssa::Account) {
+impl<'a> From<&'a mut ChildKeysPrivate> for &'a mut (KeyChain, Vec<(Identifier, nssa::Account)>) {
```
Arjentix: Looks difficult, better to create a new meaningful type as I suggested above

program_methods/guest/src/bin/privacy_preserving_circuit.rs
```diff
@@ -307,7 +307,7 @@ impl ExecutionState {
 fn compute_circuit_output(
     execution_state: ExecutionState,
     visibility_mask: &[u8],
-    private_account_keys: &[(NullifierPublicKey, SharedSecretKey)],
+    private_account_keys: &[(NullifierPublicKey, Identifier, SharedSecretKey)],
```
Arjentix, line 492: Let's also have a structure

nssa/core/src/nullifier.rs
```diff
@@ -4,18 +4,24 @@ use serde::{Deserialize, Serialize};
...
+
 #[derive(Debug, Serialize, Deserialize, PartialEq, Eq)]
 #[cfg_attr(any(feature = "host", test), derive(Clone, Hash))]
 pub struct NullifierPublicKey(pub [u8; 32]);
 
-impl From<&NullifierPublicKey> for AccountId {
-    fn from(value: &NullifierPublicKey) -> Self {
-        const PRIVATE_ACCOUNT_ID_PREFIX: &[u8; 32] =
-            b"/LEE/v0.3/AccountId/Private/\x00\x00\x00\x00";
+impl From<(&NullifierPublicKey, Identifier)> for AccountId {
```
Arjentix, line 15: Maybe let's move it to `AccountId::from_npk()` or something

wallet/src/lib.rs
```diff
@@ -256,13 +256,36 @@ impl WalletCore {
...
+            .storage
+            .user_data
+            .private_key_tree
+            .key_map
+            .get(&cci)
+            .expect("Node was just inserted")
+            .value
+            .0
+            .nullifier_public_key
+            .clone();
```
Arjentix: Ohhh, incapsulation went out. Good that I refactored stuff like that in my PR

wallet/src/cli/account.rs
```diff
@@ -91,6 +91,12 @@ pub enum NewSubcommand {
         /// Label to assign to the new account.
         label: Option<String>,
     },
+    /// Create a new receiving key (npk + vpk) to share with senders.
+    PrivateAccountsKey {
```
Arjentix, line 97: Maybe like that?

```suggestion
    PrivateAccountKey {
```

Arjentix (approved):
LGTM, nice idea

key_protocol/src/key_management/key_tree/mod.rs
schouhy: I think git diff is confusing things here. Somehow I ended up switching places for impls for `KeyTree<ChildKeysPrivate>` and `KeyTree<ChildKeysPublic>` and the diff is showing crazy. Impls are as they were before

key_protocol/src/key_management/key_tree/mod.rs
schouhy: answer [here](https://github.com/logos-blockchain/logos-execution-zone/pull/447/changes#r3135263773)

wallet/src/cli/account.rs
schouhy: ah I meant it as PrivateAccountsKey because it is now a Key to manage multiple accounts

nssa/core/src/nullifier.rs
schouhy: I agree. But I'll leave it for another PR if that's ok

program_methods/guest/src/bin/privacy_preserving_circuit.rs
schouhy: Agreed, but let's change that on another PR

key_protocol/src/key_management/key_tree/keys_private.rs
schouhy: actually we didn't even need those impls

wallet/src/chain_storage.rs
```diff
@@ -170,36 +181,89 @@ impl WalletChainStore {
...
+            let entry = entry.get_mut();
+            let identifier = entry
+                .accounts
+                .iter()
+                .find_map(|(id, _)| {
+                    (nssa::AccountId::from((&entry.key_chain.nullifier_public_key, *id))
+                        == account_id)
+                        .then_some(*id)
+                })
+                .unwrap_or(0);
```
moudyellaz: If I understand well, here the caller-supplied `identifier` (decrypted from ciphertext) is shadowed and dropped. If no existing entry in the vec matches this `account_id`, the function pushes `(0, account)`, breaking the invariant `H(npk, id) == account_id` for that vec entry. Any later spend recomputes the wrong commitment and fails.

Maybe drop the `find_map` dance and just use the `identifier` parameter directly, with `debug_assert_eq!(AccountId::from((&entry.key_chain.nullifier_public_key, identifier)), account_id)` as a guard?

wallet/src/cli/programs/native_token_transfer.rs
```diff
@@ -304,6 +312,9 @@ pub enum NativeTokenTransferProgramSubcommandShielded {
         /// `to_vpk` - valid 33 byte hex string.
         #[arg(long)]
         to_vpk: String,
+        /// Identifier for the recipient's private account.
+        #[arg(long, default_value_t = 0)]
```
moudyellaz: If I understand well, the top-level `Send` uses `Option<u128>` with `to_identifier.unwrap_or_else(rand::random)`, but the direct subcommands `Mint`, `ShieldedForeign`, `PrivateForeign`, etc. declare `#[arg(long, default_value_t = 0)]` for `to_identifier` / `holder_identifier` / `recipient_identifier`. Two unrelated senders both omitting the flag against the same `(npk, vpk)` produce identical `account_id = H(npk, 0)`, and on the recipient `chain_storage::insert_private_account_data` silently overwrites the prior account.

Maybe align all subcommands with `Option<u128>` + `unwrap_or_else(rand::random)` so omission means "fresh random id" everywhere?

wallet/src/helperfunctions.rs
```diff
@@ -178,6 +185,23 @@ pub fn produce_data_for_storage(
         }
     }
 
+    // Also persist fresh nodes that have no identifiers yet and are therefore absent from
+    // account_id_map. Without this, a restart before the first sync would lose the node
+    // entirely, making it impossible to detect incoming transfers.
+    for (chain_index, node) in &user_data.private_key_tree.key_map {
+        if !covered_private_chain_indices.contains(chain_index) {
+            let account_id = nssa::AccountId::from((&node.value.0.nullifier_public_key, 0_u128));
```
moudyellaz: This synthesizes `account_id = H(npk, 0)` for every fresh `private-accounts-key` node so it survives a restart. On reload that synthetic id gets re-inserted into `account_id_map`, which means identifier=0 is silently claimed for every empty node before any real transfer arrives. Combined with the previous comment's default-of-0, the first incoming `--to-identifier 0` lands in this placeholder and looks like a normal first receipt.

What about persisting bare nodes via a separate list (e.g. `Vec<ChainIndex>`) instead of synthesizing an `AccountId`, so the identifier-0 slot stays free?

testnet_initial_state/src/lib.rs
```diff
@@ -98,6 +98,7 @@ pub struct PrivateAccountPrivateInitialData {
     pub account_id: nssa::AccountId,
     pub account: nssa_core::account::Account,
     pub key_chain: KeyChain,
+    pub identifier: nssa_core::Identifier,
```
moudyellaz, line 100: `ChildKeysPrivate.value` also flipped from `(KeyChain, Account)` to `(KeyChain, Vec<(Identifier, Account)>)`. Both shapes feed the persisted wallet store. Without `#[serde(default)]` or a versioned envelope or a `serde(from)` migration, an existing `wallet_storage.json` from `main` will fail to deserialize on first run after this PR lands. Genesis address constants (`PRIV_ACC_A_TEXT_ADDR` etc.) also shift because the `AccountId` formula changed, so anyone with state pinned to the old addresses is locked out.

Two options worth considering: (a) `#[serde(default)]` on `identifier` plus a `From<old>` migration that wraps the legacy tuple as `vec![(0, account)]`, or (b) bump a top-level format version and announce the keystore reset in release notes.

integration_tests/tests/auth_transfer/private.rs
```diff
@@ -71,6 +72,7 @@ async fn private_transfer_to_foreign_account() -> Result<()> {
         to_label: None,
         to_npk: Some(to_npk_string),
         to_vpk: Some(hex::encode(to_vpk.0)),
+        to_identifier: Some(0),
```
moudyellaz, line 75: The PR's headline feature, two distinct identifiers under the same `(npk, vpk)` resolving to two private accounts, has zero automated coverage. Every modified integration site passes `to_identifier: Some(0)` / `holder_identifier: 0` mechanically. The new `wallet_ffi_create_private_accounts_key` tests in `wallet_ffi.rs` also use identifier 0. The KAT in `nullifier.rs` only pins `(npk, 0)`. So a future formula change (endianness flip, prefix tweak) wouldn't break any test, and the multi-sender flow in the PR description is only manually reproducible.

Could we add an integration test under this file that creates a `private-accounts-key`, sends two transfers from distinct sources with `to_identifier` = 1 and 2, runs `sync-private`, and asserts both AccountIds appear under the same `chain_index`? Plus a KAT pinning `AccountId::from((&npk, 1))` so future drift breaks loudly.

nssa/core/src/nullifier.rs
```diff
@@ -145,11 +151,11 @@ mod tests {
...
         let npk = NullifierPublicKey::from(&nsk);
         let expected_account_id = AccountId::new([
-            139, 72, 194, 222, 215, 187, 147, 56, 55, 35, 222, 205, 156, 12, 204, 227, 166, 44, 30,
-            81, 186, 14, 167, 234, 28, 236, 32, 213, 125, 251, 193, 233,
+            165, 52, 40, 32, 231, 171, 113, 10, 65, 241, 156, 72, 154, 207, 122, 192, 15, 46, 50,
+            253, 105, 164, 89, 84, 40, 191, 182, 119, 64, 255, 67, 142,
         ]);
 
-        let account_id = AccountId::from(&npk);
+        let account_id = AccountId::from((&npk, 0));
```
moudyellaz, line 158: The KAT only covers `(npk, 0)`. Since `0u128.to_le_bytes() == 0u128.to_be_bytes()`, the test would pass even if someone flipped `identifier.to_le_bytes()` to `to_be_bytes()` in the derivation. Adding a second pinned vector with a non-zero, byte-asymmetric identifier (e.g. `0x0123_4567_89AB_CDEF_FEDC_BA98_7654_3210u128`) would catch endianness or offset regressions.

docs/LEZ testnet v0.1 tutorials/token-transfer.md
```diff
@@ -218,21 +221,23 @@ Account owned by authenticated-transfer program
...
+Generated new private accounts key at path /1
 With npk 0c95ebc4b3830f53da77bb0b80a276a776cdcf6410932acc718dcdb3f788a00e
 With vpk 039fd12a3674a880d3e917804129141e4170d419d1f9e28a3dcf979c1f2369cb72
 ```
 
 > [!Tip]
-> Ignore the private account ID here and use the `npk` and `vpk` values to send to a foreign private account.
+> Ignore the account ID here and use the `npk` and `vpk` values to send to a foreign private account.
+
+### b. Send 3 tokens using the recipient’s npk and vpk
```
moudyellaz, line 240: Two inconsistencies that would confuse a first-time reader:

1. Section 4 prose still reads "The private account ID is derived from the nullifier public key." Stale, the formula is now `SHA256(prefix || npk || identifier_le)`.
2. Section 6 step b shows `wallet auth-transfer send --from ... --to-npk ... --to-vpk ... --amount 3` with no `--to-identifier`. The CLI silently picks a random one when omitted, but the tutorial never explains that, and Section 7 then says "Each sender independently chooses an identifier", leaving readers wondering whether they need to set the flag.

Worth updating Section 4 prose to mention the identifier, and adding one line in Section 6 explaining the omit-means-random behavior.

wallet/src/cli/account.rs
```diff
@@ -82,7 +82,7 @@ pub enum NewSubcommand {
         /// Label to assign to the new account.
         label: Option<String>,
     },
-    /// Register new private account.
+    /// Register new private account with a random identifier.
```
moudyellaz: `wallet account new private` (registers a random identifier and a default account) vs `wallet account new private-accounts-key` (only creates the key node, sender supplies identifiers). Both produce the same key-tree node; they differ only in whether an identifier is auto-registered. For a user who skims the help text, "new private" sounds like the general case but is actually the single-account convenience, while the multi-account workflow that this whole PR is built around lives behind a longer name.

Worth at least a doc-comment update so the help text on `private-accounts-key` leads with "Recommended for receiving from multiple senders" and the help text on `private` mentions "single-account convenience". Renaming `private` to `private-account` would also remove the "private vs private-accounts-key" cognitive overhead, but that's a bigger change.

wallet-ffi/src/account.rs
```diff
@@ -59,10 +62,12 @@ pub unsafe extern "C" fn wallet_ffi_create_account_public(
     WalletFfiError::Success
 }
 
-/// Create a new private account.
+/// Create a new private account, storing a default account entry in local storage.
 ///
-/// Private accounts use privacy-preserving transactions with nullifiers
-/// and commitments.
+/// This is the private-account equivalent of `wallet_ffi_create_account_public`.
+/// It generates a key node, assigns a random identifier, and inserts a default
```
moudyellaz, line 68: `wallet_ffi_create_account_private` is retained as the single-account convenience and internally calls `rand::random()` to pick the identifier. The resulting `AccountId` therefore depends on local RNG state and cannot be re-derived from the seed alone. The doc comment doesn't say this. A user re-importing from mnemonic on a second device will not recover the same private `AccountId`, only the same npk/vpk.

Worth one sentence in the FFI doc comment, e.g. "The random identifier is stored only in the local wallet, restoring this key from seed alone will not recover this AccountId; use `wallet_ffi_create_private_accounts_key` plus an explicitly chosen identifier for reproducible derivations."

moudyellaz (approved):
lgtm thanks, left some concerns/suggestions

testnet_initial_state/src/lib.rs
schouhy: true. I'd go for option (b) as we are not at a stage of the project where breaking changes are an issue I think

wallet/src/chain_storage.rs
schouhy: yes thanks. That code is no longer needed actually

wallet/src/cli/programs/native_token_transfer.rs
schouhy: done

wallet/src/helperfunctions.rs
schouhy: I modified `PersistentAccountDataPrivate` to hold a `Vec<Identifier>` instead of a single account Id. This makes it possible to store a chain index with an empty set of identifiers, avoiding the hacky identifier=0 default to have keys survive restarts in that edge case.

integration_tests/tests/auth_transfer/private.rs
schouhy: done

nssa/core/src/nullifier.rs
schouhy: done

wallet/src/cli/account.rs
schouhy: the NPK and VPK produced by `wallet account new private` can also be used to receive from multiple senders. I expanded the docs a bit but I didn't want to be misleading with that.

docs/LEZ testnet v0.1 tutorials/token-transfer.md
schouhy: thanks. Added

wallet-ffi/src/account.rs
schouhy: it cannot be re-derived from the seed but it would be discovered by sync-private if initialized. I added a comment on that let me know if you agree with it

docs/LEZ testnet v0.1 tutorials/token-transfer.md
schouhy: done!

key_protocol/src/key_management/key_tree/traits.rs
schouhy: done!

key_protocol/src/key_protocol_core/mod.rs
schouhy: done!

docs/LEZ testnet v0.1 tutorials/token-transfer.md
schouhy: I forgot to add a comment in the tutorial though. I'll add it

Arjentix (approved):
LGTM, thanks

PR #448, opened 2026-04-15
Fix deny by updating `rustls-webpki` and `multihash`

schouhy:
## 🎯 Purpose

Fix deny by updating `rustls-webpki` and `multihash`.
Also ignore rand as in #427


moudyellaz (approved):

Pravdyvy (approved):
lgtm

PR #449, opened 2026-04-15
feat: GroupKeyHolder for GMS key management

moudyellaz:
## 🎯 Purpose

This PR adds GroupKeyHolder to the wallet for group-owned private PDAs. A group of controllers shares a Group Master Secret (GMS). From it, each controller independently derives identical keys for any PDA the group owns, enabling private multisig and k-of-n patterns over existing PDA-based programs without changing those programs.

### key_protocol: GroupKeyHolder (core cryptography)

- Per-PDA key derivation via `SHA256(domain_prefix || gms || pda_seed)` producing NSK/NPK/VSK/VPK through the existing derivation chain.
- `epoch` counter + `ratchet(rotation_salt)` forward-hashes the GMS so removed members cannot derive future keys.
- `seal_for(recipient_vpk)` / `unseal(sealed, own_vsk)` encrypts GMS+epoch via ephemeral ECDH + domain-separated KDF + AES-256-GCM for distribution over untrusted channels.
- Debug impl redacts GMS, raw accessor named `dangerous_raw_gms` to flag intent.

### key_protocol: NSSAUserData storage

- `group_key_holders: BTreeMap<String, GroupKeyHolder>` field with `#[serde(default)]` for backward compatibility.
- `group_pda_accounts: BTreeMap<AccountId, Account>` local cache for decrypted group PDA state.
- `get_group_key_holder` / `insert_group_key_holder` accessors.

### wallet: mask-3 transaction construction

- `PrivateGroupPda { group_label, program_id, seed }` variant on `PrivacyPreservingAccount`.
- `group_pda_preparation` derives keys from GroupKeyHolder, computes `AccountId::for_private_pda`, reads PDA state from local cache.

### Test program: `group_pda_spender`

- Handles deposit (claim + balance transfer) and spend (direct balance modification + noop chained call for mask-3 binding).
- Unit tests in `nssa/src/state.rs` verify both paths at the circuit level.

### Integration test

- Demonstrates GMS sealed distribution, key agreement, and forward secrecy after ratchet.

## 🧪 How to Test

```
RISC0_DEV_MODE=1 cargo test --release -p key_protocol
RISC0_DEV_MODE=1 cargo test --release -p nssa -- group_pda
```

## 🔗 Dependencies

- PR #446 (private PDA circuit support) — merged.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


integration_tests/tests/group_pda.rs
Pravdyvy, line 1: It is an imortant test, but I don't understand why should it be integration test. You not really use neiter berock node nor indexer nor sequencer. Preconfigured accounts can be read from other places.

key_protocol/src/key_management/group_key_holder.rs
```diff
@@ -0,0 +1,601 @@
...
+/// (see `seal_for` / `unseal`). Wallets persisting a `GroupKeyHolder` must encrypt it at
+/// rest; the raw bytes are exposed only via [`GroupKeyHolder::dangerous_raw_gms`], which
+/// is intended for the sealing path exclusively.
+///
+/// # Logging safety
+///
+/// `Debug` is implemented manually to redact the GMS; formatting this value with `{:?}`
+/// will not leak the secret. Code that formats through `{:#?}` on containing types is
+/// safe for the same reason.
+#[derive(Serialize, Deserialize, Clone)]
```
Pravdyvy, line 42: `Serialize` autoderive still can show gms in human-readable format. Or is it means for protection for the case where we by mistake log this data somewhere at time of execution?

key_protocol/src/key_management/group_key_holder.rs
```diff
@@ -0,0 +1,601 @@
...
+    /// adversarial seed-grinding cannot collide two PDAs' derived keys under the same
+    /// group. Uses the codebase's 32-byte protocol-versioned domain-separation convention.
+    fn secret_spending_key_for_pda(&self, pda_seed: &PdaSeed) -> SecretSpendingKey {
+        const PREFIX: &[u8; 32] = b"/LEE/v0.3/GroupKeyDerivation/SSK";
+        let mut hasher = sha2::Sha256::new();
+        hasher.update(PREFIX);
+        hasher.update(self.gms);
+        hasher.update(pda_seed.as_ref());
+        SecretSpendingKey(hasher.finalize_fixed().into())
+    }
```
Pravdyvy, line 96: This and other functions intended to be executed by users, however we do not expose it as interface, for example, in wallet. Maybe not in this PR, but we need to expose GMS methods to users.

Pravdyvy (commented):
Thank you for PR. Left a couple of questions.

integration_tests/tests/group_pda.rs
moudyellaz: yes you are right, actually i was trying to do a more complicated toy example but some acts are still not allowed by our architecture, so i made a simpler example. Will move it to protocol_key unit tests!

key_protocol/src/key_management/group_key_holder.rs
moudyellaz: Yes, the Debug redaction protects against accidental logging during execution. Serialize intentionally includes the GMS for persistence since the doc on the struct says 'wallets persisting a GroupKeyHolder must encrypt it at rest.' A custom Serialize that redacts would break the persistence path. We could add a one-line comment above the derive to make this clearer.

key_protocol/src/key_management/group_key_holder.rs
moudyellaz: Agreed, CLI commands for group operations (create group, seal/distribute, ratchet) are planned as a follow-up. This PR focuses on the key management layer and wallet API. The CLI will build on top of these methods.

key_protocol/src/key_management/group_key_holder.rs
Pravdyvy: Ok

key_protocol/src/key_management/group_key_holder.rs
Pravdyvy: Ok

Pravdyvy (approved):
lgtm

nssa/src/state.rs
```diff
@@ -2568,6 +2568,100 @@ pub mod tests {
         assert!(matches!(result, Err(NssaError::CircuitProvingError(_))));
     }
 
+    /// Group PDA deposit: creates a new PDA and transfers balance from the
+    /// counterparty. Both accounts owned by `group_pda_spender`.
+    #[test]
+    fn group_pda_deposit() {
```
schouhy: this isn't testing anything about the `state`. Let's move it to `nssa/src/privacy_preserving_transaction/circuit.rs` maybe

key_protocol/src/key_management/group_key_holder.rs
```diff
@@ -0,0 +1,670 @@
...
+    /// `ephemeral_pubkey (33) || nonce (12) || ciphertext+tag (52)` = 97 bytes.
+    ///
+    /// Each call generates a fresh ephemeral key, so two seals of the same holder produce
+    /// different ciphertexts.
+    #[must_use]
+    #[expect(
+        clippy::arithmetic_side_effects,
+        reason = "capacity arithmetic on small constants cannot overflow"
+    )]
+    pub fn seal_for(&self, recipient_vpk: &ViewingPublicKey) -> Vec<u8> {
```
schouhy: This is using `ViewingPublicKey` type alias to derive the shared secret. Viewing keys are meant for encrypt/decrypt account data. I wouldn't use the same type alias. I would create a new one specific for this use case.

nssa/src/state.rs
schouhy, line 1: the tests added in this file aren't using the gms derivation no? Is that intentional? I'm not sure what these are really testing

key_protocol/src/key_protocol_core/mod.rs
```diff
@@ -23,6 +24,17 @@ pub struct NSSAUserData {
...
+    /// Defaults to empty for backward compatibility with wallets that predate group PDAs.
+    /// An older wallet binary that re-serializes this struct will drop the field.
+    #[serde(default)]
+    pub group_key_holders: BTreeMap<String, GroupKeyHolder>,
+    /// Cached plaintext state of group PDA accounts, keyed by `AccountId`.
+    /// Updated after each group PDA transaction by decrypting the circuit output.
+    /// The sequencer only stores encrypted commitments, so this local cache is the
+    /// only source of plaintext state for group PDAs.
+    #[serde(default)]
+    pub group_pda_accounts: BTreeMap<nssa::AccountId, nssa_core::account::Account>,
```
schouhy: These won't be recoverable from mnemonic no? I think I would expect anything derived from the wallet to be recoverable from it

schouhy (commented):
Thanks for the PR. Leaving some thoughts while I finish reviewing

key_protocol/src/key_protocol_core/mod.rs
```diff
@@ -23,6 +24,17 @@ pub struct NSSAUserData {
     pub public_key_tree: KeyTreePublic,
     /// Tree of private keys.
     pub private_key_tree: KeyTreePrivate,
+    /// Group key holders for private PDA groups, keyed by a human-readable label.
+    /// Defaults to empty for backward compatibility with wallets that predate group PDAs.
+    /// An older wallet binary that re-serializes this struct will drop the field.
+    #[serde(default)]
+    pub group_key_holders: BTreeMap<String, GroupKeyHolder>,
+    /// Cached plaintext state of group PDA accounts, keyed by `AccountId`.
```
Arjentix: Isn't it only for **private** PDA? Let's highlight it in comments

key_protocol/src/key_protocol_core/mod.rs
```diff
@@ -177,6 +191,20 @@ impl NSSAUserData {
             .copied()
             .chain(self.private_key_tree.account_id_map.keys().copied())
     }
+
+    /// Returns the `GroupKeyHolder` for the given label, if it exists.
+    #[must_use]
+    pub fn get_group_key_holder(&self, label: &str) -> Option<&GroupKeyHolder> {
```
Arjentix:
```suggestion
    pub fn group_key_holder(&self, label: &str) -> Option<&GroupKeyHolder> {
```

Let's not create more `get_` prefixes, in Rust this is [not idiomatic](https://users.rust-lang.org/t/idiomatic-naming-for-getters-setters/3581)

key_protocol/src/key_management/group_key_holder.rs
```diff
@@ -0,0 +1,670 @@
...
+use rand::{RngCore as _, rngs::OsRng};
+use serde::{Deserialize, Serialize};
+use sha2::{Digest as _, digest::FixedOutput as _};
+
+use super::secret_holders::{PrivateKeyHolder, SecretSpendingKey, ViewingSecretKey};
+
+/// Manages shared viewing keys for a group of controllers owning private PDAs.
+///
+/// The Group Master Secret (GMS) is a 32-byte random value shared among controllers.
+/// Each private PDA owned by the group gets a unique `SecretSpendingKey` derived from
```
Arjentix:
```suggestion
/// Each private PDA owned by the group gets a unique [`SecretSpendingKey`] derived from
```

Let's use doc links where possible

Arjentix: Same in other places

key_protocol/src/key_management/group_key_holder.rs
```diff
@@ -0,0 +1,670 @@
...
+    /// be a fresh 32-byte random value contributed by the member who initiates the rotation.
+    /// Reusing a salt from a previous ratchet produces the same GMS as that previous
+    /// ratchet, collapsing the key rotation. Callers must generate the salt from a CSPRNG.
+    ///
+    /// After ratcheting, all remaining controllers must receive the new `GroupKeyHolder`
+    /// via `seal_for` / `unseal`.
+    #[expect(
+        clippy::arithmetic_side_effects,
+        reason = "epoch overflow at 2^32 ratchets is not a realistic scenario"
+    )]
```
Arjentix: So maybe then do `self.epoch = self.epoch.checked_add(1).expect("...")` instead?

key_protocol/src/key_management/group_key_holder.rs
```diff
@@ -0,0 +1,670 @@
...
+    pub const fn epoch(&self) -> u32 {
+        self.epoch
+    }
+
+    /// Forward-ratchets the GMS so removed members cannot derive future keys.
+    ///
+    /// The new GMS is `SHA256(PREFIX || rotation_salt || old_gms)`. The rotation salt must
+    /// be a fresh 32-byte random value contributed by the member who initiates the rotation.
+    /// Reusing a salt from a previous ratchet produces the same GMS as that previous
+    /// ratchet, collapsing the key rotation. Callers must generate the salt from a CSPRNG.
```
Arjentix: What is CSPRNG?

key_protocol/src/key_management/group_key_holder.rs
```diff
@@ -0,0 +1,670 @@
...
+    /// Uses an ephemeral ECDH key exchange to derive a shared secret, then AES-256-GCM
+    /// to encrypt the payload. The returned bytes are
+    /// `ephemeral_pubkey (33) || nonce (12) || ciphertext+tag (52)` = 97 bytes.
+    ///
+    /// Each call generates a fresh ephemeral key, so two seals of the same holder produce
+    /// different ciphertexts.
+    #[must_use]
+    #[expect(
+        clippy::arithmetic_side_effects,
+        reason = "capacity arithmetic on small constants cannot overflow"
```
Arjentix: If something cannot happen let's better use `.expect()`

key_protocol/src/key_management/group_key_holder.rs
```diff
@@ -0,0 +1,670 @@
...
+        }
+
+        let mut gms = [0_u8; 32];
+        gms.copy_from_slice(&plaintext[..32]);
+        let epoch = u32::from_le_bytes(plaintext[32..36].try_into().unwrap());
+        Ok(Self::from_gms_and_epoch(gms, epoch))
+    }
+
+    /// Derives an AES-256 key from the ECDH shared secret via SHA-256 with a domain prefix.
+    fn seal_kdf(shared: &SharedSecretKey) -> [u8; 32] {
```
Arjentix, line 177: Probably worth creating a new type for this key?

test_program_methods/guest/src/bin/private_pda_spender.rs
```diff
@@ -0,0 +1,118 @@
...
+/// Decreases PDA balance (this program owns it), increases counterparty.
+/// Chains to a noop callee with `pda_seeds` to establish the mask-3 binding
+/// that the circuit requires for existing private PDAs.
+type Instruction = (PdaSeed, ProgramId, u128, bool);
+
+#[expect(
+    clippy::allow_attributes,
+    reason = "allow is needed because the clones are only redundant in test compilation"
+)]
+#[allow(
```
Arjentix, line 25: Why not to use `expect` instead of `allow`?

wallet/src/lib.rs
```diff
@@ -41,7 +41,7 @@ pub mod cli;
 pub mod config;
 pub mod helperfunctions;
 pub mod poller;
-mod privacy_preserving_tx;
+pub mod privacy_preserving_tx;
```
Arjentix: Hm, why? It's purpose was just some help for `send_privacy_preserving_tx_with_pre_check()` wallet method

nssa/src/state.rs
```diff
@@ -2568,6 +2568,100 @@ pub mod tests {
...
+        let program = Program::group_pda_spender();
+        let noop = Program::noop();
```
Arjentix: I don't get why we need noop

nssa/src/state.rs
Arjentix: I don't understand... I can't see any Group key holders, I see just private pda

nssa/src/state.rs
```diff
@@ -2568,6 +2568,100 @@ pub mod tests {
...
+        let (output, _proof) = result.expect("group PDA deposit should succeed");
+        // Only PDA (mask 3) produces a commitment; sender (mask 0) is public.
+        assert_eq!(output.new_commitments.len(), 1);
+    }
+
+    /// Group PDA spend binding: the noop chained call with `pda_seeds` establishes
+    /// the mask-3 binding for an existing-but-default PDA. Uses amount=0 because
+    /// testing with a pre-funded PDA requires a two-tx sequence with membership proofs.
+    #[test]
+    fn group_pda_spend_binding() {
```
Arjentix: Same here

Arjentix (commented):
Thanks for the PR. Left some small comments.

I didn't quite get how it works, probably more examples like `k/n`signatures program and an integration test demonstrating it would be nice

nssa/src/state.rs
moudyellaz: Moved.

key_protocol/src/key_protocol_core/mod.rs
moudyellaz: Fixed.

key_protocol/src/key_protocol_core/mod.rs
moudyellaz: Fixed.

key_protocol/src/key_management/group_key_holder.rs
moudyellaz: Fixed.

key_protocol/src/key_management/group_key_holder.rs
moudyellaz: Fixed.

key_protocol/src/key_management/group_key_holder.rs
moudyellaz: Replaced with "secure random source".

key_protocol/src/key_management/group_key_holder.rs
moudyellaz: Replaced with checked arithmetic and `.expect()`.

wallet/src/lib.rs
moudyellaz: Reverted to `mod`, the re-export on line 29 covers the public API.

key_protocol/src/key_management/group_key_holder.rs
moudyellaz: It IS the recipient's viewing public key from their wallet. Creating a separate type for the same key material adds confusion: the recipient doesn't generate a separate "sealing key", they use their existing VPK. The underlying secp256k1 point and ECDH operation are identical.

nssa/src/state.rs
moudyellaz: Intentional. These tests verify that `group_pda_spender` works correctly with the circuit's mask-3 constraints (claim, chained call binding, balance modification). GMS derivation is tested separately in `key_protocol` (22 tests including the full lifecycle test). The two layers are intentionally decoupled.

key_protocol/src/key_management/group_key_holder.rs
moudyellaz: It's a private helper used only within `seal_for` and `unseal`. The key never leaves the function scope, so a newtype adds ceremony without safety benefit.

test_program_methods/guest/src/bin/private_pda_spender.rs
moudyellaz: The clones are only redundant in test compilation mode (CI runs `--all-targets`), but required in normal compilation. `expect` would fail in normal mode because the lint doesn't fire. `allow` is needed because the lint fires conditionally.

nssa/src/state.rs
moudyellaz: The noop chained call establishes the mask-3 `(seed, npk)` binding that the circuit requires for existing private PDAs. Without a caller providing `pda_seeds`, the circuit's binding check rejects the account. The tests intentionally don't use `GroupKeyHolder` because they verify circuit behavior, not key derivation. GMS derivation has 22 dedicated tests in `key_protocol`.

nssa/src/state.rs
moudyellaz: Answered above.

nssa/src/state.rs
moudyellaz: Answered above.

key_protocol/src/key_protocol_core/mod.rs
moudyellaz: The GMS is shared random entropy across multiple wallets, so it can't be part of any single user's deterministic key tree. If it were mnemonic-derivable, it wouldn't be a shared secret. Recovery works through re-sealing: another group member calls `seal_for` with the restored wallet's new VPK. This is the standard recovery model for any multi-party shared secret.

moudyellaz:
`group_pda_spender` in `test_program_methods/` is the example program. A full integration test that covers deposit → spend requires membership proof resolution for existing private PDAs, which the wallet doesn't support for group PDAs yet. The unit tests in `circuit.rs` verify both paths at the circuit level. A more complete integration test will follow once the wallet can resolve group PDA membership proofs. @Arjentix

test_program_methods/guest/src/bin/private_pda_spender.rs
Arjentix: Ah, so we have something `Copy` only in tests? Let's better make it always `Copy`?

Arjentix (approved):

test_program_methods/guest/src/bin/private_pda_spender.rs
moudyellaz: The cloned types (`Account`, `AccountWithMetadata`) contain heap-allocated fields (`Data`, `Vec`), so they can't be `Copy`. The clones are only redundant in test mode because the compiler can prove ownership isn't reused, but in release mode the borrow checker needs them.

key_protocol/src/key_management/group_key_holder.rs
schouhy: viewing public keys are for encryption of lez accounts. If I understand correctly this is using them to encrypt the gms to be distributed. I think this is a different purpose. Recipient here is not associated with any particular lez account. And there's no "default wallet vpk". A  user would need to use an existing vpk for some of their private accounts, or sample a new one with `wallet account new private`. In any case it seems odd to me. I think it would be better to have a different type (or type alias like `type SealingPublicKey = Secp256k1Point`) for this to reflect the usage.

artifacts/test_program_methods/group_pda_router.bin
schouhy, line 1: there's no `group_pda_router.rs` under the test programs. is this stale?

key_protocol/src/key_management/group_key_holder.rs
```diff
@@ -0,0 +1,667 @@
...
+///
+/// # Logging safety
+///
+/// `Debug` is implemented manually to redact the GMS; formatting this value with `{:?}`
+/// will not leak the secret. Code that formats through `{:#?}` on containing types is
+/// safe for the same reason.
+#[derive(Serialize, Deserialize, Clone)]
+pub struct GroupKeyHolder {
+    gms: [u8; 32],
+    epoch: u32,
```
schouhy: not sure if I understand the need of the epoch. Doesn't seem to bound to the keys or used anywhere. In case of ratcheting, the new keys need to be distributed again. Having the epoch doesn't add much I think. With the spirit of simplifying, I would remove it.

key_protocol/src/key_management/group_key_holder.rs
```diff
@@ -0,0 +1,667 @@
...
+    ///
+    /// The new GMS is `SHA256(PREFIX || rotation_salt || old_gms)`. The rotation salt must
+    /// be a fresh 32-byte random value contributed by the member who initiates the rotation.
+    /// Reusing a salt from a previous ratchet produces the same GMS as that previous
+    /// ratchet, collapsing the key rotation. Callers must generate the salt from a secure random
+    /// source.
+    ///
+    /// After ratcheting, all remaining controllers must receive the new `GroupKeyHolder`
+    /// via `seal_for` / `unseal`.
+    pub fn ratchet(&mut self, rotation_salt: [u8; 32]) {
```
schouhy: I'm unsure how ratcheting would work in practice. It is effectively creating a new key altogether, that needs distribution. And all accounts under the new gms will need to be initialized again, resulting in different account ids even. Seems like a hard reset. Accounts under the old gms will still be usable with the old keys. If they had funds, they would need to be manually moved to the new gms accounts. Can't we simplify by removing the ratcheting mechanism and in case members need to be removed, a new ceremony needs to take place like the original one. Or why is ratcheting achieving something better?

key_protocol/src/key_protocol_core/mod.rs
```diff
@@ -23,6 +24,17 @@ pub struct NSSAUserData {
...
+    /// Defaults to empty for backward compatibility with wallets that predate group PDAs.
+    /// An older wallet binary that re-serializes this struct will drop the field.
+    #[serde(default)]
+    pub group_key_holders: BTreeMap<String, GroupKeyHolder>,
+    /// Cached plaintext state of group private PDA accounts, keyed by `AccountId`.
+    /// Updated after each group PDA transaction by decrypting the circuit output.
+    /// The sequencer only stores encrypted commitments, so this local cache is the
+    /// only source of plaintext state for group PDAs.
+    #[serde(default)]
+    pub group_pda_accounts: BTreeMap<nssa::AccountId, nssa_core::account::Account>,
```
schouhy: Why does the wallet need to know that some private accounts are pdas controlled by a group? This is a question in general about the new `PrivateGroupPda` variant introduced in this PR. From the wallet's perspective wouldn't any private group pda look exactly like a private pda? IIUC group pdas only differ by normal private pdas by the fact that the nullifier and viewing keys were generated by a key distribution mechanism instead of the normal one for private accounts. But then in terms of usage in privacy executions they are indistinguishable no?

test_program_methods/guest/src/bin/private_pda_spender.rs
schouhy, line 1: It took me some time to understand what this is actually doing. This is what I found confusing:
- It doesn't have group specific logic. I would call it `private_pda_spender`? Private group ownership is a private account concept and programs should be agnostic of them I think
- Deposits into the pda are enforced to have `is_authorized=true` explicitely with the assert. But Withdrawals from the pda are enforced to be `is_authorized=true` by the circuit and sequencer checks on chained calls pre states. This is indirect and I think it deserves a comment.
- Pdas are usually for program A to be the authority of an account owned by program B. In this example we are using the same program for both the autority and ownership, which leads to the hacky noop
  chained call. An idea in case there's time: Why not have the pda owned by the authenticated transfer program instead. That way 
  - this program only needs a pda init function (simpler) and doesnt need deposit logic: deposits into the pda can be done by calling the auth-transfer program directly.
  - withdraws from the pda need to be executed through this program (that chains a call to the auth-transfer program authorizing the pda).

wallet/src/lib.rs
```diff
@@ -201,6 +201,12 @@ impl WalletCore {
         &self.storage
     }
 
+    /// Get mutable storage (e.g. for inserting group key holders).
+    #[must_use]
+    pub const fn storage_mut(&mut self) -> &mut WalletChainStore {
+        &mut self.storage
+    }
```
schouhy: this is never called. I'd remove it and replace it when needed by some weaker version that doesn't expose the entire storage as mutable

schouhy (commented):
Hey thanks. Have some general comments about the approach

key_protocol/src/key_management/group_key_holder.rs
moudyellaz: Added `SealingPublicKey` and `SealingSecretKey` type aliases, updated `seal_for`/`unseal` signatures to use them.

artifacts/test_program_methods/group_pda_router.bin
moudyellaz: Stale, removed. Remained since I was attempting a more complicated integration test example before reverting.

key_protocol/src/key_management/group_key_holder.rs
moudyellaz: The epoch lets recipients detect stale distributions without a test derivation. After unsealing, a member can compare their epoch to the group's expected epoch to know if they missed a ratchet. Removing it saves one `u32` but loses the only lightweight staleness signal.

key_protocol/src/key_management/group_key_holder.rs
moudyellaz: You're right that ratcheting produces new account IDs and requires fund migration, same as a fresh ceremony. The advantage is operational: ratcheting is deterministic (any member can initiate without being the randomness source), and the forward secrecy is cryptographic rather than trust-based. That said, the practical difference is small. If you prefer simplicity, I'm open to removing it.

key_protocol/src/key_protocol_core/mod.rs
moudyellaz: You're right, from the wallet's perspective they're indistinguishable. Refactored `PrivateGroupPda` to `PrivatePda` with pre-resolved keys (nsk, npk, vpk, program_id, seed). The wallet no longer looks up GroupKeyHolder or knows about GMS during transaction construction.

test_program_methods/guest/src/bin/private_pda_spender.rs
moudyellaz: Good points. Renamed to `private_pda_spender`. The auth-transfer ownership approach is cleaner but requires extending authenticated_transfer to support `Claim::Pda`, which it doesn't today. Will restructure in a follow-up once we add PDA support to authenticated_transfer.

wallet/src/lib.rs
moudyellaz: Removed.

test_program_methods/guest/src/bin/private_pda_spender.rs
schouhy: ah I think the claim variants are confusing us all. This [summarizes](https://discord.com/channels/973324189794697286/1483475612780990647/1494602700694753423) it well.

authenticated transfer can be used with the current `Claim::Authorized` it already has, since the only thing that program needs to claim an account is that it has been authorized before and the caller can do that by the chained call seeds.

key_protocol/src/key_management/group_key_holder.rs
schouhy: uhm ok let's keep it. I don't insist but still unsure why not the fact that a user receives a new gms is already a signal that the old one is already stale.

key_protocol/src/key_management/group_key_holder.rs
schouhy: doesn't the salt need to be randomly chosen by ratcheter? Otherwise I don't understand how it can be initiated without randomness and at the same time kept away from the removed members. I'd personally remove to simplify things with the next prs too.

schouhy (approved):
Thanks LGTM. I'd personally remove the epoch, ratcheting and change the private_pda_spender to use the authenticated transfer. But I don't insist.

key_protocol/src/key_management/group_key_holder.rs
moudyellaz: Removed epoch and ratcheting.

key_protocol/src/key_management/group_key_holder.rs
moudyellaz: Removed ratcheting to simplify.

test_program_methods/guest/src/bin/private_pda_spender.rs
moudyellaz: Need to restructure for auth-transfer, will work on it.

PR #450, opened 2026-04-16
feat: move configurable initial data to genesis

Arjentix:
## 🎯 Purpose

After #439 it was clear that state syncing between sequencer and indexer is a bit odd in terms of initial state. Before this PR we shared common initial accounts configuration between them, which was leading to exceptional handling for genesis on indexer side and unclear world state for external viewer.

This PR is the first part of the two. It moves *configurable* initial accounts data into genesis. In the future PR we will move `testnet_initial_state` into genesis as well, so that our mainnet and testnet differ only in config files, not in code.

## ⚙️ Approach

There were a lot of changes, here are the main ones:

1. Indexer no longer have initial accounts configuration at all. It listens for genesis from bedrock as every other block.
2. Wallet no longer have preconfigured accounts. This was dangerous -- private keys stored in config file we don't even plan to encrypt. And this no longer fits into new vision. Instead `import` command should be used to import external public and private keys.
3. Did some major refactor on wallet side. Replaced `account_label` with `CliAccountMention` type to use account ids and labels interchangeably which also reduced a lot of code. Some types were a bit restructured to provide more type-safety and incapsulation. And etc.
4. Sequencer now accepts genesis configuration. Currently the only options there are supply of public and private accounts. This can be expanded in future.
5. Sequencer transforms this configuration into basic transactions by calling `genesis_supply_account` program. This program is specific to genesis (same as `clock` program). This program initialize account under `authenticated_transfer` program and then mints configured amount of native tokens.
6. To bypass some checks (like minting tokens and claiming accounts) `is_genesis` option was added to validation.
7. Sequencer no longer builds its initial state when existing storage is found, it read the state from storage instead.

## 🧪 How to Test

See CI.

## 🔗 Dependencies

None

## 🔜 Future Work

Move state hardcoding into genesis configuration.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


nssa/src/validated_state_diff.rs
```diff
@@ -47,6 +47,7 @@ impl ValidatedStateDiff {
     ) -> Result<Self, NssaError> {
         let message = tx.message();
         let witness_set = tx.witness_set();
+        let is_genesis = block_id == 1;
```
moudyellaz: The whole balance-conservation and claim-authorization bypass hangs on this single line. Two things worry me:

1. If an operator ever has to restart a sequencer against an empty/lost DB, `start_from_config` silently rebuilds genesis from the current config file (`sequencer/core/src/lib.rs` ~L66). There's nothing here that ties "block_id == 1" to the *original* genesis pinned on L1. A drifted config would produce a new, differently-hashed genesis block and skip balance conservation for it without any alarm.
2. The flag is re-derived from an integer at every call site. A future refactor that passes the wrong `block_id` (e.g. re-replaying block 1 in a test harness, or an off-by-one somewhere) would silently unlock minting.

Worth considering: pass `is_genesis` in explicitly from the block context (e.g. a named field on `Block` set once by the sequencer when it constructs the genesis block), or combine the check with `prev_block_hash == [0; 32]` so two independent invariants must hold.

sequencer/core/src/lib.rs
```diff
@@ -77,78 +71,51 @@ impl<BC: BlockSettlementClientTrait, IC: IndexerClientTrait> SequencerCore<BC, I
...
-                            nssa::program::Program::authenticated_transfer_program().id();
-
-                        (
-                            nssa_core::Commitment::new(npk, &acc),
-                            nssa_core::Nullifier::for_account_initialization(npk),
-                        )
-                    })
-                    .collect()
-            });
+        let (store, state) = SequencerStore::open_db(&db_path, signing_key.clone()).map_or_else(
```
moudyellaz: This `map_or_else` catches every `open_db` error and falls back to building a fresh genesis from the current config. That's correct for a legitimate first run, but silently wrong for every other failure mode: corrupted DB, permissions problem, transient filesystem error, partially-restored backup. In those cases the sequencer will build a new genesis block against the *current* config, which may have drifted from whatever L1 originally settled, producing a divergent chain without any operator-visible alarm.

Suggest distinguishing "DB does not exist yet" (proceed with genesis) from "DB exists but failed to open" (panic or bail so a human looks at it). The fresh-genesis path is also a good place to assert `chain_height == 0` before proceeding.

indexer/core/src/block_store.rs
```diff
@@ -20,14 +21,10 @@ pub struct IndexerStore {
...
-    ///
-    /// ATTENTION: Will overwrite genesis block.
-    pub fn open_db_with_genesis(
-        location: &Path,
-        genesis_block: &Block,
-        initial_state: &V03State,
-    ) -> Result<Self> {
-        let dbio = RocksDBIO::open_or_create(location, genesis_block, initial_state)?;
+    pub fn open_db(location: &Path) -> Result<Self> {
+        let initial_state = testnet_initial_state::initial_state();
```
moudyellaz, line 26: The indexer bootstraps from `testnet_initial_state::initial_state()` and then replays blocks from bedrock, including the genesis block. Nothing asserts that the resulting state after processing genesis matches either the hardcoded baseline or the sequencer's genesis. If `testnet_initial_state` ever drifts from what the sequencer's `build_genesis_state` produces (e.g. someone edits one but not the other, or the sequencer's config is changed), indexer and sequencer will diverge silently and only surface as incorrect query results downstream.

Cheapest mitigation: hash-compare the indexer's computed post-genesis state root against a canonical value (either pinned in config, or read from the genesis block's settlement record) and refuse to serve if they don't match.

moudyellaz (approved):
lgtm thanks, left some comments

Arjentix:
Interesting side effect -- `valid-proof-test` ci job now [takes](https://github.com/logos-blockchain/logos-execution-zone/actions/runs/24527976079/job/71703288547?pr=450) much longer. And 50 minutes are spent on setting up sequencer. That means that genesis transactions are executed pretty slowly on ci hardware.

indexer/core/src/block_store.rs
Arjentix: > the hardcoded baseline

We have no hardcoded baseline after genesis, we have only `testnet_initial_state::initial_state()` which is applied before.

> If `testnet_initial_state` ever drifts from what the sequencer's build_genesis_state produces

These things are orthogonal. Sequencer's `build_genesis_state` uses `testnet_initial_state` internally and just constructs genesis transactions and applies them on top of initial state.

> either pinned in config

We can't pin a state hash in a config, that's too verbose, hard to pre-compute and not so usable

>  or read from the genesis block's settlement record

That's bedrock and there is no such info there, just blocks. And block hash depends on data inside block, not on state it was applied on.

Our code is structured in a way that the only entrypoint to change the State is `V03State::apply_state_diff()`. So if initial state of sequencer and indexer are the same and they applied the same transactions in the same order, then they should achieve the same state. This is tested in our indexer tests. In future we'll remove this hardcoded initial state, so that the whole state could be constructed just from blocks. So I think this is the way

nssa/src/validated_state_diff.rs
Arjentix: > There's nothing here that ties "block_id == 1" to the original genesis pinned on L1.

I don't get this part. In this PR I've removed `genesis_block_id` from sequencer's config and always use `1` for it. I think there is no reason to have this configurable and it also makes the code easier.

> A future refactor that passes the wrong block_id (e.g. re-replaying block 1 in a test harness, or an off-by-one somewhere) would silently unlock minting.

Should be quite hard to mess up with this in a real code. For tests, yes, now need to remember that block id `1` is for genesis with specific rules.
My initial implementation was with `is_genesis` flag but then I've found out that I just do this integer check which I could do lower in the code. So not sure it's worth to just do it on upper code.

@schouhy , @Pravdyvy , what do you think of this?

program_methods/guest/src/bin/privacy_preserving_circuit.rs
```diff
@@ -125,10 +126,22 @@ impl ExecutionState {
...
+            // TODO: This looks hacky as hell.
+            // Derive is_genesis: true if any pre_state is a clock account at block 0.
+            // This allows the genesis supply account program to mint balance at genesis.
+            // Security: program execution is verified via env::verify(), and the genesis supply
+            // program itself panics if clock.block_id != 0, so is_genesis cannot be forged.
+            let is_genesis = program_output.pre_states.iter().any(|pre| {
+                CLOCK_PROGRAM_ACCOUNT_IDS.contains(&pre.account_id)
+                    && ClockAccountData::from_bytes(&pre.account.data.clone().into_inner()).block_id
+                        == 0
+            });
```
Pravdyvy: I think we can reduce complexity of this check.
To my understanding `is_genesis` flag exists solely for genesis supply.
So we can modify this check in a way, that flag is activated only if `program_id` is genesis suply and worry not about any possible future changes to clocks, only for correctness of genesis supply calls.

Pravdyvy (approved):
lgtm. Mostly agree, Moudy's comments are valid. Left one suggestion about `is_genesis` flag.

sequencer/core/src/lib.rs
Arjentix: Fixed & added tests

program_methods/guest/src/bin/privacy_preserving_circuit.rs
Arjentix: There are a couple of problems with that:

1. I removed `genesis_supply_account` program in favor of `Mint` instruction of `authenticated_transfer`. I think this is more simple (no need for complex chaincalls) and more incapsulated (only auth transfer manages native balance).
2. What if user just executes this program by its own? Then it would pass your program_id check IMO.
3. That approach requires hardcoding genesis_supply_account program_id or passing it another way, this is not so easy

So what I did instead is just passing `is_genesis` as input and output to/from privacy_preserving_circuit and then check `output.is_genesis` on sequencer side (it should always be `false` for user's txs). Does this make sense for you?

nssa/src/validated_state_diff.rs
Arjentix: Added always explicit `is_genesis` parameter. `block_id` check moved to the upper levels. `1` moved to a constant. See also [this discussion](https://github.com/logos-blockchain/logos-execution-zone/pull/450#discussion_r3113922692)

Pravdyvy: Ok

program_methods/guest/src/bin/privacy_preserving_circuit.rs
Pravdyvy: `genesis_supply_account` had a clock check to prevent execution after genesis. 

But it is not important, as you removed it. Forks for me.

Arjentix: 🍴

sequencer/core/src/lib.rs
```diff
@@ -48,34 +46,68 @@ impl<BP: BlockPublisherTrait> SequencerCore<BP> {
...
+                        .expect("Failed to read state from store");
+                    let genesis_block = store
+                        .get_block_at_id(store.genesis_id())
+                        .expect("Failed to read genesis block from store")
+                        .expect("Genesis block not found in store");
+                    (store, state, genesis_block)
+                }
+                Err(DbError::RocksDbError { error, .. })
+                    if error.kind() == rocksdb::ErrorKind::InvalidArgument
+                        && error.to_string().contains("does not exist") =>
```
schouhy: this is depending on the error string match. Doesn't look good as sequencer logic depends now on rocksdb error display. I'd refactor the code to avoid it.

common/src/transaction.rs
```diff
@@ -98,6 +99,36 @@ impl NSSATransaction {
...
+
+            let faucet_change_is_allowed = *post_program_owner == pre_program_owner
+                && *post_data == pre_data
+                && *post_nonce == pre_nonce
+                && *post_balance >= pre_balance;
+
+            if !faucet_change_is_allowed {
+                return Err(nssa::error::NssaError::InvalidInput(
+                    "Transaction modifies system faucet account".into(),
+                ));
```
schouhy: isn't this check redundant  with the well-behaved program checks?

configs/docker-all-in-one/indexer_config.json
schouhy, line 1: nice

indexer/core/src/block_store.rs
```diff
@@ -150,15 +144,33 @@ impl IndexerStore {
...
+                        )
+                        .context("Failed to execute genesis public transaction")?;
+                } else {
+                    transaction
+                        .clone()
+                        .transaction_stateless_check()?
+                        .execute_check_on_state(
+                            &mut state_guard,
+                            block.header.block_id,
+                            block.header.timestamp,
```
schouhy, line 177: open question. Instead of having this branched logic with the is_genesis arm only getting triggered once, couldn't the indexer have specific logic when syncing from scratch that looks for the genesis block on the bedrcock channel, does the special things for the gensis, and then start a tokio task specific for non genesis blocks? I think that way we could avoid the `is_genesis` flag and branching here.

schouhy (commented):
hey thanks! I'm leaving som comments while continue reviewing

indexer/core/src/block_store.rs
```diff
@@ -150,15 +144,33 @@ impl IndexerStore {
                 "Last transaction in block must be the clock invocation for the block timestamp"
             );
 
+            let is_genesis = block.header.block_id == 1;
```
Pravdyvy, line 153: Lets replace this 1 with `GENESIS_BLOCK_ID`.

sequencer/core/src/block_store.rs
```diff
@@ -19,25 +21,53 @@ pub struct SequencerStore {
...
 
 impl SequencerStore {
+    /// Open existing database at the given location. Fails if no database is found.
+    pub fn open_db(location: &Path, signing_key: nssa::PrivateKey) -> DbResult<Self> {
+        let dbio = Arc::new(RocksDBIO::open(location)?);
+        let genesis_id = dbio.get_meta_first_block_in_db()?;
+        let last_id = dbio.latest_block_meta()?.id;
+
+        info!("Preparing block cache");
+        let mut tx_hash_to_block_map = HashMap::new();
```
Pravdyvy, line 31: It is not really a cache. Most of usual features do not apply.
I suggest renaming it into something more fitting, like `in_memory_part_of_db`.

sequencer/core/src/lib.rs
Pravdyvy: I agree, `open_db` looks up, so it can have more structured error.

Pravdyvy (approved):
lgtm. Minor suggestions.

program_methods/guest/src/bin/authenticated_transfer.rs
schouhy, line 1: I think the authenticated transfer program is not the right fit for the program that handles the supply account. We could have a `faucet.rs` system program that can only be executed by the sequencer (enforced like the clock program) specific for that purpose I think, and dont require is_authorized=true for the sender. Maybe also reused for bridging. And leave the auth-transfer program simple and focused on user authenticated transfer only

sequencer/core/src/config.rs
```diff
@@ -10,19 +10,25 @@ use bytesize::ByteSize;
...
 use logos_blockchain_core::mantle::ops::channel::ChannelId;
+use nssa::AccountId;
 use serde::{Deserialize, Serialize};
-use testnet_initial_state::{PrivateAccountPublicInitialData, PublicAccountPublicInitialData};
 use url::Url;
 
+/// A transaction to be applied at genesis to supply initial balances.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum GenesisTransaction {
```
schouhy: these are not transactions no? I'd rename them to `GenesisAction` or something like that to avoid confusion

program_methods/guest/src/bin/vault.rs
```diff
@@ -0,0 +1,103 @@
...
+            instruction,
+        },
+        instruction_words,
+    ) = read_nssa_inputs::<Instruction>();
+
+    let pre_states_clone = pre_states.clone();
+    let post_states = unchanged_post_states(&pre_states_clone);
+
+    let chained_calls = match instruction {
+        Instruction::Initialize => {
```
schouhy: This Instruction doesnt seem to be used no? Can we remove it?

program_methods/guest/src/bin/vault.rs
```diff
@@ -0,0 +1,103 @@
...
+                ChainedCall::new(
+                    owner.account.program_owner,
+                    vec![owner_vault_for_callee],
+                    &AuthTransferInstruction::Initialize,
+                )
+                .with_pda_seeds(vec![seed]),
+            ]
+        }
+        Instruction::Transfer { amount } => {
+            let [sender, recipient, recipient_vault] = pre_states
```
schouhy: recipient here is only used by its account id. I'd pass that account id as a program instruction data instead of a pre_state. mainly because it's not an account this instruction will operate on.

schouhy (approved):
Hey thanks for the PR! It looks great thanks. I'm leaving some comments in addition to the ones I left yesterday in case you have time and energy to consider. The only part I don't agree with is the refactor of the auth-transfer program to bypass authorization for the specific use case of the system faucet. Native token transfers can be done by any program. Refactoring to use the one program that requires sender authorization feels off to me. I think a cleaner approach would have been to have the system faucet account be owned by a new faucet program only executable by the sequencer like the clock program.

common/src/transaction.rs
```diff
@@ -98,6 +99,36 @@ impl NSSATransaction {
             ));
         }
 
+        let faucet_account_id = nssa::SYSTEM_FAUCET_ACCOUNT_ID;
```
moudyellaz: Nice catch on adding the faucet protection here, this looks solid. One small thought: the check lives at the wrapper layer, and a couple of internal paths intentionally bypass it (`transition_from_public_transaction` directly for genesis and clock). That's totally fine for the current call sites, but it makes the wrapper a soft trust boundary instead of an enforced one. If someone later adds another path that calls `transition_from_public_transaction` directly with externally sourced txs, the faucet becomes drainable. Maybe a short comment near the bypass sites (or here) saying "only sequencer-constructed txs may skip `validate_on_state`"? Just so future us doesn't trip over it.

program_methods/guest/src/bin/vault.rs
```diff
@@ -0,0 +1,103 @@
...
+                .expect("Transfer requires exactly 3 accounts");
+
+            let seed = vault_core::compute_vault_seed(recipient.account_id);
+
+            let mut recipient_vault_for_callee = recipient_vault;
+            recipient_vault_for_callee.is_authorized = true;
+
+            vec![
+                ChainedCall::new(
+                    sender.account.program_owner,
```
moudyellaz, line 52: Tiny thing: the chained call uses `sender.account.program_owner` as the callee, which assumes it speaks `AuthTransferInstruction`. Works perfectly for the genesis case (faucet's owner is auth_transfer) and for normal users, but if a sender ever shows up whose owner uses a different instruction enum, this'll fail with a deserialization error rather than a clean "vault only proxies to auth_transfer" message. Not a security thing, just ergonomics. Could either assert the owner equals `authenticated_transfer_id` upfront, or drop a doc comment saying vault only works with auth-transfer-owned senders. Either's fine.

moudyellaz (approved):
lgtm, thanks. Left 2 minor comments

sequencer/core/src/lib.rs
Arjentix: > Doesn't look good as sequencer logic depends now on rocksdb error display.

Yes, I was also surprised, but we have no another way to catch this error. Look how rocksdb store the error: [here](https://docs.rs/rocksdb/latest/src/rocksdb/lib.rs.html#179-181). They just store a string and match against it, this is strange.

And when no db found we really get `InvalidArgument` kind with this string. I've moved matching into `block_store.rs` to make the code on sequencer cleaner, but still the core thing is the same.

We also have a test `start_from_config()` which checks exactly that behavior, so we're kind of safe

common/src/transaction.rs
Arjentix: I can see why it might seem so, but because of modified condition in authentificated transfer program we need to check it there.

Regarding this check in auth transfer program let's continue in another more related thread.

indexer/core/src/block_store.rs
Arjentix: Hm, I agree, having this check on hot path is not so good. However I'd like to not to make this PR even more complex and this issue seems to be minor now.

So I've created https://github.com/logos-blockchain/logos-execution-zone/issues/476 for it

indexer/core/src/block_store.rs
Arjentix: Fixed

sequencer/core/src/block_store.rs
Arjentix: Hmm, I like the current name of the structure, but I'm agree that this is not a real block cache. So let's just remove the logs.

We can make it to be more like a real in memory part when it will have more data, but currently it's really what the name suggests -- a map from tx to block.

sequencer/core/src/config.rs
Arjentix: Agree. Left this from previous iterations

program_methods/guest/src/bin/vault.rs
Arjentix: Yeah, no practical use right now, removed

program_methods/guest/src/bin/vault.rs
Arjentix: Done

common/src/transaction.rs
Arjentix: I agree, this is not solid and the name `validate_on_state()` doesn't really feel like something sequencer's transactions should skip. I'll add a comment here and will rename this in future PR related to bridge.

Probably even better would be to have more stronger types like `Transaction -validate()-> ValidatedTransaction -execute()-> ()` in future.

program_methods/guest/src/bin/vault.rs
Arjentix: Added the comment and additional `assert!(owner.is_authorized)` just in case

program_methods/guest/src/bin/authenticated_transfer.rs
Arjentix: Haven't thought about it, sounds good, implemented

sequencer/core/src/lib.rs
schouhy: uhm but why didn't we need to catch the error like this before this PR?

nssa/core/src/lib.rs
```diff
@@ -25,6 +25,10 @@ pub mod program;
 #[cfg(feature = "host")]
 pub mod error;
 
+pub const GENESIS_BLOCK_ID: BlockId = 1;
+pub const SYSTEM_FAUCET_ACCOUNT_ID: account::AccountId =
+    account::AccountId::new(*b"/LEZ/SystemFaucetAccount/0000000");
```
schouhy: This is unused, no?

sequencer/core/src/lib.rs
Arjentix: Took me some time to recap this, hah. So the problem was genesis block. It became quite expensive to construct every time. Especially considering that this should happen only once. So I implemented differentiation between existing and new db.

nssa/core/src/lib.rs
Arjentix: Ah, yes, good catch, thanks. Forgot to remove it

sequencer/core/src/lib.rs
schouhy: ok I still think we should somehow find a workaround to the string matching hah sorry. The error variant DatabaseNotFound is used by the sequencer to fallback to create from genesis. Why not revert the logic and have the sequencer first check whether the file exists. If not, goes with creation from genesis state, otherwise tries to open it and raises whatever error rocksdb throws if it can't open it

Arjentix: Hmm, so we can just check if directory exists, but not it's internal files because we have no guarantees how rocksdb organizes them... But checking directory should also work, yes. Will implement

Arjentix: Done!

schouhy: thanks!

PR #451, opened 2026-04-17
feat(wallet): add keycard support for public tx for auth-transfer

jonesmarvin8:
## 🎯 Purpose

Integrates Keycard for  public transactions for `auth-transfers`.

## ⚙️ Approach

- [X] Keycard python class
- [X] Rust wrapper for Keycard python class
- [X] Add Wallet command to check availability of Keycard and load new keys
- [X] Update Wallet `auth-transfer` and `pinata` to use Keycard

**Updated approach based on PR 450's updates**
- [X] Use `CliMention` to simplify program facade inputs for keycard
- [X] Introduce `SigningGroup` to support local wallet and keycard signatures, and allow for a single python session for keycard signatures. 

## 🧪 How to Test

Follow [setup](https://www.notion.so/Setup-34b8f96fb65c80bbbf2ced7aad3d3aa6) for Keycard before this.

In root:
- Run `wallet_with_keycard.sh` to install Wallet and the necessary python packages.
- With Keycard inserted, run `keycard_tests.sh` (in root)
- Keycard wallet commands must be ran in the appropriate python environment; root of repo directory run `source venv/bin/activate` (this command is already included in `wallet_with_keycard.sh`)

## 🔗 Dependencies

PR 452 (and PR 457) must be merged first.

## 🔜 Future Work

- PR 461 addresses Token program and privacy commands.
- **New:** Add integration tests for keycard using `MagicMock`.

## 📋 PR Completion Checklist

- [X] Complete PR description
- [X] Add/update tests


keycard_wallet/src/python_path.rs
```diff
@@ -0,0 +1,39 @@
+use std::{env, path::PathBuf};
+
+use pyo3::{prelude::*, types::PyList};
+
+/// Adds the project's `python/` directory and venv site-packages to Python's sys.path.
+pub fn add_python_path(py: Python<'_>) -> PyResult<()> {
+    let current_dir = env::current_dir().expect("Failed to get current working directory");
+
+    let paths_to_add: Vec<PathBuf> = vec![
+        current_dir.join("python"),
```
moudyellaz: two issues in the python integration:
`current_dir().join("python")` puts cwd-relative `python/` at `sys.path[0]`. after `cargo install --path wallet --force` the binary is global, so running it from any directory containing a malicious `python/keycard_wallet.py` executes that file in-process with PIN and mnemonic access.

`wallet_with_keycard.sh:14` does `git clone --branch lee-schnorr` of `bitgamma/keycard-py` unpinned. whoever pushes there controls signing for every install.

both need fixing before this lands. resolve the path from `current_exe()`, and either vendor `keycard-py` properly or pin a commit hash and verify in CI.

python/keycard_wallet.py
```diff
@@ -0,0 +1,137 @@
...
+from ecdsa import VerifyingKey, SECP256k1  
+
+from keycard.keycard import KeyCard
+
+from mnemonic import Mnemonic  
+from keycard import constants  
+  
+import keycard
+
+PIN = '123456'
```
moudyellaz: let's drop these. `setup_communication(pin=PIN, …)` and `load_mnemonic(mnemonic=DEFAULT_MNEMONIC, …)` use them as defaults, so a future caller forgetting an arg silently signs with `123456` or loads `"fashion degree mountain..."`. if test fixtures need a known mnemonic, move it to a test-only file.

wallet/src/cli/keycard.rs
```diff
@@ -0,0 +1,84 @@
...
+
+/// Represents generic chain CLI subcommand.
+#[derive(Subcommand, Debug, Clone)]
+pub enum KeycardSubcommand {
+    Available,
+    Load {
+        #[arg(short, long)]
+        mnemonic: Option<String>,
+        #[arg(short, long)]
+        pin: Option<String>,
```
moudyellaz: `--pin` on argv leaks to shell history, ps, xtrace, and stays in heap unzeroed. for a hardware key the PIN is the whole moat once the card is taken. let's read it from a TTY prompt (`rpassword`) or env var, and store it in `Zeroizing<String>`. same applies to the other subcommands taking `pin: Option<String>` (`programs/native_token_transfer.rs:40,73`, `pinata.rs:31`, `account.rs:42`, `token.rs:80,111`).

keycard_wallet/src/lib.rs
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    pub fn get_public_key_for_path(&self, py: Python, path: &str) -> PyResult<PublicKey> {
+        let public_key: Vec<u8> = self
+            .instance
+            .bind(py)
+            .call_method1("get_public_key_for_path", (path,))?
+            .extract()?;
+
+        let public_key: [u8; 32] = public_key.try_into().expect("Expect 32 bytes");
```
moudyellaz: wrong PIN, removed card, secure-channel timeout, all the everyday failure modes hit `.expect()` and panic the tokio worker. these aren't invariants. let's propagate `PyResult` (or a typed `KeycardError`) up to the CLI handler and reserve `expect` for things that can't fail.

(also `Expect a valid public key1` / `key2` look like leftover debug.)

same applies to the matching expects in `program_facades/native_token_transfer/public.rs:60,72,113,118,122`.

wallet/src/program_facades/native_token_transfer/public.rs
```diff
@@ -29,20 +32,46 @@ impl NativeTokenTransfer<'_> {
...
         let program_id = Program::authenticated_transfer_program().id();
 
-        let sign_ids = self.0.filter_owned_accounts(&[from, to]);
-
         // Fetch nonces for both accounts unconditionally
-        let nonces = self
+        let mut nonces = self
             .0
-            .get_accounts_nonces(account_ids.clone())
+            .get_accounts_nonces(vec![from])
```
moudyellaz: silent semantic change unrelated to keycard. pre-PR fetched both `[from, to]` nonces unconditionally; now `from` only, with `to` extended only if its key is local. when `to` is foreign (the normal case), `account_ids.len() == 2` but `nonces.len() == 1` and `Message::try_new(...).unwrap()` either panics or builds a malformed tx.

let's revert to `get_accounts_nonces(account_ids.clone())` and put any conditional-nonce logic in its own PR with a foreign-recipient test.

keycard_wallet/src/lib.rs
```diff
@@ -0,0 +1,143 @@
...
+
+        let signature: [u8; 64] = py_signature.try_into().map_err(|vec: Vec<u8>| {
+            PyErr::new::<pyo3::exceptions::PyValueError, _>(format!(
+                "Invalid signature length: expected 64 bytes, got {} (bytes: {:02x?})",
+                vec.len(),
+                vec
+            ))
+        })?;
+
+        Ok(Signature { value: signature })
```
moudyellaz: the 64 bytes from the card go straight into `Signature { value }` with no rust-side verify. if `keycard_wallet.py` (or `keycard-py`) ever returns a wrong signature, the only signal is a sequencer rejection, nonce burned, message digest leaked.

cheap to add: `VerifyingKey::from_bytes(pub_key.value())?.verify_prehash(message, &sig)?` after sign, refuse on failure.

keycard_wallet/src/lib.rs
```diff
@@ -0,0 +1,143 @@
...
+            .call_method1("get_public_key_for_path", (path,))?
+            .extract()?;
+
+        let public_key: [u8; 32] = public_key.try_into().expect("Expect 32 bytes");
+
+        Ok(PublicKey::try_new(public_key).expect("Expect a valid public key1"))
+    }
+
+    #[must_use]
+    pub fn get_public_key_for_path_with_connect(pin: &str, path: &str) -> PublicKey {
```
moudyellaz: each `_with_connect` does select/pair/open_secure_channel/verify_pin/.../unpair. with `get_public_key_for_path_with_connect` + `sign_message_for_path_with_connect`, that's two PIN verifications per send. the card locks after 3 bad PINs.

let's collapse to one session: connect → get_pubkey → sign → disconnect. or longer-term, hold pairing state across ops in a process-global mutex with `keycard load` setting it up once.

jonesmarvin8: Good point. I can collapse this into a single session. Thanks!

Ideally, I would hold pairing across ops state. Unfortunately, each wallet command in CLI is treated as its own instance. So, I cannot retain the python state in memory (the original approach I wanted to use). Storing the python class on disk seems problematic to me as a keycard may be removed between usage.

nssa/src/encoding/public_transaction.rs
```diff
@@ -13,6 +14,6 @@ impl PublicTransaction {
     }
 
     pub fn from_bytes(bytes: &[u8]) -> Result<Self, NssaError> {
```
moudyellaz, line 15: Heads-up, looks like a copy-paste from the `to_bytes` directly above slipped in here: `from_bytes` was changed from `?` to `.expect(...)`. The signature still says `Result`, but malformed input now crashes the process instead of returning `Err`. Could we put the `?` back?

wallet/src/program_facades/native_token_transfer/public.rs
```diff
@@ -29,20 +33,39 @@ impl NativeTokenTransfer<'_> {
...
+            if let Some(to_key_path) = to_key_path {
+                let (to_sig, to_pk) = KeycardWallet::sign_message_for_path_with_connect(
+                    &pin,
+                    to_key_path,
+                    &msg_hash,
+                )?;
+                WitnessSet::from_list(&message, &[from_sig, to_sig], &[from_pk, to_pk])
+                    .map_err(ExecutionFailureKind::TransactionBuildError)?
+            } else {
+                WitnessSet::from_list(&message, &[from_sig], &[from_pk])
```
moudyellaz: Thanks for the keycard refactor here, the 2-keypath flow was lovely to drive on the keycard. One small thing that came up during testing: when only `--from-key-path` is set and `--to` is a plain id, the wallet builds a witness with one signature but `nonces` still has two entries, so the sequencer rejects with `InvalidInput("Mismatch between number of nonces and signatures/public keys")` and the tx silently never lands.

Easiest fix is probably to also sign the recipient with the wallet-stored key when present (mirroring the non-keycard branch), so the single-keycard case still produces two sigs.

python/keycard_wallet.py
```diff
@@ -0,0 +1,125 @@
...
+from ecdsa import VerifyingKey, SECP256k1  
+
+from keycard.keycard import KeyCard
+
+from mnemonic import Mnemonic  
+from keycard import constants  
+  
+import keycard
+
+DEFAULT_PAIRING_PASSWORD = "KeycardDefaultPairing"
```
moudyellaz, line 13: Tiny thing: every wallet pairs to the card using the same hardcoded string, with no override path through `setup_communication`. That means anyone holding this wrapper (or any card it has touched) can re-pair to the card and try PIN guesses subject to the 3-attempt counter. Could the pairing secret be sourced from an env var or a generated per-install value, the way Status's own keycard tooling does?

wallet/src/cli/keycard.rs
```diff
@@ -0,0 +1,81 @@
...
+    cli::{SubcommandReturnValue, WalletSubcommand, read_pin},
+};
+
+/// Represents generic chain CLI subcommand.
+#[derive(Subcommand, Debug, Clone)]
+pub enum KeycardSubcommand {
+    Available,
+    Load {
+        #[arg(short, long)]
+        mnemonic: Option<String>,
```
moudyellaz: One for the mnemonic loading: it's currently accepted as a CLI flag, which means it lands in `argv`, `ps` output, and the user's shell history (`~/.zsh_history`, `~/.bash_history`). The PIN is already read interactively via `rpassword` in `helperfunctions.rs`, would be great to do the same for the mnemonic and wrap it in `Zeroizing<String>` until it's handed to the card.

keycard_wallet/src/lib.rs
```diff
@@ -0,0 +1,149 @@
...
+        let pub_key = self.get_public_key_for_path(py, path)?;
+        if !sig.is_valid_for(message, &pub_key) {
+            return Err(PyErr::new::<pyo3::exceptions::PyValueError, _>(
+                "keycard returned a signature that does not verify against its own public key",
+            ));
+        }
+        Ok((sig, pub_key))
+    }
+
+    pub fn sign_message_for_path_with_connect(
```
moudyellaz, line 171: While running through `keycard_tests.sh` end-to-end the card wedged into an "unpowered" state (0x80100067) after ~8 wallet operations, recovered by physically reseating. The pattern is: each `wallet keycard <cmd>` opens a fresh PC/SC session, pairs (uses a pairing slot), opens a secure channel, verifies PIN, does its work, then unpairs/disconnects. Anything that goes wrong between pair and unpair (panic, kill, sometimes just slow teardown) leaks a slot. The card only has 5.

Combined with the per-signature reopen (two-sig flows like `auth-transfer send --from-key-path A --to-key-path B` actually pair twice for one CLI command), a session of normal use easily uses up the budget.

Could the wallet hold a single keycard session for the lifetime of a CLI invocation, sign N times within it, and unpair once at the end? That would also help the per-command latency, which is currently around 30-60s per signature because of the full pair+secure-channel reopen.

wallet/src/cli/keycard.rs
```diff
@@ -0,0 +1,81 @@
...
+                    if is_connected {
+                        println!("\u{2705} Keycard is now connected to wallet.");
+                    } else {
+                        println!("\u{274c} Keycard is not connected to wallet.");
+                    }
+
+                    drop(
+                        wallet.load_mnemonic(
+                            py,
+                            &mnemonic.expect(
```
moudyellaz: Small thing: `.expect()` on the Python results crashes the wallet binary with a panic when something goes wrong (no card inserted, secure channel error, missing flag). One of those is `mnemonic.expect(...)` on `Option<String>`, which fires if a user just runs `wallet keycard load` without `--mnemonic`. Could these become `?` with a clean `anyhow::Error`? Also worth marking `mnemonic` as `required = true` on the clap attr, or reading it from stdin.

wallet_with_keycard.sh
```diff
@@ -0,0 +1,21 @@
+cargo install --path wallet --force
```
moudyellaz: One last thing on the setup side, mostly to make the next reviewer's life easier: a few small gaps tripped me up trying to run `keycard_tests.sh` end-to-end.

- The two `.cap` files (`math.cap` + the LEE keycard cap) aren't in the repo or anywhere public. Even just dropping them as a release asset on this repo would unblock anyone outside the immediate team.
- `wallet_with_keycard.sh` creates two separate venvs (one at the repo root with `pyscard/mnemonic/ecdsa/pyaes`, one inside `python/keycard-py/` with `keycard-py -e`). The wallet binary needs both deps in the same venv at runtime, so out of the box `wallet keycard available` blows up with `ModuleNotFoundError: No module named 'keycard'`. Installing `pip install -e python/keycard-py` into the outer venv fixes it; could the script just do that?
- On a freshly-flashed card, `wallet keycard load` returns `❌ Keycard is not connected` with no hint because the LEE applet needs `card.init(pin, puk, pairing_secret)` first. Either `wallet keycard load` could detect `is_initialized == False` and run init itself, or a setup README could mention the prerequisite.

moudyellaz (changes requested):
Thanks for the PR, I have tested it with the keycard and left some comments.

wallet_with_keycard.sh
jonesmarvin8: - Added cap files in the repo. Added keycard.md (portions of it) from 461 to this PR and updated directions to include cap files in repo.
- Fixed script to create a single venv.
- Freshly flashed card. Added a initialize command, and a catch for card connection to direct users to initialize. Opted to direct users with the command so it doesn't force initialization on users.

wallet_with_keycard.sh
```diff
@@ -0,0 +1,10 @@
...
```
schouhy, line 3: let's have the shebang line `#!/bin/bash` to avoid problems with other shells. Same with other scripts

docs/LEZ testnet v0.1 tutorials/keycard.md
```diff
@@ -0,0 +1,148 @@
...
+- Keycard (Blank) - a Keycard, directly, from Keycard.tech cannot (currently) be updated to support LEE.
+- Smartcard reader
+
+### Firmware installation
+Installation:
+
+1. Install math applet on your keycard; this process only needs to be done once. In the root of repo:
+    ```
+    cd python/keycard_applets
+    java -jar gp.jar --key c212e073ff8b4bbfaff4de8ab655221f --load math.cap
```
schouhy, line 21: the step downloading java and gp.jar is missing

docs/LEZ testnet v0.1 tutorials/keycard.md
```diff
@@ -0,0 +1,148 @@
...
+
+### Keycard
+
+| Command                           | Key-path options | Description                                                              |
+|-----------------------------------|------------------|--------------------------------------------------------------------------|
+| `wallet keycard available`        | —                | Checks whether a Keycard reader and card are accessible                  |
+| `wallet keycard load`             | —                | Loads a mnemonic phrase onto the Keycard                                 |
+| `wallet keycard get-private-keys` | `--key-path`     | Retrieves private account keys (nsk, vsk) for the given BIP32 path      |
+
+1. Check keycard availability
```
schouhy, line 75: I had keycard desktop open while running keycard commands and they failed to connect to the keycard. I had to close the desktop to get things working. Maybe worth mentioning

wallet/src/cli/keycard.rs
```diff
@@ -0,0 +1,72 @@
...
+
+                    if is_connected {
+                        println!("\u{2705} Keycard is now connected to wallet.");
+                    } else {
+                        println!("\u{274c} Keycard is not connected to wallet.");
+                    }
+
+                    drop(wallet.load_mnemonic(py, &mnemonic));
+
+                    drop(wallet.disconnect(py));
```
schouhy: This is running load_mnemonic and disconnect independently of the value of `is_connected`. I think it should only when it is connected

docs/LEZ testnet v0.1 tutorials/keycard.md
```diff
@@ -0,0 +1,148 @@
...
+```bash
+wallet keycard available
+
+# Output:
+✅ Keycard is available.
+```
+
+2. Load a mnemonic phrase
+```bash
+wallet keycard load --mnemonic "fashion degree mountain wool question damp current pond grow dolphin chronic then"
```
schouhy: this syntax is outdated no?

wallet/src/cli/account.rs
```diff
@@ -225,17 +222,22 @@ impl WalletSubcommand for AccountSubcommand {
...
                     &wallet_core.storage.labels,
                     &wallet_core.storage.user_data,
+                    key_path.as_deref(),
                 )?;
                 let (account_id_str, addr_kind) = parse_addr_with_privacy_prefix(&resolved)?;
 
                 let account_id: nssa::AccountId = account_id_str.parse()?;
 
+                // Add account id to the display for keycard users.
+                log::info!("Account Id: {resolved}");
```
schouhy: the rest of the information is printed and not logged.
```suggestion
                println!("Account Id: {resolved}");
```

wallet/src/cli/keycard.rs
schouhy: why these explicit drops? If it's to avoid the compiler complaining about the unused result, I think `let _ = ` does the same and it's more idiomatic

python/keycard_wallet.py
```diff
@@ -0,0 +1,125 @@
...
+            pairing_index, pairing_key = self.card.pair(password)
+            self.pairing_index = pairing_index
+
+            self.card.open_secure_channel(pairing_index, pairing_key)  
+            self.card.verify_pin(pin)
+
+            return True
+        except Exception as e:  
+            print(f"Error: {e}")  
+            return False
```
schouhy: in case of exception here we should unpair no?

docs/LEZ testnet v0.1 tutorials/keycard.md
schouhy: yes, seems like the rest is too out of sync with the current version of the code

python/keycard_applets/LEE_keycard.cap
schouhy, line 1: There should be documentation on what these cap files are and how they were generated.

wallet/src/cli/programs/native_token_transfer.rs
```diff
@@ -66,6 +65,12 @@ pub enum AuthTransferSubcommand {
         /// amount - amount of balance to move.
         #[arg(long)]
         amount: u128,
+        /// `from_key_path` (alternative to --from) uses Keycard.
```
schouhy: maybe a mouthful but it'd be nice to have the arg be explicit about it using keycard, something like `--from-keycard-key-path`.
or even better change the approach to call it like `wallet auth-transfer init --account-id <account-id> --use-keycard`. Problem is that there's currently no map from account ids to chain paths for this to resolve it.

schouhy (commented):
Hey thanks for the PR! It works, nice work! I think it's a good starting point for having keycard integration. Left some comments
Here are also a few things I don't agree with the approach to consider now or in the future.
- Wallet flow changes with keycard options in the sense that accounts are not created with "wallet account new public/private` type of command.  Also, program commands expect account id when using local storage and chain path when using keycard. I think that is confusing and the design should be such that keycard integration is more transparent.
- I'd move all files related to the keycard to a dedicated directory (except the ones that go in the wallet), including the python code. Let's remove the `python` directory from root.
- I'd have added a field to the wallet config indicating the type of key storage backend to be used, and have commands  be uniform for both backends. This could imply changing how local storage is designed too and move to something closer to how keycard manages storage.

wallet/src/program_facades/native_token_transfer/public.rs
```diff
@@ -66,16 +90,32 @@ impl NativeTokenTransfer<'_> {
...
+            WitnessSet::from_list(&message, &[signature], &[pub_key])
+                .map_err(ExecutionFailureKind::TransactionBuildError)?
+        } else {
+            let signing_key = self.0.storage.user_data.get_pub_account_signing_key(from);
+
+            let Some(signing_key) = signing_key else {
+                return Err(ExecutionFailureKind::KeyNotFoundError);
+            };
+
+            WitnessSet::for_message(&message, &[signing_key])
```
schouhy: Already mentioned on discord. But this kind of branching will only get worse with time. This requires a better code abstraction

python/keycard_wallet.py
```diff
@@ -0,0 +1,125 @@
...
+            #TODO: this appears to be the issue.
+            return True
+        except Exception as e:
+            print(f"Error during disconnect: {e}")
+            return False
+
+    def disconnect(self) -> bool:
+        try:
+            if not self.card.is_secure_channel_open:
+                return None
```
schouhy:
```suggestion
                return True
```

python/keycard_wallet.py
```diff
@@ -0,0 +1,125 @@
...
+            mnemo = Mnemonic("english")  
+            seed = mnemo.to_seed(mnemonic)  
+
+            # Load the LEE seed onto the card  
+            result = self.card.load_key(  
+                key_type = constants.LoadKeyType.LEE_SEED,  
+                lee_seed = seed  
+            )
+
+            #TODO: this appears to be the issue.
```
schouhy: what issue?

python/keycard_wallet.py
```diff
@@ -0,0 +1,125 @@
...
+            return True
+        except Exception as e:
+            print(f"Error during disconnect: {e}")
```
schouhy: I think this is not the right message to print here

python/keycard_wallet.py
jonesmarvin8: A comment that I forgot to delete after error tracking. Thanks for pointing it out!

keycard_tests.sh
schouhy, line 1: I think this script should be turn into an integration test. Otherwise CI won't test this

jonesmarvin8: Since this script depends on a smartcard reader and a keycard...how would integration test handle this?

schouhy: one way is to mock the KeycardWallet with something that behaves like it when the keycard is connected

schouhy: check out [tests](https://github.com/mmlado/keycard-py/blob/af2e600bfea86936b5ea5a5136b205ebbc52e1fa/tests/test_keycard.py) in keycard-py they seem to test it like this

jonesmarvin8: Sure, we know the mnemonic phrase. So, we can just do a catch "if keycard is not present then..."

schouhy: > Sure, we know the mnemonic phrase. So, we can just do a catch "if keycard is not present then..."

ah I didn't get this. Where would you add that catch?

jonesmarvin8: I haven't added the catch yet. Just brainstorming. I'll look at the MockMagic approach because that's likely more reasonable.

wallet/src/cli/programs/native_token_transfer.rs
```diff
@@ -257,6 +284,10 @@ pub enum NativeTokenTransferProgramSubcommand {
         /// amount - amount of balance to move.
         #[arg(long)]
         amount: u128,
+        #[arg(long)]
+        from_key_path: Option<String>,
+        #[arg(skip)]
+        to_key_path: Option<String>,
```
schouhy: Instad of adding more optional parameters a better approach is having the `from` be of an enum type that considers all different variants. Check out @Arjentix approach in PR #450 (see [here](https://github.com/logos-blockchain/logos-execution-zone/blob/4f824922283969502db05c9d10a0d55d5aa8d686/wallet/src/cli/account.rs#L28) and [here](https://github.com/logos-blockchain/logos-execution-zone/blob/4f824922283969502db05c9d10a0d55d5aa8d686/wallet/src/cli/mod.rs#L117-L120)) that already merges the label and the account id arguments

python/keycard_wallet.py
jonesmarvin8: Good catch. I reworked the logic to handle this.

python/keycard_wallet.py
jonesmarvin8: The PairingPassword is designed for protection against MITM attacks. Keycard team views this as unfeasible for Keycard. Additionally, it's possible that the secure channel setup will be simplified in the future to not rely use pairing password. Their recommendation is to leave pairing password alone.

Though, if you think it'd be beneficial to give users the option to setup their own pairing password, I can add this.

wallet/src/cli/keycard.rs
jonesmarvin8: This issue fixed by using zeroize for mnemonic phrase.

keycard_wallet/src/lib.rs
jonesmarvin8: We can maintain the same pairing parameters, but it'll have to reconnect per wallet command. This is because each wallet command is a separate instance of the wallet program. So, we can save pairing parameters on disk.

I originally thought about this, but didn't go this route since connection would need to be established each time anyways. I'm now saving the parameters since any speed up is desirable.

I've updated to only do one connection for multiple signatures.

Note: 5 slot issue has been fixed on the firmware level. So, this is less of a concern now due to the introduction of an ephemeral slot. But, the other concerns are valid.

wallet/src/program_facades/native_token_transfer/public.rs
jonesmarvin8: Thanks for catching this!

wallet/src/cli/programs/native_token_transfer.rs
jonesmarvin8: Adapted a similar approach; this will require some minor work to make sure #450 and this version mesh, but should be minor.

wallet/src/cli/programs/native_token_transfer.rs
jonesmarvin8: This has actually been addressed by the enum suggestion from 450.

docs/LEZ testnet v0.1 tutorials/keycard.md
jonesmarvin8: Should be fixed now; with all of the changes to keycard commands.

keycard_tests.sh
jonesmarvin8: Adding this to the integration tests might be a bit larger of a task than we want to get in for this specific PR. Perhaps, it can be a follow up PR (soon)? Accessing MagicMock requires some annoyance when dealing with the Rust wrapper; but I do agree that we should include it in the integration tests at some point.

wallet/src/signing.rs
```diff
@@ -0,0 +1,114 @@
...
+use nssa::{AccountId, PrivateKey, PublicKey, Signature};
+
+use crate::{WalletCore, cli::CliAccountMention};
+
+/// Groups transaction signers by type to minimise Python GIL acquisition.
+///
+/// Local signers are signed in pure Rust; all keycard signers share a single Python session
+/// with one `connect` / `close_session` pair.
+#[derive(Default)]
+pub struct SigningGroups {
```
schouhy, line 12:
```suggestion
pub struct SigningGroup {
```

wallet/src/signing.rs
```diff
@@ -0,0 +1,114 @@
+use anyhow::Result;
+use keycard_wallet::{KeycardWallet, python_path};
+use nssa::{AccountId, PrivateKey, PublicKey, Signature};
+
+use crate::{WalletCore, cli::CliAccountMention};
+
+/// Groups transaction signers by type to minimise Python GIL acquisition.
+///
+/// Local signers are signed in pure Rust; all keycard signers share a single Python session
+/// with one `connect` / `close_session` pair.
```
schouhy, line 10: I'd add the purpose of this struct in the docs too. It wasn't clear to me what this was from this description.

keycard_tests.sh
schouhy, line 1: let's move this file to keycard_wallet

wallet_with_keycard.sh
schouhy, line 1: let's move this file to keycard_wallet

schouhy (approved):
lgtm thanks. left some minor comments in case there's time to consider

keycard_wallet/python/keycard_wallet.py
```diff
@@ -0,0 +1,164 @@
...
+
+from keycard.keycard import KeyCard
+
+from mnemonic import Mnemonic
+from keycard import constants
+
+import keycard
+import secrets
+
+DEFAULT_PAIRING_PASSWORD = "KeycardDefaultPairing"
```
moudyellaz, line 13: This one didn't move this round. DEFAULT_PAIRING_PASSWORD is still a hardcoded module-level string used as the default in setup_communication, so any wallet (or card this wrapper has touched) can pair without a secret and PIN becomes the only moat. Easiest fix: drop the default and read from KEYCARD_PAIRING_PASSWORD env or a prompt, same shape as --pin now. Fine to defer.

keycard_wallet/src/lib.rs
```diff
@@ -0,0 +1,230 @@
...
+    pub fn close_session(&self, py: Python<'_>) -> PyResult<bool> {
+        self.instance
+            .bind(py)
+            .call_method0("close_session")?
+            .extract()
+    }
+
+    /// Connect using a stored pairing if available, falling back to a fresh pair.
+    /// Saves any newly established pairing to disk.
+    pub fn connect(&self, py: Python<'_>, pin: &str) -> PyResult<()> {
```
moudyellaz, line 86: Still hitting the wedge I flagged earlier: card lands in 0x80100067 ("unpowered") after ~8 wallet ops in a row and only recovers on physical reseat. connect() has no retry/reconnect path, every error from setup_communication_with_pairing or setup_communication propagates straight up. A small reset-and-retry-once on a transport error here would smooth the test/dev loop a lot. Fine as a follow-up if you'd rather merge first.

moudyellaz (approved):
lgtm, both my old concerns and the new round have been addressed. Three carryovers (pairing-password default, card wedging, broader .expect() pattern in cli/keycard.rs) noted as non-blockers, fine to land as-is or pick up in a follow-up. Thanks for the multiple rounds on this.

jonesmarvin8:
> lgtm, both my old concerns and the new round have been addressed. Three carryovers (pairing-password default, card wedging, broader .expect() pattern in cli/keycard.rs) noted as non-blockers, fine to land as-is or pick up in a follow-up. Thanks for the multiple rounds on this.

I'll add these to PR 461 as a comment. And go ahead and get this PR ready to merge. This way our attention can be turned to 461.

PR #452, opened 2026-04-20
Refactor wallet for public account changes

jonesmarvin8:
## 🎯 Purpose

Minor refactoring Wallet's handling of public accounts in preparation of Keycard integration.

## ⚙️ Approach

- [X] `KeyTreePublic` to optional storage
- [X] Reorganization for `WitnessSet` generation (specific for native token transactions).

## 🧪 How to Test

All pre-existing tests should be unaffected.

## 🔗 Dependencies

Uses [PR 457](https://github.com/logos-blockchain/logos-execution-zone/pull/457) as a base to have a single base for [PR 451](https://github.com/logos-blockchain/logos-execution-zone/pull/451) to build off of.

## 🔜 Future Work

N/A

## 📋 PR Completion Checklist

- [X] Complete PR description
- [X] Add/update tests


wallet/src/lib.rs
```diff
@@ -550,4 +547,41 @@ impl WalletCore {
...
+            private_keys.push(key);
+        }
+
+        Ok(nssa::public_transaction::WitnessSet::for_message(
+            message,
+            &private_keys,
+        ))
+    }
+
+    pub fn sign_privacy_message(
```
moudyellaz, line 602: The function never produces `Err`. Caller does `.expect("Expect a valid witness set")` on an infallible `Result`, so the wrapper is dead. Either drop the `Result` and return `WitnessSet` directly, or add the actual failure modes you anticipate when Keycard signing is wired in (e.g. `KeycardSignatureRefused`, `KeycardTimeout`).

If the `Result` is there to keep the signature symmetric with `sign_public_message`, that's worth a one-line comment explaining the intent.

wallet/src/chain_storage.rs
```diff
@@ -56,7 +56,9 @@ impl WalletChainStore {
             .expect("Malformed persistent account data, must have private root");
 
         let mut public_tree = KeyTreePublic::new_from_root(match public_root {
-            PersistentAccountData::Public(data) => data.data,
+            PersistentAccountData::Public(data) => {
+                data.data.expect("Expect valid public account keys")
```
moudyellaz: The `.expect("Expect ...")` messages here and at `chain_storage.rs:75`, `wallet/src/lib.rs:418`, and `wallet/src/program_facades/native_token_transfer/public.rs:56` are tautological. If any of these panics fire in production, the operator sees "Expect a valid witness set" with no hint about what went wrong, who provided invalid input, or what the recovery is. Worth either:

- Drop the `Result` so there's no `.expect` to write a message for (overlaps with the `sign_privacy_message` comment),
- Or replace each message with a sentence describing the upstream invariant that was broken, e.g. `expect("acc_manager.public_account_auth() never returns mismatched key counts for a transfer message")`.

wallet/src/program_facades/native_token_transfer/public.rs
```diff
@@ -26,22 +26,17 @@ impl NativeTokenTransfer<'_> {
...
-            let from_signing_key = self.0.storage.user_data.get_pub_account_signing_key(from);
-            let Some(from_signing_key) = from_signing_key else {
-                return Err(ExecutionFailureKind::KeyNotFoundError);
-            };
-            private_keys.push(from_signing_key);
-
             let to_signing_key = self.0.storage.user_data.get_pub_account_signing_key(to);
-            if let Some(to_signing_key) = to_signing_key {
-                private_keys.push(to_signing_key);
+            if let Some(_to_signing_key) = to_signing_key {
```
moudyellaz: The lookup is used purely as an existence check (`_to_signing_key` is bound and immediately discarded), then `sign_public_message` re-fetches the same key by calling `get_pub_account_signing_key` again. Two reads of the same map for one transfer.

I would either:

1. Add a `has_pub_account_signing_key(account_id) -> bool` predicate on `UserData` and use that here, so the existence check doesn't pretend to need the key.
2. Have `sign_public_message` silently skip account_ids whose key isn't found, then push `to` unconditionally; this also collapses the existence check upstream.

wallet/src/lib.rs
```diff
@@ -550,4 +547,41 @@ impl WalletCore {
     pub const fn config_overrides(&self) -> &Option<WalletConfigOverrides> {
         &self.config_overrides
     }
+
+    pub fn sign_public_message(
+        wallet: &Self,
```
moudyellaz: The first parameter is `wallet: &Self`, which makes the function callable only via `WalletCore::sign_public_message(self.0, ...)` (as the caller in `public.rs:55` does). Rust usually suggests to use `pub fn sign_public_message(&self, ...)`, which lets both `WalletCore::sign_public_message(wallet, ...)` and `wallet.sign_public_message(...)` work. The body uses `wallet.storage.user_data....` which becomes `self.storage.user_data....` with no other change.

Worth flipping to `&self` unless there's something I'm missing.

nssa/src/privacy_preserving_transaction/witness_set.rs
```diff
@@ -13,6 +13,8 @@ pub struct WitnessSet {
 
 impl WitnessSet {
     #[must_use]
+    // TODO: this generates signatures.
```
moudyellaz: I would remove this and replace it with `// TODO: swap for Keycard signing path`.

wallet/src/config.rs
```diff
@@ -23,7 +23,7 @@ use url::Url;
 pub struct PersistentAccountDataPublic {
     pub account_id: nssa::AccountId,
     pub chain_index: ChainIndex,
-    pub data: ChildKeysPublic,
+    pub data: Option<ChildKeysPublic>, // `None` when Keycard is used.
```
moudyellaz, line 26: The flip to `Option<ChildKeysPublic>` changes the serde shape of `PersistentAccountDataPublic`. Worth a small round-trip test (`Some` and `None`), and if old wallet stores need to load, a fixture confirming the legacy shape still deserializes.

wallet/src/lib.rs
jonesmarvin8: Removed. Error checking for keycard can be added later if necessary.

wallet/src/lib.rs
jonesmarvin8: Thanks.

wallet/src/program_facades/native_token_transfer/public.rs
jonesmarvin8: I opted to use suggestion 2 (new function `wallet::lib::filter_owned_accounts`) to handle this.

wallet/src/chain_storage.rs
moudyellaz: `chain_storage.rs:75` still has the original `expect("Expect valid public account keys")`. Worth the same descriptive treatment you gave to `public.rs:45`.

nssa/src/privacy_preserving_transaction/witness_set.rs
moudyellaz: Line 17 (`// However. we may need to get signatures from Keycard.`) should be removed since line 16 already covers the intent.

moudyellaz (approved):
lgtm, thanks for addressing my comments. Left 2 minor comments.

moudyellaz (approved):

nssa/src/public_transaction/witness_set.rs
```diff
@@ -8,6 +8,21 @@ pub struct WitnessSet {
...
+        let signatures_and_public_keys = signatures
+            .iter()
+            .zip(pub_keys.iter())
+            .map(|(sig, key)| (sig.clone(), key.clone()))
+            .collect();
+
+        Self {
+            signatures_and_public_keys,
+        }
+    }
```
schouhy: This seems unused. Let's remove it. Also, I think it's not a good idea to provide constructors that don't enforce valid object creation. I mean, this method would allow to create a WitnessSet instance with invalid signatures for the given public keys. In case such a method is needed for the rest I'd have it return a Result and return Err in case signatures are not valid for the give public keys, or there's a mismatch in the number of elements (instead of panicking)

wallet/src/program_facades/native_token_transfer/public.rs
```diff
@@ -22,52 +22,35 @@ impl NativeTokenTransfer<'_> {
...
+            .0
+            .get_accounts_nonces(account_ids.clone())
+            .await
+            .map_err(ExecutionFailureKind::SequencerError)?;
+
+        let message = Message::try_new(program_id, account_ids, nonces, balance_to_move).unwrap();
+
+        // Assumes this now silently skips accounts without signing keys
+        let witness_set = WalletCore::sign_public_message(self.0, &message, &sign_ids)
+            .expect("`WalletCore::sign_public_message() failed to produce a signature for a NativeTokenTransfer.");
```
schouhy:
```suggestion
            .expect("`WalletCore::sign_public_message()` failed to produce a signature for a NativeTokenTransfer.");
```

wallet/src/lib.rs
```diff
@@ -550,4 +546,48 @@ impl WalletCore {
...
+        Ok(nssa::public_transaction::WitnessSet::for_message(
+            message,
+            &private_keys,
+        ))
+    }
+
+    #[must_use]
+    pub fn sign_privacy_message(
+        message: &nssa::privacy_preserving_transaction::Message,
+        proof: &Proof,
```
schouhy: let's have `proof: Proof` as parameter so there's no hidden cloning inside this function

wallet/src/lib.rs
```diff
@@ -550,4 +546,48 @@ impl WalletCore {
...
+        }
+
+        Ok(nssa::public_transaction::WitnessSet::for_message(
+            message,
+            &private_keys,
+        ))
+    }
+
+    #[must_use]
+    pub fn sign_privacy_message(
```
schouhy, line 602: I don't understand this indireciton here. Why do we need this wrapper of the witnesset constructor?

schouhy (commented):
Thanks for the PR. I'm finding hard to review as this looks like some minor adjustments for a bigger PR and I'd like to understand the bigger scope to see if I agree with the design choices.

nssa/src/public_transaction/witness_set.rs
jonesmarvin8: This function is necessary for Keycard. Keycard produces a signature within its logic. Though, it makes sense to add checks for:
- Check signature with corresponding public key.
- Enforce length.

wallet/src/lib.rs
jonesmarvin8: I separated the logic in `sign_privacy_message` to `send_privacy_preserving_tx_with_pre_check` since Keycard support will introduce more complex logic for signatures.

nssa/src/public_transaction/witness_set.rs
jonesmarvin8: I added tests for this constructor as well. Let me know what you think.

jonesmarvin8:
The changes introduced in this PR were incorporated in a larger PR 451. This specific PR, itself, was deemed unnecessary as it was difficult to see the motivation/logic behind the changes introduced without the full context of PR 451.

PR #456, opened 2026-04-22
Indexer query API

Pravdyvy:
## 🎯 Purpose

Introduces query API into indexer FFI through RPC client.

## ⚙️ Approach

Added `IndexerClient` to `IndexerServiceFFI` struct.

- [x] Added `IndexerClient` to `IndexerServiceFFI` struct.
- [x] Added indexer types and their ffi conversions.
- [x] Implemented queries.
- [x] Correct freeing of data

## 🧪 How to Test

Modified integration tests to use query API.

## 🔗 Dependencies

None

## 🔜 Future Work

- Update indexer module with that.
- Move ffi types into common ffi crate to share between all our ffi's.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


indexer/ffi/Cargo.toml
Arjentix, line 1: Let's move this crate to into `indexer/ffi`?

indexer_ffi/Cargo.toml
```diff
@@ -5,7 +5,13 @@ name = "indexer_ffi"
 version = "0.1.0"
 
 [dependencies]
+nssa.workspace = true
 indexer_service.workspace = true
+sequencer_core.workspace = true
+indexer_service_rpc.workspace = true
```
Arjentix: I think it should require at least one of `server` or `client` feature enabled. In fact we have a check for that in this crate. Probably it compiles fine for you because of feature unification, but I still suggest to add a concrete feature here

indexer/ffi/indexer_ffi.h
```diff
@@ -7,11 +7,25 @@ typedef enum OperationStatus {
...
+typedef enum FfiBedrockStatus {
+  Pending = 0,
+  Safe,
+  Finalized,
+} FfiBedrockStatus;
+
 typedef struct IndexerServiceFFI {
   void *indexer_handle;
   void *runtime;
+  void *indexer_client;
```
Arjentix, line 28: Hm, shouldn't it be separate?

indexer/ffi/src/api/types/mod.rs
```diff
@@ -0,0 +1,165 @@
...
+
+pub mod account;
+pub mod block;
+pub mod transaction;
+pub mod vectors;
+
+/// 32-byte array type for `AccountId`, keys, hashes, etc.
+#[repr(C)]
+#[derive(Clone, Copy, Default)]
+pub struct FfiBytes32 {
```
Arjentix, line 11: I think in future we could move such common types to a separate crate. And the same for `.h` file

indexer/ffi/src/api/types/account.rs
```diff
@@ -0,0 +1,82 @@
...
+// Helper functions to convert between Rust and FFI types
+
+impl From<&nssa::AccountId> for FfiBytes32 {
+    fn from(id: &nssa::AccountId) -> Self {
+        Self::from_account_id(id)
+    }
+}
+
+impl From<nssa::Account> for FfiAccount {
+    fn from(value: nssa::Account) -> Self {
```
Arjentix, line 33: Let's apply destructuring pattern here. So that if new fields added, we catch that at compile time

indexer/ffi/src/api/types/account.rs
```diff
@@ -0,0 +1,82 @@
...
+            data,
+            data_len,
+            data_cap,
+            nonce: value.nonce.0.into(),
+        }
+    }
+}
+
+impl From<FfiAccount> for indexer_service_protocol::Account {
+    fn from(value: FfiAccount) -> Self {
```
Arjentix, line 58: Same suggestion here. Probably not so important here but still. I suggest looking at other conversions as well.

Arjentix (approved):
LGTM, thanks, left some comments

indexer_ffi/src/api/types/transaction.rs
```diff
@@ -0,0 +1,521 @@
...
+                },
+                kind: FfiTransactionKind::Public,
+            },
+            Transaction::PrivacyPreserving(priv_tx) => Self {
+                body: FfiTransactionBody {
+                    public_body: std::ptr::null_mut(),
+                    private_body: Box::into_raw(Box::new(priv_tx.into())),
+                    program_deployment_body: std::ptr::null_mut(),
+                },
+                kind: FfiTransactionKind::Public,
```
schouhy:
```suggestion
                kind: FfiTransactionKind::Private,
```

indexer_ffi/src/api/types/account.rs
```diff
@@ -0,0 +1,82 @@
+use indexer_service_protocol::ProgramId;
+
+use crate::api::types::{FfiBytes32, FfiProgramId, FfiU128};
+
+/// Account data structure - C-compatible version of nssa Account.
+///
+/// Note: `balance` and `nonce` are u128 values represented as little-endian
+/// byte arrays since C doesn't have native u128 support.
+#[repr(C)]
+#[derive(Clone)]
```
schouhy: how does this Clone behave data pointer? looks like the clone will have pointers to the same memory.

schouhy (commented):
hey thanks for the PR. have a few comments

indexer/ffi/src/api/types/transaction.rs
```diff
@@ -0,0 +1,521 @@
...
+pub unsafe extern "C" fn free_ffi_transaction_vec(val: FfiVec<FfiTransaction>) {
+    let ffi_tx_std_vec: Vec<_> = val.into();
+    for tx in ffi_tx_std_vec {
+        unsafe {
+            free_ffi_transaction(tx);
+        }
+    }
+}
+
+fn cast_validity_window(window: ValidityWindow) -> [u64; 2] {
```
schouhy, line 527: ValidityWindow boundaries are Option<u64>. Doesn't seem like it can be fully parametrized by [u64; 2]

indexer/ffi/Cargo.toml
Pravdyvy: Moved

indexer_ffi/Cargo.toml
Pravdyvy: Added `client` feature

indexer/ffi/indexer_ffi.h
Pravdyvy: More convenient this way, user can not call this directly, there is query methods for that. Also needs runtime for queries.

indexer/ffi/src/api/types/mod.rs
Pravdyvy: Agree

indexer/ffi/src/api/types/account.rs
Pravdyvy: Applied

indexer/ffi/src/api/types/account.rs
Pravdyvy: Applied everywhere in `indexer_ffi`, where possible and needed.

indexer_ffi/src/api/types/transaction.rs
Pravdyvy: Fixed

indexer_ffi/src/api/types/account.rs
Pravdyvy: Yes, you are right. Removed this `Clone`.

indexer/ffi/src/api/types/transaction.rs
Pravdyvy: It is `indexer_service_protocol::ValidityWindow`. It can be cast only from `BlockValidityVindow`, so, considering, that our genesis is 1 it is safe to treat left `None` boundary as 0 and right `None` boundary as `u64::MAX`.
Also this should work for timestaps as well.

schouhy (approved):
LGTM thanks

moudyellaz (approved):
lgtm, thanks.

PR #457, opened 2026-04-24
Refactor signatures

jonesmarvin8:
## 🎯 Purpose

Refactor signatures for compatibility with Keycard. Both Keycard and LEE are BIP-340 compliant; however, the proposal was updated in 2022. Keycard is pre-2022 BIP-340 and LEE (currently) is post-2022. Post 2022 allowed for signatures to sign messages of variable length.

In this PR, we pre-hash messages before signing. Thus, enforcing 32-byte length for messages.

**This PR updates how messages are signed. This induces a breaking change to LEE.**
   - Changes to Signature breaks verification of previous transactions. Thus, current state from history cannot be verified without introducing versions (for Signature).
       - **Mitigation**: Generate a signature based on current testnet state, and re-initialize genesis block based on thee current state. 
   - This does not affect spendability or retrievability of public accounts. Public accounts are retrieved from the sequencer, and new transactions (using this PR) will generate and verify signatures based on the logic included in this PR. 

## ⚙️ Approach

- [X] Introduce functions for `Message` (public and privacy version) for to hash a message (`hash_message()`).
- [X] Update `Signature::new` logic to only accept `[u8;32]` message.
- [X] `Signature::new_with_aux_rand` updated to `sign_prehash_with_rand`
- [X] Comment out varied length test vectors for BIP-340 added in 2022.

## 🧪 How to Test

Refactoring. All previous tests still work (except the test vectors noted above)

## 🔗 Dependencies

N/A

## 🔜 Future Work

None.

## 📋 PR Completion Checklist

- [X] Complete PR description
- [X] Add/update tests
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -


nssa/src/privacy_preserving_transaction/message.rs
```diff
@@ -118,6 +118,11 @@ impl Message {
             timestamp_validity_window: output.timestamp_validity_window,
         })
     }
+
+    #[must_use]
+    pub fn hash_message(&self) -> [u8; 32] {
+        Sha256::digest(self.to_bytes()).into()
```
moudyellaz: `Sha256(borsh::to_vec(self))` with no domain tag, in three contexts (privacy tx here, public tx at `nssa/src/public_transaction/message.rs:70`, block at `common/src/block.rs:90` via `OwnHasher` which is also SHA256). Borsh has no type tag in output, so cross-context collision resistance depends on schemas never overlapping. Worth a tagged hash like `tagged_hash("LEE/v0/PrivacyMessage", to_bytes())`, distinct tag per context.

nssa/src/signature/mod.rs
```diff
@@ -37,22 +37,22 @@ impl FromStr for Signature {
...
-        message: &[u8],
+        message: &[u8; 32],
         aux_random: [u8; 32],
     ) -> Self {
         let value = {
             let signing_key = k256::schnorr::SigningKey::from_bytes(key.value())
                 .expect("Expect valid signing key");
             signing_key
-                .sign_raw(message, &aux_random)
+                .sign_prehash_with_aux_rand(message, &aux_random)
```
moudyellaz, line 57: PR body says "induces a breaking change to LEE" but no migration plan: testnet reset, genesis re-sign, wallet store invalidation. Any block or tx persisted under the old `sign_raw(to_bytes)` won't verify under `sign_prehash(hash_message)`. Worth spelling out in the body whether testnet wipe is required and whether wallet storage formats need versioning.

nssa/src/signature/mod.rs
```diff
@@ -37,22 +37,22 @@ impl FromStr for Signature {
...
     #[must_use]
-    pub fn new(key: &PrivateKey, message: &[u8]) -> Self {
+    pub fn new(key: &PrivateKey, message: &[u8; 32]) -> Self {
         let mut aux_random = [0_u8; 32];
         OsRng.fill_bytes(&mut aux_random);
         Self::new_with_aux_random(key, message, aux_random)
     }
 
     pub(crate) fn new_with_aux_random(
         key: &PrivateKey,
```
moudyellaz, line 49: `new` and `new_with_aux_random` were tightened to `&[u8; 32]` but `is_valid_for` at line 64 still takes `&[u8]`. Asymmetric: signer can't pass non-32, verifier can. All callers feed `&message_hash` (32 bytes), but a future caller could pass arbitrary length and verification would happily compute over those bytes. Tighten to `&[u8; 32]` for parity.

nssa/src/signature/bip340_test_vectors.rs
```diff
@@ -276,89 +290,95 @@ pub fn test_vectors() -> Vec<TestVector> {
...
-                hex::decode("99999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999").unwrap(),
-            ),
-            signature: Signature {
-                value: hex_to_bytes(
-                    "403B12B0D8555A344175EA7EC746566303321E5DBFA8BE6F091635163ECA79A8585ED3E3170807E7C03B720FC54C7B23897FCBA0E9D0B4A06894CFD249F22367",
-                ),
-            },
-            verification_result: true,
-        },
+        // Test with invalid message length (0); valid test for BIP-340 post 2022.
```
moudyellaz, line 304: What about these commented tests?

nssa/src/signature/mod.rs
```diff
@@ -37,22 +37,22 @@ impl FromStr for Signature {
 
 impl Signature {
     #[must_use]
-    pub fn new(key: &PrivateKey, message: &[u8]) -> Self {
+    pub fn new(key: &PrivateKey, message: &[u8; 32]) -> Self {
```
moudyellaz, line 42: `Signature::new(key, message: &[u8; 32])` has no doc comment. A future reader sees the 32-byte constraint with no context. Worth a `///` line referencing pre-2022 BIP-340 / Keycard compat.

nssa/src/signature/mod.rs
```diff
@@ -114,7 +114,7 @@ mod tests {
     #[test]
     fn signature_verification_from_bip340_test_vectors() {
         for (i, test_vector) in bip340_test_vectors::test_vectors().into_iter().enumerate() {
-            let message = test_vector.message.unwrap_or(vec![]);
+            let message = test_vector.message.unwrap_or([0_u8; 32]);
```
moudyellaz: No active vector has `message: None` (the only ones that did are now commented out), so this branch is unreachable. Also semantically different from the old `unwrap_or(vec![])`: a 32-zero-byte hash is a real value, not absence. Either drop the `Option` and require `[u8; 32]`, or keep it and document why None is no longer expected.

nssa/src/public_transaction/message.rs
```diff
@@ -63,4 +64,9 @@ impl Message {
             instruction_data,
         }
     }
+
+    #[must_use]
+    pub fn hash_message(&self) -> [u8; 32] {
+        Sha256::digest(self.to_bytes()).into()
```
moudyellaz: No KAT pinning `hash_message` output here or at `nssa/src/privacy_preserving_transaction/message.rs:124`. A field reorder or schema tweak in `to_bytes` would silently invalidate every signature without any test failing. Worth a small pinned-vector test for each: build a known Message, assert `hash_message()` equals a fixed 32-byte hex.

nssa/src/signature/bip340_test_vectors.rs
jonesmarvin8: These tests were added in 2022 to support arbitrary message lengths. For the original version of BIP-340, these tests would fail to build (since the messages are defined to be [u8;32]).

nssa/src/signature/mod.rs
jonesmarvin8: Added

moudyellaz (approved):
lgtm, thanks for addressing my comments

nssa/src/privacy_preserving_transaction/message.rs
```diff
@@ -118,22 +120,33 @@ impl Message {
             timestamp_validity_window: output.timestamp_validity_window,
         })
     }
+
+    #[must_use]
+    pub fn hash_message(&self) -> [u8; 32] {
```
schouhy: I'd call this method just `hash`, otherwise calling it like `message.hash_message()` feels a bit redundant

nssa/src/public_transaction/message.rs
```diff
@@ -63,4 +66,74 @@ impl Message {
             instruction_data,
         }
     }
+
+    #[must_use]
+    pub fn hash_message(&self) -> [u8; 32] {
```
schouhy: same as above
```suggestion
    pub fn hash(&self) -> [u8; 32] {
```

nssa/src/privacy_preserving_transaction/message.rs
```diff
@@ -118,22 +120,33 @@ impl Message {
...
+        let mut bytes = Vec::with_capacity(
+            PREFIX
+                .len()
+                .checked_add(self.to_bytes().len())
```
schouhy: let's save `self.to_bytes` in a variable to avoid calling it twice in this method.

nssa/src/signature/mod.rs
```diff
@@ -37,22 +37,24 @@ impl FromStr for Signature {
 
 impl Signature {
     #[must_use]
-    pub fn new(key: &PrivateKey, message: &[u8]) -> Self {
+    /// This function expects the incoming message to be prehashed to be pre-2022 BIP-340/Keycard
+    /// compatible.
```
schouhy, line 40: I think it's more common to place docs before attributes like #[must_use]

common/src/block.rs
schouhy, line 1: there's a `block_hash` method in this file that got out of sync with the new hashing formula for blocks. I think it's unused so I'd remove it

schouhy (approved):
Thanks LGTM! Left some minor comments

PR #458, opened 2026-04-25
Fix clock transaction validation in Indexer

Arjentix:
## 🎯 Purpose

Fixes the problem we encountered on stage when 100th block failed on indexer side.

## ⚙️ Approach

- Apply the same validation rules in indexer as we have in sequencer. More precisely it means to skip stateless validation for clock tx
- Adjust tests 

## 🧪 How to Test

**Tested on stage: it works:** ✅

## 🔗 Dependencies

None

## 🔜 Future Work

Probably better unify transaction execution so that we don't have to repeat code on both sides.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests

### DIFF
--- a/storage/src/indexer/mod.rs
+++ b/storage/src/indexer/mod.rs
@@ -1,6 +1,9 @@
 use std::{path::Path, sync::Arc};
 
-use common::block::Block;
+use common::{
+    block::Block,
+    transaction::{NSSATransaction, clock_invocation},
+};
 use nssa::V03State;
 use rocksdb::{
     BoundColumnFamily, ColumnFamilyDescriptor, DBWithThreadMode, MultiThreaded, Options,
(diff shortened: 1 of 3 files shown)


moudyellaz (approved):

storage/src/indexer/mod.rs
```diff
@@ -238,6 +271,17 @@ mod tests {
         AccountId::from(&PublicKey::new_from_private_key(&acc2_sign_key()))
     }
 
+    fn produce_dummy_block_with_clock(
```
Pravdyvy: Small nitpick: every block should have clock invocation, so I suppose we should rename it into `produce_dummy_block` or replace `common::test_utils::produce_dummy_block` with it. 
Just to avoid confusion for future people who would try to write new tests.

Pravdyvy (approved):
lgtm. Left small suggestion.

storage/src/indexer/mod.rs
Arjentix: Done

PR #461, opened 2026-04-28
feat(wallet): add keycard support for public accounts for public/privacy txs for program facades

jonesmarvin8:
## 🎯 Purpose

Integration for Keycard. Add supports for shielded transactions for `auth-transfers` as well as public/shielded transactions for token program and amm. Additionally, adds a subcommand to `Keycard` to get private account secret keys from Keycard (demonstration that this feature can be used later).

## ⚙️ Approach

- [X] Add python function, Rust wrapper and appropriate command for querying private account secret keys.
- [X] Add logic (similar to auth-transfer) to support Keycard usage for public accounts for AMM, ATA and Token Program.

**Updated**:
- ~~Use `SigningGroup` to support mixture of local/keycard accounts.~~ Replaced with AccountManager 
- Abstract privacy transactions with `send_privacy_preserving_tx_with_pre_check` and `send_privacy_preserving_tx`.
- Private account secret keys can be retrieved; this is locked behind `keycard-debug` feature.

## 🧪 How to Test

**Disclaimer:** Keycard must be setup following these [directions](https://www.notion.so/Setup-34b8f96fb65c80bbbf2ced7aad3d3aa6).

All previous tests should work. Additionally, keycard can be tested manually using the files `keycard_tests_1.sh`, `keycard_tests_2.sh` and `keycard_test_3.sh`.

**For convenience, I have made additional test files `keycard_tests_2_1.sh` (Token), `keycard_tests_2_2.sh` (AMM) and `keycard_tests_2_3.sh` (ATA). These files consist of all of the same tests and key paths from `keycard_tests_2.sh`. These separate files may be easier to run in case of hardware issues with keycard. Each file must be ran consecutively as accounts used in `keycard_tests_2_3.sh` may have been initialized in `keycard_tests_2_1.sh`.**

## 🔗 Dependencies

Builds off of [PR 451](https://github.com/logos-blockchain/logos-execution-zone/pull/451)

## 🔜 Future Work

Potentially, add logic to use Keycard for deshielded and private accounts. This is deferred till wallet logic is cleaned up as discussed on [discord](https://discord.com/channels/973324189794697286/1500860069661638786).

## 📋 PR Completion Checklist

- [X] Complete PR description
- [X] Add/update tests


wallet/src/cli/keycard.rs
```diff
@@ -43,6 +49,15 @@ impl WalletSubcommand for KeycardSubcommand {
 
                 Ok(SubcommandReturnValue::Empty)
             }
+            Self::GetPrivateKeys { key_path } => {
+                let pin = read_pin()?;
+                let (nsk, vsk) =
+                    KeycardWallet::get_private_keys_for_path_with_connect(&pin, &key_path)
+                        .map_err(anyhow::Error::from)?;
+                println!("nsk: {}", hex::encode(nsk));
```
moudyellaz: Thanks for adding this, super useful for debugging Keycard flows. One thing to flag before merge though: as written, `GetPrivateKeys` prints both NSK and VSK to stdout every time, with no confirmation step. Verified live on a flashed Keycard:

```
$ wallet keycard get-private-keys --key-path "m/44'/60'/0'/0/0"
nsk: 55e505bf925e536c843a12ebc08c41ca5f4761eeeb7fa33725f0b44e6f1ac2e4
vsk: 30f798893977a7b7263d1f77abf58e11e014428c92030d6a02fe363cceb41ffa
```

Since NSK gates nullifier creation and VSK gates note decryption, having either of them land in terminal scrollback, tmux logs, or a recorded screen share is enough to fully compromise that account's privacy. A couple of small additions would help here:
- Gate the disclosure behind something like `--reveal` (with a short stderr warning right before the print).
- Wrap the returned `(nsk, vsk)` in `Zeroizing` from `keycard_wallet::lib` through to the CLI.

Also, since the PR description frames this as "demonstration that this feature can be used later", would it make sense to put it behind a `cfg(feature = "keycard-debug")` for now?

wallet/src/program_facades/amm.rs
```diff
@@ -289,8 +357,37 @@ impl Amm<'_> {
...
+            })?;
+            let (sig_1, pk_1) = keycard_wallet::KeycardWallet::sign_message_for_path_with_connect(
+                &pin, kp_a, &msg_hash,
+            )?;
+            let (sig_2, pk_2) = keycard_wallet::KeycardWallet::sign_message_for_path_with_connect(
+                &pin, kp_b, &msg_hash,
+            )?;
+            nssa::public_transaction::WitnessSet::from_list(
+                &message,
+                &[sig_1, sig_2],
```
moudyellaz: Two things came up trying to run `swap-exact-output` with a keycard. First, the CLI is currently a dead end: in `wallet/src/cli/programs/amm.rs:154` the variant declares `user_holding_a: String` (mandatory), and `user_holding_a_key_path` has `conflicts_with = "user_holding_a"`. So passing either alone fails (`required` error, or `cannot be used with` error), and the keycard path can never be reached from the CLI. The fix that `SwapExactInput` and `AddLiquidity` already use, `Option<String>` + `required_unless_present_any`, would close that.

Second, even if the CLI is unblocked, the function itself has a sigs-vs-nonces mismatch: `nonces` only contains the seller's nonce (length 1), but the keycard branch signs both holdings and builds a 2-sig witness, so `WitnessSet::from_list`'s length check would reject. Same fix that `send_swap_exact_input` already got: pick a single `seller_key_path` based on `definition_token_a_id == token_definition_id_in` and sign only that one.

wallet/src/program_facades/token.rs
```diff
@@ -200,8 +244,23 @@ impl Token<'_> {
...
+                ExecutionFailureKind::KeycardError(pyo3::PyErr::new::<PyRuntimeError, _>(
+                    e.to_string(),
+                ))
+            })?;
+            let (signature, public_key) = KeycardWallet::sign_message_for_path_with_connect(
+                &pin,
+                &sender_key_path,
+                &message.hash(),
+            )?;
+            WitnessSet::from_list(&message, &[signature], &[public_key])
```
moudyellaz: Came up while testing on the keycard: with the keycard sender at path 4 (LEZ holding) and a fresh keycard recipient at path 8 (never initialized), the sequencer rejected the tx with:

```
ERROR sequencer_core Transaction with hash fb6c36e5b058...
  failed execution check with error:
  InvalidProgramBehavior(ClaimedUnauthorizedAccount { account_id: CJwKfrb3DFMmFvujQSB5ARcRTAa8EdP6eWm2hmSkF7Rb })
```

`CJwKfrb...` is path 8. The token program claimed it Authorized (because the recipient holding is default per `programs/token/src/transfer.rs:109`), but the keycard branch only signs the sender. The non-keycard branch correctly adds the recipient signature when the wallet has its key, so could the keycard branch do the same? The CLI already has `--to-key-path` at `wallet/src/cli/programs/token.rs:142`, would be great to thread it through and sign with both when present.

wallet/src/program_facades/token.rs
```diff
@@ -563,8 +726,23 @@ impl Token<'_> {
...
+                ExecutionFailureKind::KeycardError(pyo3::PyErr::new::<PyRuntimeError, _>(
+                    e.to_string(),
+                ))
+            })?;
+            let (signature, public_key) = KeycardWallet::sign_message_for_path_with_connect(
+                &pin,
+                definition_key_path,
+                &message.hash(),
+            )?;
+            WitnessSet::from_list(&message, &[signature], &[public_key])
```
moudyellaz: Same shape as the Transfer one above, also came up while testing on the keycard: with the keycard definition at path 0 and a fresh keycard holder at path 9 (never initialized), the sequencer rejected every attempt with `ClaimedUnauthorizedAccount` on path 9. Mint claims the holder Authorized when it's default (`programs/token/src/mint.rs:70`), but the keycard branch only signs the definition. Adding a `holder_key_path: Option<&str>` and signing both when set would fix it.

wallet/src/program_facades/amm.rs
```diff
@@ -58,43 +64,30 @@ impl Amm<'_> {
...
-        {
-            private_keys.push(signing_key_lp);
+        if lp_sk.is_some() {
             let lp_nonces = self
                 .0
                 .get_accounts_nonces(vec![user_holding_lp])
                 .await
                 .map_err(ExecutionFailureKind::SequencerError)?;
-            nonces.extend(lp_nonces);
+            if lp_nonces.is_empty() {
```
moudyellaz: Small nit: `get_accounts_nonces` always returns one `Nonce` per requested id (defaulting to `Nonce(0)` for non-existent accounts), so the `is_empty()` branch is unreachable. Could probably collapse to `nonces.extend(lp_nonces)`.

wallet/src/lib.rs
```diff
@@ -600,16 +661,77 @@ impl WalletCore {
...
+    pub fn sign_privacy_message_with_keycard(
+        message: &nssa::privacy_preserving_transaction::Message,
+        proof: Proof,
+        key_paths: &[String],
+    ) -> Result<nssa::privacy_preserving_transaction::witness_set::WitnessSet, ExecutionFailureKind>
+    {
+        let mut signatures_and_public_keys = Vec::<(Signature, PublicKey)>::new();
+
+        for path in key_paths {
+            let pin = crate::helperfunctions::read_pin().map_err(|e| {
```
moudyellaz: Two PIN-handling things to flag, both visible in `wallet/src/lib.rs`:
- `sign_privacy_message_with_keycard` (line 703) calls `read_pin()` inside the loop (line 712), so for a multi-path sign the user would be prompted N times for the same PIN.
- For a single shielded transfer that goes through `send_privacy_preserving_tx_with_pre_check`, the PIN is read at line 421 to derive the account id and then again inside `sign_privacy_message` at line 673 for the actual sign. Two prompts for one tx.

Could the PIN be read once and passed down as `&Zeroizing<String>`?

wallet/src/cli/programs/amm.rs
```diff
@@ -260,6 +388,8 @@ impl WalletSubcommand for AmmProgramAgnosticSubcommand {
                                 user_holding_lp,
                                 balance_a,
                                 balance_b,
+                                user_holding_a_key_path.as_deref(),
+                                user_holding_b_key_path.as_deref(),
```
moudyellaz: The CLI accepts `--user-holding-lp-key-path` (line 86, with full clap config) and uses it at line 361 to resolve the LP account id from the keycard, but the call to `send_new_definition` at line 385 only threads `user_holding_a_key_path` and `user_holding_b_key_path`, the LP key path is silently dropped. So a user who sets `--user-holding-lp-key-path` expecting keycard signing for the LP holding will hit `ClaimedUnauthorizedAccount` on the LP when it's fresh (LP mint claims it Authorized per `programs/amm/src/new_definition.rs`). Either thread the LP key path into `send_new_definition` and have the facade sign with it, or reject the flag at the CLI with a clear "not yet supported" error.

keycard_wallet/python/keycard_wallet.py
```diff
@@ -122,4 +122,25 @@ def sign_message_for_path(self, message: bytes, path: str = "m/44'/60'/0'/0/0")
...
+                path
+            )
+
+            nsk = private_keys.lee_nsk
+            vsk = private_keys.lee_vsk
+
+            return (nsk, vsk)
+        
+        except Exception as e:
+            import traceback
```
moudyellaz: Tiny one to round out the secret-hygiene story: this `traceback.print_exc()` dumps frame locals on failure, and those locals can include `nsk`, `vsk`, or intermediate bytes from a partial export. Came up while testing on the keycard: a failed export produced a multi-frame Python stack trace to stderr including function names from the secure channel internals. No secret bytes in that particular trace because the export aborted early, but the leak path is real. The other methods in this file use a plain `print(f"Error: ...")`. Could we match that pattern here?

wallet/src/cli/programs/token.rs
```diff
@@ -835,12 +1066,14 @@ impl WalletSubcommand for TokenProgramSubcommandPublic {
...
+                definition_key_path,
             } => {
                 Token(wallet_core)
                     .send_mint_transaction(
                         definition_account_id.parse().unwrap(),
                         holder_account_id.parse().unwrap(),
                         amount,
+                        definition_key_path.as_deref(),
                     )
                     .await?;
```
moudyellaz, line 779: Small UX bug spotted while testing on the keycard: `MintToken` (and `BurnToken` at line 1055) calls the facade and discards the result. No tx hash printed, no `poll_native_token_transfer`. Compare to `TransferToken` at line 1043 which does both. When the sequencer rejects a keycard mint (e.g. the `ClaimedUnauthorizedAccount` case in the comment on `token.rs:741`), the user sees nothing and the command exits 0. Worth at least binding the tx hash, printing it, and calling the poller so failures surface.

moudyellaz (changes requested):
Thanks for the PR, I have tested it with the keycard and left some comments.

wallet/src/cli/keycard.rs
jonesmarvin8: I can add zeroize within Rust, but the weak point is still Python. This is an issue we'll need to figure out before fully integrating Keycard with private transactions.

For the feature flag, @schouhy . what do you think?

wallet/src/cli/programs/token.rs
jonesmarvin8: Added this for amm and ata as well.

wallet/src/program_facades/amm.rs
jonesmarvin8: Appears to be fixed by a combination of PR 450 (already merged) and SigningGroups.

wallet/src/program_facades/token.rs
jonesmarvin8: Handled by 450 and SigningGroups.

wallet/src/program_facades/token.rs
jonesmarvin8: Handled by 450 and SigningGroups.

wallet/src/lib.rs
jonesmarvin8: Fixed.

wallet/src/cli/programs/amm.rs
jonesmarvin8: Cleaned up with 450 and SigningGroups.

jonesmarvin8:
(keycard_wallet.py) This one didn't move this round. DEFAULT_PAIRING_PASSWORD is still a hardcoded module-level string used as the default in setup_communication, so any wallet (or card this wrapper has touched) can pair without a secret and PIN becomes the only moat. Easiest fix: drop the default and read from KEYCARD_PAIRING_PASSWORD env or a prompt, same shape as --pin now. Fine to defer.

jonesmarvin8:
- (keycard_wallet/src/lib.rs) Still hitting the wedge I flagged earlier: card lands in 0x80100067 ("unpowered") after ~8 wallet ops in a row and only recovers on physical reseat. connect() has no retry/reconnect path, every error from setup_communication_with_pairing or setup_communication propagates straight up. A small reset-and-retry-once on a transport error here would smooth the test/dev loop a lot. Fine as a follow-up if you'd rather merge first.

I wasn't able to generate the unpowered issue locally. I added branching logic in case this is detected, and a test file `keycard_wallet/tests/keycard_power_recovery_tests.sh` that forces unpowered states.

jonesmarvin8:
expect tautology issues (cli/keycard.rs)

Pravdyvy (approved):
lgtm. I can not test keycard part, from Rust all seems fine. Conflicts should be trivial, this PR changes should be taken.

lez/wallet/Cargo.toml
```diff
@@ -46,6 +46,9 @@ optfield = "0.4.0"
 url.workspace = true
 derive_more = { workspace = true, features = ["display"] }
 
+[features]
+keycard-debug = []
```
Arjentix, line 52: Let's add a comment here explaining what this feature does

wallet/src/lib.rs
```diff
@@ -80,6 +81,20 @@ pub enum ExecutionFailureKind {
...
+    #[must_use]
+    #[expect(
+        clippy::needless_pass_by_value,
+        reason = "used as a method reference in map_err"
+    )]
+    pub fn from_anyhow(e: anyhow::Error) -> Self {
+        Self::KeycardError(pyo3::PyErr::new::<pyo3::exceptions::PyRuntimeError, _>(
+            e.to_string(),
+        ))
+    }
```
Arjentix: `anyhow::Error` is very common thing and it's not clear from method name that it binds this error  to `KeycardError`, so let's better do it like this:


```suggestion
    pub fn keycard_error(e: impl ToString) -> Self {
        Self::KeycardError(pyo3::PyErr::new::<pyo3::exceptions::PyRuntimeError, _>(
            e.to_string(),
        ))
    }
```

wallet/src/lib.rs
```diff
@@ -577,27 +590,30 @@ impl WalletCore {
...
                     .collect(),
                 output,
-            )
-            .unwrap();
+            )?;
+
+        let message_hash = message.hash();
+        let signatures_public_keys = acc_manager
+            .sign_message(message_hash)
+            .map_err(ExecutionFailureKind::from_anyhow)?;
```
Arjentix: Aha, I can see the error can also be not related to keycard. Let's better create `SignError` or check if existing variants might work

lez/wallet/src/account_manager.rs
```diff
@@ -11,8 +12,15 @@ use nssa_core::{
...
 #[derive(Clone)]
-pub enum PrivacyPreservingAccount {
+pub enum AccountIdentity {
     Public(AccountId),
+    /// A public account without signing. Would not try to sign, even if account is owned.
+    PublicNoSign(AccountId),
+    /// A public account from keycard. Mandatory signing.
+    PublicKeycard {
+        account_id: AccountId,
+        key_path: String,
```
Arjentix, line 22: Can't key path be described in a more strongly typed format? Not a strong opinion, just a suggestion

wallet/src/account_manager.rs
```diff
@@ -289,21 +354,66 @@ impl AccountManager {
...
+            .fold(vec![], |mut acc, state| match state {
+                State::Private(_) | State::Public { .. } => acc,
+                State::PublicKeycard {
+                    account: _,
+                    key_path,
+                } => {
+                    acc.push(key_path.as_str());
+                    acc
+                }
+            });
```
Arjentix: I think this could be implemented with `.map().collect()` which will be easier to read

wallet/src/account_manager.rs
```diff
@@ -289,21 +354,66 @@ impl AccountManager {
...
+
+        if let Some(pin) = self.pin.clone() {
+            pyo3::Python::with_gil(|py| -> pyo3::PyResult<()> {
+                python_path::add_python_path(py)?;
+                let wallet = KeycardWallet::new(py)?;
+                wallet.connect(py, &pin)?;
+                for path in keycard_paths {
+                    sigs.push(wallet.sign_message_for_path(py, path, &message_hash)?);
+                }
+                drop(wallet.close_session(py));
```
Arjentix: Looks a bit strange, why do you have to explicitly drop the return value of this function?

wallet/src/program_facades/native_token_transfer/public.rs
```diff
@@ -22,110 +17,57 @@ impl NativeTokenTransfer<'_> {
         from_mention: &CliAccountMention,
         to_mention: &CliAccountMention,
```
Arjentix: I lost the moment when we started having both `from` and `from_mention` as well as `to` and `to_mention`. What's the meaning of this? I think to identify an account there should be only one argument, maybe `AccountIdentity`?

Arjentix: Same idea in other places

wallet/src/signing.rs
```diff
@@ -1,114 +1,36 @@
...
-        &mut self,
-        mention: &CliAccountMention,
-        account_id: AccountId,
-        wallet_core: &WalletCore,
-    ) -> Result<()> {
-        if let CliAccountMention::KeyPath(path) = mention {
-            self.keycard.push((account_id, path.clone()));
-            return Ok(());
+    pub fn get_or_connect<'py>(
+        &'py mut self,
```
Arjentix:
```suggestion
        &mut self,
```

I think this should also work and makes more sense to me. I can't see why self should depend on `'py`

wallet/src/signing.rs
```diff
@@ -1,114 +1,36 @@
...
-
-    /// Account IDs that require a nonce (every non-foreign signer).
-    #[must_use]
-    pub fn signing_ids(&self) -> Vec<AccountId> {
-        self.local
-            .iter()
-            .map(|(id, _)| *id)
-            .chain(self.keycard.iter().map(|(id, _)| *id))
-            .collect()
+        Ok(self.wallet.as_ref().unwrap())
```
Arjentix: Oh, that's a good place for [`get_or_try_insert_with()`](https://doc.rust-lang.org/std/option/enum.Option.html#method.get_or_try_insert_with) but it's not stabilized yet...

Btw, let's replace `unwrap()` with `expect()`

wallet/src/signing.rs
```diff
@@ -1,114 +1,36 @@
...
-                for (_, path) in &self.keycard {
-                    sigs.push(wallet.sign_message_for_path(py, path, hash)?);
-                }
-                drop(wallet.close_session(py));
-                Ok(())
-            })
-            .map_err(anyhow::Error::from)?;
+    pub fn close(self, py: Python<'_>) {
+        if let Some(w) = self.wallet {
+            drop(w.close_session(py));
```
Arjentix: I think `drop()` here does nothing

lez/wallet/src/signing.rs
```diff
@@ -1,114 +1,36 @@
...
-/// Groups transaction signers by type to minimise Python GIL acquisition.
-///
-/// Local signers are signed in pure Rust; all keycard signers share a single Python session
-/// with one `connect` / `close_session` pair.
-#[derive(Default)]
-pub struct SigningGroups {
-    local: Vec<(AccountId, PrivateKey)>,
-    keycard: Vec<(AccountId, String)>,
+/// Lazily opens and reuses a single Keycard session for all keycard signers in one transaction.
+pub struct KeycardSessionContext {
```
Arjentix, line 5: I think it could be better if we make session depend on python lifetime and store python:

```rust
pub struct KeycardSessionContext<'py> {
    python: Python<'py>,
    ...
}
```

That way you don't need to pass around `Python` every time and you can even remove `close()` and move it's logic into `impl Drop`.

wallet/src/program_facades/ata.rs
```diff
@@ -17,45 +16,39 @@ impl Ata<'_> {
         &self,
         owner_id: AccountId,
         definition_id: AccountId,
+        owner_mention: &CliAccountMention,
```
Arjentix: Yeah, as I wrote above, I don't think this is a good idea to have duplicated information about account: `owner_id` and `owner_mention`. Better to somehow resolve it before and pass a singe argument there

lez/wallet/src/cli/programs/token.rs
```diff
@@ -680,33 +690,78 @@ impl WalletSubcommand for TokenProgramSubcommandPublic {
...
-                Token(wallet_core)
+                let sender = sender_account_id.resolve(wallet_core.storage())?;
+                let recipient = recipient_account_id.resolve(wallet_core.storage())?;
+                let (
+                    AccountIdWithPrivacy::Public(sender_id),
+                    AccountIdWithPrivacy::Public(recipient_id),
+                ) = (sender, recipient)
+                else {
+                    anyhow::bail!("Only public accounts supported for token transfer");
+                };
```
Arjentix, line 721: Wait, why? Looks like before it was working for private accounts as well

Arjentix: And same problem in other places. Looks like we restricting everything just because of keycard...

keycard_wallet/tests/keycard_tests.sh
```diff
@@ -79,3 +83,44 @@ wallet account get --account-id "m/44'/60'/0'/0/0"
...
+  --to     "m/44'/60'/0'/0/1" \
+  --amount 5
+echo "Deshielded transfer of 5: $PRIV_SENDER → keycard path 1"
+
+sleep 15
+
+echo "priv-sender state (balance should have decreased by 5):"
+wallet account get --account-id "$PRIV_SENDER"
+echo "Keycard path 1 state (balance should have increased by 5):"
+wallet account get --account-id "m/44'/60'/0'/0/1"
```
Arjentix: Let's add a new line here

lez/keycard_wallet/tests/keycard_tests.sh
Arjentix, line 1: These tests look like real e2e tests. Let's think about how can we introduce this in our CI

Arjentix: Ah, wait, they don't assert anything sadly. But I think they could?

Arjentix: Oh, I think they require keycard connected... Yeah, that's complicated

lez/keycard_wallet/src/lib.rs
```diff
@@ -190,11 +207,65 @@ impl KeycardWallet {
...
+    pub fn get_private_keys_for_path_with_connect(
+        pin: &str,
+        path: &str,
+    ) -> PyResult<PrivateKeyPair> {
+        Python::with_gil(|py| {
+            python_path::add_python_path(py)?;
+            let wallet = Self::new(py)?;
+            wallet.connect(py, pin)?;
+            let result = wallet.get_private_keys_for_path(py, path);
+            drop(wallet.disconnect(py));
```
Arjentix, line 265: I really don't understand what this should do, lol

Arjentix (commented):
Hey, thanks for the PR. Left some comments, my main concern is account arguments duplication, introduced in previous PR and in this one.

lez/wallet/src/signing.rs
jonesmarvin8: `keycard-py` will be replaced with a Rust SDK (currently in development). Due to this, I think it's fine to avoid a complex refactoring for python when python will ultimately be removed altogether.

lez/keycard_wallet/tests/keycard_tests.sh
jonesmarvin8: For the python SDK this is not possible. Hopefully, with the eventual Rust SDK this will be possible. Otherwise, we risk silent breaks to keycard support.

lez/wallet/src/account_manager.rs
jonesmarvin8: This might make more sense to change with the (eventual) Rust-SDK. Currently, we're relying on the SDK to parse the string (and return the error).

lez/keycard_wallet/src/lib.rs
jonesmarvin8: This establishes a connection with Keycard (and promptly closes it) to retrieve the set of private keys for a given private account's key path. This is a proof of concept to demonstrate that we can retrieve private keys in case we want to integrate this into the wallet's workflow.

lez/wallet/src/cli/programs/token.rs
jonesmarvin8: The anyhow bail is bad, will update it.

However, this branching path is only for public accounts (and is that way in [main](https://github.com/logos-blockchain/logos-execution-zone/blob/b77cfc29c22e150882d7219b1e2ef37a86500566/wallet/src/cli/programs/token.rs#L672) )

wallet/src/signing.rs
jonesmarvin8: Drop was used here to satisfy CI. Lint gets upset when either `.ok` or `_ = ` is used.

wallet/src/program_facades/native_token_transfer/public.rs
jonesmarvin8: This was added in this PR as a work around of just using id's, but fixed by using AccountIdentity.

wallet/src/program_facades/ata.rs
jonesmarvin8: Simplified.

wallet/src/signing.rs
Arjentix: Oh, so that's a result type? I thought it doesn't return anything. Then I think in general we do it like: `let _res = w.close_session(py))`

lez/wallet/src/cli/programs/token.rs
Arjentix: Oh, yes, indeed, thanks

Arjentix (approved):
LGTM, thanks

wallet/src/account_manager.rs
jonesmarvin8: CI requires drop here. Lint complains if I use `_ = ...` or `.. .ok()?`.

wallet/src/signing.rs
jonesmarvin8: That works.

wallet/src/account_manager.rs
jonesmarvin8: Applied suggested fix below.

docs/LEZ testnet v0.1 tutorials/keycard.md
```diff
@@ -60,17 +60,41 @@ Unset it when done:
...
+## Pairing password
+
+The pairing password is used to establish a secure channel between the wallet and the card. It is set permanently on the card during `wallet keycard init` and must match on every subsequent re-pair.
+
+The default password (`KeycardDefaultPairing`) is [recommended](https://docs.keycard.tech/en/developers/core) for most users. Wallet CLI allows advance users the flexibility to set their own pairing password.
+
+To use a custom pairing password, set it before `init`:
+
+```bash
+export KEYCARD_PAIRING_PASSWORD=my-custom-password
```
schouhy: suggesting this command will make the users's password end up in the shell history in plaintext. One way to avoid that is to add a space before the command, which tells the terminal to not add it to the history
```suggestion
# Note: Keep the leading space before this command.
# Leading space prevents this command from being stored in shell history
# (when HISTCONTROL=ignorespace is enabled).
 export KEYCARD_PAIRING_PASSWORD=my-custom-password
```

schouhy (approved):
lgtm thanks

PR #462, opened 2026-04-28
feat: strong-type PrivacyPreservingCircuitInput with per-account enum

### LINKED ISSUE #455: Strong-type PrivacyPreservingCircuitInput with a per-account enum
## Motivation

`PrivacyPreservingCircuitInput` currently carries parallel vectors indexed by per-account position:

- `visibility_mask: Vec<u8>`
- `private_account_keys: Vec<(NullifierPublicKey, SharedSecretKey)>`
- `private_account_nsks: Vec<NullifierSecretKey>`
- `private_account_membership_proofs: Vec<Option<MembershipProof>>`

The circuit guest iterates them in lock-step and asserts lengths, with per-mask branches deciding which vectors to consume. The lock-step is implicit, the per-account invariants are runtime panics, and several `Option` values inside the membership-proof vector encode whole separate code paths in the guest.

## Proposal

Collapse into a single `Vec<PrivacyPreservingCircuitInputAccount>` aligned 1:1 with the program's pre_states (including `Public` slots), where each variant carries exactly the fields the corresponding mask path needs:

```rust
pub struct PrivacyPreservingCircuitInput {
    pub program_outputs: Vec<ProgramOutput>,
    /// One entry per pre_state, in the same order as the program's pre_states.
    /// Length must equal the number of pre_states derived from `program_outputs`.
    pub accounts: Vec<PrivacyPreservingCircuitInputAccount>,
    pub program_id: ProgramId,
}

pub enum PrivacyPreservingCircuitInputAccount {
    Public,
    PrivateAuthorizedInit {
        ssk: SharedSecretKey,
        nsk: NullifierSecretKey,
    },
    PrivateAuthorizedUpdate {
        ssk: SharedSecretKey,
        nsk: NullifierSecretKey,
        membership_proof: MembershipProof,
    },
    PrivateUnauthorized {
        npk: NullifierPublicKey,
        ssk: SharedSecretKey,
    },
    PrivatePdaInit {
        npk: NullifierPublicKey,
        ssk: SharedSecretKey,
    },
    PrivatePdaUpdate {
        ssk: SharedSecretKey,
        nsk: NullifierSecretKey,
        membership_proof: MembershipProof,
    },
}
```

Three design decisions worth calling out:

1. `PrivateAuthorized` and `PrivatePda` are split into `Init` / `Update` variants instead of carrying `membership_proof: Option<MembershipProof>`. In the guest, `Some` and `None` of the proof are two unrelated code paths (commitment-based update nullifier vs init nullifier with `pre_state == Account::default()` precondition). Splitting makes the two paths structurally distinct and removes the runtime "membership proof must be None" assert.

2. `npk` is omitted from variants that carry `nsk`. The guest already re-derives and asserts `npk == NullifierPublicKey::from(nsk)`. Carrying both lets the host submit an inconsistent pair whose only outcome is a panic; dropping `npk` removes that footgun.

3. The `accounts` vector is documented as 1:1 with the pre_state order across all variants including `Public`. The guest's `private_pda_npk_by_position` and `private_pda_bound_positions` rely on this position alignment; a length assert at the top of `compute_circuit_output` makes drift loud.

## Out of scope

- No behavioral changes. Pure refactor of the input shape.
- Callers (wallet, tests) migrate from the parallel-vec API to constructing the enum directly.
- The privacy circuit ELF image_id will change once. Expected.

moudyellaz:
## 🎯 Purpose

Refactor `PrivacyPreservingCircuitInput` from four parallel vectors (`visibility_mask`, `private_account_keys`, `private_account_nsks`, `private_account_membership_proofs`) into a single `Vec<InputAccountIdentity>` enum, eliminating implicit lock-step iteration and runtime length-coupling assertions in the privacy circuit guest. Closes #455.

## ⚙️ Approach

- [x] Add `InputAccountIdentity` enum in `nssa/core/src/circuit_io.rs` with six variants: `Public`, `PrivateAuthorizedInit`, `PrivateAuthorizedUpdate`, `PrivateUnauthorized`, `PrivatePdaInit`, `PrivatePdaUpdate`.
- [x] Replace parallel-vec fields in `PrivacyPreservingCircuitInput` with a single `account_identities: Vec<...>` field aligned 1:1 with the program's pre_states.
- [x] Rewrite `compute_circuit_output` and `derive_from_outputs` in the privacy circuit guest to consume `account_identities` directly, dropping the four lock-step iterators and trailing length asserts.
- [x] Rewrite `execute_and_prove` host signature to take `account_identities`.
- [x] Migrate ~50 call sites in `nssa/src/state.rs`, `nssa/src/privacy_preserving_transaction/circuit.rs`, sequencer test, and integration test.
- [x] Refactor `AccountManager` (`wallet/src/privacy_preserving_tx.rs`) to expose `account_identities()` directly, eliminating the parallel-vec accessors and the wallet-side iterator-flatten that would have replicated the same anti-pattern at the wallet/circuit boundary.
- [x] Rebake `privacy_preserving_circuit.bin`.

## 🧪 How to Test

`RISC0_DEV_MODE=1 cargo test --release` on `nssa-core` and `nssa`.

## 🔗 Dependencies

None.

## 🔜 Future Work

The "authorized init of a private PDA" sub-case (an authorized PDA pre-state that is still default) is structurally allowed by the guest's private-PDA paths today but is not produced by any wallet code. If a future PR enables it via `(seed, owner)` side input for top-level reuse, the variant set will need to be revisited.

## 🗑️ Removed tests

Seven tests have been removed because each one exercised a parallel-vec length mismatch that the new `Vec<InputAccountIdentity>` makes structurally impossible:

- `circuit_fails_if_insufficient_nonces_are_provided`
- `circuit_fails_if_insufficient_keys_are_provided`
- `circuit_fails_if_insufficient_commitment_proofs_are_provided`
- `circuit_fails_if_insufficient_auth_keys_are_provided`
- `circuit_should_fail_with_too_many_nonces`
- `circuit_should_fail_with_too_many_private_account_keys`
- `circuit_should_fail_with_too_many_private_account_auth_keys`

Each variant of `InputAccountIdentity` carries exactly the fields the corresponding circuit code path needs, so over-supply or under-supply of any per-account field is a compile-time error rather than a runtime panic.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


nssa/core/src/circuit_io.rs
```diff
@@ -12,23 +12,86 @@ use crate::{
...
+    /// rely on this position alignment.
+    pub accounts: Vec<PrivacyPreservingCircuitInputAccount>,
     /// Program ID.
     pub program_id: ProgramId,
 }
 
+/// Per-account input to the privacy-preserving circuit. Each variant carries exactly the fields
+/// the guest needs for that account's code path.
+#[derive(Serialize, Deserialize, Clone)]
+pub enum PrivacyPreservingCircuitInputAccount {
```
Pravdyvy: I think this naming is ambiguous on the size of variable. We have now 2 accounts types, one holds keys, other holds data and balance. I propose to rename this enum into something like `PrivacyPreservingCircuitInputAccountIdentity`, for example `PrivacyPreservingCircuitInputAccountIdentity` and name variables of this enum `account_identity`.

Pravdyvy (commented):
Thank you for PR. Left one suggestion.

nssa/core/src/circuit_io.rs
schouhy: Agree. Also I think we can drop the `PrivacyPreservingCircuit` prefix as it already is defined inside `circuit_io`

nssa/core/src/circuit_io.rs
```diff
@@ -12,23 +12,86 @@ use crate::{
...
+    /// Public account. The guest reads pre/post state from `program_outputs` and emits no
+    /// commitment, ciphertext, or nullifier.
+    Public,
+    /// Init of an authorized standalone private account (mask 1, no membership proof). The
```
schouhy: I'd remove all mentions to mask as that's now legacy and would be confusing for someone without context of previous implementations

nssa/src/state.rs
```diff
@@ -1868,137 +1862,24 @@ pub mod tests {
...
-            vec![Some((0, vec![]))],
-            &program.into(),
-        );
-
-        assert!(matches!(result, Err(NssaError::CircuitProvingError(_))));
-    }
+    // The four `circuit_fails_if_insufficient_*_are_provided` tests have been removed in this
+    // refactor: each variant of `PrivacyPreservingCircuitInputAccount` carries exactly the fields
+    // the corresponding circuit code path needs, so the parallel-vec length mismatches those
+    // tests exercised are now compile-time impossibilities.
```
schouhy: agreed with removing the tests. But let's not add this comment here. Let's put it in the PR description instead

schouhy (approved):
Thanks for the PR. This is a major improvement for readability of the code. Left minor comments

nssa/core/src/circuit_io.rs
moudyellaz: Done, renamed the enum to InputAccountIdentity, variant-of-this-enum variables to account_identity, and the struct field to account_identities.

moudyellaz: Done, dropped the PrivacyPreservingCircuit prefix. The type is now InputAccountIdentity, re-exported from nssa_core::lib.rs.

nssa/core/src/circuit_io.rs
moudyellaz: Done.

nssa/src/state.rs
moudyellaz: Done, removed the in-code comments and added a Removed tests section to the PR description.

Pravdyvy (approved):
lgtm

PR #464, opened 2026-05-01
feat!: Diversify private pdas by identifier

schouhy:
## 🎯 Purpose

Adds an `identifier` parameter to private PDA address derivation, giving each (program_id, seed, npk) tuple a family of 2^128 distinct addresses instead of a single fixed one. This unblocks use cases where a program needs to manage multiple independent private PDAs per user. The main use case for this is private multisig executions.

## ⚙️ Approach
The` AccountId::for_private_pda` formula now takes a fourth input, identifier, which is hashed into the address. To let the receiver reconstruct the correct `AccountId` after decryption, the ciphertext header is replaced by a fixed sized `PrivateAccountKind` payload that encodes either `Regular(identifier)` or` Pda { program_id, seed, identifier }`.

  - Add identifier param to `AccountId::for_private_pda`
  - Add `PrivateAccountKind` enum with fixed-length `to_header_bytes` / `from_header_bytes`.
  - Update circuit: `private_pda_bound_positions` becomes a `HashMap<usize, (ProgramId, PdaSeed)>` so the circuit can inject the correct `PrivateAccountKind::Pda` into the output ciphertext
  - `InputAccountIdentity::PrivatePdaInit` and `PrivatePdaUpdate` carry the identifier
  - Refactor wallet `PrivacyPreservingAccount`: replace `PrivatePda { nsk, program_id, seed, ... }` with `PrivatePdaOwned(AccountId)` and `PrivatePdaForeign { account_id, npk, vpk, identifier }`
  - Update wallet sync to reconstruct `AccountId` from the decrypted `PrivateAccountKind` instead of assuming the regular formula
  - Add `auth_transfer_proxy` test program for funding and spending private PDAs through a chained call

## ⚠️ Breaking changes
- `AccountId::for_private_pda` takes a new identifier argument. Programs that compute private PDA addresses must now pass an explicit identifier.
- Existing encrypted private account data is unreadable. The ciphertext format changed.

## 🧪 How to Test

The key new test is `private_pda_family_members_receive_and_spend` in `integration_tests/tests/private_pda.rs`. It funds two PDAs from the same seed and npk, with identifier 0 and 1. Then syncs and spends from both.

## 🔗 Dependencies

None

## 🔜 Future Work

Make it possible to use private pdas in an execution without a chained call or a `Claim::Pda`. This PR uses the new test program for both funding a withdrawing, but it would be better to have the possibility to fund the private pda by a direct execution of the auth-transfer program.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


nssa/core/src/program.rs
```diff
@@ -43,6 +48,89 @@ impl AsRef<[u8]> for PdaSeed {
...
+                let identifier = Identifier::from_le_bytes(bytes[65..81].try_into().unwrap());
+                Some(Self::Pda {
+                    program_id,
+                    seed,
+                    identifier,
+                })
+            }
+            _ => None,
+        }
+    }
```
Arjentix, line 107: What about bincode?

nssa/core/src/program.rs
```diff
@@ -65,34 +153,51 @@ impl AccountId {
...
                 .try_into()
                 .expect("Hash output must be exactly 32 bytes long"),
         )
     }
+
+    /// Derives the [`AccountId`] for a private account from the nullifier public key and kind.
+    #[must_use]
+    pub fn for_private_account(npk: &NullifierPublicKey, kind: &PrivateAccountKind) -> Self {
+        match kind {
+            PrivateAccountKind::Regular(identifier) => Self::from((npk, *identifier)),
```
Arjentix: Maybe time to change this `from` conversion to a proper constructor?

wallet/src/lib.rs
```diff
@@ -538,24 +542,36 @@ impl WalletCore {
...
+                                    program_id,
+                                    seed,
+                                    identifier,
+                                } => nssa::AccountId::for_private_pda(
+                                    program_id,
+                                    seed,
+                                    npk,
+                                    *identifier,
+                                ),
+                            };
```
Arjentix: I think you have `AccountId::for_private_account()` which does the same matching

wallet/src/privacy_preserving_tx.rs
```diff
@@ -20,15 +19,16 @@ pub enum PrivacyPreservingAccount {
         vpk: ViewingPublicKey,
         identifier: Identifier,
     },
-    /// A private PDA with externally-provided keys. The caller resolves the keys
-    /// (e.g. via `GroupKeyHolder::derive_keys_for_pda`) before constructing this variant.
-    /// The wallet computes the `AccountId` via `AccountId::for_private_pda(program_id, seed, npk)`.
-    PrivatePda {
-        nsk: NullifierSecretKey,
+    /// An owned private PDA: wallet holds the nsk/npk; `account_id` was derived via
+    /// `AccountId::for_private_pda`. Produces visibility mask 3.
```
Arjentix: Let's drop `mask` terminology. And also let's use `[``]` for links

Arjentix (approved):
LGTM, thanks!

The only thing that worries me is the complexety of user keychain and account ids in general grows very fast in recent time. But looks like there is no better option

nssa/core/src/program.rs
schouhy: I wouldn't use it here. I'm not sure how well specified bincode is. But switching to borsh to reduce code overhead.

program_methods/guest/src/bin/privacy_preserving_circuit.rs
```diff
@@ -353,18 +358,20 @@ impl ExecutionState {
...
-                            let pda = AccountId::for_private_pda(&program_id, &seed, npk);
+                            let pda =
+                                AccountId::for_private_pda(&program_id, &seed, npk, *identifier);
                             assert_eq!(
                                 pre_account_id, pda,
                                 "Invalid private PDA claim for account {pre_account_id}"
                             );
-                            self.private_pda_bound_positions.insert(pre_state_position);
+                            self.private_pda_bound_positions
+                                .insert(pre_state_position, (program_id, seed));
```
moudyellaz: Hey, quick thought on this: the HashMap silently overwrites on duplicate keys. If both binding paths land on the same position, whichever runs last wins, and the `(program_id, seed)` we end up writing into the ciphertext header might not match what was used to derive `pre_account_id`. I'm pretty sure both paths always agree (the equality assertion above weeds out the bad cases), but it took me a minute to convince myself. Mind either asserting equality on duplicate insert, or leaving a comment proving they can't diverge? Same spot at line 480.

wallet/src/privacy_preserving_tx.rs
```diff
@@ -20,15 +19,16 @@ pub enum PrivacyPreservingAccount {
...
-    /// (e.g. via `GroupKeyHolder::derive_keys_for_pda`) before constructing this variant.
-    /// The wallet computes the `AccountId` via `AccountId::for_private_pda(program_id, seed, npk)`.
-    PrivatePda {
-        nsk: NullifierSecretKey,
+    /// An owned private PDA: wallet holds the nsk/npk; `account_id` was derived via
+    /// [`AccountId::for_private_pda`].
+    PrivatePdaOwned(AccountId),
+    /// A foreign private PDA: wallet knows the recipient's npk/vpk but not their nsk.
+    /// Uses a default (uninitialised) account.
+    PrivatePdaForeign {
```
moudyellaz, line 27: Tiny nit: with the old `PrivatePda` the wallet computed the AccountId itself, so a mismatch was caught early. Now we trust whatever the caller passes and a bad AccountId only surfaces inside the circuit. Could we add a doc-comment spelling out that the caller has to pass `account_id == for_private_pda(program_id, seed, npk, identifier)`? Just to make the contract obvious.

nssa/core/src/program.rs
```diff
@@ -43,6 +58,55 @@ impl AsRef<[u8]> for PdaSeed {
...
+    pub fn to_header_bytes(&self) -> [u8; Self::HEADER_LEN] {
+        let mut bytes = [0_u8; Self::HEADER_LEN];
+        let serialized = borsh::to_vec(self).expect("borsh serialization is infallible");
+        bytes[..serialized.len()].copy_from_slice(&serialized);
+        bytes
+    }
+
+    #[cfg(feature = "host")]
+    #[must_use]
+    pub fn from_header_bytes(bytes: &[u8; Self::HEADER_LEN]) -> Option<Self> {
```
moudyellaz, line 105: Super minor: `deserialize(&mut bytes.as_ref())` doesn't enforce full-buffer consumption. That's actually what we want (trailing zero padding should be ignored), but `try_from_slice` does the same and also rejects truncated input. Either's fine, just flagging it.

moudyellaz (approved):
lgtm, thanks. Left some minor comments but i do not insist.

Pravdyvy (approved):
lgtm

program_methods/guest/src/bin/privacy_preserving_circuit.rs
schouhy: sure!

wallet/src/privacy_preserving_tx.rs
schouhy: ah I followed the same pattern as in standard private and public variants that only hold the account id. The account id is then used to fetch the keys

nssa/core/src/program.rs
schouhy: ah I thought try_from_slice enforces full buffer consumption

nssa/core/src/program.rs
Arjentix: Oh, yes, I mean borsh in my mind

PR #460, opened 2026-04-27
feat(wallet)!: wallet group commands with shared account derivation

moudyellaz:
## 🎯 Purpose

Adds wallet CLI commands for group key management and support for shared private accounts (both PDA and regular), building on the `GroupKeyHolder` primitive from PR #449.

## ⚙️ Approach

- [x] Add `derive_keys_for_shared_account` method with distinct domain separator for regular shared accounts
- [x] Add `derive_keys_for_pda` now includes `program_id` in derivation for per-program key separation
- [x] Add `PrivateShared` wallet variant for regular shared accounts and `PrivatePda` for PDA shared accounts
- [x] Add `SharedAccountEntry` struct with group label, identifier, optional PDA seed/program_id for sync tracking
- [x] Add 5 group CLI commands: `new`, `list`, `remove`, `invite`, `join`
- [x] Add `new-sealing-key` command for dedicated GMS distribution key pair (replaces ViewingKey reuse)
- [x] Add `PrivateGms` CLI variant on `wallet account new` for shared account creation (both regular and PDA)
- [x] Add `resolve_private_account` on `WalletCore` to resolve shared accounts across all program facades
- [x] Extend `sync_private_accounts_with_tx` to scan shared accounts with GMS-derived keys
- [x] Restructure and rename `private_pda_spender` to `auth_transfer_proxy`, delegating to `authenticated_transfer` via chained calls
- [x] Persist `group_key_holders`, `shared_private_accounts`, and `sealing_secret_key` in `PersistentStorage`
- [x] Add encapsulation methods on `NSSAUserData` (no direct field access)
- [x] Remove `Serialize`/`Deserialize` derives from `NSSAUserData`
- [x] Remove unsafe `export`/`import` commands (all GMS distribution goes through sealed `invite`/`join`)
- [x] Remove mask terminology from comments and docs

## 🧪 How to Test

- `cargo test -p key_protocol` — unit tests for derivation, serde, and domain separation
- `RISC0_DEV_MODE=1 cargo test -p nssa --release -- private_pda_init private_pda_withdraw shared_account_receives` — circuit tests
- `RISC0_DEV_MODE=1 cargo test -p integration_tests -- shared_accounts` — integration tests (2 pass, `fund_shared_account_from_public` ignored pending auth flow)
- CLI: `wallet group new-sealing-key`, `wallet group new <name>`, `wallet account new private-gms <group>`

## 🔗 Dependencies

- PR #449 (merged)

## ⚠️ Breaking Changes

[...]

PR #466, opened 2026-05-05
chore(ci): adopt conventional commits and auto-generated changelog

moudyellaz:
## 🎯 Purpose

Currently breaking changes ship without explicit communication, so downstream consumers (notably the LEZ programs in the separate repo) discover them at build time. This PR adopts Conventional Commits + auto-generated CHANGELOG.md so each release surfaces breaking changes in a single, parseable place.

## ⚙️ Approach

- [x] Add `cliff.toml` (`git-cliff` config) at repo root, grouping commits by type with breaking changes flagged via `**[BREAKING]**`.
- [x] Add `.github/workflows/pr-title.yml`, enforcing Conventional Commit PR titles via `amannn/action-semantic-pull-request@v5`. Combined with squash-merge + PR-title-as-subject, the `main` log stays conventional by construction.
- [x] Add `.github/workflows/release.yml`, firing on `v*` tag push: generates the latest-section markdown, regenerates `CHANGELOG.md`, commits it back to `main`, and creates a GitHub Release with the new section.
- [x] Add `CONTRIBUTING.md` documenting the format, allowed types, and the breaking-change marker conventions.

## 🧪 How to Test

- Push a non-conventional PR title on a throwaway branch and confirm the new `PR Title` check fails. Rename to a conventional title and confirm it passes.
- After this lands, push a test tag `v0.0.0-test` on a side branch (or trigger via `workflow_dispatch`) and confirm `CHANGELOG.md` regenerates and a GitHub Release is created with the latest-section content.
- Confirm the Releases page surfaces breaking-change entries (e.g. mark a follow-up commit with `feat(scope)!:` and verify it shows under `[BREAKING]`).

## 🔗 Dependencies

N/A.

## 🔜 Future Work

- Repo setting: enforce squash-merge with PR title as commit subject (out-of-tree GitHub setting, not a code change).
- Adopt `cargo-release` to automate version bumps + tag pushes per workspace crate.
- Backfill `CHANGELOG.md` for the previous release that included PR #446 (manual one-shot, then auto-generated from this point on).

## 📋 PR Completion Checklist

- [x] Complete PR description
- [ ] Add/update tests

### DIFF
--- a/.github/workflows/pr-title.yml
+++ b/.github/workflows/pr-title.yml
@@ -1,0 +1,26 @@
+name: PR Title
+
+on:
+  pull_request:
+    types: [opened, edited, synchronize, reopened]
+
+jobs:
+  validate:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: amannn/action-semantic-pull-request@v5
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          types: |
+            feat
+            fix
+            chore
+            docs
+            test
+            refactor
+            perf
+            build
+            ci
+            revert
+          requireScope: false
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,49 @@
+name: Release
+
+on:
+  push:
+    tags: ['v*']
+
+permissions:
+  contents: write
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Generate this-release section
+        uses: orhun/git-cliff-action@v3
+        id: cliff_latest
+        with:
+          config: cliff.toml
+          args: --latest --strip header
+
+      - name: Generate full CHANGELOG.md
+        uses: orhun/git-cliff-action@v3
+        with:
+          config: cliff.toml
+          args: --output CHANGELOG.md
+
+      - name: Commit CHANGELOG.md to main
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+          git fetch origin main
+          git checkout main
+          git add CHANGELOG.md
+          if git diff --cached --quiet; then
+            echo "No changelog changes"
+          else
+            git commit -m "chore: update CHANGELOG.md for ${{ github.ref_name }}"
+            git push origin main
+          fi
+
+      - name: Create GitHub Release
+        uses: softprops/action-gh-release@v2
+        with:
+          body: ${{ steps.cliff_latest.outputs.content }}
+          token: ${{ secrets.GITHUB_TOKEN }}
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,15 @@
+# Contributing
+
+## Commit message format
+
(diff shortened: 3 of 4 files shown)


.github/workflows/pr-title.yml
```diff
@@ -0,0 +1,26 @@
...
+            fix
+            chore
+            docs
+            test
+            refactor
+            perf
+            build
+            ci
+            revert
+          requireScope: false
```
0x-r4bbit, line 26: Not sure this is desired.
If this enforces you to also apply commit conventions to PR titles, then maybe you want to relax this a bit.

Sometimes you have a PR with multiple (semantic) commits, but the PR itself has a summary description of both.

moudyellaz: suggest what is the best practice and will accept the suggestion

Arjentix (approved):
LGTM, thanks.

Thought about forcing squashing for some time as well.

.github/workflows/pr-title.yml
0x-r4bbit: I'd say, give this a try and see if this gets in the way. Then you can still remove it.

I personally wouldn't force the conventions on PR titles/descriptions.
What matter is the commits. It easily happens to have multiple atomic commits per PR, in which case you don't want to be blocked just because your PR title doesn't follow some convention.

0x-r4bbit:
> Thought about forcing squashing for some time as well.

@Arjentix I guess this is up to you guys to decide, but this is also something I'd personally not do. It would prevent you from having multiple atomic commits in a single PR, because you'd be forced to squash them all into one (which is not what you want in this case).

So unless you're happy with needing to put up a PR for every single commit (can make reviewing easier as well), I would not force squashing.

People can easily squash locally and update their PRs, regardless of how many atomic commits should end up in the history.

.github/workflows/pr-title.yml
moudyellaz: @0x-r4bbit @Arjentix @Pravdyvy @schouhy  I would do this:

- I would leave the PR as is without merging it for now.
- I would use the PR title convention only
- I would add a section to the PR template "breaking changes"

Check [PR460](https://github.com/logos-blockchain/logos-execution-zone/pull/460) for example. Does it work for you like this @0x-r4bbit ?

Pravdyvy: I agree, there is no point in controlling PR titles. Commits hold all the necessary data.

Pravdyvy (approved):
lgtm. Deny looks unfixable now, need to add another item into deny.toml.

schouhy:
> > Thought about forcing squashing for some time as well.
> 
> @Arjentix I guess this is up to you guys to decide, but this is also something I'd personally not do. It would prevent you from having multiple atomic commits in a single PR, because you'd be forced to squash them all into one (which is not what you want in this case).
> 
> So unless you're happy with needing to put up a PR for every single commit (can make reviewing easier as well), I would not force squashing.
> 
> People can easily squash locally and update their PRs, regardless of how many atomic commits should end up in the history.

I second this. I would try to have better commit habits to avoid having "wip" commits merged to main before forcing squashing.

.github/workflows/pr-title.yml
schouhy: > [@0x-r4bbit](https://github.com/0x-r4bbit) [@Arjentix](https://github.com/Arjentix) [@Pravdyvy](https://github.com/Pravdyvy) [@schouhy](https://github.com/schouhy) I would do this:
> 
> * I would leave the PR as is without merging it for now.
> * I would use the PR title convention only

without title convention being enforced by the CI you mean?
> * I would add a section to the PR template "breaking changes"

sounds good

Arjentix:
The problem with not-squashed commit is that it's hard to find in git history the exact PR which introduced this commits to see the full picture. However I agree that it breaks atomic commits in one PR. Not a strong opinion, I'm fine with either way.

But I'd suggest forcing rebasing instead of merging to keep history clean.

Arjentix:
Closed in favor of #483

PR #471, opened 2026-05-07
Add getAccountAtBlock RPC to indexer

Thompsonmina:
## 🎯 Purpose

Here's the draft PR as promised @Arjentix.

The indexer's public RPC currently exposes only *current* account state via `getAccount` — there's no way to query an account's state at a historical block. This PR adds `getAccountAtBlock` to close that gap.

The internal capability already exists (`BlockStore::get_state_at_block` reconstructs the public state at any block); this PR just plumbs it through to the public RPC surface in a per-account-shaped query, following your preference for `getAccountAtBlock`-style methods over a bulk `getStateAtBlock` that returns raw `V03State`.

The new method is broadly useful for explorers, analytics, and any consumer reasoning about historical account state.

## ⚙️ Approach

A new RPC method on the indexer that, for a given `(account_id, block_id)`, reconstructs the public state at `block_id` (via the existing `BlockStore::get_state_at_block`) and returns the looked-up `Account` — a default `Account` is returned for ids not present in the public state at that block, matching the behavior of `V03State::get_account_by_id`.

- [x] Add `getAccountAtBlock` to the indexer RPC trait (`indexer/service/rpc/src/lib.rs`)
- [x] Implement on `IndexerService` (`indexer/service/src/service.rs`)
- [x] Implement on `MockIndexerService` — returns current state regardless of `block_id` (`indexer/service/src/mock_service.rs`)
- [x] Add `account_state_at_block` helper on `IndexerStore`, parallel to the existing `account_current_state` (`indexer/core/src/block_store.rs`)
- [x] Add unit test exercising historical balance lookups across genesis, mid-chain, and tip blocks

**Performance note:** the implementation is intentionally naive — each call reconstructs the full `V03State` at the requested block, then indexes by `AccountId`. This was previously not a use case the indexer was tuned for, so there's no per-account historical side index to lean on.

## 🧪 How to Test

Run the new unit test alongside the existing block-store tests:

```sh
RISC0_DEV_MODE=1 cargo nextest run -p indexer_core block_store::tests
```

Expected: 3 tests pass — `correct_startup`, `state_transition`, and the new `account_state_at_block`. The new test builds a chain with native-token transfers across 8 blocks and asserts that historical balance lookups at genesis, mid-chain (block 5), and tip (block 9) return the correct values.

End-to-end RPC verification (manual, optional):
1. Run an indexer service locally that has at least a few blocks with state changes.
2. JSON-RPC `getAccountAtBlock` with `(account_id, block_id)` of an account whose state is known to have changed.
3. Confirm the returned `Account` matches the expected historical state.

## 🔗 Dependencies

None.

## 🔜 Future Work

Intentionally out of scope for this PR; happy to follow up if there's appetite:

- Performance: a per-account state side index (mirroring the `(account_id → tx_list)` index that backs `getTransactionsByAccount`) would turn the current O(N) state reconstruction into an O(log) read.
- A `getAccountsAtBlock(account_ids: Vec<AccountId>, block_id)` batch variant — useful for snapshot workflows that walk many accounts at the same block; lets a single state reconstruction serve many lookups.
- A `getAccountAtBlockHash(account_id, block_hash)` variant parallel to `getBlockByHash` if hash-based block addressing is wanted on this query.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests

### DIFF
--- a/indexer/core/src/block_store.rs
+++ b/indexer/core/src/block_store.rs
@@ -135,6 +135,12 @@ impl IndexerStore {
             .get_account_by_id(*account_id))
     }
 
+    pub fn account_state_at_block(&self, account_id: &AccountId, block_id: u64) -> Result<Account> {
+        Ok(self
+            .get_state_at_block(block_id)?
+            .get_account_by_id(*account_id))
+    }
+
     pub async fn put_block(&self, mut block: Block, l1_header: HeaderId) -> Result<()> {
         {
             let mut state_guard = self.current_state.write().await;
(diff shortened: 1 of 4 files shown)


Pravdyvy (approved):
lgtm. 
Deny chech can be fixed with similar modifications to #456 in file `.deny.toml`.

Arjentix (approved):
LGTM, thanks

Arjentix:
Regarding further work I'd suggest implement it if you'll notice performance issues. For now I'd like to keep things simple

PR #472, opened 2026-05-07
docs: Add specs

schouhy:
## 🎯 Purpose

Add specifications of the system to `docs/specs.md`

## ⚙️ Approach

N/A

## 🧪 How to Test

N/A

## 🔗 Dependencies

None

## 🔜 Future Work

None

## 📋 PR Completion Checklist

- [x] Complete PR description
~~- [ ] Implement the core functionality~~
~~- [ ] Add/update tests~~


Pravdyvy (approved):
lgtm

docs/specs.md
```diff
@@ -0,0 +1,1629 @@
...
+```rust
+/// ASCII "/LEE/v0.3/AccountId/Public/" zero-padded to 32 bytes
+PUBLIC_ACCOUNT_ID_PREFIX: [u8; 32] = b"/LEE/v0.3/AccountId/Public/\x00\x00\x00\x00\x00"
+```
+
+**Private account ID:**
+
+$$\mathsf{AccountId} = \mathsf{SHA256}(\mathsf{PRIVATE\_ACCOUNT\_ID\_PREFIX} \;||\; \mathsf{Npk} \;||\; \mathsf{identifier\_le})$$
+
+The hash input is 80 bytes: 32-byte prefix + 32-byte `Npk` + 16-byte little-endian `identifier`. Each `(Npk, identifier)` pair yields a distinct account ID, so a single keypair controls up to $2^{128}$ private accounts.
```
jonesmarvin8: For little-endian. It might be good to mention that LEE/LEZ uses little-endian except for key protocol (due to BIP-032).

docs/specs.md
```diff
@@ -0,0 +1,1629 @@
...
+    }
+}
+```
+
+### Nonce field
+
+The nonce is a 128-bit integer value. It has different uses depending on the visibility of the account:
+
+- **Public accounts:** The nonce counts the number of transactions in which the associated public key of the account appears as a signer. This serves as a sequence number to prevent replay of transactions involving this account.
+- **Private accounts:** A pseudorandom value used to provide entropy for the account's commitment, making it unconditionally hiding. The initial nonce is derived from the account ID, and subsequent nonces are derived from the nullifier secret key and the previous nonce.
```
jonesmarvin8: Could be good to mention: like the public variant, private account's nonce is iteratively produced. It's implicitly mentioned, but might be good to be explicit.

docs/specs.md
```diff
@@ -0,0 +1,1629 @@
+# LEZ v0.3 specifications
```
jonesmarvin8: Could we divide these into separate files to be easier for parsing? Additionally, maybe we could provide context with respect to repo?

E.g., features/primitives/logic we could specify which crate these are defined in? Which file might be more detailed than we want. But that could be useful.

docs/specs.md
```diff
@@ -0,0 +1,1629 @@
...
+### Balance field
+
+The number of native tokens held by the account. It's represented as a 128-bit number. That means, the total supply of the system should never exceed $2^{128} - 1$. As long as that is guaranteed, transfer operations on the balance will not overflow. This is because under normal transfer operations, no account can end up with a balance greater than the total supply of the system. NSSA prevents overflow exploits on the balance field by upcasting each term to u256 when computing the total balances before and after an execution.
+
+### Data field
+
+An arbitrary byte string field to be managed by the `program_owner` program. The content and interpretation of this field are defined by the program logic. The maximum size of it is defined by the constant `DATA_MAX_LENGTH_IN_BYTES`.
+
+Private accounts use a *nullifier public key* (`Npk`) as a core identifier. The next two subsections derive `Npk` and define the account ID formats — both are prerequisites for the nonce, commitment, and nullifier fields that follow.
+
```
jonesmarvin8, line 150: The flow of this document defines `Account` and, seemingly, defines each entry `Program owner field`, `Balance field` and `Data field`. But `Nonces` is shifted.

I think it might make sense to have the `Nonces field` explained here. Of course, `Npk`/`nsk` aren't defined at this, point but maybe we can add a link for a reader to reference these?

docs/specs.md
```diff
@@ -0,0 +1,1629 @@
...
+The hash input is 80 bytes: 32-byte prefix + 32-byte `Npk` + 16-byte little-endian `identifier`. Each `(Npk, identifier)` pair yields a distinct account ID, so a single keypair controls up to $2^{128}$ private accounts.
+
+```rust
+/// ASCII "/LEE/v0.3/AccountId/Private/" zero-padded to 32 bytes
+PRIVATE_ACCOUNT_ID_PREFIX: [u8; 32] = b"/LEE/v0.3/AccountId/Private/\x00\x00\x00\x00"
+```
+
+```rust
+impl AccountId {
+    fn from_private(npk: &NullifierPublicKey, identifier: Identifier) -> Self {
```
jonesmarvin8, line 199: Might be good to emphasize here that `identifier` inclusion ensures that the same set of private account keys can be used for multiple private accounts.

docs/specs.md
```diff
@@ -0,0 +1,1629 @@
...
+### Account default value
+
+```rust
+Account {
+    balance: 0,
+    program_owner: [0; 8],
+    data: Data::default(),
+    nonce: 0,
+}
+```
```
jonesmarvin8: Can we include this at the initial definition of `Account`?

docs/specs.md
```diff
@@ -0,0 +1,1629 @@
...
+
+where `Data::default()` is the unique zero-length byte array.
```
jonesmarvin8: Maybe include this as a comment in the pseudocode?

docs/specs.md
```diff
@@ -0,0 +1,1629 @@
...
+    let total_post = WrappedBalanceSum::from_balances(post_states.iter().map(|p| p.account.balance));
+    if total_pre != total_post {
+        return Err(ExecutionValidationError::MismatchedTotalBalance { .. });
+    }
+
+    Ok(())
+}
+```
+
+The `is_authorized` flag in `AccountWithMetadata` indicates whether the account owner has provided authorization. It is the program's responsibility to check it where needed. Authorization is granted via different mechanisms depending on the account type:
```
jonesmarvin8, line 583: Should mention that PDAs are authorized in a different manner.

docs/specs.md
```diff
@@ -0,0 +1,1629 @@
...
+                AccountPostState::new_claimed_if_default(recipient_post, Claim::Authorized),
+            ]
+        }
+    };
+
+    (pre_states, post_states, vec![])
+}
+```
+
+### Piñata program
```
jonesmarvin8: Should we include pinata token program as well?

docs/specs.md
```diff
@@ -0,0 +1,1629 @@
...
+    /// A printed instance of an NFT.
+    NftPrintedCopy { definition_id: AccountId, owned: bool },
+}
+
+pub struct TokenMetadata {
+    pub definition_id: AccountId,
+    pub standard: MetadataStandard,
+    pub uri: String,
+    pub creators: Vec<...>,
+    pub primary_sale_date: u64,
```
jonesmarvin8: At the beginning of the specs, a list of custom data types are defined (Balance, Timestamp etc). It would make sense to replace the various u128, u64, etc with the appropriate custom data type. E.g., `pub primary_sale_date: Timestamp`.

docs/specs.md
```diff
@@ -0,0 +1,1629 @@
...
+    reserve_a: u128,
+    reserve_b: u128,
+    fees: u128,
+    active: bool,
+}
+```
+
+**Initialize pool:**
+
+Creates the pool definition, two vaults (one per token), and an LP token definition. Deposits the initial liquidity through chained Token Program calls.
```
jonesmarvin8: Might be good to mention that each token pair generate a unique pool.

docs/specs.md
```diff
@@ -0,0 +1,1629 @@
...
+    PrivatePdaUpdate {
+        ssk: SharedSecretKey,
+        nsk: NullifierSecretKey,
+        membership_proof: MembershipProof,
+        identifier: Identifier,
+    },
+}
+```
+
+The `ssk` (shared secret key) is the 32-byte x-coordinate of the ECDH shared point between the sender's ephemeral key and the recipient's viewing public key. It is used to encrypt the post-state.
```
jonesmarvin8: `ssk` might be confusing since the key protocol uses `ssk` as `spending secret key` which generates `nsk` and `vsk`.

jonesmarvin8: The shared secret key is generated two ways:
- ephemeral secret key + recipient's viewing public key, or
- ephemeral public key + recipient's viewing secret key.

docs/specs.md
```diff
@@ -0,0 +1,1629 @@
...
+| Purpose | Domain separator |
+|---------|-----------------|
+| Commitment | `b"/LEE/v0.3/Commitment/\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00"` |
+| Nullifier — initialization | `b"/LEE/v0.3/Nullifier/Initialize/\x00"` |
+| Nullifier — update | `b"/LEE/v0.3/Nullifier/Update/\x00\x00\x00\x00\x00"` |
+| Account ID — public | `b"/LEE/v0.3/AccountId/Public/\x00\x00\x00\x00\x00"` |
+| Account ID — private | `b"/LEE/v0.3/AccountId/Private/\x00\x00\x00\x00"` |
+| Account ID — public PDA | `b"/NSSA/v0.2/AccountId/PDA/\x00\x00\x00\x00\x00\x00\x00"` |
+| Account ID — private PDA | `b"/LEE/v0.3/AccountId/PrivatePDA/\x00"` |
+| Nullifier public key (from NSK) | `b"LEE/keys"` + nsk + `[7]` + `[0; 23]` |
```
jonesmarvin8, line 1016: What about the domain separators used for other account keys?

docs/specs.md
```diff
@@ -0,0 +1,1629 @@
...
+    for (nullifier, digest) in &message.new_nullifiers {
+        assert!(!nssa_state.private_state.1.contains(nullifier));
+        assert!(nssa_state.private_state.0.is_current_or_previous_digest(digest));
+    }
+}
+```
+
+### Note on replay attacks
+
+Replay attacks are not possible for privacy transactions. Any accepted transaction either adds commitments or nullifiers to the state. A replay attempt is rejected by commitment freshness or nullifier uniqueness checks.
```
jonesmarvin8, line 1290: For shielded transaction, the public account's nonce protects against replay attack.

jonesmarvin8 (approved):
Looks good. I left comments that may improve this document.

docs/specs.md
schouhy: done

docs/specs.md
schouhy: done

docs/specs.md
schouhy: ah I wouldn't specify the crates in the specs. I think specs should be independent of our rust implementation. It's the implementation that should reference the specs where important. Ideally specs should be general enough to implement the system in different languages. The rust-like pseudocode in the specs in sync with the code is imo a good balance to improve code auditing of our implementation.
Splitting will have the effect that files are either not self contained or repetitive. I like the one file approach for simplicity of navigation. But if it's an unpopular opinion let's split it.

docs/specs.md
schouhy: true, thanks. I changed it

docs/specs.md
schouhy: done

docs/specs.md
schouhy: done

docs/specs.md
schouhy: done. refactored it a bit

docs/specs.md
schouhy: thanks. done

docs/specs.md
schouhy: I wouldn't include it tbh. Actually I think we should remove all programs soon from the specs

docs/specs.md
schouhy: done

docs/specs.md
schouhy: done. I also mention that this uniqueness is per namespace (defined by npk and identifier) in the private state

docs/specs.md
schouhy: true, but we have `ssk` in our code too. Added some comments to this

schouhy: done

docs/specs.md
schouhy: what do you mean other account keys?

docs/specs.md
schouhy: what I meant is that replay attacks are not possible because of commitments and nullifier freshness checks. Even a shielded transaction would add commitments and nullifiers. That alone already prevents replay attacks

docs/specs.md
```diff
@@ -0,0 +1,1565 @@
...
+            &call.pre_states,
+            &call.instruction_data,
+        );
+
+        // Compute the set of public PDAs the callee is authorized to mutate via the
+        // caller's pda_seeds. For the top-level call (caller_program_id == None) this is
+        // always the empty set.
+        let authorized_pdas =
+            compute_public_authorized_pdas(caller_program_id, &call.pda_seeds);
+        let is_authorized = |account_id: &AccountId| {
```
moudyellaz: Quick question on the authorization model here. The lambda uses signer_account_ids for the recursive check, but the actual code at validated_state_diff.rs:108-156 threads a per-call caller_data.authorized_accounts set down the chain, populated from the parent's program_output.pre_states.filter(is_authorized). The two agree for public accounts (only signers can ever be authorized), but for PDAs the code allows multi-hop propagation while the spec restricts to one hop from the immediate caller.

docs/specs.md
```diff
@@ -0,0 +1,1565 @@
...
+| `PrivateUnauthorized` | `from_private(npk, ident)` | `nonce_init(account_id)` | init nullifier | none | **not enforced** (skipped) |
+| `PrivatePdaInit` | `for_private_pda(prog, seed, npk, ident)` | `nonce_init(account_id)` | init nullifier | pda_seeds/Claim::Pda | `is_authorized` must be `true` |
+| `PrivatePdaUpdate` | `for_private_pda(prog, seed, npk(nsk), ident)` | `nonce_increment(nsk)` | update nullifier | nsk + membership proof | `is_authorized` must be `true` |
+
+For each private account the post-state commitment is `Commitment::new(account_id, post_account)` (with the new nonce applied). The ciphertext is `EncryptionScheme::encrypt(post_account, kind, ssk, commitment, output_index)`.
+
+The chain-of-calls logic and `validate_execution` rules are identical to the public execution path. The claiming rules diverge for **standalone private accounts only**: `Claim::Authorized` on a `Regular` private kind is allowed unconditionally (no `is_authorized` check), because operating these accounts already requires possession of the corresponding `nsk`. For public accounts and private PDAs, `Claim::Authorized` enforcement matches the public path.
```
moudyellaz: The chain-of-calls logic is shared, but the privacy circuit guest at program_methods/guest/src/bin/privacy_preserving_circuit.rs has additional re-seen-account semantics (the Vacant vs Occupied branches in ExecutionState::push_states, plus resolve_authorization_and_record_bindings for accounts seen at multiple chain levels). The current text positions the privacy path as a strict subset of the public one, but the re-seen-account rule belongs here too. A sentence or short paragraph would close the gap.

docs/specs.md
```diff
@@ -0,0 +1,1565 @@
...
+
+**Workflow:**
```
moudyellaz: The privacy circuit also enforces MAX_NUMBER_CHAINED_CALLS, but the constraint is only mentioned in the public acceptance criteria. One-line cross-reference in the workflow steps would close the gap.

docs/specs.md
```diff
@@ -0,0 +1,1565 @@
...
+    Ok(())
+}
+```
+
+The `is_authorized` flag in `AccountWithMetadata` indicates whether the account owner has provided authorization. It is the program's responsibility to check it where needed. Authorization is granted via different mechanisms depending on the account type:
+
+- For public accounts: through digital signatures.
+- For private accounts: through knowledge proofs of the corresponding nullifier secret key.
+
+### Account authority vs program ownership
```
moudyellaz, line 589: A one-liner that is_authorized is recomputed per chained-call boundary from the parent's validated output, not taken verbatim from chained_call.pre_states (which the caller controls), would make the authorization model self-evident here.

moudyellaz (approved):
lgtm, thanks. Left some minor comments.

docs/specs.md
jonesmarvin8: I'm fine with this. Especially since you mentioned potentially removing programs from the document. At very least separating those out would reduce the file's length.

docs/specs.md
jonesmarvin8: I agree. At least, put programs in their own document for the time being.

docs/specs.md
jonesmarvin8: I think `ssk` in the key protocol as spending secret key probably should be renamed for clarity. This would have the least impact on the code base too.

docs/specs.md
jonesmarvin8: Viewing secret key uses `b"LEE/keys" + ssk + [2] + [i_u32] + [0;19]`,
nullifier secret key uses `b"LEE/keys" + ssk + [1] + [i_u32] + [0;19]`

docs/specs.md
jonesmarvin8: I think this comment was for `npk` and `identifier` producing unique AccountId and allowing the same account keys to be used for multiple private accounts.

Doesn't appear that this comment was addressed.

docs/specs.md
schouhy: ah yes that went outdated with the last merge from main thanks

docs/specs.md
schouhy: ah I added it right after the the AMM program title [here](https://github.com/logos-blockchain/logos-execution-zone/blob/5a6ae2a3ff8a4782827ad86110d468ccc9a81f20/docs/specs.md?plain=1#L852)

docs/specs.md
schouhy: ah right but those are not system specifications. Those are wallet standards for generating keys. I mean I think there's no constraint on those keys that should have that form

docs/specs.md
schouhy: ah I didn't understand what you mean here

docs/specs.md
```diff
@@ -0,0 +1,1565 @@
...
+- **Pre-state consistency:** Each `program_output.pre_states[i]` must equal the current state of `pre_states[i].account_id` (or the diffed value from earlier chained calls). The `is_authorized` flag in each pre-state must match the actual authorization status.
+- **Program identity consistency:** Each program output's `self_program_id` must equal the program ID it was invoked under, and its `caller_program_id` must equal the caller (or `None` for the top-level call).
+- **Validity window enforcement:** For *each* `ProgramOutput` in the chain (top-level and chained calls), the current `block_id` must fall within `program_output.block_validity_window` and the current `timestamp` must fall within `program_output.timestamp_validity_window`. Out-of-window outputs are rejected.
+- **Maximum chained-call depth:** The total number of program executions (including the top-level call) must not exceed `MAX_NUMBER_CHAINED_CALLS` (10).
```
moudyellaz: Code has an off-by-one. validated_state_diff.rs:125 and execution_state.rs:131 check `chain_calls_counter <= MAX_NUMBER_CHAINED_CALLS` with counter starting at 0 and incrementing after execution, allowing 11 executions. Should be `<`. Affects public path and circuit.

docs/specs.md
```diff
@@ -0,0 +1,1565 @@
...
+- `balance_to_move == 0` with a single pre-state: claims the pre-state account on behalf of the caller (a self-init).
+- `balance_to_move != 0` with two pre-states `[sender, recipient]`: standard transfer. The recipient is claimed only if its `program_owner` is currently the default — pre-existing accounts owned by other programs keep their owner.
+
+```rust
+/// Instruction data: u128 amount to transfer (0 means initialize-account).
+fn transfer_authorized(
+    self_program_id: ProgramId,
+    caller_program_id: Option<ProgramId>,
+    pre_states: Vec<AccountWithMetadata>,
+    balance_to_move: u128,
```
moudyellaz: Spec describes the auth-transfer instruction as a u128, but on-chain it's `enum Instruction { Transfer { amount: u128 }, Initialize }` (programs/authenticated_transfer/core/src/lib.rs), and the guest at program_methods/guest/src/bin/authenticated_transfer.rs matches on the enum variants. Spec line 712 pseudocode (`instruction: balance_to_move (u128)`) and line 707 description need to be updated to reflect the borsh-encoded enum.

docs/specs.md
```diff
@@ -0,0 +1,1565 @@
...
+    self_program_id: ProgramId,
+    caller_program_id: Option<ProgramId>,
+    pre_states: Vec<AccountWithMetadata>,
+    solution: u128,
+) -> (Vec<AccountWithMetadata>, Vec<AccountPostState>, Vec<ChainedCall>) {
+    let [pinata, winner] = <[_; 2]>::try_from(pre_states.clone()).unwrap();
+
+    let challenge = Challenge::parse(&pinata.account.data);
+    if !challenge.validate_solution(solution) {
+        // No valid solution: emit unchanged states and return.
```
moudyellaz: Spec pseudocode at line 762-769 has the program emit a no-op ProgramOutput (unchanged pre/post states, no chained calls) when the solution fails validation. Code at program_methods/guest/src/bin/pinata.rs early-returns from `fn main()` without calling `ProgramOutput::new(...).write()`, so no journal is committed and the transaction fails proof verification on the host. Spec describes a recoverable no-op, code rejects.

docs/specs.md
```diff
@@ -0,0 +1,1565 @@
...
+### Witness set
+
+- `signatures_and_public_keys`: BIP-340 Schnorr signature pairs for any public accounts requiring authorization.
+- `proof`: A borsh-serialized `risc0_zkvm::InnerReceipt` proving correct execution of the privacy-preserving circuit.
+
+## The privacy-preserving execution circuit
+
+The circuit is a RISC-V program proven with the risc0 zkVM. It is executed entirely off-chain by the transaction sender; the sequencer only verifies the proof. It is not a built-in program in the NSSA sense — it wraps the execution of built-in programs inside a ZK proof.
+
+The circuit supports a single chain of tail-calls (each program invocation may chain to at most one subsequent program).
```
moudyellaz: Circuit at execution_state.rs:124-184 iterates `program_output.chained_calls` via VecDeque + push_front, allowing multiple chained calls per invocation. Spec sentence is wrong.

docs/specs.md
```diff
@@ -0,0 +1,1565 @@
...
+        Self { account, claim: is_default.then_some(claim) }
+    }
+}
+```
+
+After execution, the runtime processes each post-state's optional `claim`:
+
+- `Claim::Authorized` — sets `program_owner = executing_program_id`, but only if the account currently has `DEFAULT_PROGRAM_ID`. The authorization precondition depends on the account's kind:
+  - **Public account or private PDA:** the account must be authorized — i.e. its `is_authorized` flag must be `true`. Authorization comes from a signature (public) or from a caller's `pda_seeds` matching the PDA derivation under the appropriate npk (private PDA).
+  - **Standalone private account** (the `Regular` kinds — `PrivateAuthorizedInit`, `PrivateAuthorizedUpdate`, `PrivateUnauthorized`): the privacy circuit **does not enforce** the `is_authorized` precondition for `Claim::Authorized`. This is intentional: any party producing a valid update for a standalone private account already needs the corresponding `nsk` (to compute the update nullifier), so the authorization is implicit in the proof. The claim is therefore allowed unconditionally for these kinds, even when `pre_state.is_authorized == false`.
```
moudyellaz: Spec line 616 says the circuit "does not enforce the is_authorized precondition" for Regular private kinds (PrivateAuthorizedInit, PrivateAuthorizedUpdate, PrivateUnauthorized). Code at program_methods/guest/src/bin/privacy_preserving_circuit/output.rs:51-54 (Init) and 88-91 (Update) does `assert!(pre_state.is_authorized, ...)` unconditionally. A wallet building a PrivateAuthorizedInit/Update with `is_authorized=false` would be valid per spec, rejected by code.

docs/specs.md
schouhy: thanks, fixed the phrasing. But I think the pseudocode part is ok no? If MAX_NUMBER_CHAINED_CALLS was 0, it should still allow the top-level execution, so we need a <= in that assert in the pseudocode I think

docs/specs.md
schouhy: I can't find these lines you mention. auth-transfer program is no longer in the specs.

docs/specs.md
schouhy: fixed

docs/specs.md
schouhy: fixed in new builtin_programs.rs file

docs/specs.md
schouhy: fixed in new builtin_programs.rs file

docs/specs.md
moudyellaz: Following up now that the authorization propagation landed. What I meant: is_authorized for a child call is not taken verbatim from the caller-supplied chained_call.pre_states, it is recomputed by the node from caller_data.authorized_accounts unioned with the parent's verified authorized pre-states, then the child's reported flags are checked for exact equality against that recomputation. So a caller cannot forge or drop authorization across the boundary. A one-line note to that effect here would make the trust boundary explicit, since the current text reads as if the flag flows through unchanged.

docs/specs.md
```diff
@@ -0,0 +1,1382 @@
...
+- `nonces`: One nonce per signer (public account). Each must match the current nonce of the corresponding account in state.
+- `instruction_data`: Parameters encoded as a list of `u32` words.
+
+### Witness set
+
+A list of `(signature, public_key)` pairs. Each pair must produce a valid BIP-340 Schnorr signature over the message hash. The accounts derived from each `public_key` (via `AccountId::from(&public_key)`) form the *authorized signer set*. The program receives `is_authorized = true` for any pre-state account in this set (programs use the flag to decide whether to permit user-driven actions like transfers). Programs may also gain authorization for additional accounts via the PDA mechanism in chained calls. Authorization propagates down the call chain monotonically: the authorized set passed to each child call is the union of the parent's own authorized set and the parent's verified authorized pre-states, so an account authorized at any hop remains authorized for all subsequent calls even if an intermediate hop does not include it in its pre-states. Upon acceptance, each signing account's nonce is incremented by 1.
```
moudyellaz: Confirming intent: the union means authorization propagates down the whole chain for PDAs too, not only signer accounts. Once a program marks its own PDA authorized and includes it in a pre-state, that account id enters the set and every later call, including calls into unrelated programs, sees is_authorized true for it. For signer accounts this is clearly right since the signature covers the whole tx. For PDAs it means a downstream program can observe another program's PDA as authorized. The private circuit already behaves this way (resolve_authorization_and_record_bindings is sticky), so it is consistent, not a regression. Worth one line here stating explicitly that PDA authorization is also monotonic, so readers do not assume it is scoped to the immediate callee.

schouhy: done!

nssa/src/validated_state_diff.rs
Arjentix, line 281: Wait, why? It may be not desirable. For example a program may want to disable authorization for some account if another program it calls via a chain call doesn't need authorization or behaves not in a desired way if account is authorized.

docs/builtin_programs.md
```diff
@@ -0,0 +1,300 @@
...
+
+LEZ v0.3 supports the following built-in programs, loaded at genesis. They are immutable and identified by unique `ProgramId` values.
+
+- Authenticated transfer (`AUTHENTICATED_TRANSFER_ID`)
+- Token (`TOKEN_ID`)
+- AMM (`AMM_ID`)
+- Associated Token Account (`ASSOCIATED_TOKEN_ACCOUNT_ID`)
+- Clock (`CLOCK_ID`)
+- Vault (`VAULT_ID`)
+- Faucet (`FAUCET_ID`)
```
Arjentix, line 11: Also bridge already exists

docs/builtin_programs.md
```diff
@@ -0,0 +1,300 @@
...
+
+### Authenticated transfer program
+
+Moves native tokens from a source account to a destination account, requiring the source account to be authorized. Has two instruction variants:
+
+- `Initialize` with a single pre-state: claims the pre-state account (which must be `Account::default()`) on behalf of the caller. Authorization is enforced by the runtime when processing `Claim::Authorized`.
+- `Transfer { amount }` with two pre-states `[sender, recipient]`: standard transfer. Sender must be authorized. The recipient is claimed only if its `program_owner` is currently the default — pre-existing accounts owned by other programs keep their owner.
+
+```rust
+pub enum Instruction {
```
Arjentix, line 23: Ah, supporting this spec will be a pain

docs/builtin_programs.md
```diff
@@ -0,0 +1,300 @@
+# Built-in programs
+
+LEZ v0.3 supports the following built-in programs, loaded at genesis. They are immutable and identified by unique `ProgramId` values.
+
+- Authenticated transfer (`AUTHENTICATED_TRANSFER_ID`)
+- Token (`TOKEN_ID`)
+- AMM (`AMM_ID`)
```
Arjentix, line 7: Do you think it makes sense to add it here? I think it's already moved to a separate repo and won't be a part of default programs, no?

docs/specs.md
```diff
@@ -0,0 +1,1382 @@
+# LEE v0.3 specifications
+
+## LEE v0.3 basic types and constants
+
+```rust
+type AccountId = [u8; 32];
+type NativeToken = u128;
+type ProgramId = [u32; 8];
+struct PdaSeed([u8; 32]);
+type Data = List<u8>;
```
Arjentix, line 10: Looks like everything else is pure Rust, maybe change this to `Vec` then?

docs/specs.md
```diff
@@ -0,0 +1,1382 @@
...
+            program_owner: [0; 8],
+            balance: 0,
+            data: [],
+            nonce: 0,
+        }
+    }
+}
+```
+
+### Program owner field
```
Arjentix, line 95: Maybe better to move description closer to a describing type? Although maybe it won't be handy to write in markdown, not sure

Arjentix (commented):
Thanks for the PR. My main concern is this change in program validation.

Also just an idea. Since a big part of the spec is just Rust code, we may consider adding it as crate documentation (and later post some of our crates to crates.io). That way when someone would go to `docs.rs/our-crate` they will see it

schouhy:
Specs have been moved [here](https://lip.logos.co/blockchain/raw/lez/lee-v0.3-specifications.html) and the fixes in this PR have been addressed [here](https://github.com/logos-blockchain/logos-execution-zone/pull/504)

PR #474, opened 2026-05-12
feat(key-protocol/nssa)!: PQ resistance primitives for vsk/vpk for shared key agreement

jonesmarvin8:
## 🎯 Purpose

Shifts LEE to a PQ secure scheme, Kyber-768, for the key agreement. This requires changing the viewing keys from `secp256k1` related data structures to byte arrays that can be used to produce ML-KEM-768 encapsulation and decapsulation keys. Updated [key protocol](https://www.notion.so/UPDATED-Hierarchical-Deterministic-wallets-in-LEE-ovk-removed-unified-logic-for-private-account-p-3678f96fb65c808a8860fcf152a29109) for viewing key changes.

## ⚙️ Approach

- [X] Update key protocol to change `vsk` and `vpk` generation.
- [X] Update shared key generation logic to use Kyber-768.
- [X] Update `specs.md` portions related to key agreement (only these sections are included).

## 🧪 How to Test

Refactored to replace ECIES with Kyber. As such, all pre-existing tests should still work. Additionally, added `kem_to_chacha20_round_trip` to demonstrate encapsulation and decapsulation does not distort account data.

## 🔗 Dependencies

None.

## 🔜 Future Work

None.

## 📋 PR Completion Checklist

- [X] Complete PR description
- [X] Add/update tests


Pravdyvy (commented):
I see, that some keys grow in magnitude. It will be very interesting to look at performance. 
Unfortunately still not in CI, but can be called easily with `just bench` after merging main.

jonesmarvin8:
> I see, that some keys grow in magnitude. It will be very interesting to look at performance. Unfortunately still not in CI, but can be called easily with `just bench` after merging main.

What do you mean they grow in magnitude?
`vsk`, `vpk` should be fixed sizes. And the resulting ephemeral public key from a key agreement should be fixed size as well.

`vsk` is 64-bytes
`vpk` is 1kb
ephemeral public key is 1kb.

Pravdyvy:
> > I see, that some keys grow in magnitude. It will be very interesting to look at performance. Unfortunately still not in CI, but can be called easily with `just bench` after merging main.
> 
> What do you mean they grow in magnitude? `vsk`, `vpk` should be fixed sizes. And the resulting ephemeral public key from a key agreement should be fixed size as well.
> 
> `vsk` is 64-bytes `vpk` is 1kb ephemeral public key is 1kb.

I mean that they grew in comparison to main. `vpk` before was 33 bytes. It is interesting to see how it affects performance.

nssa/core/src/encryption/shared_key_derivation.rs
```diff
@@ -1,78 +1,181 @@
...
+
+impl ViewingPublicKey {
+    #[must_use]
+    pub fn to_bytes(&self) -> &[u8] {
+        &self.0
+    }
+
+    /// Derive the ML-KEM-768 encapsulation key from the FIPS 203 seed halves `d` and `r`.
+    /// Allows any crate to construct a VPK from raw seed bytes without importing
+    /// `key_protocol::ViewingSecretKey`.
```
schouhy: `key_protocol::ViewingSecretKey` doesn't seem right, no? I'd remove the comment

key_protocol/src/key_management/secret_holders.rs
```diff
@@ -19,8 +17,13 @@ pub struct SeedHolder {
...
 #[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
 pub struct SecretSpendingKey(pub [u8; 32]);
-
-pub type ViewingSecretKey = Scalar;
+/// Viewing secret key: the KEM seed split into its two 32-byte halves `d` and `r` (= z in
+/// FIPS 203), from which the ML-KEM 768 decapsulation key is derived deterministically.
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub struct ViewingSecretKey {
+    pub d: [u8; 32],
+    pub r: [u8; 32],
```
schouhy: why not call it z as in FIPS 203?

lez/wallet/src/cli/programs/native_token_transfer.rs
schouhy, line 1: wallet command docs still read `valid 33 byte hex string.` for the vpk argument.

nssa/core/src/encryption/shared_key_derivation.rs
```diff
@@ -1,78 +1,181 @@
...
-pub struct Secp256k1Point(pub Vec<u8>);
-
-impl std::fmt::Debug for Secp256k1Point {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        let hex: String = self.0.iter().fold(String::new(), |mut acc, b| {
-            write!(acc, "{b:02x}").expect("writing to string should not fail");
-            acc
-        });
-        write!(f, "Secp256k1Point({hex})")
+pub struct ViewingPublicKey(pub Vec<u8>);
```
schouhy: would it be a hassle to remove the `pub` in the inner field and have it constructed from a `from_bytes` constructor that checks the correct length? (It already has a `to_bytes` companion)

key_protocol/src/key_management/group_key_holder.rs
```diff
@@ -153,18 +148,16 @@ impl GroupKeyHolder {
...
-    /// Each call generates a fresh ephemeral key, so two seals of the same holder produce
+    /// Each call generates a fresh KEM encapsulation, so two seals of the same holder produce
     /// different ciphertexts.
     #[must_use]
     pub fn seal_for(&self, recipient_key: &SealingPublicKey) -> Vec<u8> {
-        let mut ephemeral_scalar: Scalar = [0_u8; 32];
-        OsRng.fill_bytes(&mut ephemeral_scalar);
-        let ephemeral_pubkey = Secp256k1Point::from_scalar(ephemeral_scalar);
-        let shared = SharedSecretKey::new(ephemeral_scalar, &recipient_key.0);
+        let vpk = ViewingPublicKey(recipient_key.0.clone());
```
schouhy: this doesn't look good. ViewingPublicKeys have a different semantic meaning and shouldn't be used in a function not related to account encryption. I think a better approach is to rename what's now called `ViewingPublicKey` to `MlKem768EncapsulationKey` (or something along those lines), make `pub type ViewingPublicKey = MlKem768EncapsulationKey`, and the same for the sealing keys. That way the code for sealing a recipient key does not involve viewing public keys.

schouhy (changes requested):
Hey thanks for the PR!
I think tutorials need updating (and with it probably a sensible solution to the problem of passing a 1kb sized vpk to a cli command)

I'm also leaving other minor comments.

key_protocol/src/key_management/secret_holders.rs
jonesmarvin8: Thanks for pointing this out. I misremembered which variable name was used (r vs z).

lee/state_machine/core/src/encryption/shared_key_derivation.rs
```diff
@@ -1,78 +1,181 @@
...
+    ///
+    /// The KEM randomness is derived as `SHA-256(message_hash || output_index_le)`,
+    /// making the ciphertext reproducible from the same `(vpk, message_hash, output_index)`
+    /// triple. Use a distinct `output_index` for each private account output in the same
+    /// transaction to ensure per-output EPK uniqueness.
     #[must_use]
-    pub fn from_scalar(value: Scalar) -> Self {
-        let x_bytes: FieldBytes = value.into();
-        let x = k256::Scalar::from_repr(x_bytes).unwrap();
+    pub fn encapsulate_deterministic(
```
moudyellaz, line 95: Small hardening suggestion on this one. Since the KEM message is derived purely from public inputs (message_hash, output_index), the shared secret has no secret entropy, so anyone who knows the recipient vpk could recompute it without the decapsulation key. That is totally fine for the tests that use it today, but it would defeat the KEM if it ever reached an on-chain path, and the spec says production uses a random message. Could we gate it `#[cfg(test)]` or rename it to make "test only, never on-chain" clear, so it cannot get wired into production by accident? And if it stays public, maybe a small test for the reproducibility the doc mentions.

nssa/core/src/encryption/shared_key_derivation.rs
```diff
@@ -1,78 +1,181 @@
...
+    /// `d` and `r` are the two 32-byte halves of the FIPS 203 `ViewingSecretKey` seed.
+    #[must_use]
+    pub fn decapsulate(ciphertext: &EphemeralPublicKey, d: &[u8; 32], r: &[u8; 32]) -> Self {
+        let mut seed = Seed::default();
+        seed[..32].copy_from_slice(d);
+        seed[32..].copy_from_slice(r);
+        let dk = ml_kem::DecapsulationKey768::from_seed(seed);
+        let ss = dk
+            .decapsulate_slice(&ciphertext.0)
+            .expect("EphemeralPublicKey must be 1088 bytes (ML-KEM-768 ciphertext)");
```
moudyellaz: One robustness thing worth tightening here: decapsulate does `.expect("EphemeralPublicKey must be 1088 bytes")`, and the epk comes straight from chain data. Both wallet scan paths reach this after only the 1-byte view-tag filter (sync_private_accounts_with_tx via calculate_shared_secret_receiver, and sync_shared at lib.rs:773), so about 1 in 256 outputs get here, and someone who knows a victim's npk/vpk could craft a match on purpose. A borsh-decoded epk of the wrong length would then panic the scanning wallet. Could we return a Result/Option and just skip malformed outputs instead? The canonical from_cursor path is already length-safe, so it is only the borsh path that needs this.

docs/specs.md
```diff
@@ -0,0 +1,159 @@
+# LEZ v0.3 specifications for Key Agreement
```
moudyellaz: A few small things while reading this through:
- The title says LEZ v0.3 but the body uses LEE (the view tag) and NSSA/v0.2 (the KDF), so three names in one short doc. Could we settle on LEE/v0.3 throughout?
- Line 27: small typo in the decapsulate formula, `vks.r` should be `vsk.r`.
- Line 117: the discovery note still says "avoiding expensive ECDH on irrelevant outputs", but it is ML-KEM decapsulation now, not ECDH.
- The InputAccountIdentity block here does not have the `seed: Option<(PdaSeed, ProgramId)>` field on PrivatePdaInit/Update that the external-seed work added, so it reads a little stale.

key_protocol/src/key_management/mod.rs
```diff
@@ -69,21 +69,16 @@ impl KeyChain {
     pub fn calculate_shared_secret_receiver(
         &self,
         ephemeral_public_key_sender: &EphemeralPublicKey,
-        index: Option<u32>,
+        _index: Option<u32>,
```
moudyellaz: Minor cleanup: now that calculate_shared_secret_receiver decapsulates with the keychain's own stored vsk, the `_index` parameter is never used (the keychain is already account-specific, so this is correct, just vestigial). Could we drop the param so callers are not misled into thinking the index matters here? Related, ViewingSecretKey.d/.r are pub fields (secret_holders.rs:24-25), same as the from_bytes hygiene you raised on ViewingPublicKey, so a constructor there would be consistent.

key_protocol/src/key_management/key_tree/keys_private.rs
```diff
@@ -59,16 +59,18 @@ impl ChildKeysPrivate {
...
-        #[expect(clippy::arithmetic_side_effects, reason = "TODO: fix later")]
-        let parent_pt =
-            Scalar::from_repr(self.value.0.private_key_holder.nullifier_secret_key.into())
-                .expect("Key generated as scalar, must be valid representation")
-                * Scalar::from_repr(self.value.0.private_key_holder.viewing_secret_key.into())
-                    .expect("Key generated as scalar, must be valid representation");
-        let mut input = vec![];
+        let mut parent_hash = sha2::Sha256::new();
+        parent_hash.update(b"LEE/keys");
+        parent_hash.update([0_u8; 16]);
```
moudyellaz: Last small one: the child-key derivation hashes in `[0;16]` then `[9]` as domain bytes before nsk and the vsk halves. They look intentional but unexplained, and since this defines how child keys are derived, anyone reimplementing the wallet would need to match them exactly. Could we add a short comment on what these constants are, and document the child-derivation in the spec so it is reproducible?

moudyellaz (changes requested):
lgtm, thanks. Left some comments worth looking at.

nssa/core/src/encryption/shared_key_derivation.rs
jonesmarvin8: Added tests to check malformed epk's `decapsulate_returns_none_for_malformed_epk` and shared key `calculate_shared_secret_receiver_returns_none_for_malformed_epk`

key_protocol/src/key_management/key_tree/keys_private.rs
jonesmarvin8: Thanks for bringing this back to my attention. I initially included these as a means to avoid collisions, but that's when this logic was further downstream. Where it is now, it is unnecessary.

jonesmarvin8: I added documentation to briefly explain `parent_hash` and `input`.

docs/specs.md
```diff
@@ -0,0 +1,159 @@
...
+#### KDF
+
+```rust
+fn kdf(
+    shared_secret: &SharedSecretKey,    // 32-byte output of the KEM
+    commitment: &Commitment,            // 32-byte output commitment
+    output_index: u32,                  // index of this output within the tx (LE)
+) -> [u8; 32] {
+    let mut bytes = Vec::new();
+    bytes.extend_from_slice(b"NSSA/v0.2/KDF-SHA256/");
```
moudyellaz: you are still using NSSA here

moudyellaz (approved):
lgtm, thanks for addressing previous comments. Left a tiny nit as you are still using "NSSA" in one place.

integration_tests/tests/auth_transfer/private.rs
```diff
@@ -32,6 +35,7 @@ async fn private_transfer_to_owned_account() -> Result<()> {
         to: Some(private_mention(to)),
         to_npk: None,
         to_vpk: None,
+        to_keys: None,
```
schouhy, line 42: maybe for another PR, but is there a reason why to_keys can't be another variant of CliAccountMention factored into the `to` argument?

schouhy (approved):
lgtm thanks for considering my previous comments.

docs/specs.md
jonesmarvin8: Thanks.

integration_tests/tests/auth_transfer/private.rs
jonesmarvin8: We'll save this for a fewer PR.

PR #475, opened 2026-05-13
fix: Private shared account preparation

schouhy:
## 🎯 Purpose

Fixes a bug where shared private accounts were incorrectly treated as non-existent during transaction preparation when their local state was empty. This caused private transactions involving shared accounts to fail or produce incorrect proofs.

## ⚙️ Approach

- Fix private_shared_acc_preparation to always set exists = true and always include the nsk key, regardless of whether the local account state is populated yet.
- Fix get_private_account_commitment to look up shared private accounts in addition to regular private accounts.
- Add a comment to get_private_account clarifying it does not cover shared private accounts.
- Enable the previously ignored fund_shared_account_from_public integration test and add a private account sync step before funding.

## 🧪 How to Test

Run the integration test that was previously ignored:

`RISC0_DEV_MODE=1 cargo nextest run -p integration_tests fund_shared_account_from_public`

## 🔗 Dependencies

None

## 🔜 Future Work

None

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests

### DIFF
--- a/integration_tests/tests/shared_accounts.rs
+++ b/integration_tests/tests/shared_accounts.rs
@@ -155,7 +155,6 @@ async fn group_invite_join_key_agreement() -> Result<()> {
 /// Fund a shared account from a public account via auth-transfer, then sync.
 /// TODO: Requires auth-transfer init to work with shared accounts (authorization flow).
 #[test]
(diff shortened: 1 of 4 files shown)


Pravdyvy (approved):
lgtm

moudyellaz (approved):

PR #478, opened 2026-05-14
feat: add cycle_bench tool for executor, prove, PPE, and verify cycle measurements

moudyellaz:
## 🎯 Purpose

Add a benchmarking tool that produces the four cost inputs the LEZ fee model needs: `G_executor` (per-program Risc0 user cycles), `G_prove` (single-program prover wall time), `G_verify` (verifier wall time per outer receipt), and
`S_agg` (the borsh-serialized outer-proof byte size). Also captures the PPE composition tax and a chained-call depth sweep so the model can fit a linear cost-per-call coefficient.

## ⚙️ Approach

New package at `tools/cycle_bench` (not a workspace dep of anything else).
Three feature flags:

- default: executor only (fast).
- `prove`: adds `--prove`; calls `prover.prove(env, elf)` per program.
- `ppe`: adds `--ppe` (composition cost + `chain_caller` depth=1/3/5/9) and `--verify` (generates one receipt, times `Receipt::verify(...)` over N iters).

Wall-clock samples are reported as `best / mean ± stdev (n=N)` for executor and verify; prove cases are single-shot since each prove is minutes long. Long-running tables also print seconds alongside ms.

- [x] Add `tools/cycle_bench` package (`Cargo.toml`, `src/main.rs`, `src/ppe.rs`, `src/stats.rs`, `README.md`).
- [x] Add `docs/benchmarks/cycle_bench.md` with results from an M2 Pro dev box.
- [x] Add `docs/benchmarks/README.md` index.
- [x] Register `tools/cycle_bench` in workspace `members`.

## 🧪 How to Test

```sh
# Executor cycles only (fast, ~seconds)
cargo run --release -p cycle_bench

# + real proving (slow, ~minutes per program)
cargo run --release -p cycle_bench --features prove -- --prove

# + PPE composition + chain-call sweep (very slow, ~hour)
cargo run --release -p cycle_bench --features ppe -- --prove --ppe

# + verifier microbench (G_verify, 1000 iters)
cargo run --release -p cycle_bench --features ppe -- --verify --verify-iters 1000
```

Numbers will vary by host; expect the relative ordering between programs to be preserved. Output is also written to `target/cycle_bench.json` for regression diffs. See `docs/benchmarks/cycle_bench.md` for the full result tables and findings from one reference run.

## 🔗 Dependencies

None.

## 🔜 Future Work

None.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests *(bench tool with no automated tests; verified by running each invocation; outputs in `docs/benchmarks/cycle_bench.md`)*


tools/cycle_bench/src/main.rs
```diff
@@ -0,0 +1,660 @@
...
+    #[arg(long, default_value_t = 1000)]
+    verify_iters: usize,
+
+    /// Iterations for executor wall-time sampling per case. First iter is
+    /// discarded as warmup, remaining N feed the stats.
+    #[arg(long, default_value_t = 5)]
+    exec_iters: usize,
+}
+
+const AMM_PROGRAM_ID: ProgramId = [42; 8];
```
Pravdyvy: I think we can use program ids from nssa here. After all, we already use elfs from there.

Pravdyvy (approved):
lgtm. Left minor comment.

tools/cycle_bench/src/main.rs
moudyellaz: done

docs/benchmarks/cycle_bench.md
```diff
@@ -0,0 +1,98 @@
...
+| case | auth_transfer Transfer in PPE |
+| proof_bytes (S_agg) | 223,551 |
+| journal_bytes | 412 |
+| verify_ms (best / mean ± stdev, n=1000) | 11.71 / 12.06 ± 1.99 |
+
+## Findings
+
+- Proving cost scales with po2-bucketed `total_cycles`, not raw `user_cycles`. Trimming user_cycles only helps if it crosses a 2^N boundary.
+- Single-program PPE composition tax on M2 Pro CPU: ≈ 48 s (61.5 − 13.7).
+- Chained-call cost is linear at ≈ 53 s per call. A max-depth chain (10) would take ≈ 600 s standalone on this CPU.
```
Arjentix, line 81: Wow, a minute per a chain call, that's not fast...

tools/cycle_bench/src/main.rs
```diff
@@ -0,0 +1,656 @@
...
+    clippy::print_stderr,
+    clippy::print_stdout,
+    clippy::redundant_type_annotations,
+    clippy::ref_option,
+    clippy::ref_patterns,
+    clippy::similar_names,
+    clippy::single_call_fn,
+    clippy::single_match_else,
+    clippy::std_instead_of_alloc,
+    clippy::std_instead_of_core,
```
Arjentix: Hmm, this should not be forbidden by our workspace lints. Can you replace `allow` with `expect` and check if some lints are unfulfilled?

tools/cycle_bench/src/main.rs
```diff
@@ -0,0 +1,656 @@
...
+//!
+//! Runs each guest ELF through the Risc0 executor (no proving) with realistic inputs
+//! drawn from the existing per-program unit tests, then prints a table and writes a
+//! JSON dump for regression comparison.
+//!
+//! Run with `cargo run --release -p cycle_bench`. `RISC0_DEV_MODE` has no effect on
+//! executor cycle counts.
+
+#![allow(
+    clippy::arbitrary_source_item_ordering,
```
Arjentix: Let's not allow this and follow it. Should be simple and the code will look more like other files

tools/cycle_bench/src/main.rs
```diff
@@ -0,0 +1,656 @@
...
+    clippy::wildcard_enum_match_arm,
+    reason = "Bench tool: matches test-style fixture code"
+)]
+
+use std::{path::PathBuf, time::Instant};
+
+mod ppe;
+mod stats;
+
+use amm_core::{PoolDefinition, compute_liquidity_token_pda, compute_pool_pda, compute_vault_pda};
```
Arjentix, line 26: This is why I wrote the comment above, haha

tools/cycle_bench/src/main.rs
```diff
@@ -0,0 +1,656 @@
...
+    for iter in 0..total {
+        let mut env_builder = ExecutorEnv::builder();
+        env_builder
+            .write(&self_program_id)?
+            .write(&caller_program_id)?
+            .write(&pre_states)?
+            .write(&instruction_words)?;
+        let env = env_builder.build()?;
+
+        let started = Instant::now();
```
Arjentix: For benchmarking consider [`quanta`](https://crates.io/crates/quanta). Not a strong opinion, just a suggestion

tools/cycle_bench/src/main.rs
```diff
@@ -0,0 +1,656 @@
...
+    results.push(run_case(
+        "authenticated_transfer",
+        "Transfer",
+        AUTHENTICATED_TRANSFER_ELF,
+        AUTHENTICATED_TRANSFER_ID,
+        authenticated_transfer_transfer(),
+        &transfer_amount,
+        prove,
+        exec_iters,
+    )?);
```
Arjentix: Let's refactor to have something like this:

```rust
struct Case {
    ...
}

impl Case {
    pub fn run() -> Result<BentchResult> {...}
}

let results = [Case {...}, Case {...}, ...].map(Case::run).collect();
```

tools/cycle_bench/src/stats.rs
```diff
@@ -0,0 +1,54 @@
...
+
+use serde::Serialize;
+
+#[derive(Debug, Serialize, Clone, Copy, Default)]
+pub struct Stats {
+    pub n: usize,
+    pub best_ms: f64,
+    pub mean_ms: f64,
+    pub stdev_ms: f64,
+}
```
Arjentix, line 22: Let's add docs to the fields, I don't quite get what is `stdev_ms`

tools/cycle_bench/src/stats.rs
```diff
@@ -0,0 +1,54 @@
...
+        }
+    }
+
+    /// Format as `best / mean ± stdev (n=N)` for table display.
+    pub fn format(&self) -> String {
+        format!(
+            "{:.2} / {:.2} ± {:.2} (n={})",
+            self.best_ms, self.mean_ms, self.stdev_ms, self.n,
+        )
+    }
```
Arjentix: Better do `impl Display for Stats`

Arjentix (approved):
LGTM, left some minor comments

docs/benchmarks/cycle_bench.md
moudyellaz: Yeah, fair. This is single-CPU M2 Pro running the PPE composition through `execute_and_prove` (mock-proving, no aggregation). Real prover infra (GPU, Bonsai) plus the aggregation scheme will land this much lower; the number here is meant as the standalone composition ceiling for the fee model handoff, not a target prod cost.

tools/cycle_bench/src/main.rs
moudyellaz: done

tools/cycle_bench/src/main.rs
moudyellaz: done

tools/cycle_bench/src/main.rs
moudyellaz: done

tools/cycle_bench/src/main.rs
moudyellaz: Going to defer this to the upcoming criterion follow-up PR, which moves the wall-time measurements onto criterion's own high-precision timer (plus per-iteration outlier detection, etc). Quanta is a good suggestion if we stayed hand-rolled, but if we're moving to criterion anyway it'll handle this for us.

tools/cycle_bench/src/main.rs
moudyellaz: done with one deviation from the sugg: since prove / exec_iters are constant across cases, the iterator uses a closure .map(|c| c.run(prove, exec_iters)) instead of .map(Case::run). Storing them per-Case would be wasteful.

tools/cycle_bench/src/stats.rs
moudyellaz: done

tools/cycle_bench/src/stats.rs
moudyellaz: done

Arjentix (approved):

PR #479, opened 2026-05-15
fix: Bug in faucet account protection mechanism

schouhy:
## 🎯 Purpose

The faucet account protection had a gap: it only blocked transactions that directly invoked the faucet program. A user could deploy a custom program that chain-calls the faucet, bypassing the check and draining faucet funds. This PR closes that gap.
Also fixes an unrelated issue in the privacy-preserving circuit where authorization was not propagating transitively through chained calls like in the public execution.

## ⚙️ Approach

- Change faucet protection from "reject if top-level program is faucet" to "reject if faucet account state changed after execution", so it catches both direct and indirect modifications.
- Fix authorization propagation in the circuit: add a `HashSet<AccountId>` to track which accounts have been authorized, so authorization persists transitively through chained calls (matches the behavior already in the public execution path)
- Remove `can_transfer_funds_to_system_faucet_account` test. This is no longer a valid operation.
- Add `faucet_chain_caller` test program that simulates the attack by chain-calling the faucet
- Minor: fix wrong error message in `vault.rs` ("exactly 3 accounts" → "exactly 2 accounts")

## 🧪 How to Test

Added integration tests for both public and private transactions: `user_tx_that_chain_calls_faucet_is_dropped` and `ppt_that_chain_calls_faucet_is_dropped`

Both tests deploy the faucet_chain_caller program, send a transaction that tries to drain the faucet via a chain call, and assert the transaction is dropped and balances are unchanged.

## 🔗 Dependencies

None

## 🔜 Future Work

None

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


common/src/transaction.rs
```diff
@@ -111,6 +102,16 @@ impl NSSATransaction {
             ));
         }
 
+        let faucet_id = nssa::system_faucet_account_id();
+        if public_diff
+            .get(&faucet_id)
+            .is_some_and(|post| *post != state.get_account_by_id(faucet_id))
+        {
+            return Err(nssa::error::NssaError::InvalidInput(
+                "Transaction modifies system faucet account".into(),
```
Arjentix, line 111: So no transfers to faucet is allowed, okay. But I think I had an integration tests which checks exactly that, it probably should be removed

Arjentix (approved):
Hey, thanks, good catch!

common/src/transaction.rs
schouhy: ah yes, removed it. We can eventually relax the condition inside `is_some_and` to allow funding if we need so

Pravdyvy (approved):
lgtm

PR #486, opened 2026-05-19
feat!: Add new path for pda seed input

schouhy:
## 🎯 Purpose

Private PDAs previously could only be used through a program that used `Claim::Pda` or `pda_seeds` to establish the binding between the PDA `account_id` and the npk. This meant you couldn't call auth_transfer directly to fund a private PDA, you always needed a wrapper. This PR removes that limitation.

## ⚙️  Approach

- Add an optional seed: `Option<(PdaSeed, ProgramId)> ` field to the `PrivatePdaInit` and `PrivatePdaUpdate` variants of `InputAccountIdentity`. When the seed is provided, the circuit verifies `AccountId::for_private_pda(pda_program_id, seed, npk, identifier) == pre_state.account_id` directly, establishing the binding without needing a `Claim::Pda` or caller `pda_seeds`. It also verifies `is_authoried` is set to false. 
- Add binding logic in `execution_state.rs` for when an external seed is supplied
- Remove the authorization check on private pda claiming in the ppt circuit. Authorization checks only remains on public account claiming.
- Rename `pda_fund_spend_proxy` to `pda_spend_proxy`. The fund path is now handled directly via auth_transfer.
- Rewrite `fund_private_pda` in integration tests to use auth_transfer directly with the external seed

## 🧪 How to Test

RISC0_DEV_MODE=1 cargo test

## 🔗 Dependencies

None

## 🔜 Future Work

Update wallet commands to use the external seed path when appropiate.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


Pravdyvy (approved):
lgtm

program_methods/guest/src/bin/privacy_preserving_circuit/execution_state.rs
```diff
@@ -348,14 +399,11 @@ impl ExecutionState {
...
+                    // Private accounts: don't enforce the claim semantics. Unauthorized private
+                    // claiming is intentionally allowed
                     match claim {
-                        Claim::Authorized => {
-                            assert!(
-                                pre_is_authorized,
-                                "Cannot claim unauthorized private PDA {pre_account_id}"
-                            );
-                        }
+                        Claim::Authorized => {}
```
moudyellaz, line 417: The PR body should mention that this also drops the per-tx `pre_is_authorized` check for `Claim::Authorized` on private PDAs (here at lines 402-406), not only the output.rs Update gate. Reason: reading the body alone, the only behavioral relaxation appears to be guarded by `external_seed.is_some()`, but `Claim::Authorized` on a private PDA is now an unconditional no-op regardless of the input variant.

wallet/src/privacy_preserving_tx.rs
```diff
@@ -252,11 +252,13 @@ impl AccountManager {
                         nsk,
                         membership_proof,
                         identifier: pre.identifier,
+                        seed: None,
```
moudyellaz, line 255: Small nit on the PR body: `account_identities()` hardcodes `seed: None` on both `PrivatePdaUpdate` and `PrivatePdaInit`, so the only callers exercising the new path are hand-built test inputs. Could we either wire the wallet to populate `seed` when it knows the `(seed, pda_program_id)` for an owned PDA, or move that to `Future Work`? Happy with either; just want to make sure the feature has a user-facing entry point before we close the loop on this PR series.

program_methods/guest/src/bin/privacy_preserving_circuit/execution_state.rs
```diff
@@ -305,6 +305,57 @@ impl ExecutionState {
...
+                            pda_program_id,
+                            seed,
+                        );
+                        assert_family_binding(
+                            &mut self.pda_family_binding,
+                            pda_program_id,
+                            seed,
+                            pre_account_id,
+                        );
+                    }
```
moudyellaz, line 369: The external-seed binding only fires on `Entry::Vacant`. That's fine, since subsequent encounters of the same PDA inherit the binding via `private_pda_bound_positions`, but it took a re-read to convince myself. A one-line note in the Vacant branch (something like "external seed is only consulted the first time the account is seen; subsequent calls inherit binding via `private_pda_bound_positions`") would save future readers the trace.

program_methods/guest/src/bin/privacy_preserving_circuit/execution_state.rs
```diff
@@ -383,10 +431,6 @@ impl ExecutionState {
                             );
                         }
                     }
-                } else {
-                    // Standalone private accounts: don't enforce the claim semantics.
-                    // Unauthorized private claiming is intentionally allowed since operating
-                    // these accounts requires the npk/nsk keypair anyway.
                 }
 
                 post.account_mut().program_owner = program_id;
```
moudyellaz, line 447: When `fund_private_pda` runs through `auth_transfer` with `seed: Some((seed, pda_program_id))`, the resulting PDA has `program_owner == auth_transfer_id`, not `pda_program_id`. That's mechanically consistent (this line sets `program_owner = program_id` of the current call after a claim), and private PDA gating happens via the `for_private_pda` binding rather than `program_owner`, so funds aren't at risk. But two observable consequences worth pinning down: (1) a subsequent top-level `Claim::Pda(seed)` from `pda_program_id` fails the `post.account().program_owner == DEFAULT_PROGRAM_ID` precondition at lines 365-369, so the owning program can never cleanly "take ownership" of a PDA that was first touched via auth_transfer; always-external-seed access is the only path after that. (2) The `program_owner` field carries no useful information for these PDAs (it's "whoever last wrote to the account"), which is a small inconsistency vs how everyone reasons about `program_owner` for non-PDA accounts. Could we add either a test that pins the "Claim::Pda fails after auth_transfer fund" behavior so we don't accidentally regress it, or a doc-comment on `PrivatePdaInit`'s `seed` field saying "the resulting account's `program_owner` will be the executing program, not `pda_program_id`, by design"? Whichever you prefer.

moudyellaz (approved):
lgtm thanks, left some minor comments casue i struggled a bit.

moudyellaz:
One more thing on commit hygiene: the PR title carries `feat!:` but none of the individual commits (`b37e3ca8`...`e4a71961`) use the Conventional Commits `!` marker or a `BREAKING CHANGE:` footer. If the merge is a squash-merge that lifts the PR title, we'll be fine on the changelog side, but if any of these land as-is (e.g., on a rebase merge), the breaking nature gets lost.

Could you add a `BREAKING CHANGE:` footer to `b37e3ca8` (the type-shape change) and mark it with `!`, and the same on `eb5f54d7` for the rename? That way the convention is honored regardless of merge strategy.

nssa/core/src/circuit_io.rs
```diff
@@ -69,6 +76,12 @@ pub enum InputAccountIdentity {
         nsk: NullifierSecretKey,
         membership_proof: MembershipProof,
         identifier: Identifier,
+        /// When `Some((seed, pda_program_id))`, the circuit binds this position via the
+        /// external derivation check
+        /// `AccountId::for_private_pda(pda_program_id, seed, npk, identifier) ==
+        /// pre_state.account_id` rather than requiring a caller `pda_seeds` to establish
+        /// the binding. The `pre_state` must have `is_authorized == false`.
```
Arjentix, line 83: But in the comment above there's a statement that authorization should be `true`

program_methods/guest/src/bin/privacy_preserving_circuit/execution_state.rs
Arjentix, line 417: I don't get why we need this change...

Arjentix: I mean, now you can claim other users pda-derived accounts, no?

schouhy: yes sorry I forgot to mention it. Private PDAs are not vulnerable to griefing attacks. It seemed ok to relax that "authorization on claiming" check, which was necessary for the new flow with PrivatePdaInit with external seed.

wallet/src/privacy_preserving_tx.rs
schouhy: thanks I'll fix this.

program_methods/guest/src/bin/privacy_preserving_circuit/execution_state.rs
schouhy: uhm I'm not getting what you mean here. The `pda_program_id` is not meant to be the program owner of the account. It would be an account owned by the auth-transfer program whose authority is the `pda_program_id`. Maybe I should call this variable `authority_program_id`
- About consequence (1), any claim to an initialized account will fail and that's ok. And PDAs are not meant to be owned by the authority program. Also, the authority program can chain a call to the transfer program to spend the funds like in the test, and that doesn't use the external-seed path. Which would actually be the only path to debit the funds in this case. I didn't get what you meant here we should note.
- About (2) I don't think this is true. The program owner is the only program that can modify the account. So the pda in this case is owned by the auth-transfer program. Only that program can debit funds from that account. But no user can grant authorization to do so, only the authority program.

It seems to me that there's some confusion in terms of program ownership and authority program for pdas.

nssa/core/src/circuit_io.rs
schouhy: thanks I'll fix it

program_methods/guest/src/bin/privacy_preserving_circuit/execution_state.rs
schouhy: we need to remove the authorization constraint on claiming private pdas because otherwise there will be no way for bob to send some funds to alice in a (newly initialized) private pda for which bob doesn't know the nsk. It's the same reason we don't impose authorization on claiming a regular private account. The surface area for griefing attacks is reduced compared to public pdas because private ones depend on the npk and an identifier
.
But yes, you could claim other users pdas for a specific npk and identifier.

program_methods/guest/src/bin/privacy_preserving_circuit/execution_state.rs
moudyellaz: Reading `pda_program_id`, I took it as "the program that owns this PDA" because that's how it works for public PDAs (claimer == `program_owner`). So when I saw `program_owner` end up as auth-transfer after a fund I read it as a divergence, and built the two consequences on top of that. Thanks for the `authority_program_id` rename, it makes it clearer.

Arjentix (approved):
LGTM, thanks

PR #487, opened 2026-05-19
feat!: add integration_bench tool for end-to-end scenario latency, block, and tx-byte measurements

moudyellaz:
## 🎯 Purpose

Add a benchmark tool for the full end-to-end LEZ stack: scenarios driven through the wallet against a real Bedrock node, in-process sequencer, and indexer, sharing a single docker-compose session across all scenarios in a run. Records per-step wall time (as `Duration` internally, f64 seconds on the wire), Bedrock L1-finality latency, borsh-serialized block sizes, and per-tx-variant wire sizes. Complements `cycle_bench` (program-level cycles) and `crypto_primitives_bench` (wallet-side primitives).

## ⚙️ Approach

New standalone package at `tools/integration_bench`. Bench uses `test_fixtures::TestContext` (extracted from `integration_tests` in this PR) to bring up the same docker-compose stack the integration tests use, one shared bedrock + sequencer + indexer session for the whole run. Five scenarios:

- `token_onboarding`: sequential public token Send + one shielded recipient setup.
- `amm_swap_flow`: pool create, swap, add liquidity, remove liquidity (all public).
- `multi_recipient_fanout`: one sender to N recipients, sequential.
- `private_chained_flow`: shielded, deshielded, private to private (PPE-bearing).
- `parallel_fanout`: N senders submit concurrently into one block.

Two modes: `RISC0_DEV_MODE=1` for fast latency-only runs, or unset for real STARK proving. JSON dumps default to `target/integration_bench_{dev,prove}.json`; override with `--json-out`. Per-step timing uses a `ScenarioOutput::step` closure helper instead of hand-rolled `Instant::now()` pairs, and end-to-end submit timeouts use `tokio::time::timeout`.

- [x] Add `tools/integration_bench` package with 5 scenarios, harness, and shared TestContext setup.
- [x] Extract `test_fixtures` crate from `integration_tests` so the bench depends on the shared fixtures, not the test crate.
- [x] Add `docs/benchmarks/integration_bench.md` with canonical dev-mode and real-proving result tables from the docker-compose sweep.
- [x] Add `integration_bench` row to `docs/benchmarks/README.md`.
- [x] Register `tools/integration_bench` and `test_fixtures` in workspace `members`.

## 🧪 How to Test

Docker must be running.

```sh
# Dev mode, all 5 scenarios (~17 min on M2 Pro)
RISC0_DEV_MODE=1 cargo run --release -p integration_bench -- --scenario all

# Real proving for private (~13 min)
cargo run --release -p integration_bench -- --scenario private

# Real proving for amm (~9 min); override json path since default collides with private
cargo run --release -p integration_bench -- --scenario amm --json-out /tmp/integration_bench_prove_amm.json
```

See `docs/benchmarks/integration_bench.md` for the canonical result tables (M2 Pro dev box, CPU only) including the per-PPE-step submit mean (~127 s) and `ppe_tx_bytes` (225,728, matching `cycle_bench` `S_agg`).

## 🔗 Dependencies

None.

## 🔜 Future Work

- Criterion-based regression detection (tracked separately).

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests *(bench tool with no automated tests; verified by running the binary across all scenarios; canonical numbers in `docs/benchmarks/integration_bench.md`)*


docs/benchmarks/e2e_bench.md
```diff
@@ -0,0 +1,125 @@
...
+| private_chained_flow | 354,843 | 354.84 | 23,778 | 23.78 | +274.14 s (≈ 91 s per PPE step × 3) |
+
+Per-step breakdown for `private_chained_flow` in real proving:
+
+| Step | submit_ms | inclusion_ms | total_ms | total_s |
+|---|---:|---:|---:|---:|
+| token_new_fungible (public) | 1.1 | 20,276.0 | 20,291.2 | 20.29 |
+| shielded_transfer (PPE) | 111,683.3 | 1.0 | 111,730.4 | 111.73 |
+| deshielded_transfer (PPE) | 111,454.7 | 1.1 | 111,511.2 | 111.51 |
+| private_to_private (PPE) | 111,237.0 | 1.1 | 111,293.0 | 111.29 |
```
Arjentix: Does having `total_ms` and `total_s` here gives us something? Looks like seconds should be enough?

docs/benchmarks/e2e_bench.md
```diff
@@ -0,0 +1,125 @@
...
+- PPE transactions are ≈ 225 KB on the wire in real proving, dominated by the outer succinct proof. Dev mode emits a ≈ 2 KB stub that does not represent the L1 payload — fee-model storage gas inputs must come from a real-proving run.
+- Per-PPE-step prove cost on M2 Pro CPU is ≈ 110-120 s, paid on the wallet side at submit time (not on the sequencer). For a single-program chained flow the cost stacks linearly.
+- Empty clock-only ticks set the per-block fixed-cost baseline at ≈ 334 bytes across all scenarios and both modes.
+- Bedrock L1 finality stays around 20 s regardless of proving mode, because finality is paced by L1 cadence, not the LEZ prover.
+
+## Reproduce
+
+```sh
+export LEZ_BEDROCK_BIN=/path/to/logos-blockchain/target/release/logos-blockchain-node
+export LEZ_BEDROCK_CONFIG_DIR=/path/to/bedrock/configs
```
Arjentix: I suspect the configuration file used here may change the bench results. Or am I wrong? Anyway worth to note somewhere if it does or not.

docs/benchmarks/e2e_bench.md
```diff
@@ -0,0 +1,125 @@
...
+- Public-only scenarios converge between dev mode and real proving in both latency and byte counts. Either mode is suitable to characterize them.
+- PPE transactions are ≈ 225 KB on the wire in real proving, dominated by the outer succinct proof. Dev mode emits a ≈ 2 KB stub that does not represent the L1 payload — fee-model storage gas inputs must come from a real-proving run.
+- Per-PPE-step prove cost on M2 Pro CPU is ≈ 110-120 s, paid on the wallet side at submit time (not on the sequencer). For a single-program chained flow the cost stacks linearly.
+- Empty clock-only ticks set the per-block fixed-cost baseline at ≈ 334 bytes across all scenarios and both modes.
+- Bedrock L1 finality stays around 20 s regardless of proving mode, because finality is paced by L1 cadence, not the LEZ prover.
+
+## Reproduce
+
+```sh
+export LEZ_BEDROCK_BIN=/path/to/logos-blockchain/target/release/logos-blockchain-node
```
Arjentix: Currently not sure about that... So benches run this as a binary? Why not to run it once manually and just pass port to the benchmarks?

docs/benchmarks/e2e_bench.md
```diff
@@ -0,0 +1,125 @@
...
+
+`amm_swap_flow` is byte-identical between dev and real (no proof payload). `private_chained_flow`'s `ppe_tx_bytes` matches the cycle_bench `S_agg` measurement (≈ 225 KB borsh InnerReceipt). The `block_bytes` max (226,578) is the block containing the largest PPE transaction.
+
+## Findings
+
+- Public-only scenarios converge between dev mode and real proving in both latency and byte counts. Either mode is suitable to characterize them.
+- PPE transactions are ≈ 225 KB on the wire in real proving, dominated by the outer succinct proof. Dev mode emits a ≈ 2 KB stub that does not represent the L1 payload — fee-model storage gas inputs must come from a real-proving run.
+- Per-PPE-step prove cost on M2 Pro CPU is ≈ 110-120 s, paid on the wallet side at submit time (not on the sequencer). For a single-program chained flow the cost stacks linearly.
+- Empty clock-only ticks set the per-block fixed-cost baseline at ≈ 334 bytes across all scenarios and both modes.
+- Bedrock L1 finality stays around 20 s regardless of proving mode, because finality is paced by L1 cadence, not the LEZ prover.
```
Arjentix: I think this is exactly what is changed by bedrock config, so not sure it makes sense to paste this here

Cargo.toml
```diff
@@ -75,6 +76,7 @@ faucet_core = { path = "programs/faucet/core" }
 vault_core = { path = "programs/vault/core" }
 test_program_methods = { path = "test_program_methods" }
 testnet_initial_state = { path = "testnet_initial_state" }
+integration_tests = { path = "integration_tests" }
```
Arjentix: I think there is something wrong here. It's strange to depend on `integration_tests`. If you need things like `TestContext` it's better to separate it to a separate crate

tools/e2e_bench/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+Required env vars (no defaults):
+
+```sh
+export LEZ_BEDROCK_BIN=/path/to/logos-blockchain/target/release/logos-blockchain-node
+export LEZ_BEDROCK_CONFIG_DIR=/path/to/bedrock/configs
+# optional: LEZ_BEDROCK_PORT (default 18080)
+```
+
+The config dir must contain `node-config.yaml` and a `deployment-settings.yaml` template with the literal string `PLACEHOLDER_CHAIN_START_TIME` (rewritten per launch).
```
Arjentix: Mmmm, not sure about that. Look how our `integration_tests` do stuff. Config is implemented in the code. And probably supplying different configs from user doesn't make much sense for benchmarks. So instead of relying on placeholder string I'd suggest to implement configuration construction directly in code

tools/integration_bench/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+common.workspace = true
+tempfile.workspace = true
+borsh.workspace = true
+chrono.workspace = true
+
+anyhow.workspace = true
+tokio = { workspace = true, features = ["macros", "rt-multi-thread", "time"] }
+clap.workspace = true
+serde.workspace = true
+serde_json.workspace = true
```
Arjentix, line 24: New line split is a bit strange here. We usually split our crates from external crates. So `jsonrpsee`, `tempfile`, `borsh` and `chrono` should be on the second list

tools/e2e_bench/src/main.rs
```diff
@@ -0,0 +1,229 @@
...
+//! Run examples:
+//!   RISC0_DEV_MODE=1 cargo run --release -p e2e_bench -- --scenario all.
+//!   cargo run --release -p e2e_bench -- --scenario amm.
+//!
+//! `RISC0_DEV_MODE=1` skips proving and produces latency-only numbers in
+//! ~minutes; omitting it produces realistic proving-inclusive numbers but
+//! the run takes much longer.
+
+#![expect(
+    clippy::arbitrary_source_item_ordering,
```
Arjentix: Let's not allow that

tools/e2e_bench/src/main.rs
```diff
@@ -0,0 +1,229 @@
...
+        let disk_before = ctx.disk_sizes();
+        let mut result = run_scenario(name, setup_ms, &mut ctx).await?;
+        result.disk_before = Some(disk_before);
+        result.disk_after = Some(ctx.disk_sizes());
+        result.bedrock_finality_ms = Some(measure_bedrock_finality(&ctx).await?);
+        harness::print_table(&result);
+        all_results.push(result);
+
+        drop(ctx);
+        drop(bedrock);
```
Arjentix: Consider direct lifetime bounds `{ }` instead of explicit drops

tools/e2e_bench/src/main.rs
```diff
@@ -0,0 +1,229 @@
...
+        ScenarioName::Fanout => scenarios::fanout::run(ctx).await?,
+        ScenarioName::Private => scenarios::private::run(ctx).await?,
+        ScenarioName::Parallel => scenarios::parallel::run(ctx).await?,
+        ScenarioName::All => unreachable!("dispatched above"),
+    };
+    Ok(ScenarioResult { setup_ms, ..result })
+}
+
+fn elapsed_ms(t: std::time::Instant) -> f64 {
+    t.elapsed().as_secs_f64() * 1_000.0
```
Arjentix: That kind of calculation eliminates the reason to use `ms`, no?

tools/e2e_bench/src/main.rs
```diff
@@ -0,0 +1,229 @@
...
+        .await
+        .context("connect indexer WS")?;
+    let sequencer_tip = ctx.sequencer_client().get_last_block_id().await?;
+
+    let started = std::time::Instant::now();
+    let deadline = started + Duration::from_secs(60);
+    loop {
+        match indexer_ws.get_last_finalized_block_id().await {
+            Ok(Some(b)) if b >= sequencer_tip => {
+                return Ok(started.elapsed().as_secs_f64() * 1_000.0);
```
Arjentix: Same

Arjentix: Also better to return `Duration` instead of `f64` to increase readability

tools/e2e_bench/src/bench_context.rs
```diff
@@ -0,0 +1,205 @@
...
+use tempfile::TempDir;
+use wallet::{WalletCore, config::WalletConfigOverrides};
+
+const DEFAULT_BEDROCK_ADDR: &str = "127.0.0.1:18080";
+
+#[expect(
+    clippy::partial_pub_fields,
+    reason = "Internal TempDirs are kept alive via private fields for RAII; \
+              client and wallet are public for scenarios to drive."
+)]
```
Arjentix: Let's remove pubs and create getters instead

tools/e2e_bench/src/bench_context.rs
```diff
@@ -0,0 +1,205 @@
...
+    pub wallet_password: String,
+    sequencer_handle: Option<SequencerHandle>,
+    indexer_handle: IndexerHandle,
+    temp_indexer_dir: TempDir,
+    temp_sequencer_dir: TempDir,
+    temp_wallet_dir: TempDir,
+}
+
+impl BenchContext {
+    pub async fn new() -> Result<Self> {
```
Arjentix: Ah, that's a copy-paste from `TestContext`. I think there should be a way to make `TestContext` more configurable and use it instead

tools/e2e_bench/src/bedrock_handle.rs
```diff
@@ -0,0 +1,147 @@
...
+    env,
+    net::SocketAddr,
+    path::PathBuf,
+    process::{Child, Command, Stdio},
+    time::{Duration, Instant},
+};
+
+use anyhow::{Context as _, Result, bail};
+
+pub struct BedrockHandle {
```
Arjentix: So it runs a binary. But what was wrong with docker compose? I'm sure it doesn't create any difference in terms of speed, while keeps all runtime dependencies isolated in `bedrock` dir and in docker images their team provides. One day they may require an SQLIte for example and that could be isolated inside docker setup without touching our tests.

Also the beauty of `integration_tests::TestContext` is that it's fully self-contained: everything in one instance is isolated from other instances. That means you could even possibly run benchmarks in parallel with it, and I wonder if it will create any difference or not

Arjentix: So the idea was to isolate the thing we don't own -- bedrock, and to be explicit about things we own -- sequencer, indexer, wallet.

tools/e2e_bench/src/bench_context.rs
```diff
@@ -0,0 +1,205 @@
...
+pub struct BenchContext {
+    pub sequencer_client: SequencerClient,
+    pub wallet: WalletCore,
+    #[expect(
+        dead_code,
+        reason = "Retained for parity with TestContext; may be needed later."
+    )]
+    pub wallet_password: String,
+    sequencer_handle: Option<SequencerHandle>,
+    indexer_handle: IndexerHandle,
```
Arjentix: Hmm, now I see that... It's kind of not what e2e is IMO. In real e2e you would run released binaries directly (like you do right now with bedrock). While I don't expect that it will affect bench results much...

Yeah, current approach with `TestContext` makes more sense in terms of simplicity, but I'd say these are integration benchmarks, not e2e benchmarks. I'd even say that I'm not sure we need real e2e benchmarks.

So I'd suggest to rename to `integration_bench`

tools/e2e_bench/src/harness.rs
```diff
@@ -0,0 +1,297 @@
...
+}
+
+#[derive(Debug, Serialize, Default)]
+pub struct ScenarioResult {
+    pub name: String,
+    pub setup_ms: f64,
+    pub steps: Vec<StepResult>,
+    pub total_ms: f64,
+    /// Disk sizes (sequencer / indexer / wallet tempdirs) sampled at scenario start.
+    pub disk_before: Option<crate::bench_context::DiskSizes>,
```
Arjentix: DiskSizes should be probably defined in this module

tools/e2e_bench/src/harness.rs
```diff
@@ -0,0 +1,297 @@
...
+/// let started = Instant::now();
+/// let ret = wallet::cli::execute_subcommand(ctx.wallet_mut(), cmd).await?;
+/// let step = finalize_step("label", started, ret, ctx).await?;
+/// ```
+/// Begin a timed step. Capture this *before* submitting the wallet operation
+/// so we can later subtract it from the post-submit block height to detect
+/// when the chain has advanced past the tx's block.
+pub async fn begin_step(ctx: &BenchContext) -> Result<u64> {
+    Ok(ctx.sequencer_client().get_last_block_id().await?)
+}
```
Arjentix: The comment is for wrong function

tools/e2e_bench/src/harness.rs
```diff
@@ -0,0 +1,297 @@
...
+        if let SubcommandReturnValue::PrivacyPreservingTransfer { tx_hash } = ret {
+            tx_hash_str = Some(format!("{tx_hash}"));
+        }
+        let started_inclusion = Instant::now();
+        wait_for_chain_advance(ctx, pre_block_id, 2).await?;
+        inclusion_ms = Some(started_inclusion.elapsed().as_secs_f64() * 1_000.0);
+
+        let started_sync = Instant::now();
+        sync_wallet_to_tip(ctx).await?;
+        wallet_sync_ms = Some(started_sync.elapsed().as_secs_f64() * 1_000.0);
```
Arjentix: Same strange thing about ms counting

tools/e2e_bench/src/harness.rs
```diff
@@ -0,0 +1,297 @@
...
+
+pub async fn finalize_step(
+    label: impl Into<String>,
+    started: Instant,
+    pre_block_id: u64,
+    ret: &SubcommandReturnValue,
+    ctx: &mut BenchContext,
+) -> Result<StepResult> {
+    let label = label.into();
+    let submit_ms = started.elapsed().as_secs_f64() * 1_000.0;
```
Arjentix: And here

tools/integration_bench/src/harness.rs
```diff
@@ -0,0 +1,297 @@
...
+            Err(err) => eprintln!("get_last_block_id error (continuing poll): {err:#}"),
+        }
+        if Instant::now() > deadline {
+            bail!(
+                "chain did not advance from {from_block_id} to at least {target} within {TX_INCLUSION_TIMEOUT:?}"
+            );
+        }
+        tokio::time::sleep(TX_INCLUSION_POLL_INTERVAL).await;
+    }
+}
```
Arjentix, line 208: Better use tokio's [`timeout`](https://docs.rs/tokio/latest/tokio/time/fn.timeout.html). And I think I've already saw another place with this kind of code, same suggestion

Pravdyvy (approved):
lgtm

tools/e2e_bench/src/scenarios/token.rs
```diff
@@ -0,0 +1,127 @@
...
+use wallet::cli::{
+    Command, SubcommandReturnValue,
+    account::{AccountSubcommand, NewSubcommand},
+    programs::token::TokenProgramAgnosticSubcommand,
+};
+
+use crate::harness::{ScenarioResult, finalize_step};
+
+pub async fn run(ctx: &mut crate::bench_context::BenchContext) -> Result<ScenarioResult> {
+    let mut result = ScenarioResult::new("token_onboarding");
```
Arjentix: I don't really like using `result` naming not for `Result` type. Let's call it `ScenarioOutput` and use `output` for variable

tools/integration_bench/src/scenarios/token.rs
```diff
@@ -0,0 +1,127 @@
...
+                definition_account_id: public_mention(definition_id),
+                supply_account_id: public_mention(supply_id),
+                name: "BenchToken".to_owned(),
+                total_supply: 1_000_000,
+            }),
+        )
+        .await?;
+        let step = finalize_step("token_new_fungible", started, pre_block, &ret, ctx).await?;
+        result.push(step);
+    }
```
Arjentix: Hmm, looks like a good place for RAII. Or some closure. There are different ideas, for example I could imagine something like this:

```rust
let mut scenario = Scenario::new("name");
scenario.make_step("step_name", || {
    // Step code here
});
// Other steps here

// Getting output
scenario.output()
```

Arjentix (changes requested):
Hey, thanks for the PR.

Left some comments. My main concerns are:

1. Usage of bedrock binary (I suggest to still use with docker)
2. Not using binaries of our sequencer, indexer and wallet and using their library code instead. It makes all of this `integration_bench` not `e2e_bench`. IMO it's alright, just don't call it `e2e_bench`.
3. Depending on `integration_tests` and mimicing `TestContext` with duplicated code, while we could possibly expand `TestContext`, move it to a separate crate and reuse here.

I understand that these are just benches and not the main code. But this is still a part of our project which we have to support in future (even after moving to criterion). So I'm suggesting to refactor some places to make them easier to work with.

test_fixtures/src/lib.rs
```diff
@@ -0,0 +1,500 @@
+//! Shared test/bench fixtures: spins up bedrock + sequencer + indexer + wallet
+//! end-to-end against docker-compose, exposes a `TestContext` callers can drive.
+//!
+//! Originally lived under `integration_tests`; split out so non-test consumers
+//! (e.g. `integration_bench`) can depend on the fixtures without pulling in the
+//! `integration_tests` test files.
```
Arjentix: Let's not leave history comments in code, we have `git` for history. I've noticed that Claude Code loves to do it for some reason

docs/benchmarks/e2e_bench.md
moudyellaz: Dropped `total_ms`, kept seconds only.

docs/benchmarks/e2e_bench.md
moudyellaz: The bench now runs against a single docker-compose stack via TestContext, so the config is the standard integration-tests config and that's documented in the new README.

docs/benchmarks/e2e_bench.md
moudyellaz: Pivoted to docker-compose via TestContext, with a single shared session across all scenarios in a run.

docs/benchmarks/e2e_bench.md
moudyellaz: Removed the machine table.

Cargo.toml
moudyellaz: Extracted a `test_fixtures` crate from `integration_tests`; the bench depends on that instead.

tools/e2e_bench/README.md
moudyellaz: Config is now in-code via TestContext; placeholder templating is gone.

tools/integration_bench/Cargo.toml
moudyellaz: Reordered, workspace crates first, then externals.

tools/e2e_bench/src/main.rs
moudyellaz: Blanket allow removed.

tools/e2e_bench/src/main.rs
moudyellaz: Moot, that block was removed when `BenchContext` went away.

tools/e2e_bench/src/main.rs
moudyellaz: Switched to `Duration` internally and seconds f64 on the JSON wire.

tools/e2e_bench/src/main.rs
moudyellaz: Same, `Duration` internally and seconds f64 on the wire (so the returned values are `Duration`, not `f64`).

tools/e2e_bench/src/bench_context.rs
moudyellaz: Moot, `BenchContext` is deleted entirely.

tools/e2e_bench/src/bench_context.rs
moudyellaz: Made `TestContext` configurable in test_fixtures and reused it; the bench-specific copy is gone.

tools/e2e_bench/src/bedrock_handle.rs
moudyellaz: Pivoted to docker-compose; `bedrock_handle.rs` and the binary-spawn path are deleted. Docker now isolates the whole stack including bedrock.

tools/e2e_bench/src/bench_context.rs
moudyellaz: Renamed the tool to `integration_bench` and pivoted to docker-compose, same as the integration tests.

tools/e2e_bench/src/harness.rs
moudyellaz: `DiskSizes` is now produced by `TestContext::disk_sizes()` in test_fixtures; harness consumes it via that getter.

tools/e2e_bench/src/harness.rs
moudyellaz: Docstring fixed; the harness rewrite shifted the affected function as well.

tools/e2e_bench/src/harness.rs
moudyellaz: Switched to `Duration` internally and seconds f64 on the wire.

tools/e2e_bench/src/harness.rs
moudyellaz: Switched to `Duration` internally and seconds f64 on the wire.

tools/integration_bench/src/harness.rs
moudyellaz: Switched to `tokio::time::timeout`.

tools/e2e_bench/src/scenarios/token.rs
moudyellaz: Renamed `ScenarioResult` to `ScenarioOutput`, with the variable now `output`.

tools/integration_bench/src/scenarios/token.rs
moudyellaz: Added a `ScenarioOutput::step` closure helper; scenarios now call `output.step("label", || async { ... }).await`.

Arjentix (approved):
Thanks for addressing my comments. Looks much better to me now.

Cargo.toml
Arjentix: I think now we can remove `integration_tests` from workspace dependencies as no one should depend on them

PR #488, opened 2026-05-19
feat: migrate bench tools to criterion harness

moudyellaz:
## 🎯 Purpose

Migrate the two microbenchmark tools (`crypto_primitives_bench` and `cycle_bench`'s verify path) from hand-rolled `Instant::now()` timing loops to the criterion harness. Gives statistical CIs, outlier flagging, and a richer comparison story for future PR gating.

## ⚙️ Approach

`crypto_primitives_bench`: replaced the custom timing loop with a criterion bench under `tools/crypto_primitives_bench/benches/primitives.rs`. The binary entry point is gone (no `src/`), the crate is now bench-only with `harness = false`.

`cycle_bench`: split into a library + binary so benches can reuse the helpers. The hand-rolled `--verify` CLI flag, `--verify-iters` argument, and `run_verify(iters)` helper are removed. A new criterion target at `tools/cycle_bench/benches/verify.rs` reuses `prove_auth_transfer_in_ppe` (extracted from the binary into the new `cycle_bench` library) to set up one PPE receipt outside the timed loop, then samples `Receipt::verify(PRIVACY_PRESERVING_CIRCUIT_ID)` under criterion's statistical sampler (sample_size 100, warm_up 2 s, measurement 15 s, noise_threshold 5%). Also aligns the auth_transfer call with the current `authenticated_transfer_core::Instruction::Transfer { amount }` enum API.

- [x] Migrate `crypto_primitives_bench` to criterion (replace bin with bench target).
- [x] Refactor `cycle_bench` into lib + binary; drop `--verify` CLI flag and `run_verify` helper.
- [x] Add criterion verify bench for `G_verify` at `tools/cycle_bench/benches/verify.rs`.
- [x] Add Justfile recipe `just bench` that runs both criterion targets.
- [x] Update `docs/benchmarks/cycle_bench.md` and `tools/cycle_bench/README.md`.

## 🧪 How to Test

```sh
# Run the criterion benches locally:
cargo bench -p crypto_primitives_bench --bench primitives
cargo bench -p cycle_bench --features ppe --bench verify
# Or both via the Justfile recipe:
just bench
```

Criterion output lands under `target/criterion/<group>/<bench>/`. The `cycle_bench` verify bench runs one full PPE prove for setup (minutes) before the timed loop, so expect a long warmup on first run.

## 🔗 Dependencies

None.

## 🔜 Future Work

Add a `bench-regression` GitHub workflow that uses `boa-dev/criterion-compare-action` to gate `crypto_primitives_bench` regressions on PRs. The workflow needs the base branch to already have the criterion bench, so it ships as a follow-up after this PR merges.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


Pravdyvy (approved):
lgtm

Justfile
```diff
@@ -23,6 +23,12 @@ test:
     @echo "🧪 Running tests"
     RISC0_DEV_MODE=1 cargo nextest run --no-fail-fast
 
+# Run criterion benches: fast crypto primitives, then the slow PPE verify (real proving setup).
+bench:
```
Arjentix, line 27: Nice

tools/crypto_primitives_bench/Cargo.toml
```diff
@@ -8,11 +8,12 @@ publish = false
...
-[dependencies]
+[dev-dependencies]
 key_protocol.workspace = true
 nssa_core = { workspace = true, features = ["host"] }
-
-anyhow.workspace = true
-serde.workspace = true
-serde_json.workspace = true
 rand = { workspace = true }
+criterion = { workspace = true, features = ["html_reports"] }
```
Arjentix: Hm, I wonder why we need this feature here. So in order for `cargo bench` to generate html we need to activate this feature in every benchmark?

tools/crypto_primitives_bench/benches/primitives.rs
```diff
@@ -0,0 +1,90 @@
...
+}
+
+fn bench_encryption(c: &mut Criterion) {
+    // One-time setup: a fixed Account/Commitment and a SharedSecretKey to bench
+    // encrypt/decrypt over a representative note. ESK gen is excluded from the
+    // measured loop (covered by the SharedSecretKey bench above).
+    let recipient_kc = KeyChain::new_os_random();
+    let vpk = recipient_kc.viewing_public_key;
+    let account = Account::default();
+    let account_id = AccountId::new([7; 32]);
```
Arjentix: Let's derive account id from key, so that it will be a more realistic scenario

tools/cycle_bench/README.md
```diff
@@ -4,6 +4,8 @@ Per-program Risc0 cycle counts, prover wall time, PPE composition cost, and veri
 
 ## Run
 
+The binary handles executor cycles, prover wall time, and PPE composition cost:
```
Arjentix, line 7: So why do we still need this thing as a binary? Can't it also be just a criterion benchmark?

Arjentix (commented):
Thanks for the PR, I like new approach. My main question is why not everything was moved to criterion? Also I can see that integration benches where not touched

tools/crypto_primitives_bench/Cargo.toml
moudyellaz: Lifted the feature into the workspace-level criterion declaration so every consumer inherits HTML reports without repeating it.

tools/crypto_primitives_bench/benches/primitives.rs
moudyellaz: Switched to deriving the account_id from the recipient keychain's nullifier public key via `AccountId::for_regular_private_account(&npk, 0)`.

tools/cycle_bench/README.md
moudyellaz: The binary measures things criterion doesn't fit. Per-program executor cycle counts come from the executor as numeric counts on one run, not wall time. The per-program prove and PPE composition sweep (chain_caller depth 1, 3, 5, 9) each take minutes; criterion's 50-100 sample loop would multiply that into hours. Output is a JSON dump under `target/cycle_bench.json` consumed by the fee-model analysis, not the CI/outlier comparison criterion is built around. Verify was the only tight-loop microbench in here and that's the one we moved.

Arjentix: Hmm, I see, good

Arjentix (approved):
Cool. Maybe also rename `tools` to `benches`?

PR #489, opened 2026-05-19
fix(nssa): audit 91 issue fix

jonesmarvin8:
## 🎯 Purpose

The purpose of this PR is to address the security vulnerability highlighted in [audit-issue 91](https://github.com/logos-co/audit-reports/issues/91). Specifically, a pair of malicious programs can be used to drain a victim's account; P1 injects a victim's account (and authorization) in a chained-call, P2 uses victim's account (with authorization) to call `auth-transfer` with victim's account as the sender. The fix ensures that a chained call can only propagate authorization that the caller's own `program_output.pre_states` already carry, never elevate beyond what the caller was authorized for.

## ⚙️ Approach

[X] - Updates `authorized_accounts` (in validate_state_diff.rs, line 268) to to get `pre_states` from `program_output` (validated by circuit logic) rather than unvalidated `chained_call`.
[X] - Add pair of test-program-methods: `malicious_injector` and `malicious_launderer`.
[X] - Add test to `validate_state_diff` using the malicious programs demonstrating the vulnerability has been patched.

## 🧪 How to Test

Run tests `malicious_programs_cannot_drain_victim_without_signature` and `privacy_malicious_programs_cannot_drain_public_victim`
- The test passes provided the victim's balance is unaffected by the execution.

## 🔗 Dependencies

N/A

## 🔜 Future Work

N/A

## 📋 PR Completion Checklist

- [X] Complete PR description
- [X] Add/update tests


schouhy (commented):
thanks very cool. Are private txs safe? Can we add a test that checks there's no vulnerability in a private transaction too?

jonesmarvin8:
> thanks very cool. Are private txs safe? Can we add a test that checks there's no vulnerability in a private transaction too?

The privacy txs are safe from this attack against public accounts. The individual inner proofs for each chained call proofs are generated, but the outer proof is rejected by the validator. I've added a similar test for privacy transaction using the malicious programms.

nssa/src/validated_state_diff.rs
```diff
@@ -488,3 +488,253 @@ fn n_unique<T: Eq + Hash>(data: &[T]) -> usize {
...
+    use nssa_core::account::{AccountId, Nonce};
+
+    use crate::{
+        PrivateKey, PublicKey, V03State,
+        program::Program,
+        public_transaction::{Message, WitnessSet},
+        validated_state_diff::ValidatedStateDiff,
+    };
+
+    /// Privacy-path version of the authorization-injection attack.
```
schouhy: IMO the wording here makes it look like the vulnerability still exists

schouhy (approved):
LGTM thanks!

Pravdyvy (approved):
lgtm

nssa/src/validated_state_diff.rs
```diff
@@ -488,3 +488,253 @@ fn n_unique<T: Eq + Hash>(data: &[T]) -> usize {
...
+    ///
+    /// `execute_and_prove` succeeds: all inner receipts are valid, and the outer circuit
+    /// honestly commits `victim(is_authorized=true)` to its journal.
+    /// `from_privacy_preserving_transaction` rejects the proof because the NSSA validator
+    /// independently reconstructs `public_pre_states` from chain state using
+    /// `signer_account_ids.contains(victim_id) = false`, producing `victim(is_authorized=false)`.
+    /// The committed journal and the expected output diverge, so `receipt.verify` fails.
+    #[test]
+    fn privacy_malicious_programs_cannot_drain_public_victim() {
```
moudyellaz, line 522: Quick question on coverage, not a blocker. The two new tests nail public attacker → public victim and attacker-with-private-account → public victim, which is great. The case attacker → private victim isn't exercised, and I wanted to double check the reasoning since the fix lives on the host path while the circuit has a similar chained_call.pre_states → caller authorized set shape. From reading program_methods/guest/src/bin/privacy_preserving_circuit.rs my read is: Vacant branch in ExecutionState::push_states accepts pre.is_authorized verbatim on first sight, but visibility_mask=1 then requires a sealing-key proof (the "Pre-state not authorized for authenticated private account" assertion), and mask=0 routes the account into public_pre_states where the host fix catches it. If that matches your intent, could you add a one-liner in the PR body, or even better a small test for the private-victim path so the property is locked in? Happy either way.

nssa/src/validated_state_diff.rs
```diff
@@ -488,3 +488,253 @@ fn n_unique<T: Eq + Hash>(data: &[T]) -> usize {
...
+        )
+        .unwrap();
+
+        let witness_set = WitnessSet::for_message(&message, &[&attacker_key]);
+        let tx = crate::PublicTransaction::new(message, witness_set);
+
+        let result = ValidatedStateDiff::from_public_transaction(&tx, &state, 1, 0);
+
+        assert!(
+            result.is_err(),
```
moudyellaz: Small suggestion: would you mind pinning the expected error variant here instead of just is_err()? As written either test would still pass if a future refactor rejected the tx for an unrelated reason (balance, nonce, proof shape), and we'd lose the audit-91 regression signal. Something like `assert!(matches!(result, Err(NssaError::InvalidProgramBehavior(InvalidProgramBehaviorError::InvalidAccountAuthorization { account_id })) if account_id == victim_id))` for the public test, and matching on the proof-verification error for the privacy one. Not a blocker.

nssa/src/validated_state_diff.rs
```diff
@@ -265,7 +265,7 @@ impl ValidatedStateDiff {
                 state_diff.insert(pre.account_id, post.account().clone());
             }
 
-            let authorized_accounts: HashSet<_> = chained_call
+            let authorized_accounts: HashSet<_> = program_output
```
moudyellaz, line 272: Tiny nit, feel free to ignore. Would you mind dropping a short comment right at this line pinning down the why, so a future reader doesn't accidentally undo it? Something like: `// Source from program_output.pre_states, not chained_call.pre_states: the loop above already gates program_output's is_authorized via the !pre.is_authorized || is_indeed_authorized check, while chained_call.pre_states is caller-controlled and can be forged (audit-issue 91).` A pointer to the audit issue plus the gating invariant is probably enough.

moudyellaz (approved):
lgtm, left some comments. Thanks for the fix again

nssa/src/validated_state_diff.rs
moudyellaz: Just want to second schouhy's note here, the current phrasing tripped me up the same way on first read. Maybe something like: "execute_and_prove succeeds because each inner receipt is individually valid and the outer circuit faithfully commits whatever the attacker's program output says, including victim(is_authorized=true). The host-side validator is what catches the attack: it independently reconstructs public_pre_states from chain state using signer_account_ids.contains(victim_id) = false, so it expects victim(is_authorized=false). The committed journal and the reconstructed expected output diverge, and receipt.verify fails." Feel free to trim further, just wanted to help the "host validator is the gatekeeper, not the circuit" point land more obviously. Thanks for adding the privacy version of the test, it's a really nice addition.

moudyellaz:
One tiny writeup ask, totally optional. Could you add a sentence to the PR description or commit body spelling out the invariant the fix locks in? Something like "a chained call can only propagate authorization that the caller's own program_output.pre_states already carry, never elevate beyond what the caller was authorized for." It would save the next person who looks at this from having to re-derive the property from the diff. Really nice fix overall, thanks for taking this on.

nssa/src/validated_state_diff.rs
Arjentix, line 272: I see... So my initial mistake was that I've assumed that `chained_call.pre_states == program_output.pre_states`. But turns out it's not. But shouldn't we better check that?

nssa/src/validated_state_diff.rs
jonesmarvin8: Added a test `privacy_malicious_programs_cannot_drain_private_victim` that handles the host mask=0 possibility. In the comments for this test, I add why mask=1 is not a concern (attacker would need the private account's nsk).

nssa/src/validated_state_diff.rs
jonesmarvin8: Could this be addressed in a future PR if necessary?

jonesmarvin8: Saving this for a future PR as on-going [conversation](https://discord.com/channels/973324189794697286/1506354177712849037).

PR #429, opened 2026-04-02
feat: add wallet ffi function for arbitrary public transactions

adklempner:
## Purpose

Add a generic FFI function for sending arbitrary public transactions to any deployed program, removing the need to add per-program FFI wrappers.

Related: logos-blockchain/logos-execution-zone#422

## Approach

- [x] Add `wallet_ffi_send_public_transaction` C FFI function that accepts a program ID, account list, raw instruction bytes, and a signer
- [x] Add `WalletCore::send_public_transaction` async method that builds, signs, and submits a `PublicTransaction`
- [x] Add `InvalidArgument` error variant for input validation (e.g. instruction length not a multiple of 4)
- [x] Generate updated C header (`wallet_ffi.h`)

## How to Test

1. Build the wallet-ffi crate: `cargo build -p wallet-ffi`
2. Call `wallet_ffi_send_public_transaction` from a logos-core module (see companion PR in logos-execution-zone-module)
3. Verify the transaction is accepted by the sequencer and state is updated

## Dependencies

None.

## Future Work

- Support arbitrary private (shielded) transactions via FFI (the other half of logos-blockchain/logos-execution-zone#422)

## PR Completion Checklist

- [x] Complete PR description
- [x] Implement the core functionality
- [ ] Add/update tests
- [ ] Add/update documentation and inline comments

PR #491, opened 2026-05-20
feat(wallet_ffi): wallet ffi generic transactions

Pravdyvy:
## 🎯 Purpose

Extension of `wallet_ffi` to support custom public and privacy-preserving transactions.

## ⚙️ Approach

Exposes functions to resolve public and private accounts.

- [x] Added functions `wallet_ffi_resolve_public_account`  and `wallet_ffi_resolve_private_account`.
- [x] Generic public transactions added to FFI.
- [x] Generic private transactions added to FFI.
- [x] Added integration tests for generic cases.

## 🧪 How to Test

Added integration tests for generic cases.

## 🔗 Dependencies

None

## 🔜 Future Work

PDA, shared accounts, keycard, etc.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


wallet-ffi/src/types.rs
```diff
@@ -266,3 +318,224 @@ impl TryFrom<&FfiPublicAccountKey> for nssa::PublicKey {
...
+                let vpk_len = vpk_vec.len();
+                let vpk_data = if vpk_len > 0 {
+                    let vpk_data_boxed = vpk_vec.into_boxed_slice();
+                    Box::into_raw(vpk_data_boxed) as *const u8
+                } else {
+                    ptr::null()
+                };
+
+                Self {
+                    kind: FfiAccountIdentityKind::PrivateShared,
```
moudyellaz: Nice work threading all eight identity kinds through here. One thing in the PrivatePdaShared arm: kind is set to FfiAccountIdentityKind::PrivateShared (variant 6) rather than PrivatePdaShared (variant 7). The arm still writes account_id, but on the way back TryFrom matches PrivateShared and rebuilds AccountIdentity::PrivateShared, which has no account_id, so a shared PDA quietly loses its account_id and flips variant on the round trip. resolve_private_account hands back PrivatePdaShared for shared PDAs and it flows straight through this .into(), so it is reachable in practice. Setting kind to PrivatePdaShared should be all it needs. Could you also add a tiny round-trip test over the kinds to lock this down?

lez/wallet-ffi/src/keys.rs
```diff
@@ -250,3 +252,117 @@ pub unsafe extern "C" fn wallet_ffi_account_id_from_base58(
...
+    out_account_identity: *mut FfiAccountIdentity,
+) -> WalletFfiError {
+    let resolved_account = if needs_sign {
+        AccountIdentity::Public(account_id.into())
+    } else {
+        AccountIdentity::PublicNoSign(account_id.into())
+    };
+
+    unsafe {
+        *out_account_identity = resolved_account.into();
```
moudyellaz, line 286: Small safety nit on the two resolve functions: wallet_ffi_resolve_public_account writes through out_account_identity without a null check, and wallet_ffi_resolve_private_account does the same at line 329. The rest of the crate guards its out pointers and returns NullPointer, for example wallet_ffi_get_public_account_key and wallet_ffi_get_private_account_keys. Mind adding the same is_null guard to both so they match the house style?

wallet-ffi/src/generic_transaction.rs
```diff
@@ -0,0 +1,449 @@
...
+        }
+    };
+
+    let mut accounts = Vec::with_capacity(account_identities_size);
+    let mut instruction_data = Vec::with_capacity(instruction_words_size);
+
+    // Alignment will be different, we need to read elements one-by-one
+    for i in 0..account_identities_size {
+        accounts.push(
+            match match unsafe { account_identities.add(i).as_ref() }
```
moudyellaz: account_identities.add(i).as_ref() is UB once the size exceeds the real array, and since as_ref only null-checks, the "does not match actual size" branch can never fire. The size is caller-trusted anyway, so building the slice once with slice::from_raw_parts and documenting that contract would be cleaner. Same block lives in the private fn at :365, so a shared helper would fix both at once. Minor: the null messages say "output pointer" for what are input pointers.

wallet-ffi/src/generic_transaction.rs
```diff
@@ -0,0 +1,449 @@
...
+                    .into_boxed_slice();
+                let secrets_data = Box::into_raw(boxed_slice) as *const FfiBytes32;
+
+                (*out_result).secrets_size = secrets_size;
+                (*out_result).secrets_data = secrets_data;
+            }
+            WalletFfiError::Success
+        }
+        Err(e) => {
+            print_error(format!("Public send failed: {e:?}"));
```
moudyellaz: Tiny copy-paste: this is the private path but the message says "Public send failed".

wallet-ffi/src/keys.rs
```diff
@@ -250,3 +252,117 @@ pub unsafe extern "C" fn wallet_ffi_account_id_from_base58(
...
+/// Resolve private account.
+///
+/// # Parameters
+/// - `handle`: Valid wallet handle
+/// - `account_id`: 32 bytes of the public account ID
+/// - `out_account_identity`: valid pointer, where output will be written
+///
+/// # Returns
+/// - `Success` on successful retrieval
+/// - `InternalError` if wailed to lock wallet
```
moudyellaz: A few small typos in the new docs while you're here:
keys.rs:296 "wailed to lock" should be "failed to lock"
keys.rs:297 "failed to found account" should be "if the account is not found"
keys.rs:260 "does account needs signing" should be "whether the account needs signing"
keys.rs:267 and :301, and types.rs:182, reference FfiAccountManagerAccountIdentity, which should be FfiAccountIdentity
types.rs:195 "representing of account identity ... at intialization" should be "representing an account identity ... at initialization"
generic_transaction.rs:129 "Public transaction have" should be "Public transactions have"

moudyellaz (approved):

wallet-ffi/src/types.rs
Pravdyvy: Fixed , added roundtrip test.

lez/wallet-ffi/src/keys.rs
Pravdyvy: Unified with rest of crate.

wallet-ffi/src/generic_transaction.rs
Pravdyvy: Replaced with `slice::from_raw_parts`. Fixed null pointer error message.

wallet-ffi/src/generic_transaction.rs
Pravdyvy: Fixed

wallet-ffi/src/keys.rs
Pravdyvy: Fixed

nssa/src/program.rs
```diff
@@ -22,8 +22,8 @@ const MAX_NUM_CYCLES_PUBLIC_EXECUTION: u64 = 1024 * 1024 * 32; // 32M cycles
 
 #[derive(Clone, Debug, PartialEq, Eq, BorshSerialize, BorshDeserialize)]
 pub struct Program {
-    id: ProgramId,
-    elf: Vec<u8>,
+    pub id: ProgramId,
+    pub elf: Vec<u8>,
```
schouhy: why making them pub if we already have the getter methods? Can we revert this

wallet-ffi/wallet_ffi.h
```diff
@@ -200,6 +218,61 @@ typedef struct FfiAccount {
   struct FfiU128 nonce;
 } FfiAccount;
 
+typedef struct SerializationHelperResult {
```
schouhy: why not call this `FfiInstructionWords` or something along those lines?

lez/wallet-ffi/src/generic_transaction.rs
```diff
@@ -0,0 +1,420 @@
...
+/// - `input_instruction_data_size`: Size of `input_instruction_data`
+///
+/// # Returns
+/// - `Success` on successful creation
+/// - Error code on failure
+///
+/// # Safety
+/// - `input_instruction_data` must be a valid pointer
+#[no_mangle]
+pub unsafe extern "C" fn wallet_ffi_serialization_helper(
```
schouhy, line 158: shouldn't there be a free function for the res_ptr here?

lez/wallet-ffi/src/keys.rs
```diff
@@ -250,3 +252,127 @@ pub unsafe extern "C" fn wallet_ffi_account_id_from_base58(
...
+            viewing_public_key_len,
+            identifier: _,
+        } = *account_identity;
+
+        if !viewing_public_key.is_null() {
+            let slice = std::slice::from_raw_parts_mut(
+                viewing_public_key.cast_mut(),
+                viewing_public_key_len,
+            );
+            drop(Box::from_raw(std::ptr::from_mut::<[u8]>(slice)));
```
schouhy, line 376: don't we need to make viewing_public_key = nullptr before returning? to avoid a second call from double freeing the memory

wallet/src/account_manager.rs
```diff
@@ -10,7 +10,7 @@ use nssa_core::{
 
 use crate::{ExecutionFailureKind, WalletCore};
 
-#[derive(Clone)]
+#[derive(Debug, Clone, PartialEq, Eq)]
```
schouhy: this will make variants holding secret keys print them in logs. Safest side is to manually implement Debug and omit secret keys.

nssa/src/program.rs
schouhy: making them public allows the constructor with the { id: , elf: } syntax. a program with the wrong id can be constructed then

schouhy (commented):
thanks for the PR. leaving some comments

wallet/src/account_manager.rs
Pravdyvy: Implemented `Debug` manually.

lez/wallet-ffi/src/generic_transaction.rs
Pravdyvy: Added free function for `FfiInstructionWords`

wallet-ffi/wallet_ffi.h
Pravdyvy: Renamed

nssa/src/program.rs
Pravdyvy: Removed `pub`.
I needed to consume it in the cast into ffi type. This cast is for integration tests purposes, so rewrote it with cloning, should be fine.

lez/wallet-ffi/src/keys.rs
Pravdyvy: No, we don't. Dropping raw pointer does nothing, even more `drop(Box::from_raw(std::ptr::from_mut::<[u8]>(slice)));` already sets `viewing_public_key` as null pointer, added a test for this for clarity.

Arjentix:
Btw, I think this might be related: https://github.com/logos-blockchain/logos-execution-zone/pull/429

I wonder if 429 will be not needed after this PR

Pravdyvy:
> Btw, I think this might be related: #429
> 
> I wonder if 429 will be not needed after this PR

It is related, it would not be needed anymore, furthermore, author of #429 knows about its existence.
We can freely close #429 after merging.

s-tikhomirov:
Will this (or another related) PR also supercede https://github.com/logos-blockchain/logos-execution-zone-module/pull/16 (i.e. expose send_public_transaction in the wallet module)?

Pravdyvy:
> Will this (or another related) PR also supercede [logos-blockchain/logos-execution-zone-module#16](https://github.com/logos-blockchain/logos-execution-zone-module/pull/16) (i.e. expose send_public_transaction in the wallet module)?

Yes, precisely this one: https://github.com/logos-blockchain/logos-execution-zone-module/pull/19 will supercede 16.

schouhy (approved):
lgtm thanks!

PR #493, opened 2026-05-21
feat: Implement deposit operation for bridge

Arjentix:
## 🎯 Purpose

This PR implements deposit flow of L1 -> L2 bridge.

## ⚙️ Approach

- Added bridge program with `Deposit` instruction
- Bridge account is deposited from genesis via new special genesis action
- Sequencer listens for `Deposit` event from Bedrock
- Then event happens sequencer creates `bridge_core::Deposit` instruction and pushes it to the mempool
- As this instruction should only be executed by sequencer now we store `TransactionOrigin` to skip validation for sequencer transactions
- Added tests, including full integration check with Bedrock

## 🧪 How to Test

```bash
RUST_LOG=info RISC0_DEV_MODE=1 cargo nextest run -p integration_tests -- bridge  
```

## 🔗 Dependencies

- https://github.com/logos-blockchain/logos-blockchain/pull/2776

## 🔜 Future Work

- Withdraw flow
- Indexer side validation

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


common/src/transaction.rs
```diff
@@ -184,3 +170,21 @@ pub fn clock_invocation(timestamp: clock_core::Instruction) -> nssa::PublicTrans
         nssa::public_transaction::WitnessSet::from_raw_parts(vec![]),
     )
 }
+
+fn validate_doesnt_modify_account(
```
schouhy, line 174: nice

bedrock/deployment-settings.yaml
```diff
@@ -67,7 +67,9 @@ cryptarchia:
         - opcode: 17
           payload:
             channel_id: '0000000000000000000000000000000000000000000000000000000000000000'
-            inscription: '67656e65736973'
+            # chain_id_len=12 (u64_le), chain_id=logos-devnet (utf-8),
+            # genesis_time=2026-01-10T07:47:56Z (u64_le), epoch_nonce=[0u8; 32]
+            inscription: '0c000000000000006c6f676f732d6465766e65742c046269000000000000000000000000000000000000000000000000000000000000000000000000'
```
schouhy, line 72: how is this value derived?

sequencer/core/src/lib.rs
```diff
@@ -235,25 +285,41 @@ impl<BP: BlockPublisherTrait> SequencerCore<BP> {
...
+                    let NSSATransaction::Public(public_tx) = &tx else {
+                        panic!("Sequencer may only generate Public transactions, found {tx:#?}");
+                    };
+                    self.state
+                        .transition_from_public_transaction(
+                            public_tx,
+                            new_block_height,
+                            new_block_timestamp,
+                        )
+                        .context("Failed to execute sequencer-generated transaction")?;
```
schouhy, line 320: propagating the error with `?` would kill the sequencer main loop in case one of these transactions originated by the sequencer fails. That sounds scary

sequencer/core/src/lib.rs
```diff
@@ -235,25 +285,41 @@ impl<BP: BlockPublisherTrait> SequencerCore<BP> {
...
-                new_block_height,
-                new_block_timestamp,
-            ) {
-                Ok(diff) => diff,
-                Err(err) => {
-                    error!(
-                        "Transaction with hash {tx_hash} failed execution check with error: {err:#?}, skipping it",
-                    );
-                    continue;
+            match origin {
```
schouhy, line 292: this difference in handling transactions depending on their origin shouldn't also be present on the indexer block processing too? The `calculate_state_for_id` seems to treat every tx as if they were originated by users?
I think the indexer just silently errors without crashing and indexer and sequencer states diverge

test_fixtures/src/config.rs
```diff
@@ -143,7 +143,12 @@ pub fn genesis_from_accounts(
             balance: account.balance,
         });
 
-    public_genesis.chain(private_genesis).collect()
+    let supply_bridge_account = GenesisAction::SupplyBridgeAccount { balance: 1_000_000 };
```
schouhy, line 146: this supply bridge account should also be added to the sequencer testnet config, no?

common/src/transaction.rs
```diff
@@ -90,26 +90,12 @@ impl NSSATransaction {
...
-            .get(&faucet_id)
-            .is_some_and(|post| *post != state.get_account_by_id(faucet_id))
-        {
-            return Err(nssa::error::NssaError::InvalidInput(
-                "Transaction modifies system faucet account".into(),
-            ));
+        let system_accounts = nssa::CLOCK_PROGRAM_ACCOUNT_IDS.iter().copied().chain([
+            nssa::system_faucet_account_id(),
+            nssa::system_bridge_account_id(),
+        ]);
```
schouhy, line 96: it would be nice if there was a `sequencer::system_accounts()` method that returns these as defined by the sequencer. But I guess there's a dependency cycle issue if the common crate depends on the sequencer.

schouhy: maybe make `validate_on_state` take the system accounts as parameter, and have sequencer and indexer call it by passing system accounts defined by a sequencer method. That way the sequencer crate fully defines the system behavior

program_methods/guest/src/bin/faucet.rs
```diff
@@ -57,6 +62,29 @@ fn main() {
                 .with_pda_seeds(vec![faucet_core::compute_faucet_seed()]),
             ]
         }
+        Instruction::TransferDirect { amount } => {
```
schouhy: This is specific to fund the bridge account on genesis no? In that case I'd call it `Instruction::InitialBridgeAccountFund` (or something along those liens) and have it check the recipient is the bridge account

sequencer/core/src/lib.rs
schouhy: Also, in case of failure for whatever reason, either by killing the sequencer or skipping the tx in the block, the deposit on the L1 would be locked forever no? If that's the case, the sequencer should have a fallback function in case of error to revert the deposit on the L1 when something fails on LEZ. That way deposits are "atomic" from the L1 perspective.

mempool/src/lib.rs
```diff
@@ -48,6 +48,14 @@ pub struct MemPoolHandle<T> {
...
 }
 
+impl<T> Clone for MemPoolHandle<T> {
+    fn clone(&self) -> Self {
+        Self {
+            sender: self.sender.clone(),
+        }
+    }
+}
+
```
schouhy, line 58: isn't this what the derived impl would do?

schouhy (commented):
Hey thanks for the PR! I left some comments to chat about.

One question more related to the zone sdk than to this PR I think. Do the `ZoneSequencer` deposit events have a fallback mechanism to revert them in case something goes wrong on LEZ bridge execution (software or hardware failure)?

bedrock/deployment-settings.yaml
Arjentix: Honestly no idea, I was just getting errors from docker bedrok setup. I figured out the problem was inscription format in this line and then just took the line from bedrock repo [here](https://github.com/logos-blockchain/logos-blockchain/blob/5d29a6bd980847e06d739164409f3e6b4ec74500/nodes/node/binary/src/config/deployment/devnet/deployment.yaml). And it just worked. They should probably make this thing more clear...

sequencer/core/src/lib.rs
Arjentix: I don't think there is a way to deny deposit from L2. No way to inform L1 we can't do it. So I think we should just assume that Sequencer always make deposit or immediately fails (meaning there is a bug).

If sequencer was killed the next restart should check if all deposits where processed and if not, then try to process them. I think I need to implement this recovery, right, will do

mempool/src/lib.rs
Arjentix: No, because derived impl would also restrict: `T: Clone`

program_methods/guest/src/bin/faucet.rs
Arjentix: Well, in fact yes, but the functionality is pretty generic. It could be any account, maybe there will be more in future. Wouldn't mentioning bridge here break abstraction and incapsulation?

sequencer/core/src/lib.rs
Arjentix: Hmm, yes, you're right. But we decided that @Pravdyvy will work on Indexer side. But you're right to highlight that after this PR the state may diverge because of deposits.

Not sure currently how should Indexer decide if it's user tx or no... I think it can detect Deposit event from bedrock and then just assume the next Deposit tx is from Sequencer, as no one else is able to submit it.

test_fixtures/src/config.rs
Arjentix: Hmm, yes, but it should be synced with Bedrock setup. Although I think we can start with 1 million and adjust later. Will add to configs

sequencer/core/src/lib.rs
schouhy: can't the sequencer do a withdrawal back to the L1 in case of failing to process the deposit?

Arjentix: Sounds good, but we have no information about sender address on L1 to transfer funds back

sequencer/core/src/lib.rs
schouhy: ah right. Maybe the indexer can flag a tx that touches the bridge account and see if it is consistent with L1 events

sequencer/core/src/lib.rs
Arjentix: Also then it's technically hard for indexer to distinguish full-fledged withdraw from a malicious one (in future with untrusted decentralized sequencers)

program_methods/guest/src/bin/faucet.rs
schouhy: ah true. I'm fine either way. It's true that doing what I proposed would make the faucet program dependent on the bridge program. The advantage is that it reduces the surface area for exploits. A faucet program with a general transfer function is something we need to keep a close eye on.

sequencer/core/src/lib.rs
schouhy: > Sounds good, but we have no information about sender address on L1 to transfer funds back

oh really? are deposits anonymous?

common/src/transaction.rs
Arjentix: Hmm, I see you want to encapsulate system accounts and logic just inside sequencer. But in fact this accounts and ids are used in initial state as well, which is a part of `nssa` anyway...

Arjentix: And I don't think decoupling state constructor from these thing is good, because it's needed for both: sequencer and indexer

Arjentix: What do you think about https://github.com/logos-blockchain/logos-execution-zone/issues/432? Maybe we can store this info in `nssa_state`?

sequencer/core/src/lib.rs
schouhy: > Also then it's technically hard for indexer to distinguish full-fledged withdraw from a malicious one (in future with untrusted decentralized sequencers)

yes. It's not clear how the indexer validates sequencer actions. I can think of a few things like "check that each withdrawal can be explained by a withdrawal execution or a sibling deposit on the same block". But it feels more like a heuristic-based approach.Also without the L1 address where these deposits come from is hard to do anything

Arjentix: All we have is this: https://github.com/logos-blockchain/logos-blockchain/blob/master/zone-sdk/src/state.rs#L58

Maybe as a part of the solution they could include sender address in this struct

schouhy (approved):
LGTM thanks!

sequencer/core/src/lib.rs
Arjentix: Feels like we need not a L1 Withdraw as a refund mechanism, but a special instruction like `DepositRefund`. That way it will be simpler for Indexer to understand what's going on

program_methods/guest/src/bin/faucet.rs
Arjentix: I see. Well, currently this instruction doesn't need a special validation and existing one fully covers it. So let's wait for a third opinion here.

@Pravdyvy @moudyellaz

common/src/transaction.rs
schouhy: #432 proposes to separate the state from the state transition function no? I'm not sure that's something I'd do. I think they both define the state machine and neither makes sense without the other. So it makes sense to me to keep them together. And also keep them separate from any initial configuration or initial state specifically used by the sequencer. A general constructor to define the initial state for any system based on that state machine should be enough.
The sequencer on the other hand defines a system based on that state machine, adding some protected accounts managed only by it, and operates at a blockchain level as an L2 on top of Bedrock. That's why it doesn't seem right to me to have a method in the `common` crate to determine which accounts are managed by the sequencer. That's something I'd leave for the sequencer.

Also about the initial state. Maybe we can remove it entirely and express it as genesis actions? Integration tests relying on private accounts would be more annoying but there should be a workaround

sequencer/core/src/lib.rs
Pravdyvy: Ok, I may be missing something, but why include this transaction in block at all? If sequencer stumbles on it, then indexer would as well. 
Indexer should reflect sequencer state, not validate it, we would have validators for this purpose.

program_methods/guest/src/bin/faucet.rs
Pravdyvy: I think current checks is enough, but naming may be misleading, considering that this will fail everywhere exept genesis. Is suggest renaming it into `Instruction::GenesisTransferDirect` or similar for clarity. Also it would look more natural in tests if genesis instruction works only in genesis.

Pravdyvy (approved):
In general lgtm, interesting how threads will progress.

sequencer/core/src/lib.rs
Arjentix: If I understand you right and you're talking about refunding, then yes, we discussed in Discord that we shouldn't implement that.

For a general problem "How to make sure deposit event will be processed if sequencer process dies for any reason" I'll implement crash-safety

Arjentix: Let's move crach-tolerancy to a next PR. This PR is already big and got approvals while this new feature takes some lines as well

common/src/transaction.rs
Arjentix: Hmm, okay, I see.. I also think state transition should live in the same place as state. In fact I'd better make it state method only. And maybe introduce type-safety like `StateValidatedTransaction`. We've already discussed some flaws in our current naming and abstractions, so looks like we can address them in a another PR, because having this accounts defined in common was not introduced in this PR.

So my plan is:

- Merge this PR
- Implement crash-tolerant deposits
- Refactor validation, accounts and etc as we discussed

Does this make sense for you?

program_methods/guest/src/bin/faucet.rs
Arjentix: I see, okay, will do

PR #504, opened 2026-05-28
fix!: several protocol fixes

schouhy:
## 🎯 Purpose

Specs updating revealed several implementation errors. This PR fixes them.

## ⚙️ Approach

- Fix nonce init (`PrivateAuthorizedInit`): the initial nonce for `PrivateAuthorizedInit` accounts was incorrectly computed.
- Fix nullifier deduplication bug: sequencer checks nullifieres are not duplicated by looking at nullifiers only (not (nullifier, root) pairs).
- Add authorization bidirectional check in the circuit: programs must now set `is_authorized = true` for every authorized account in pre-states, not just avoid marking unauthorized ones as authorized.
- Mirror authorization propagation in the circuit with the approach of the public execution. The authorized accounts set grows as the union across the call chain instead of being reset at each hop.

## 🧪 How to Test

CI should pass

## 🔗 Dependencies

None

## 🔜 Future Work

None

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


Pravdyvy (approved):
lgtm

jonesmarvin8 (approved):
Lgtm. Artifacts need to be updated, but everything else looks good.

PR #506, opened 2026-05-29
feat(sequencer): make bridge deposits fault-tolerant

Arjentix:
## 🎯 Purpose

Previously in #493 we discovered that there are ways we can forget about deposit event happened on L1. This PR addressed this issue.

## ⚙️ Approach

- Update to a new Zone SDK, where checkpoints are more stable. With that we should never receive same deposit event twice which wasn't a guarantee before.
- Store pending deposits in rocksdb together with optional block id, which is filled when tx fulfilling deposit on LEZ is submitted to L1 but not yet finalized
- When tx is finalized, pending deposits are cleared based on block id
- When starting sequencer it checks for unfulfilled pending deposits and submits transactions for them
- Also store deposit op id in `bridge_core::Instruction::Deposit` for:
  - Sequencer needs
  - More transparent binding to L1 tx in our L2 block history

## 🧪 How to Test

Added test:

```bash
cargo test --package sequencer_core --all-features -- tests::start_from_config_replays_unfulfilled_deposit_events_from_db
```

## 🔗 Dependencies

None

## 🔜 Future Work

None

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


moudyellaz (commented):
Nice work on the crash tolerance. Left a few questions, mostly around the submitted/finalized lifecycle and the publish error path.

lez/sequencer/core/src/lib.rs
```diff
@@ -217,30 +253,47 @@ impl<BP: BlockPublisherTrait> SequencerCore<BP> {
...
 
         // TODO: Remove msg_id from store.update — it is no longer needed now that
         // zone-sdk manages L1 settlement state via its own checkpoint.
         let placeholder_msg_id = [0_u8; 32];
 
-        if let Err(err) = self.block_publisher.publish_block(&block).await {
-            error!("Failed to publish block to Bedrock with error: {err:#}");
-        }
+        self.block_publisher
+            .publish_block(&block)
```
moudyellaz, line 283: produce_new_block used to log and continue on a publish failure, now it propagates with `?` and aborts block production. For the fault-tolerance goal that is arguably more correct, you do not want to advance local state if the block never reached Bedrock. But it does mean a transient Bedrock publish error now stops block production entirely rather than retrying next tick. Is that the intended trade-off, and is there a retry/backoff around the caller so a blip does not wedge the sequencer?

lez/sequencer/core/src/lib.rs
```diff
@@ -217,30 +253,47 @@ impl<BP: BlockPublisherTrait> SequencerCore<BP> {
...
+            .await
+            .context("Failed to publish block to Bedrock")?;
+
         self.store.update(&block, placeholder_msg_id, &self.state)?;
 
+        let updated_deposits = self
+            .store
+            .mark_unfulfilled_deposit_events_submitted(&deposit_event_ids, block.header.block_id)?;
```
moudyellaz, line 291: One lifecycle edge case: submitted_in_block_id is set here right after publish, and removal only happens once that block id is finalized (remove keeps records where submitted_id > finalized, mod.rs:305). Replay only re-queues records where submitted_in_block_id.is_none(). So if a block is produced and marked submitted but that exact block never finalizes (reorg, or the node rebuilds from an earlier height), the record is stuck: it is not None so it never replays, and its submitted_id is below the finalized tip so it is never removed either. The deposit would then be silently stranded. Is that scenario possible in the Bedrock/zone-sdk model, and if so should mark-submitted be reverted to None when a produced block fails to finalize?

lez/sequencer/core/src/lib.rs
```diff
@@ -311,6 +364,20 @@ impl<BP: BlockPublisherTrait> SequencerCore<BP> {
                     let NSSATransaction::Public(public_tx) = &tx else {
                         panic!("Sequencer may only generate Public transactions, found {tx:#?}");
                     };
+
+                    if public_tx.message.program_id == Program::bridge().id() {
+                        let instruction: bridge_core::Instruction =
+                            risc0_zkvm::serde::from_slice(&public_tx.message.instruction_data)
```
moudyellaz, line 384: This pulls the risc0-zkvm dep into sequencer_core just to re-deserialize every bridge tx's instruction in the block-building loop to recover l1_deposit_op_id. It works, but it means re-parsing instruction bytes on the host hot path. Since the sequencer is the one that built these deposit txs in the first place, could the op id be carried alongside the tx (or in block metadata) so you avoid the per-tx serde here and the extra dependency?

lez/sequencer/core/src/lib.rs
Arjentix: Yes, I thought about this and decided that we should fail here. So looking into how Zone SDK works, it should never actually fail. If e.g. network blinked then it shouldn't fail here, it should retry tx next time.
The only way it can fail here: if something is wrong with Zone SDK Sequencer Actor. In that case I think we can do nothing else than fail our Sequencer as well.

lez/sequencer/core/src/lib.rs
Arjentix: So we expect that if a block was submitted to Zone SDK it will always be finalized. There should be no stale deposit events. Ideally though I'd later have some metrics published in grafana dashboard to keep an eye on it

lez/sequencer/core/src/lib.rs
Arjentix: I would argue that this is not a hot path. Real hot path is origin check. But after we're sure that it's `TransactionOrigin::Sequencer` and program is bridge, decoding shouldn't be so much of a pain.

I think implementing another way of storing deposits included in the block is more complicated, because we never know how much transactions would fit into the block size beforehand.

moudyellaz (approved):
lgtm, thanks

Pravdyvy (approved):
lgtm

PR #508, opened 2026-06-01
feat(sequencer): implement bridge withdraw flow

Arjentix:
## 🎯 Purpose

This PR implements Withdrawal Flow of the Bridge for Sequencer.

## ⚙️ Approach

- Added `Withdraw` instruction for `bridge` program. It can be executed by anyone, but only publicly (we could relax this and allow to just lose funds with private tx)
- Sequencer creates withdraw operations and atomically submits them together with built block
- Added `wallet bridge withdraw` command
- Added bedrock events validation on sequencer side. This just gives us logging and nothing more
- Added tests

## 🧪 How to Test

```bash
RUST_LOG=info RISC0_DEV_MODE=1 cargo test --package integration_tests --test bridge --all-features -- bedrock_deposit_claim_and_withdraw_round_trip_succeeds --exact --nocapture --include-ignored
```

## 🔗 Dependencies

None

## 🔜 Future Work

Add bedrock withdraw events validation on Indexer side

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


lee/state_machine/src/state.rs
```diff
@@ -4725,4 +4725,90 @@ pub mod tests {
 
         assert_eq!(state.get_account_by_id(recipient_id).balance, amount);
     }
+
+    #[test]
+    fn private_bridge_withdraw_invocation_is_dropped() {
```
schouhy: this test passes only because the transaction message is constructed with a nonce but there's no signing key associated with it in the witness set. This change makes the test fail

```diff
             .into(),
         );

-        let bridge_nonce = bridge_pre.account.nonce;
+        let _bridge_nonce = bridge_pre.account.nonce;

         let (output, proof) = execute_and_prove(
             vec![sender_pre, bridge_pre],
...
         let message = Message::try_from_circuit_output(
             vec![bridge_account_id],
-            vec![bridge_nonce],
+            vec![],
             vec![(
                 sender_keys.npk(),
                 sender_keys.vpk(),
```

I think I'd allow private executions of the bridge

Arjentix: Good catch! Indeed it was impossible to test this behavior in this crate, so moved to sequencer crate, fixed that problem and now test works normally.

> I think I'd allow private executions of the bridge

Why? I think from user standpoint it may look valid to try to execute this and then they will be surprised it didn't work as expected.

schouhy: ah I think you still need to remove the one in state.rs no?

lez/common/src/transaction.rs
```diff
@@ -115,6 +117,50 @@ impl LeeTransaction {
...
+            balance: post_balance,
+            program_owner: post_program_owner,
+            data: post_data,
+            nonce: post_nonce,
+        } = post;
+
+        let only_balance_increased = post_balance >= pre_balance
+            && post_program_owner == pre_program_owner
+            && post_data == pre_data
+            && post_nonce == pre_nonce;
```
schouhy: If we ever add a field to the account struct and don't update how this variable is computed it could silently introduce a bug. Maybe something along the lines of the following is more robust
```rust
only_balance_increased = { 
  let mut expected_pre = post.clone();
  expected_pre.balance = pre.balance; 
  (expected_pre == pre) && (pre.balance <= post.balance) 
};
```

lez/common/src/transaction.rs
```diff
@@ -90,14 +90,16 @@ impl LeeTransaction {
             }
         }?;
 
-        let system_accounts = lee::CLOCK_PROGRAM_ACCOUNT_IDS.iter().copied().chain([
-            lee::system_faucet_account_id(),
-            lee::system_bridge_account_id(),
-        ]);
+        let system_accounts = lee::CLOCK_PROGRAM_ACCOUNT_IDS
```
schouhy: the name of this variable is a bit misleading now no? the bridge account is a system account not in this list.

lez/sequencer/core/src/block_publisher.rs
```diff
@@ -127,16 +142,33 @@ impl BlockPublisherTrait for ZoneSdkPublisher {
         })
     }
 
-    async fn publish_block(&self, block: &Block) -> Result<()> {
+    async fn publish_block(&self, block: &Block, withdraws: Vec<WithdrawArg>) -> Result<()> {
```
schouhy:
```suggestion
    async fn publish_block(&self, block: &Block, withdrawals: Vec<WithdrawArg>) -> Result<()> {
```

lez/sequencer/core/src/lib.rs
```diff
@@ -499,6 +547,13 @@ impl<BP: BlockPublisherTrait> SequencerCore<BP> {
 struct BlockWithMeta {
     block: Block,
     deposit_event_ids: Vec<HashType>,
+    withdraws: Vec<WithdrawArg>,
+}
+
+#[derive(Debug, Clone, Copy)]
+struct WithdrawReconciliationKey {
```
schouhy:
```suggestion
struct WithdrawalReconciliationKey {
```

schouhy (commented):
Hey thanks for the PR! I'm trying to understand how fault tolerant is this implementation and what would happen under a crash with the unseen withdrawal counts and so on. Could you expand a bit more on that?

Leaving some minor comments

lez/sequencer/core/src/lib.rs
```diff
@@ -228,59 +270,67 @@ impl<BP: BlockPublisherTrait> SequencerCore<BP> {
...
         let BlockWithMeta {
             block,
             deposit_event_ids,
-        } = block_with_meta;
+            withdraws,
+        } = self
+            .build_block_from_mempool()
+            .context("Failed to build block from mempool transactions")?;
+
+        for withdraw in &withdraws {
```
moudyellaz: record_unseen_withdraw bumps the counter before publish_block (line 333), and the two are not atomic. If we crash or publish fails in between, that +1 is never consumed and the unseen-withdraw count drifts permanently (or double-counts if the tx is re-mined). The deposit side avoids this with the PendingDepositEventRecord lifecycle. Could we record after a successful publish, or lean on the atomicity publish_atomic_withdraw already provides?

lez/sequencer/core/src/lib.rs
```diff
@@ -665,6 +720,97 @@ fn build_bridge_deposit_tx_from_event(event: &PendingDepositEventRecord) -> Resu
...
+            .ok()?;
+
+    match instruction {
+        bridge_core::Instruction::Withdraw {
+            amount,
+            bedrock_account_pk,
+        } => {
+            let recipient_pk =
+                logos_blockchain_key_management_system_service::keys::ZkPublicKey::from(
+                    BigUint::from_bytes_le(&bedrock_account_pk),
```
moudyellaz, line 741: bedrock_account_pk is interpreted little-endian here (from_bytes_le), but the wallet CLI hex-decodes it big-endian and the deposit path uses the same hex as an opaque bedrock wallet id. If bedrock's ZkPublicKey hex form is big-endian, this reverses it and the withdraw pays a different L1 key than the depositor. Can you confirm the byte order round-trips end to end? The round-trip test does not currently catch this.

integration_tests/tests/bridge.rs
```diff
@@ -446,9 +464,91 @@ async fn bedrock_deposit_mints_to_vault_then_claim_succeeds() -> anyhow::Result<
...
-        recipient_balance_before + 1,
+        recipient_balance_before + u128::from(amount),
         "Recipient balance should increase by claimed amount"
     );
 
+    // Withdraw back to Bedrock and wait for finalized withdraw event.
+    let sender_id = recipient_id;
+
+    let observer = create_zone_indexer_observer(ctx.bedrock_addr())?;
+    let observe_fut = wait_for_finalized_withdraw_op(&observer, amount);
```
moudyellaz: This only matches on amount, so the test passes even if the withdraw paid the wrong L1 recipient. To truly close the loop, and to settle the byte-order question above, assert the bedrock balance of bedrock_account_pk returns to its pre-deposit value after the withdraw finalizes.

programs/bridge/core/src/lib.rs
```diff
@@ -17,7 +17,20 @@ pub enum Instruction {
         l1_deposit_op_id: [u8; 32],
         vault_program_id: ProgramId,
         recipient_id: AccountId,
-        amount: u128,
+        amount: u64,
```
moudyellaz, line 20: Deposit amount goes from u128 to u64 here (the guest now casts with u128::from). That is a borsh wire change to the instruction from #506. Assuming it is intentional to match bedrock's u64 amount and line up with Withdraw, just want to make sure L1 amounts cannot exceed u64.

moudyellaz (dismissed):
lgtm, thanks. I left some comments and questions to address before merging.

lee/state_machine/src/state.rs
Arjentix: Yeap, missed that, removed

lez/common/src/transaction.rs
Arjentix: Hm, isn't destructuring pattern enough? If a new field is added to Account compilation will fail here and dev should manually type this type in destructuring pattern. And then to introduce a bug they should manually ignore this field this should be pretty strange and we can catch it on review.

Your code works but kind of hard to understand, don't you find?

lez/common/src/transaction.rs
Arjentix: Agree, renamed to `restricted_modification_accounts`

lez/sequencer/core/src/block_publisher.rs
Arjentix: Done

lez/sequencer/core/src/lib.rs
Arjentix: Indeed. Fixed by moving withdrawals increment into db update. They're now committed in the same write batch, so state saving and withdrawals increments are now atomic

lez/sequencer/core/src/lib.rs
Arjentix: This code removed in favor of borsh (de-)serialization in atomic update. This is the serialization style used for every other db cell

integration_tests/tests/bridge.rs
Arjentix: Fixed

programs/bridge/core/src/lib.rs
Arjentix: Yes, it's fixed by bedrock everythere in their api as well. If they someday change that our code will break and we'll be able to rethink it

Arjentix:
> I'm trying to understand how fault tolerant is this implementation and what would happen under a crash with the unseen withdrawal counts and so on. Could you expand a bit more on that?

So Moudy suggested to enhance one place if we crash between incrementing withdrawals count and publishing block, I fixed this place and now we atomically update db for both block and withdrawals (just like zone sdk, lol).

So currently we have this:

1. User submits withdraw tx
2. It's placed in mempool
3. Block is executed and submitted to the bedrock
4. Db state (both block apply and withdrawals count) updated, incrementing unseen withdraw count for exact key (amount, pub key).
5. When sequencer gets a withdraw event from bedrock it decrements the count.

Overall this counting doesn't do anything special, it's just extra validation and logging, nothing more.

So withdrawals are as fault-tolerant as any other transaction. If we crash before publishing it just never be submitted to bedrock and user will have to resubmit it after restart.

moudyellaz (approved):
lgtm, thanks. and thanks for addressing my comments!

lez/common/src/transaction.rs
schouhy: sure I'm not saying it will inevitable be missed. And we are most probably not adding new fields to accounts. But I do think it could be easily missed in a big PR introducing this change among others, specially if eventually a refactor removes the destructuring pattern for some reason, which is the main protection, or the code is moved around over time. I'd personally go with the option that doesn't need maintenance. Trusting future devs to not miss something is generally not a great idea.

schouhy (approved):
lgtm thanks

lez/common/src/transaction.rs
Arjentix: Ok, I see the point, don't have a strong opinion here, so implemented your approach

schouhy: thanks!

PR #509, opened 2026-06-03
feat(wallet-ffi): expose stable identity lookup primitives

RomThpt:
## Summary

Adds two `wallet-ffi` entries (and matching `WalletCore` accessors) that let a client resolve a stable cryptographic identity from a persisted wallet without rotating it at every restart.

- `wallet_ffi_get_first_private_accounts_key(handle, out_keys)` — returns the first private accounts key in the wallet (BTreeMap-ordered, deterministic across reopens for the same persisted storage).
- `wallet_ffi_get_private_accounts_key_by_chain_index(handle, chain_index_str, out_keys)` — returns the private accounts key at an explicit chain index using the wallet-CLI string format (`/`, `/0`, `/0/1`, …).

Purely additive; no behaviour change for existing callers.

## Why

Currently, `wallet-ffi` exposes:

- `wallet_ffi_create_private_accounts_key` — derives a **new** chain at the next free layered index each time it is called, rotating the NPK at every call.
- `wallet_ffi_get_private_account_keys(account_id)` — returns a **sub-account** key chain (the per-account NPK derived under a parent), not the parent chain master.

Neither lets a downstream application restore the **same agent-level NPK** after a wallet reopen. The workaround today is to cache the NPK in a sidecar file at the application layer, which is fragile and couples the application to file-layout assumptions that live outside the wallet.

The new accessors close that gap: by anchoring identity on a known chain index (`"/"` for the seed-root node) the application gets a deterministic identity for the lifetime of the wallet seed.

## Motivation

I am building an [LP-0008 (autonomous AI module)](https://github.com/logos-co/lambda-prize/blob/main/prizes/LP-0008.md) submission. Agent identity is the agent's NPK; without this patch, the agent's identity rotates every time Basecamp restarts the host process, breaking AgentCards announced to peers, owner-channel routing, and demo reproducibility.

## Implementation

- `wallet/src/lib.rs`: two new accessors on `WalletCore`. They thin-wrap existing `UserKeyChain::private_account_key_chains` (BTreeMap iter, deterministic) and `UserKeyChain::private_account_key_chain_by_index`.
- `wallet-ffi/src/keys.rs`: matching FFI entry points. They use the same `FfiPrivateAccountKeys` out-shape and lifetime management as the existing `wallet_ffi_get_private_account_keys`.
- `wallet-ffi/src/lib.rs`: bump `c_str_to_string` visibility to `pub(crate)` so the new entry can reuse it.
- `wallet-ffi/wallet_ffi.h`: regenerated by cbindgen as part of the wallet-ffi build.

## Test plan

- [x] `cargo build --release -p wallet-ffi --no-default-features` (used downstream in the agent module).
- [x] Symbol present in the built dylib: `nm libwallet_ffi.dylib | grep wallet_ffi_get_private_accounts_key_by_chain_index`.
- [x] Manual round-trip from the agent module: persisted wallet reopened across module restarts returns the identical NPK via the `"/"` lookup. Verified with NPK `40b5854e7b498cdf91461ed66da64df23e1a47389f752f4f392a5d19c3218371`.
- [ ] Unit tests for the new accessors (happy to add if reviewers want them; existing `wallet-ffi` crate does not have an integration test bench I could plug into directly).

## Notes

The header file changes are mechanical (cbindgen output). The visibility bump on `c_str_to_string` is the minimal change needed to reuse the existing helper; happy to inline a private copy instead if the reviewer prefers.

### DIFF
--- a/wallet-ffi/Cargo.toml
+++ b/wallet-ffi/Cargo.toml
@@ -19,6 +19,8 @@ sequencer_service_rpc = { workspace = true, features = ["client"] }
 tokio.workspace = true
 key_protocol.workspace = true
 serde_json.workspace = true
+k256.workspace = true
+sha2 = "0.10"
 
 [build-dependencies]
 cbindgen = "0.29"
--- a/wallet-ffi/src/keys.rs
+++ b/wallet-ffi/src/keys.rs
@@ -145,6 +145,172 @@ pub unsafe extern "C" fn wallet_ffi_get_private_account_keys(
     WalletFfiError::Success
 }
 
+/// Return the keys of the first private accounts key chain in the wallet.
+///
+/// The first chain is determined by BTreeMap ordering over chain indices,
+/// which is deterministic — calling this function on a wallet persisted to
+/// disk and reopened later returns the same NPK as long as no preceding
+/// chain index was inserted in between.
+///
+/// Intended for clients (e.g. agent runtimes) that need a stable
+/// cryptographic identity derived from the wallet seed without rotating
+/// it at every call to `wallet_ffi_create_private_accounts_key`.
+///
+/// # Parameters
+/// - `handle`: Valid wallet handle
+/// - `out_keys`: Output pointer for the key material
+///
+/// # Returns
+/// - `Success` on success
+/// - `AccountNotFound` if the wallet has no private accounts key yet
+/// - Error code on other failures
+///
+/// # Memory
+/// The keys must be freed with `wallet_ffi_free_private_account_keys()`.
+///
+/// # Safety
+/// - `handle` must be a valid wallet handle from `wallet_ffi_create_new` or `wallet_ffi_open`
+/// - `out_keys` must be a valid pointer to a `FfiPrivateAccountKeys` struct
+#[no_mangle]
+pub unsafe extern "C" fn wallet_ffi_get_first_private_accounts_key(
+    handle: *mut WalletHandle,
+    out_keys: *mut FfiPrivateAccountKeys,
+) -> WalletFfiError {
+    let wrapper = match get_wallet(handle) {
+        Ok(w) => w,
+        Err(e) => return e,
+    };
+
+    if out_keys.is_null() {
+        print_error("Null pointer argument");
+        return WalletFfiError::NullPointer;
+    }
+
+    let wallet = match wrapper.core.lock() {
+        Ok(w) => w,
+        Err(e) => {
+            print_error(format!("Failed to lock wallet: {e}"));
+            return WalletFfiError::InternalError;
+        }
+    };
+
+    let Some(key_chain) = wallet.first_private_accounts_key_chain() else {
+        print_error("Wallet has no private accounts key");
+        return WalletFfiError::AccountNotFound;
+    };
+
+    let npk_bytes = key_chain.nullifier_public_key.0;
+
+    let vpk_bytes = key_chain.viewing_public_key.to_bytes();
+    let vpk_len = vpk_bytes.len();
+    let vpk_vec = vpk_bytes.to_vec();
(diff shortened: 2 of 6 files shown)


wallet/src/lib.rs
```diff
@@ -248,6 +248,34 @@ impl WalletCore {
...
+    /// look the same key back up across restarts deterministically.
+    #[must_use]
+    pub fn private_accounts_key_chain_by_index(
+        &self,
+        chain_index: &ChainIndex,
+    ) -> Option<&KeyChain> {
+        self.storage
+            .key_chain()
+            .private_account_key_chain_by_index(chain_index)
+    }
```
Arjentix, line 277: IMO both of this could be achieved by accessing `wallet.storage().key_chain()` and no additional wrappers needed.

Currently we indeed have some wrappers in this wallet, but I think this is not a good pattern and it's better not to expand this. Maybe we'll work on reducing it as well

wallet-ffi/Cargo.toml
```diff
@@ -12,13 +12,15 @@ crate-type = ["rlib", "cdylib", "staticlib"]
...
-lee_core.workspace = true
+nssa.workspace = true
+nssa_core.workspace = true
 sequencer_service_rpc = { workspace = true, features = ["client"] }
 
 tokio.workspace = true
 key_protocol.workspace = true
 serde_json.workspace = true
+k256.workspace = true
+sha2 = "0.10"
```
Arjentix, line 23: We have this in workspace, better use it

wallet-ffi/src/keys.rs
```diff
@@ -0,0 +1,536 @@
...
+        clippy::as_conversions,
+        reason = "We need to convert the boxed slice into a raw pointer for FFI"
+    )]
+    let vpk_ptr = Box::into_raw(vpk_boxed) as *const u8;
+
+    unsafe {
+        (*out_keys).nullifier_public_key.data = npk_bytes;
+        (*out_keys).viewing_public_key = vpk_ptr;
+        (*out_keys).viewing_public_key_len = vpk_len;
+    }
```
Arjentix, line 218: I can see this intialization became a repeated pattern in existing codebase (this is 3rd time I think). Let's better encapsulate this to a separate function

wallet-ffi/src/keys.rs
```diff
@@ -0,0 +1,536 @@
...
+
+    let Ok(chain_index_str) = crate::c_str_to_string(chain_index_str, "chain_index_str") else {
+        return WalletFfiError::InvalidUtf8;
+    };
+
+    let Ok(chain_index) = ChainIndex::from_str(&chain_index_str) else {
+        print_error(format!("Failed to parse chain index: {chain_index_str}"));
+        return WalletFfiError::InvalidKeyValue;
+    };
```
Arjentix, line 278: Looks like it's not a key problem

wallet-ffi/src/keys.rs
```diff
@@ -0,0 +1,536 @@
...
+    use k256::elliptic_curve::rand_core::{OsRng, RngCore as _};
+    OsRng.fill_bytes(&mut aux_random);
+
+    let signature = match signing_key.sign_prehash_with_aux_rand(&prehash, &aux_random) {
+        Ok(s) => s,
+        Err(e) => {
+            print_error(format!("Schnorr signing failed: {e}"));
+            return WalletFfiError::InternalError;
+        }
+    };
```
Arjentix, line 419: I don't understand why you need to manually perform cryptographic operations. I think you should use existing functionality of our keys. Or expand this functionality and use it here.

moudyellaz:
Closing this, @RomThpt. This landed via the key protocol rather than a wrapper layer: dev has `private_account_key_chain_by_index`, and the FFI surface has changed substantially since.

PR #510, opened 2026-06-04
Pravdyvy/programs elfs deployments

Pravdyvy:
## 🎯 Purpose

Adding program deployments and exposes elfs for testing.

## ⚙️ Approach

- [x] Added program deployment as ffi function.
- [x] Added couple of elfs as ffi functions.
- [x] New zones API adaptation.

## 🧪 How to Test

Integration tests will do.

## 🔗 Dependencies

None

## 🔜 Future Work

- After merging dependant C++ PRs needs to be updated. 
- Renaming.
- Remove logos-blockchain circuits dependency from benchmarks.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [ ] Add/update tests


moudyellaz (approved):
lgtm, thanks

Cargo.toml
```diff
@@ -138,13 +138,13 @@ tokio-retry = "0.3.0"
...
+logos-blockchain-common-http-client = { git = "https://github.com/logos-blockchain/logos-blockchain.git", rev = "d8711bbc3d43d3ef9755ef9b73af32fd0f703160" }
+logos-blockchain-key-management-system-service = { git = "https://github.com/logos-blockchain/logos-blockchain.git", rev = "d8711bbc3d43d3ef9755ef9b73af32fd0f703160" }
+logos-blockchain-core = { git = "https://github.com/logos-blockchain/logos-blockchain.git", rev = "d8711bbc3d43d3ef9755ef9b73af32fd0f703160" }
+logos-blockchain-chain-broadcast-service = { git = "https://github.com/logos-blockchain/logos-blockchain.git", rev = "d8711bbc3d43d3ef9755ef9b73af32fd0f703160" }
+logos-blockchain-chain-service = { git = "https://github.com/logos-blockchain/logos-blockchain.git", rev = "d8711bbc3d43d3ef9755ef9b73af32fd0f703160" }
+logos-blockchain-zone-sdk = { git = "https://github.com/logos-blockchain/logos-blockchain.git", rev = "d8711bbc3d43d3ef9755ef9b73af32fd0f703160" }
+logos-blockchain-http-api-common = { git = "https://github.com/logos-blockchain/logos-blockchain.git", rev = "d8711bbc3d43d3ef9755ef9b73af32fd0f703160" }
```
Arjentix, line 148: I've read the conversations you had with Gusto and now I understand why you needed to bump this.

So this version should no longer require circuits. I think then we also need to remove circuits installation from CI. And maybe even check the whole repo for docs and etc referencing logos circuits.

flake.nix
```diff
@@ -13,8 +13,14 @@
 
     crane.url = "github:ipetkov/crane";
 
+    # Must stay in sync with the lbc-* tags in logos-blockchain/Cargo.toml.
     logos-blockchain-circuits = {
-      url = "github:logos-blockchain/logos-blockchain-circuits";
+      url = "github:logos-blockchain/logos-blockchain-circuits/2846ee7a4cfa24458bb8063412ab2e753b344d2f";
+    };
+
+    # Must stay in sync with the rust-rapidsnark rev in Cargo.toml.
```
Arjentix:
```suggestion
    # Must stay in sync with the rust-rapidsnark rev in Cargo.lock.
```

lez/sequencer/core/src/block_publisher.rs
```diff
@@ -47,9 +55,16 @@ pub trait BlockPublisherTrait: Clone {
 }
 
 /// Real block publisher backed by zone-sdk's `ZoneSequencer`.
+///
+/// The sequencer is owned exclusively by the drive task — the new zone-sdk
+/// `SequencerHandle` is a `&mut self` borrow, so any out-of-task access
+/// requires routing requests through a channel. We send serialized
+/// inscriptions over a bounded mpsc; the drive task `tokio::select!`s
+/// between `next_event()` and the inbox, calling `sequencer.handle().publish(...)`
+/// inline.
```
Arjentix: This looks like a Claude message about differences between previous and current implementation details, let's remove this, we have git for that

lez/sequencer/core/src/block_publisher.rs
```diff
@@ -80,49 +95,78 @@ impl BlockPublisherTrait for ZoneSdkPublisher {
...
+                        // Drain external publish requests by calling the
+                        // borrowing handle — `&mut sequencer` is only
+                        // available here.
+                        Some(data) = publish_rx.recv() => {
+                            if let Err(e) = sequencer.handle().publish(data) {
+                                warn!("zone-sdk publish failed: {e:?}");
+                            }
+                        }
+                        event = sequencer.next_event() => {
+                            let Some(event) = event else { continue };
```
Arjentix:
```suggestion
                        Some(event) = sequencer.next_event() => {
```

Arjentix (commented):
Good to finally have SDK updated

flake.nix
Pravdyvy: Fixed

lez/sequencer/core/src/block_publisher.rs
Pravdyvy: Removed

lez/sequencer/core/src/block_publisher.rs
Pravdyvy: Updated

Pravdyvy: Un-updated, I think pattern matching in tokio::select! works differently, this should be the same, but leads to deadlock.

Cargo.toml
Pravdyvy: Removed all found entries of logos-blockchain circuits mentions.

lez/sequencer/core/src/block_publisher.rs
Arjentix: Wow, interesting

.github/actions/install-logos-blockchain-circuits/action.yaml
Arjentix, line 1: I think re can remove this action at all now

Arjentix (approved):
LGTM, thanks

.github/actions/install-logos-blockchain-circuits/action.yaml
Pravdyvy: Let's do it little later, I will open an issue. Right now it breaks benchmarks in CI. I suppose it tries to run them on main in the same job and fails, because circuits are not set.

PR #515, opened 2026-06-08
fix: reject malformed privacy-preserving proof instead of panicking

moudyellaz:
## 🎯 Purpose

Fixes an authenticated denial-of-service in privacy-preserving transaction validation. A `PrivacyPreservingTransaction` carrying malformed proof bytes panics during validation instead of being rejected, which unwinds the sequencer's block-production loop and halts the chain.

## ⚙️ Approach

`Proof::is_valid_for` deserializes the proof with `borsh::from_slice(&self.0).unwrap()`. Because `Proof` is a plain `Vec<u8>` at the borsh level, `from_bytes` accepts any proof content, so a validly-signed message with garbage proof bytes passes every upstream check (commitments, nonces, signatures, validity window) and reaches this unwrap, which panics. The sequencer's `validate_on_state` call site only handles `Err` (log + skip), so the panic escapes and kills the `main_loop` task.

- [x] Fix `Proof::is_valid_for` to return `false` on a malformed receipt instead of unwrapping
- [x] Add regression test asserting a garbage proof is rejected with `InvalidPrivacyPreservingProof`, not a panic

## 🧪 How to Test

```
RISC0_DEV_MODE=1 cargo test -p lee privacy_garbage_proof_is_rejected_not_panic
```

Passes with the fix. Reverting the one-liner makes it panic at `circuit.rs:34` (the unwrap), confirming the test guards the real path.

## 🔗 Dependencies

None.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests

### DIFF
--- a/lee/state_machine/src/privacy_preserving_transaction/circuit.rs
+++ b/lee/state_machine/src/privacy_preserving_transaction/circuit.rs
@@ -31,10 +31,6 @@ impl Proof {
     }
 
     pub(crate) fn is_valid_for(&self, circuit_output: &PrivacyPreservingCircuitOutput) -> bool {
-        // A malformed proof must be rejected as invalid, not panic: these bytes are
-        // attacker-controlled (`Proof` is a plain `Vec<u8>` at the borsh level, so
-        // `from_bytes` accepts any content), and a panic here would unwind the
-        // sequencer's block-production loop.
         let Ok(inner) = borsh::from_slice::<InnerReceipt>(&self.0) else {
             return false;
         };
--- a/lee/state_machine/src/privacy_preserving_transaction/circuit.rs
+++ b/lee/state_machine/src/privacy_preserving_transaction/circuit.rs
@@ -31,7 +31,13 @@ impl Proof {
     }
 
     pub(crate) fn is_valid_for(&self, circuit_output: &PrivacyPreservingCircuitOutput) -> bool {
-        let inner: InnerReceipt = borsh::from_slice(&self.0).unwrap();
+        // A malformed proof must be rejected as invalid, not panic: these bytes are
+        // attacker-controlled (`Proof` is a plain `Vec<u8>` at the borsh level, so
+        // `from_bytes` accepts any content), and a panic here would unwind the
+        // sequencer's block-production loop.
+        let Ok(inner) = borsh::from_slice::<InnerReceipt>(&self.0) else {
+            return false;
+        };
         let receipt = Receipt::new(inner, circuit_output.to_bytes());
         receipt.verify(PRIVACY_PRESERVING_CIRCUIT_ID).is_ok()
     }
--- a/lee/state_machine/src/validated_state_diff.rs
+++ b/lee/state_machine/src/validated_state_diff.rs
@@ -930,4 +930,61 @@ mod tests {
             "recipient should receive nothing"
         );
     }
+
+    /// Regression test: a `PrivacyPreservingTransaction` carrying a structurally invalid
+    /// proof must be rejected with a clean `Err`, never a panic.
+    ///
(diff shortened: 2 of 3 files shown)


lee/state_machine/src/privacy_preserving_transaction/circuit.rs
```diff
@@ -31,7 +31,13 @@ impl Proof {
     }
 
     pub(crate) fn is_valid_for(&self, circuit_output: &PrivacyPreservingCircuitOutput) -> bool {
-        let inner: InnerReceipt = borsh::from_slice(&self.0).unwrap();
+        // A malformed proof must be rejected as invalid, not panic: these bytes are
+        // attacker-controlled (`Proof` is a plain `Vec<u8>` at the borsh level, so
+        // `from_bytes` accepts any content), and a panic here would unwind the
+        // sequencer's block-production loop.
```
Arjentix: Let's remove this notice about historical implementation

lee/state_machine/src/validated_state_diff.rs
```diff
@@ -930,4 +930,61 @@ mod tests {
...
+    ///
+    /// `Proof` is a plain `Vec<u8>` at the borsh level, so `from_bytes` accepts arbitrary
+    /// proof content. A validly-signed message with garbage proof bytes passes every
+    /// upstream check (commitments non-empty, no duplicates, nonces, signatures, validity
+    /// window) and reaches `Proof::is_valid_for`, which used to `unwrap()` the borsh
+    /// deserialization and panic — unwinding the sequencer's block-production loop. The fix
+    /// maps the deserialization failure to `false`, surfacing as
+    /// `InvalidPrivacyPreservingProof`. Before the fix this test panicked instead of failing.
+    #[test]
+    fn privacy_garbage_proof_is_rejected_not_panic() {
```
Arjentix: Same, let's just:

```suggestion
    /// Regression test: a `PrivacyPreservingTransaction` carrying a structurally invalid
    /// proof must be rejected with a clean `Err`.
    #[test]
    fn privacy_garbage_proof_is_rejected() {
```

lee/state_machine/src/validated_state_diff.rs
```diff
@@ -930,4 +930,61 @@ mod tests {
...
+            },
+        };
+
+        let state = V03State::new_with_genesis_accounts(&[], vec![], 0);
+
+        // Minimal message that passes every check up to proof verification: a single
+        // commitment satisfies the non-empty requirement, no signers makes the
+        // nonce/signature checks vacuously true, and unbounded validity windows are valid
+        // for any block/timestamp.
+        let commitment = Commitment::new(&AccountId::new([1_u8; 32]), &Account::default());
```
Arjentix: Let's properly form account it from key. I think this `new()` constructor should be forbidden someday

Arjentix (approved):
Good catch, thanks!

Actually would be good someday to forbid `unwraps()` in the code (except tests) via clippy and go through them replacing with `expect()` or proper error handling.

lee/state_machine/src/privacy_preserving_transaction/circuit.rs
moudyellaz: removed

lee/state_machine/src/validated_state_diff.rs
moudyellaz: applied

lee/state_machine/src/validated_state_diff.rs
moudyellaz: done, now derived from a key: `AccountId::from(&PublicKey::new_from_private_key(&PrivateKey::try_new([1; 32]).unwrap()))`

Pravdyvy (approved):
lgtm

PR #520, opened 2026-06-10
fix(indexer): bypass system account guards

erhant:
## 🎯 Purpose

The new deposit transactions are sequencer-originated, NOT user-originated; and they upgrade the system accounts. Such transactions are prevented by the system account guard of the `indexer` within state validations; but not on the sequencer side. This causes a state-divergence after a deposit happens.

While a proper fix & handling is required, this PR creates a hot-fix for the sake of allowing testnet v0.2 to work with the indexer & sequencer.

## ⚙️ Approach

- [x] Change two `execute_check_on_state` calls to the new `execute_without_system_accounts_check_on_state` within `indexer`
- [x] Add `execute_without_system_accounts_check_on_state`, akin to `execute_check_on_state` but does not check system account differences
- [x] Add `compute_state_diff` (shared between `execute_*_on_state` functions to compute the state difference)
- [x] Add indexer vs sequencer system account state comparison at the end of `bedrock_deposit_mints_to_vault_then_claim_succeeds` integration test

## 🧪 How to Test

The command below runs the happy-path end-to-end bridge test: 

```sh
RUST_LOG=info RISC0_DEV_MODE=1 cargo test -p integration_tests --test bridge -- bedrock_deposit_mints_to_vault_then_claim_succeeds --exact --nocapture --include-ignored
```

The test itself as it is written in this branch should fail in target `main` branch, because indexer will not allow the deposit transactions. Seeing that the test fails on `main` but passes here demonstrates that indexer hot-fix is correct. The failure looks like a bunch of error logs followed by the final assertion panic:

```sh
assertion `left == right` failed: Indexer and sequencer diverged for account 891294mzQ2pn5JrKtFc3NuDPUYQgbjqB4333fAcrCx2x after deposit
  left: Account { program_owner: ProgramId([2172491596, 648287714, 3443481193, 1757114575, 2729968860, 2157376487, 835305942, 2223466012]), balance: 10000, data: Data([]), nonce: 1 }
 right: Account { program_owner: ProgramId([2172491596, 648287714, 3443481193, 1757114575, 2729968860, 2157376487, 835305942, 2223466012]), balance: 10001, data: Data([]), nonce: 2 }
```

There are no error logs in this branch's version, nor a failure.

## 🔗 Dependencies

None

## 🔜 Future Work

A proper fix is required, in place of this hot-fix. Relevant TODO comments are placed to remove the hot-fix related code.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests

### DIFF
--- a/lez/common/src/transaction.rs
+++ b/lez/common/src/transaction.rs
@@ -137,7 +137,7 @@ impl LeeTransaction {
     /// tag that lets the sequencer bypass the guard is not carried in the block,
     /// so the indexer cannot yet distinguish deposit txs from user txs.
     ///
-    /// REMOVE ME when the indexer can authenticate deposit transactions
+    /// REMOVE ME when the indexer can authenticate deposit transactions.
     pub fn execute_unchecked_on_state(
         self,
         state: &mut V03State,
--- a/lez/common/src/transaction.rs
+++ b/lez/common/src/transaction.rs
@@ -94,7 +94,7 @@ impl LeeTransaction {
 
     /// Computes the validated state diff without enforcing the system-account
     /// restriction. Shared by [`Self::validate_on_state`] and
-    /// [`Self::execute_unchecked_on_state`].
+    /// [`Self::execute_without_system_accounts_check_on_state`].
     fn compute_state_diff(
         &self,
         state: &V03State,
@@ -138,7 +138,7 @@ impl LeeTransaction {
     /// so the indexer cannot yet distinguish deposit txs from user txs.
     ///
     /// REMOVE ME when the indexer can authenticate deposit transactions.
-    pub fn execute_unchecked_on_state(
+    pub fn execute_without_system_accounts_check_on_state(
         self,
         state: &mut V03State,
         block_id: BlockId,
--- a/lez/common/src/transaction.rs
+++ b/lez/common/src/transaction.rs
@@ -78,18 +78,9 @@ impl LeeTransaction {
         block_id: BlockId,
         timestamp: Timestamp,
     ) -> Result<ValidatedStateDiff, lee::error::LeeError> {
-        let diff = match self {
-            Self::Public(tx) => {
-                ValidatedStateDiff::from_public_transaction(tx, state, block_id, timestamp)
-            }
-            Self::PrivacyPreserving(tx) => ValidatedStateDiff::from_privacy_preserving_transaction(
-                tx, state, block_id, timestamp,
-            ),
-            Self::ProgramDeployment(tx) => {
-                ValidatedStateDiff::from_program_deployment_transaction(tx, state)
-            }
-        }?;
+        let diff = self.compute_state_diff(state, block_id, timestamp)?;
 
+        // system accounts guard
         let system_accounts = lee::CLOCK_PROGRAM_ACCOUNT_IDS.iter().copied().chain([
             lee::system_faucet_account_id(),
             lee::system_bridge_account_id(),
@@ -99,6 +90,28 @@ impl LeeTransaction {
         }
 
         Ok(diff)
+    }
+
+    /// Computes the validated state diff without enforcing the system-account
+    /// restriction. Shared by [`Self::validate_on_state`] and
+    /// [`Self::execute_unchecked_on_state`].
+    fn compute_state_diff(
+        &self,
+        state: &V03State,
+        block_id: BlockId,
+        timestamp: Timestamp,
+    ) -> Result<ValidatedStateDiff, lee::error::LeeError> {
+        match self {
+            Self::Public(tx) => {
+                ValidatedStateDiff::from_public_transaction(tx, state, block_id, timestamp)
+            }
+            Self::PrivacyPreserving(tx) => ValidatedStateDiff::from_privacy_preserving_transaction(
+                tx, state, block_id, timestamp,
+            ),
+            Self::ProgramDeployment(tx) => {
+                ValidatedStateDiff::from_program_deployment_transaction(tx, state)
+            }
+        }
     }
 
     /// Validates the transaction against the current state, rejects modifications to clock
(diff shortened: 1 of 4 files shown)


erhant:
@Arjentix @Pravdyvy I got a Rust-specific question about the following code

https://github.com/logos-blockchain/logos-execution-zone/blob/73f540fa0a662601613bacfa37ad9c5f394ed97a/integration_tests/tests/bridge.rs#L458-L474

Here notice that I'm getting the indexer's account info with:

```rs
let indexer_account = indexer_service_rpc::RpcClient::get_account(ctx.indexer_client().deref(), account_id.into()).await?; 
```

because there is an overlapping trait for the `Rpc` and you have to [disambugate](https://doc.rust-lang.org/rust-by-example/trait/disambiguating.html). and if instead you do:

```rs
let indexer_account = ctx.indexer_client().get_account(account_id).await?; 
```

this causes trait resolution to go with `use sequencer_service_rpc::RpcClient`, which fails with:

```sh
Error: Parse error: invalid type: string "69yxuBKC8HLo56NQKKcLwqdgCiN3AmfiPqQsd1hUFxDq", expected an array of length 8 at line 1 column 63

Caused by:
    invalid type: string "69yxuBKC8HLo56NQKKcLwqdgCiN3AmfiPqQsd1hUFxDq", expected an array of length 8 at line 1 column 63
```

because the parsing logic is different among indexer and sequencer, in particular the `program_owner` field of `Account`.

**Sequencer** uses the re-export from `lee_core` which is a native type and is later given impls of `Serialize, Deserialize, BorshSerialize, BorshDeserialize` via `Account`:

https://github.com/logos-blockchain/logos-execution-zone/blob/73f540fa0a662601613bacfa37ad9c5f394ed97a/lee/state_machine/core/src/program.rs#L15

**Indexer** uses a `newtype` over that same native type, with a different `SerializeDisplay, DeserializeFromStr`:

https://github.com/logos-blockchain/logos-execution-zone/blob/73f540fa0a662601613bacfa37ad9c5f394ed97a/lez/indexer/service/protocol/src/lib.rs#L53-L56

This difference causes the parsing error above due to trait ambiguity.

I think there is something smelly here that we could fix, would love to hear your opinion!

Pravdyvy:
> @Arjentix @Pravdyvy I got a Rust-specific question about the following code
> 
> https://github.com/logos-blockchain/logos-execution-zone/blob/73f540fa0a662601613bacfa37ad9c5f394ed97a/integration_tests/tests/bridge.rs#L458-L474
> 
> Here notice that I'm getting the indexer's account info with:
> 
> ```rust
> let indexer_account = indexer_service_rpc::RpcClient::get_account(ctx.indexer_client().deref(), account_id.into()).await?; 
> ```
> 
> because there is an overlapping trait for the `Rpc` and you have to [disambugate](https://doc.rust-lang.org/rust-by-example/trait/disambiguating.html). and if instead you do:
> 
> ```rust
> let indexer_account = ctx.indexer_client().get_account(account_id).await?; 
> ```
> 
> this causes trait resolution to go with `use sequencer_service_rpc::RpcClient`, which fails with:
> 
> ```shell
> Error: Parse error: invalid type: string "69yxuBKC8HLo56NQKKcLwqdgCiN3AmfiPqQsd1hUFxDq", expected an array of length 8 at line 1 column 63
> 
> Caused by:
>     invalid type: string "69yxuBKC8HLo56NQKKcLwqdgCiN3AmfiPqQsd1hUFxDq", expected an array of length 8 at line 1 column 63
> ```
> 
> because the parsing logic is different among indexer and sequencer, in particular the `program_owner` field of `Account`.
> 
> **Sequencer** uses the re-export from `lee_core` which is a native type and is later given impls of `Serialize, Deserialize, BorshSerialize, BorshDeserialize` via `Account`:
> 
> https://github.com/logos-blockchain/logos-execution-zone/blob/73f540fa0a662601613bacfa37ad9c5f394ed97a/lee/state_machine/core/src/program.rs#L15
> 
> **Indexer** uses a `newtype` over that same native type, with a different `SerializeDisplay, DeserializeFromStr`:
> 
> https://github.com/logos-blockchain/logos-execution-zone/blob/73f540fa0a662601613bacfa37ad9c5f394ed97a/lez/indexer/service/protocol/src/lib.rs#L53-L56
> 
> This difference causes the parsing error above due to trait ambiguity.
> 
> I think there is something smelly here that we could fix, would love to hear your opinion!

I don't think it is really a problem. You should get some problems if you decide to use a trait from a different crate. It is really can be a problem only in cases if you use a couple of Rpc traits, which is the case only for integration tests. And again, compiler would object to direct usage if you import both.

So I would not try to fix this behaviour.

lez/common/src/transaction.rs
```diff
@@ -115,6 +128,28 @@ impl LeeTransaction {
...
+    /// Similar to [`Self::execute_check_on_state`], but skips the system-account guard.
+    ///
+    /// FIXME: HOT FIX (testnet v0.2): the indexer replays blocks the sequencer already
+    /// accepted, including sequencer-generated deposit transactions that
+    /// legitimately modify the bridge account. The `TransactionOrigin::Sequencer`
+    /// tag that lets the sequencer bypass the guard is not carried in the block,
+    /// so the indexer cannot yet distinguish deposit txs from user txs.
+    ///
+    /// REMOVE ME when the indexer can authenticate deposit transactions.
+    pub fn execute_unchecked_on_state(
```
Pravdyvy: It is not exactly unchecked, it is more of a check, but without system accounts.
Better rename it for clarity. 
It is up to you, possible candidate is `execute_without_system_accounts_check_on_state`.

Pravdyvy (approved):
lgtm, left minor comment

lez/common/src/transaction.rs
erhant: Agreed, renamed as such. (https://github.com/logos-blockchain/logos-execution-zone/pull/520/commits/4076de05c09b1c9803a9deba978792a3b3c1b3d8)

schouhy (approved):
LGTM thanks!

moudyellaz (approved):
lgtm, thanks

PR #521, opened 2026-06-11
feat(cycle_bench): add public-execution ms calibration for fee model

moudyellaz:
## 🎯 Purpose

The tokenomics team needs public-program execution reported in milliseconds, not only Risc0 user cycles, so public and private execution sit on one common ms axis (the private path already reports `G_verify` in ms). This adds the ms measurement and a pinned-hardware calibration so the fee model has a single unit for both paths.

## ⚙️ Approach

Fit executor wall-time against the deterministic user-cycle counts and report the per-instruction ms alongside the existing cycles.

- [x] Add `Calibration`: ordinary least squares of `best_ms = intercept + slope · user_cycles` across the standalone cases, deriving a throughput constant (cycles/ms) and the host-side fixed per-call overhead (intercept ms). Best-of-N is used so one OS scheduling spike cannot tilt the slope.
- [x] Add two derived per-instruction fields: `calibrated_ms` (= slope · cycles, compute only, the number for the common public/private axis) and `net_compute_ms` (= best_ms − intercept, measured compute with overhead stripped).
- [x] Print a calibration block and add the calibration object to `target/cycle_bench.json`.
- [x] Document the methodology and the finding that the fixed ~30 ms overhead is paid per transaction today (public execution at `lee/state_machine/src/program.rs:56-87` parses the ELF per call with no image cache), so the raw per-tx cost is overhead-dominated and caching the parsed image would drop it to `calibrated_ms`.

## 🧪 How to Test

`cargo run --release -p cycle_bench -- --exec-iters 50` prints the per-instruction table (user_cycles, exec_ms, calib_ms, net_ms) and the calibration block (throughput, fixed overhead, R²). `cargo test -p cycle_bench` runs three unit tests over the OLS fit (known-line recovery, fewer-than-two-points, degenerate identical-x guard). On an M2 Pro the fit reproduces at throughput ≈ 33,500 cycles/ms, fixed overhead ≈ 30.4 ms, R² ≈ 0.93.

## 🔗 Dependencies

None.

## 🔜 Future Work

Optionally cache the parsed `MemoryImage` per `ProgramId` in the public-execution path to remove the ~30 ms per-tx parse overhead, which would bring per-tx public cost down to the calibrated compute figure.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests

### DIFF
--- a/docs/benchmarks/cycle_bench.md
+++ b/docs/benchmarks/cycle_bench.md
@@ -14,21 +14,37 @@ Per-program Risc0 cycle counts, prover wall time, PPE composition cost, and veri
 | Profile | release |
 | GPU acceleration | none |
 
-## Executor cycles
+## Executor cycles and public-execution ms
 
-`SessionInfo::cycles()` per instruction. Deterministic across runs. Wall time is `best / mean ± stdev` over 5 timed iterations (1 warmup discarded).
+`SessionInfo::cycles()` per instruction. Deterministic across runs. Wall time is `best / mean ± stdev` over the timed iterations (1 warmup discarded; `--exec-iters` sets the count, 50 below). `calib_ms` and `net_ms` are the public-execution time in milliseconds, on the same axis as the private `G_verify` so the fee model has one common unit for both paths. See the calibration block below for how they are derived.
 
-| Program | Instruction | user_cycles | segments | exec_ms (best / mean ± stdev) |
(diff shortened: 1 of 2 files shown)


Pravdyvy (approved):
lgtm

schouhy (approved):
looks great thanks

PR #523, opened 2026-06-11
fix: bind the EncryptedAccountData to instance

agureev:
## 🎯 Purpose

Currently, for private transfers with no public accounts involved, the only binding part of the transaction is the shielding proof. However, the proof does not bind the epk or the view tag, making them malleable. By changing the epk the malicious actor could thereby burn any such transfer by getting rid of the information needed for ss recovery.

## ⚙️ Approach

The approach of the PR is basic: simply feed both the epk and the view_tag to the circuit as additional read bytes and then write the `EncryptedAccountData`. Note: this still does **not** grant that either the view tag or the epk are well-formed/connected to other witness info. E.g. the user can still feed a random ephemeral key 

- [x] Expose `EncryptedAccountData` to guest
- [x] Change the private account identities to accept the extra encryption info
- [x] Change the shielding program to write the full `EncryptedAccountData` to journal
- [x] Update tests and ELFs

## 🧪 How to Test

Simulate a user sending in a transaction with only private state, try to tamper with the encrypted account data and try to submit the malformed tx. It should be rejected on the proof verification stage.

## 🔗 Dependencies

None

## 🔜 Future Work

The fact that there is 

1) no actual binding between the note-owner and the shared secret and
2) that there is currently no binding between the epk and the shared secret

makes me think that the most elegant further way to go is to actually skip generation of the ciphertext in-guest and instead let the user just feed arbitrary bytes to the program, spitting them out in journal for binding.

However this would change the discovery mechanism in a major way, so instead of trying it on my own I will delegate the decision regarding this to further discussion.

TO COMPLETE IF APPLICABLE

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


schouhy (approved):
looks great thanks!

moudyellaz (approved):
lgtm, thanks

PR #524, opened 2026-06-11
Decouple last parts of LEZ from LEE

Arjentix:
## 🎯 Purpose

After recent refactor with lee/lez renaming some parts of lez still persisted in lee. This PR aims to fix this problem. Also done some other refactoring on the way.

## ⚙️ Approach

- Fixed docker builds
- Keep LEZ and LEE programs isolated, create a separate crate `programs` for LEZ system programs.
- Remove artifacts for test programs as this makes no sense, we only need determenistic binaries for production programs
- Refactor wallet `send_pub_tx()` to accept program_id instead of program, as this was redundant
- Store `Cow<'static, [u8]>` as program elf to make our system programs to be lightweight 

## 🧪 How to Test

See CI.

## 🔗 Dependencies

None

## 🔜 Future Work

No related ideas for now

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


lee/state_machine/src/state.rs
```diff
@@ -3104,14 +2844,14 @@ pub mod tests {
...
-        let auth_transfer = Program::authenticated_transfer_program();
+        let chain_caller = crate::test_methods::chain_caller();
+        let simple_transfer = crate::test_methods::simple_balance_transfer();
         let from_key = PrivateKey::try_new([1; 32]).unwrap();
         let from = AccountId::from(&PublicKey::new_from_private_key(&from_key));
         let initial_balance = 100;
         let initial_data = [(from, initial_balance)];
-        let mut state =
-            V03State::new_with_genesis_accounts(&initial_data, vec![], 0).with_test_programs();
+        let mut state = V03State::new(public_state_from_balances(&initial_data), vec![], vec![])
```
erhant: it seems `V03State::new({something}, vec![], vec![])` is a rather common pattern in tests, maybe we can put it within the VO3State `impl` itself, gated with a `#[cfg(test)]`, to not bother with the two `vec![]`s?

lee/state_machine/src/state.rs
```diff
@@ -3387,101 +3124,101 @@ pub mod tests {
...
-        let tx = PublicTransaction::new(message, witness_set);
-        state.transition_from_public_transaction(&tx, 1, 0).unwrap();
-
-        let winner_token_holding_post = state.get_account_by_id(winner_token_holding_id);
-        assert_eq!(
-            winner_token_holding_post,
-            expected_winner_token_holding_post
-        );
-    }
+    // #[test]
```
erhant: perhaps a `// TODO:` comment on when to uncomment this?

lez/sequencer/service/Dockerfile
```diff
@@ -26,11 +26,11 @@ RUN ARCH=$(uname -m); \
     else \
         echo "Using manual build for $ARCH"; \
         git clone --depth 1 --branch release-3.0 https://github.com/risc0/risc0.git; \
-        git clone --depth 1 --branch r0.1.91.0 https://github.com/risc0/rust.git; \
+        git clone --depth 1 --branch risc0-1.94.1 https://github.com/risc0/rust.git; \
```
erhant: could we share these steps accross the project with a common Dockerfile, e.g. `Dockerfile.risc0` or something in those lines?

test_programs/src/lib.rs
```diff
@@ -0,0 +1,88 @@
+#![expect(
+    clippy::no_effect_underscore_binding,
+    reason = "This way we can remove warnings about unused path constants"
```
erhant, line 3: why do we have to `use` and then unuse imports here? risc0 stuff probably but I was not aware 🤔

erhant (approved):
LGTM, awesome work; looks so much cleaner and readable i think

just few comments here and there

lee/state_machine/src/state.rs
Arjentix: Good idea, replaced with builder pattern, now looks cleaner

lee/state_machine/src/state.rs
Arjentix: Thanks for noticing, I've forgot to fix this. Now fixed moved to `lez/sequencer/core/lib.rs` as these tests requires lez programs

lez/sequencer/service/Dockerfile
Arjentix: Good idea, but I don't think it's worth it, because after we updated to a new zone sdk (I think it's already on main) we no longer need logos circuits. Which means we can finally emulate x86_64 architecture on mac. Which means we no longer need manual installation here. So I'll remove it.

I think I'll do this changes and probably move this commit to another PR to merge it faster

test_programs/src/lib.rs
Arjentix: Hah, no, it's not risc0. So the thing is that we do `include!` code generated at `build.rs`. But now we include them privately. Which means everything defined in this generated `methods.rs` must be used in some way. So not using `*_PATH` constant generates a warning, so that's why I assign it in a variable and don't use it later.

As an alternative I could just allow `dead_code` lint or something like that. But I don't like that allowing this would also allow to ignore `*_ELF` and `*_ID` constants which are meaningful. For example if some of this variable is not used that would mean someone placed a new risc0 program here, but forgot to add a constructor-function for it.

lez/wallet-ffi/src/generic_transaction.rs
```diff
@@ -214,7 +216,7 @@ pub unsafe extern "C" fn wallet_ffi_send_generic_public_transaction(
     account_identities_size: usize,
     instruction_words: *const u32,
     instruction_words_size: usize,
-    program_with_dependencies: *const FfiProgramWithDependencies,
+    program_id: *const ProgramId,
```
schouhy: This needs to use FfiProgramId?

moudyellaz: I agree, the ProgramId is a bare [u32;8] alias with no cbindgen typedef, so wallet_ffi.h won't compile. I think you should use *const FfiProgramId.

moudyellaz (approved):
lgtm, thanks. one thing to note is that the program crate renames regenerate ELFs, changing all built-in ProgramIds and genesis-derived account ids.

lez/wallet-ffi/src/generic_transaction.rs
Arjentix: Actually FfiProgramId works well here, thanks, updated

PR #529, opened 2026-06-15
fix(lee): Fix redundant `ELF.to_vec()` on `Program::*().id()` calls, and avoid redundant hash recomutations

### LINKED ISSUE #528: `Program::*().id()` allocates and copies the full program ELF just to read a const `ProgramId`
It appears that `Program::bridge()`, `Program::faucet()`, etc. (all `Program::*()` constructors in `lee/state_machine/src/program.rs`) eagerly copy the program's entire ELF binary into a freshly heap-allocated `Vec<u8>` (`elf: BRIDGE_ELF.to_vec()` etc.).

This is a slight waste when used within the common pattern `Program::bridge().id()` which we can see across the state machine, the sequencer, the indexer, and tests. This means around 400-530 KB copy is allocated, memcpy'd, and immediately dropped, when all the caller wanted was the `ProgramId`, which is **already a compile-time constant**.

For tests / one-time constructors and such this is okay, but this call sits on the per-transaction validation hot path, when we are iterating over system accounts during state transition validation.

## Background / root cause

The build-generated `program_methods/mod.rs` already exposes both pieces as `statics`/`consts`:

https://github.com/logos-blockchain/logos-execution-zone/blob/6744e2f8c3a380acd9fe29f380df590195ec8fcd/lee/state_machine/src/lib.rs#L36-L38

```rust
pub const BRIDGE_ELF: &[u8] = include_bytes!(r#"{YOUR_DIR}/logos/logos-execution-zone/lee/state_machine/../../artifacts/program_methods/bridge.bin"#);
#[expect(clippy::unreadable_literal, reason = "Generated image IDs from risc0 are cryptographic hashes represented as u32 arrays")]
pub const BRIDGE_ID: [u32; 8] = [2578452121, 40021765, 1586464073, 2631595208, 4266510835, 1545795964, 2598872627, 1407648414];
// ...
```

The `Program` constructors then wrap them:

https://github.com/logos-blockchain/logos-execution-zone/blob/6744e2f8c3a380acd9fe29f380df590195ec8fcd/lee/state_machine/src/program.rs#L169-L175

Yet `id()` just returns the const field:

https://github.com/logos-blockchain/logos-execution-zone/blob/6744e2f8c3a380acd9fe29f380df590195ec8fcd/lee/state_machine/src/program.rs#L40-L43

So `Program::{{*}}().id()` does a full-ELF allocate-copy-free just for the sake of getting the id.

## Cost

ELF sizes (`artifacts/program_methods/`), copied in full on each constructor call:

| Program                  | ELF size copied per call |
| ------------------------ | ------------------------ |
| amm                      | 533 KB                   |
| token                    | 483 KB                   |
| associated_token_account | 451 KB                   |
| bridge                   | 425 KB                   |
| faucet                   | 422 KB                   |
| vault                    | 414 KB                   |
| authenticated_transfer   | 405 KB                   |
| clock                    | 404 KB                   |
| pinata / pinata_token    | ~394-400 KB              |

Each `Program::X().id()` therefore costs roughly one ~400-530 KB allocation + memcpy + free. On Apple Silicon that is on the order of **15-50 µs per call** (the memcpy alone is ~10-15 µs; allocations this large go through the allocator's large-object path, so malloc/free + page-fault overhead can dominate).

Most call sites are in tests and don't matter. The one that does is the
**per-transaction validation path**, hit by both the sequencer and the indexer (inside `validate_on_state`, runs per tx):

https://github.com/logos-blockchain/logos-execution-zone/blob/6744e2f8c3a380acd9fe29f380df590195ec8fcd/lez/common/src/transaction.rs#L84-L87

Each of those two helpers rebuilds a full `Program` purely to read its `id`, and then *also* recomputes a hash:

https://github.com/logos-blockchain/logos-execution-zone/blob/6744e2f8c3a380acd9fe29f380df590195ec8fcd/lee/state_machine/src/state.rs#L400-L407

That is **~850 KB of needless heap churn plus two hash computations per transaction**, on top of the allocator pressure it creates.

In *relative* terms this is likely <1% of public-transaction validation time, because `compute_state_diff` runs the Risc0 executor (milliseconds), which dwarfs it. But it is:
- 100% wasted work (no correctness or functional purpose),
- on the hottest shared path in the node,
- and a source of large-allocation churn that pressures the allocator under load (e.g. ~3-10% of one core at 1k tx/s just for the copies, before counting allocator/GC-of-the-OS effects).

It is **cheap** to remove entirely.

## Proposed fix

We have two options:

1. **Add const id accessors that never touch the ELF.** For each program, expose something like:

 ```rust
 #[must_use]
 pub const fn bridge_id() -> ProgramId { ProgramId::from(BRIDGE_ID) }
 ```

and switch `.id()`-only call sites to use them. This removes the copy at every such site (including all the test sites, which speeds up the test suite too).

2. **Cache the system account ids.** Make `system_faucet_account_id` and `system_bridge_account_id` back onto `LazyLock<AccountId>` statics. This eliminates *both* the ELF copy and the repeated `compute_*_account_id` hashing on the per-tx path, reducing the per-transaction cost to a pointer read:

```rust
static FAUCET_ACCOUNT_ID: LazyLock<AccountId> = LazyLock::new(|| faucet_core::compute_faucet_account_id(Program::faucet_id()));
   
#[must_use]
pub fn system_faucet_account_id() -> AccountId { *FAUCET_ACCOUNT_ID }
```

(Same for bridge.)

Option 2 alone removes essentially all of the real-world steady-state waste, since everything else is test code or one-time initialization. Option 1 generalizes the fix and cleans up the rest.

erhant:
## 🎯 Purpose

- Fixes #528 (see issue for details)

## ⚙️ Approach

`Program::*().id()` is optimized with alternative `id` only calls, `SYSTEM_FAUCET_ACCOUNT_ID` and `SYSTEM_BRIDGE_ACCOUNT_ID` now uses `LazyLock` to avoid hashing each time

- [x] Change `Program::*().id()` instances to `Program::*_id()` instead, skipping the redundant `ELF.to_vec()` on each call
- [x] Change `system_faucet_account_id` and `system_bridge_account_id` to use a static LazyLock'ed result instead of recomputing the same thing on each call

## 🧪 How to Test

```sh
# test specific package
cargo test -p lee --lib

# checks
cargo check --workspace --all-targets --all-features

# unit tests
RISC0_DEV_MODE=1 cargo nextest run --workspace --exclude integration_tests --no-fail-fast
```

## 🔗 Dependencies

None

## 🔜 Future Work

None

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests

### DIFF
--- a/lee/state_machine/src/program.rs
+++ b/lee/state_machine/src/program.rs
@@ -113,65 +113,97 @@ impl Program {
     #[must_use]
     pub fn authenticated_transfer_program() -> Self {
         Self {
-            id: AUTHENTICATED_TRANSFER_ID,
+            id: Self::authenticated_transfer_program_id(),
             elf: AUTHENTICATED_TRANSFER_ELF.to_vec(),
         }
+    }
+    #[must_use]
+    pub const fn authenticated_transfer_program_id() -> ProgramId {
+        AUTHENTICATED_TRANSFER_ID
     }
 
     #[must_use]
     pub fn token() -> Self {
         Self {
-            id: TOKEN_ID,
+            id: Self::token_id(),
             elf: TOKEN_ELF.to_vec(),
         }
+    }
+    #[must_use]
+    pub const fn token_id() -> ProgramId {
+        TOKEN_ID
     }
 
     #[must_use]
     pub fn amm() -> Self {
         Self {
-            id: AMM_ID,
+            id: Self::amm_id(),
             elf: AMM_ELF.to_vec(),
         }
+    }
+    #[must_use]
+    pub const fn amm_id() -> ProgramId {
+        AMM_ID
     }
 
     #[must_use]
     pub fn clock() -> Self {
         Self {
-            id: CLOCK_ID,
+            id: Self::clock_id(),
             elf: CLOCK_ELF.to_vec(),
         }
+    }
+    #[must_use]
+    pub const fn clock_id() -> ProgramId {
+        CLOCK_ID
     }
 
     #[must_use]
     pub fn ata() -> Self {
         Self {
-            id: ASSOCIATED_TOKEN_ACCOUNT_ID,
+            id: Self::ata_id(),
             elf: ASSOCIATED_TOKEN_ACCOUNT_ELF.to_vec(),
         }
+    }
+    #[must_use]
+    pub const fn ata_id() -> ProgramId {
+        ASSOCIATED_TOKEN_ACCOUNT_ID
     }
 
     #[must_use]
     pub fn vault() -> Self {
         Self {
-            id: VAULT_ID,
+            id: Self::vault_id(),
             elf: VAULT_ELF.to_vec(),
         }
+    }
+    #[must_use]
+    pub const fn vault_id() -> ProgramId {
+        VAULT_ID
     }
 
     #[must_use]
     pub fn faucet() -> Self {
         Self {
-            id: FAUCET_ID,
+            id: Self::faucet_id(),
             elf: FAUCET_ELF.to_vec(),
         }
+    }
+    #[must_use]
+    pub const fn faucet_id() -> ProgramId {
+        FAUCET_ID
     }
 
     #[must_use]
     pub fn bridge() -> Self {
         Self {
-            id: BRIDGE_ID,
+            id: Self::bridge_id(),
             elf: BRIDGE_ELF.to_vec(),
         }
+    }
+    #[must_use]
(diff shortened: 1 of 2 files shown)


erhant:
Note that #524 uses `Cow<'static, [u8]>` for the `elf` field, which does solve the problem. This PR can be closed in favor of that one.

PR #533, opened 2026-06-15
refactor(integration_tests): cleaning up long functions

jonesmarvin8:
## 🎯 Purpose

This PR introduces helper functions to reduce duplicated code in `integration_tests`. This refactor aims to shorten long functions in `integration_tests`.

## ⚙️ Approach

Simple refactor.

## 🧪 How to Test

No change in functionality. All previous tests should pass.

## 🔗 Dependencies

N/A

## 🔜 Future Work

N/A

## 📋 PR Completion Checklist

- [X] Complete PR description
- [X] Add/update tests

### DIFF
--- a/integration_tests/src/lib.rs
+++ b/integration_tests/src/lib.rs
@@ -169,9 +169,6 @@ pub fn assert_public_account_restored(ctx: &TestContext, account_id: AccountId, 
         .unwrap_or_else(|| panic!("{label} should be restored"));
 }
 
-/// Maximum time to wait for the indexer to catch up to the sequencer.
-pub const L2_TO_L1_TIMEOUT: Duration = Duration::from_mins(7);
-
 /// Poll the indexer until its last finalized block id reaches the sequencer's
 /// current last block id or until [`L2_TO_L1_TIMEOUT`] elapses.
 /// Returns the last indexer block id observed.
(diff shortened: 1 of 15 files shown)


integration_tests/src/lib.rs
```diff
@@ -6,12 +6,169 @@
...
+
+/// Fetch the current commitment for `account_id` and assert it is present in the sequencer state.
+pub async fn assert_private_commitment_in_state(
+    ctx: &TestContext,
+    account_id: AccountId,
+    label: &str,
+) -> Result<()> {
+    let commitment = ctx
+        .wallet()
+        .get_private_account_commitment(account_id)
```
moudyellaz, line 134: Coverage is preserved overall, one note on this helper. It fetches the wallet's stored commitment via `get_private_account_commitment` rather than the independently computed post-spend value the inline asserts used (for example `post_spend_commitment_0` in `private_pda`), so it checks wallet-vs-chain consistency instead of expected-vs-chain. Fine if intentional, just flagging that a few of these checks are marginally weaker than before.

jonesmarvin8: The helper function matches with the main's approach for these functions (wallet-vs-chain). Perhaps, expected-vs-chain was changed to wallet-vs-chain in a previous PR?

Pravdyvy (approved):
lgtm

schouhy (approved):
lgtm thanks

moudyellaz (approved):
lgtm, thanks! (resolved the discussion, you were right!)

PR #542, opened 2026-06-17
feat!(cross-zone): async LEZ to LEZ cross-zone messaging

moudyellaz:
## 🎯 Purpose

Adds async cross-zone messaging between LEZ instances: one zone sends a message to another over Bedrock, delivered and verified without trusting the source sequencer. Ships two demos on one shared spine, a ping (Demo 1) and a wrapped-token bridge that locks on zone A and mints on zone B (Demo 2), plus an interactive two-zone chat app to try it live.

## ⚙️ Approach

An emitter program on the source zone records an outbound message, the destination zone's sequencer runs a watcher that reads the source zone's finalized Bedrock blocks and injects the delivery as a sequencer-origin transaction, and the destination indexer independently re-derives that delivery from the same finalized block and rejects any forgery or replay (Option B, indexer-verified).

- [x] Six builtin programs under lez/programs/ (cross_zone_outbox, cross_zone_inbox, ping_sender, ping_receiver, bridge_lock, wrapped_token), each a guest-pure core plus guest
- [x] Host glue crate lez/cross_zone (emission extraction, dispatch builder, genesis account builders, sequencer-only guard)
- [x] Sequencer watcher and indexer Option B verifier, with peer block-signing key pinning
- [x] Genesis registration in testnet_initial_state, per-zone inbox config and bridge-lock holdings seeded consistently on sequencer and indexer
- [x] Ingress guard so user transactions cannot invoke the sequencer-only inbox
- [x] Interactive chat demo app (tools/cross_zone_chat)

## 🧪 How to Test

Dev mode, needs Docker for Bedrock:
- just demo-cross-zone-ping (indexer-verified ping A to B)
- just demo-cross-zone-bridge (lock on A, mint on B)
- just cross-zone-chat (interactive two-column web UI, open the printed localhost URL)

Automated: cargo nextest run -p integration_tests --test cross_zone_state_machine / cross_zone_ingress_guard / cross_zone_verified / cross_zone_bridge (RISC0_DEV_MODE=1; on macOS set DYLD_FALLBACK_FRAMEWORK_PATH).

## 🔗 Dependencies

None.

## 🔜 Future Work

This PR is the first part: a working cross-zone messaging spine with the ping and bridge demos. A broader follow-up will generalize it into full async cross-LEZ messaging, expanding the design (generalized message envelopes, an emitter-agnostic watcher that reacts to outbox activity, refund handling, and the wider trust and discovery model) beyond the two demo flows here.

## ⚠️ Breaking Changes

Genesis state root changes. Registers the six cross-zone builtin programs and seeds their genesis accounts. Building the new cores alongside the existing builtins enables serde/alloc for the shared programs build, which regenerates every builtin program image id. All nodes must upgrade together.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


lee/state_machine/src/state/mod.rs
```diff
@@ -194,6 +194,14 @@ impl V03State {
         self.programs.insert(program.id(), program);
     }
 
+    /// Seeds a single genesis account that is not produced by any transaction
+    /// (e.g. the cross-zone inbox config or a bridge-lock holding). Lets the
+    /// sequencer and indexer seed identical zone-specific state after building
+    /// the shared initial state.
+    pub fn insert_genesis_account(&mut self, account_id: AccountId, account: Account) {
+        self.public_state.insert(account_id, account);
+    }
```
Arjentix, line 203: Hm, why existing constructor `with_public_accounts()` is not enough? Ideally I'd only modify state via transitions from transactions

lez/indexer/core/Cargo.toml
```diff
@@ -16,6 +16,9 @@ common.workspace = true
 logos-blockchain-zone-sdk.workspace = true
 lee.workspace = true
 lee_core.workspace = true
+cross_zone.workspace = true
+cross_zone_inbox_core.workspace = true
```
Arjentix, line 20: I don't understand why indexer now depends on cross zone... I thought for this feature we don't touch our main code?

lez/sequencer/core/Cargo.toml
```diff
@@ -20,6 +20,8 @@ bridge_core.workspace = true
 vault_core.workspace = true
 programs.workspace = true
 system_accounts.workspace = true
+cross_zone.workspace = true
+cross_zone_inbox_core.workspace = true
```
Arjentix, line 24: Same question... I thought we make just a non-invasive example above our services

lez/testnet_initial_state/src/lib.rs
```diff
@@ -267,6 +288,12 @@ fn initial_programs() -> Vec<Program> {
         programs::vault(),
         programs::faucet(),
         programs::bridge(),
+        programs::cross_zone_outbox(),
+        programs::cross_zone_inbox(),
+        programs::ping_sender(),
+        programs::ping_receiver(),
+        programs::bridge_lock(),
+        programs::wrapped_token(),
```
Arjentix, line 300: This is the function for our "production" programs. I think this cross zone implementation serves more like an example and these programs shouldn't be there.

In your case I'd consider two options:
1. Manually submit every program you need during cross zone communication setup
2. Add a GenesisAction::DeployProgram and use it in genesis config

lez/testnet_initial_state/src/lib.rs
```diff
@@ -254,9 +254,30 @@ fn initial_public_accounts() -> HashMap<AccountId, Account> {
...
+        Account {
+            program_owner: wrapped_token_id,
+            data: wrapped_token_core::minter_bytes(programs::cross_zone_inbox().id())
+                .to_vec()
+                .try_into()
+                .expect("minter id fits in account data"),
+            ..Default::default()
+        },
+    )
+}
```
Arjentix, line 283: Same proposal. This function is for production system accounts only. Let's better submit this account in cross zone setup via wallet or with genesis config specifically for cross zone setup

Arjentix (changes requested):
I partially reviewed the PR and decided to leave comments now, because all further comments will be about the same thing.

My main concern is that this feature touched code in Indexer and Sequencer. To my understanding this should have been a feature on top of them, not inside them.

Currently to me it looks like a very narrow case which requires a lot of core changes.

1) I don't see a reason to not to have separate Indexers for separate blockchains. Zone B just should have it's own Indexer with custom rules and etc, we should not hardcode this logic into ours Indexer.
2) Sequencer IMO should also only follow our own rules. Supporting every other possible chain is a ton of work and security risk. If someone wants to create a bridge to our zone they could deploy programs and use oracles/relayers approach or do verification entirely on Chain B. Kind of what everybody do I suppose

Arjentix:
After some thinking and talking with @agureev I now see the point.

So seems like we completely don't want **trustful** implementation with relayers. We want a more trustless approach where each chain watches other chain. In that case there is no way to avoid new code in core services.

But as I understand current implementation heavily relies on the fact that chain B is the same LEZ. To me looks like that assumption will never be met in reality. So instead I'd plan how we can make this briding expandable through some traits.

So not hardcode things for this example in sequencer, but have a very generic trait usage there. And move all chain B related verification and etc to this trait implementation in another crate.

I imagine structure like that:

```
lee/
lez/
  bridging/
    src/
      lib.rs    <--- This is where trait defined
  sequencer/
    core/      <--- This is where trait used
third_party_chain_bridges/
  chain_b_bridge/
    src/
      lib.rs    <--- This is where trait implemented
```

lez/indexer/core/Cargo.toml
moudyellaz: the verifier (indexer) and watcher (sequencer) run inside those processes, so they depend on the cross-zone glue crate `lez/cross_zone`, which is where the zone B (what you call chain) logic actually lives. As you can see, it is opt-in: `cross_zone` is an Option in the config, off by default, so a node with no cross-zone config is unaffected. This is the unavoidable core touch of the trustless model, but the chain-specific logic itself is not hardcoded in the service, it is in the separate crate.

lez/testnet_initial_state/src/lib.rs
moudyellaz: Ok fair point. They are here now only because the programs must be registered as builtins for the demo to run. I will move them out of the production list in a follow-up, either gated behind the cross-zone setup or via a `GenesisAction::DeployProgram` as you suggest.

lez/testnet_initial_state/src/lib.rs
moudyellaz: Same as above, I will move it into a cross-zone-specific genesis config in the follow-up.

lee/state_machine/src/state/mod.rs
moudyellaz: These are zone-specific genesis accounts (the inbox config and bridge-lock holdings) that depend on the node's cross-zone config and are not produced by any transaction, so they are seeded after the shared initial_state is built, which with_public_accounts is part of and does not see the per-zone config. I agree that modifying state through a transaction is cleaner, and I will move this to a genesis Configure transaction in the follow-up.

moudyellaz:
> After some thinking and talking with @agureev I now see the point.
> 
> So seems like we completely don't want **trustful** implementation with relayers. We want a more trustless approach where each chain watches other chain. In that case there is no way to avoid new code in core services.
> 
> But as I understand current implementation heavily relies on the fact that chain B is the same LEZ. To me looks like that assumption will never be met in reality. So instead I'd plan how we can make this briding expandable through some traits.
> 
> So not hardcode things for this example in sequencer, but have a very generic trait usage there. And move all chain B related verification and etc to this trait implementation in another crate.
> 
> I imagine structure like that:
> 
> ```
> lee/
> lez/
>   bridging/
>     src/
>       lib.rs    <--- This is where trait defined
>   sequencer/
>     core/      <--- This is where trait used
> third_party_chain_bridges/
>   chain_b_bridge/
>     src/
>       lib.rs    <--- This is where trait implemented
> ```

The problem with relayers is that it would be a trustful option, and the whole point here is that each zone verifies the other itself, which is why the watcher and verifier have to live in the sequencer and indexer.

On the one assumption you flagged, that chain B is the same LEZ (zone B): that is the intended scope, not a limitation. 

This feature is cross-LEZ messaging, zone to zone between LEZ instances on the same Bedrock. Zone B being another LEZ is the premise that makes Option B verification possible at all, both sides run identical LEZ code, so zone B's indexer can re-derive zone A's dispatch byte for byte and reject anything that does not match. Bridging to a non-LEZ zone is a different and larger problem, and it is out of scope for this feature.

That is also where your trait idea fits. A generic bridging trait with chain-specific impls in their own crate is the right shape if we ever need to support a second, non-LEZ zone type. This feature does not need it: v1 is LEZ to LEZ, one implementation, so a trait with a single impl would be indirection with no payoff. I would leave it out for now and only revisit it if a concrete non-LEZ-zone requirement shows up. Worth noting we are already partway to the separation you want anyway: the Zone-B logic lives in its own crate, lez/cross_zone, and the sequencer and indexer only call it, and cross-zone is opt-in, an Option in config that is off by default, so a zone with no cross-zone config behaves exactly as today.

On the genesis and state-seeding touches (programs in testnet_initial_state, insert_genesis_account): fair points. I explained the current choice in each thread, and I am happy to clean these up in a follow-up.

lez/indexer/core/src/lib.rs
```diff
@@ -137,11 +164,30 @@ impl IndexerCore {
 
                     info!("Indexed L2 block {}", block.header.block_id);
 
+                    // Option B: re-derive and verify every cross-zone dispatch
+                    // before applying the block. A forged or replayed dispatch
+                    // halts ingestion rather than persisting an invalid state.
+                    if let Some(verifier) = &self.verifier
+                        && let Err(err) = verifier.verify_block(&block).await
```
agureev, line 263: So, question, what if we have a subscription chain A -> B -> C? (i.e. the token send chain is C -> B -> A) Then a malicious C can write a malicious block minting 100 BTC on B alongside making a tx on B which sends said 100 tokens to A. Now, the watcher of B will put this in a block (even if the indexer will later reject it via option B). 

However, the tx still gets posted on Bedrock and hence A's watcher (which has B as its peer) will read it and actually print 100 malicious BTC which its indexer will recognize, as it only verifiers A's blocks, not C's.

lez/indexer/core/src/cross_zone_verifier.rs
```diff
@@ -0,0 +1,500 @@
...
+///
+/// For every cross-zone dispatch in a block it re-derives the transaction from
+/// the peer's finalized block and rejects it if the bytes differ (a forgery) or
+/// the message was already delivered (a replay), so delivery no longer relies on
+/// trusting the sequencer.
+#[derive(Clone)]
+pub struct CrossZoneVerifier {
+    self_zone: ZoneId,
+    /// Pinned block-signing key per peer zone, enforced during re-derivation.
+    peer_pubkeys: HashMap<ZoneId, PublicKey>,
```
agureev, line 75: This probably deserves some comment about handling it for the decentralization, i.e. where there will be several possible keys per zone (or at least they will have to be replaced based on some criterion in the struct).

agureev: If we trust Bedrock indexing we can I guess just skip the signature part since to write you have to already have the appropriate signature for your turn?

lez/programs/cross_zone_inbox/src/main.rs
```diff
@@ -0,0 +1,125 @@
...
+        "Target program is not allowed for this peer"
+    );
+
+    let key = message_key(&msg.src_zone, msg.src_block_id, msg.src_tx_index);
+    let mut shard =
+        SeenShard::from_bytes(&seen.account.data.clone().into_inner()).expect("seen shard decodes");
+    let already_seen = shard.contains(&key);
+
+    // On replay this is a no-op: the seen shard is untouched and no call is made.
+    let (seen_post, chained_calls) = if already_seen {
```
agureev, line 76: This seems to differ from the Option B semantics, where `verify_block` fails on a seen shard.

lez/indexer/core/src/cross_zone_verifier.rs
```diff
@@ -0,0 +1,500 @@
...
+    /// first forged or replayed dispatch. The caller halts ingestion on error.
+    pub async fn verify_block(&self, block: &Block) -> Result<()> {
+        for tx in &block.body.transactions {
+            let Some(msg) = Self::decode_dispatch(tx) else {
+                continue;
+            };
+
+            let key = message_key(&msg.src_zone, msg.src_block_id, msg.src_tx_index);
+            if self.seen.read().await.contains(&key) {
+                bail!(
```
agureev, line 125: This is the point where it seems we should `continue` if we are to abide by the program semantics

lez/cross_zone/src/lib.rs
```diff
@@ -0,0 +1,189 @@
...
+/// watcher and verifier both use this so they agree on what a given source tx
+/// emits.
+#[must_use]
+pub fn extract_emission(program_id: ProgramId, instruction_data: &[u32]) -> Option<Emission> {
+    if program_id == programs::ping_sender().id() {
+        let ping_core::SenderInstruction::Send {
+            target_zone,
+            target_program_id,
+            target_accounts,
+            payload,
```
agureev, line 49: If I understand everything correctly, there is currently nothing stopping the ping program instruction to write `wrapped_token::Mint{..}` here is there? The serde just allows to decode whatever and the program allows to write whatever during the ping. Of course this is a ping-specific bug but probably worth addressing here.

agureev (commented):
Thank you for the PR! Left a couple of comments/questions.

lez/indexer/core/src/lib.rs
agureev: Worth saying that this is not currently exploitable in the program (lock tokens are gotten on genesis rn iiuc), but may be something to consider as we go further because a general program will probably allow for much more general functionality.

agureev: Actually, I think I am wrong about B's indexer's catching it. `verify_block` does not actually re-run the transactions, does it? And based on the description the indexer for B only catches B's own wrong transactions by replay. Do we have any place where we actually rerun the sending zone's transactions? Or do we simply trust it right now?

lez/cross_zone/src/lib.rs
```diff
@@ -0,0 +1,189 @@
...
+    program_id == programs::cross_zone_inbox().id()
+}
+
+/// Extracts the cross-zone emission from a source transaction.
+///
+/// Recognizes the known emitter programs (`ping_sender`, `bridge_lock`). The
+/// watcher and verifier both use this so they agree on what a given source tx
+/// emits.
+#[must_use]
+pub fn extract_emission(program_id: ProgramId, instruction_data: &[u32]) -> Option<Emission> {
```
jonesmarvin8, line 43: Should this be verified with a signature to ensure this is in fact from the emitter program?

lez/programs/wrapped_token/core/src/lib.rs
```diff
@@ -0,0 +1,121 @@
...
+//! this by reading the authorized minter from a genesis-seeded config account.
+
+use lee_core::{
+    account::AccountId,
+    program::{PdaSeed, ProgramId},
+};
+use serde::{Deserialize, Serialize};
+
+const CONFIG_SEED_DOMAIN: [u8; 32] = *b"/LEZ/v0.3/WrappedTokenConfig/00/";
+const HOLDING_SEED_DOMAIN: [u8; 32] = *b"/LEZ/v0.3/WrappedTokenHold/00000";
```
jonesmarvin8, line 12: For `CONFIG_SEED_DOMAIN` you end with `/` but don't for `HOLDING_SEED_DOMAIN`. I think these should be consistent. Usually, we just pad with 0's, so seems that the `/` in `CONFIG_SEED_DOMAIN` is unnecessary.

lez/sequencer/core/src/config.rs
```diff
@@ -25,6 +26,11 @@ pub enum GenesisAction {
     SupplyBridgeAccount {
         balance: u128,
     },
+    /// Seeds a bridge-lock holder's initial bridgeable balance into genesis state.
+    SupplyBridgeLockHolding {
+        holder: AccountId,
+        amount: u128,
```
jonesmarvin8, line 32: Should we use `Balance` (or another alias data type) to avoid hardcoding types? I know Sergio prefers this for flexibility in the coding, and it makes for cleaner documentation for specs.

lez/indexer/core/src/lib.rs
moudyellaz: You are right on both counts. verify_block does not re-run the sending zone's transactions. It re-derives the delivery from the peer's finalized Bedrock block, byte-identical to what the peer's outbox recorded, and compares, it does not re-execute the source program. So for a direct single-hop peer we trust the peer's finalized block: the delivery came from the peer's own emitter tx in a Bedrock-final block, and the peer is an honest LEZ that verifies its own blocks by replay. That is sound for one hop.

The A to B to C case you describe is the real gap: A verifies B's block, but a message B relays may have originated at C, whom A does not verify, so A transitively trusts whoever B trusts. Handling that provenance chain, not just the immediate peer, is transitive trust, and I will handle it as part of the general layer. As you note it is not exploitable today, holdings are genesis-fixed and the demos are two-zone with direct peers only. A follow-up PR will also address the genesis and config points Arjentix raised, and the clarifying comment on this trust boundary (we trust the peer's finalized block, no source re-execution) will land with that follow-up.

lez/indexer/core/src/cross_zone_verifier.rs
moudyellaz: Rotation is tied to decentralized sequencing: while a zone has a single sequencer, one pinned key is sufficient, and key sets with rotation come in with decentralized sequencing. On skipping the signature: you are right that Bedrock's turn-based write authorization already makes the block-signature pin largely redundant, so it is kept optional and off by default. A comment noting the single-key limitation will land in the cleanup follow-up.

lez/indexer/core/src/cross_zone_verifier.rs
moudyellaz: Agreed, good catch. The inbox treats a replay as an idempotent no-op, so a replayed dispatch is a valid transaction on chain, but the verifier bails on a seen key, which does not match. The verifier's seen-set is redundant with the inbox's on-chain idempotency and it can halt an indexer on a legitimate replay, for example after a sequencer restart re-injects, since the watcher cursor is in memory. Fix is to continue on an already-seen key rather than bail, so the block is accepted and the inbox no-ops it, exactly as you say. I will make that change in the cleanup follow-up.

lez/programs/cross_zone_inbox/src/main.rs
moudyellaz: Same point as verifier.rs:125, the on-chain no-op is correct and the verifier should continue rather than bail on a seen key. Fixing in the cleanup follow-up.

lez/cross_zone/src/lib.rs
moudyellaz: Good catch, you have it right. ping_sender lets the emitter name any target_program_id and payload, so nothing in the ping program itself stops it from emitting a wrapped_token::Mint, and the inbox would deliver it as a chained call whose caller is the inbox, which is what wrapped_token accepts as its minter. What actually stops it is the inbox's per-peer allowed_targets allowlist: a peer can only invoke the targets it is allow-listed for, and in the demos the ping peer allows only ping_receiver, so it is not reachable as configured. You are right that the program-level authorization is loose and the safety rests on the allowlist being tight. The proper hardening is for a value-bearing target like wrapped_token to verify the source of the message, that it came from a bridge_lock lock, not just that the caller is the inbox. That belongs in the general layer.

lez/indexer/core/src/lib.rs
agureev: Fair enough, I did not know that execution-level trust is assumed. This resolves my comment.

lez/cross_zone/src/lib.rs
agureev: Hm, but suppose that we are doing the bridge demo, in the bridge demo, the allowed target is the `wrapped_token_id`. However, the zones all are configured with the same `extract_emission` function, which recognizes the `ping_sender`. In other words, even for the purposes of the demo where I only allow the wrapped token as a target, I can permissionlessly deploy a ping sender to Zone A and send a `Mint` message through it to the `wrapped_token_id` via ` ping_sender::Send{ target_program_id: wrapped_token, payload: <Mint{...}> }`, no? The watcher will recognize the program through `extract_emission` and relay the dispatch. The allowed targets are satisfied, the sender is not checked afterwards iiuc based on your comment and the code. So this allows - even for the demo - to maliciously mint tokens without ever registering the ping sender as a receiver.

Hopefully I'm getting everything correctly here. As you already mention, the proper hardening of source-verification would resolve this. But seems as things stand, this is a possible attack during the demo.

If we recognize this as an attack surface to be fixed in the future that is reasonable for the demo, this is perfectly fine, just want to be explicit about the attack scenario.

program_methods/guest/src/bin/bridge_lock.rs
Arjentix, line 1: Binaries from this directory are old-style. After my refactor program's main code should be located in `programs/your_program/src/main.rs`. That way they're close to their appropriate `lib.rs`

lez/cross_zone/src/lib.rs
moudyellaz: oh you're right, i somehow missed that! when i answered i was only thinking of the ping demo where the allowed target is harmless. so it is demo-grade and not production-safe, and the fix is source-verification: a value-bearing target like wrapped_token must verify the message originated from bridge_lock, not just that the caller is the inbox, which needs the source program id threaded into the dispatch and checked at the target. I have captured the attack and the fix in #611 (general-layer, out of scope for the cleanup itself), and I will mark the bridge demo as not production-safe in the cleanup PR.

agureev (approved):
Approved given all the issues mentioned have been recorded in #611 to be further developed after the initial merge

Thank you for addressing all the questions!

lez/sequencer/core/src/lib.rs
```diff
@@ -609,14 +621,18 @@ fn build_genesis_state(config: &SequencerConfig) -> (lee::V03State, Vec<LeeTrans
...
-            } => build_supply_account_genesis_transaction(account_id, *balance),
+            } => Some(build_supply_account_genesis_transaction(
+                account_id, *balance,
+            )),
             GenesisAction::SupplyBridgeAccount { balance } => {
-                build_supply_bridge_account_genesis_transaction(*balance)
+                Some(build_supply_bridge_account_genesis_transaction(*balance))
             }
+            // Force-inserted below: bridge_lock has no mint transaction.
+            GenesisAction::SupplyBridgeLockHolding { .. } => None,
```
Arjentix, line 654: Why? Maybe let's do it with transaction? So that the whole state is visible in block history

Arjentix (approved):
LGTM considering there will be future work on this.

From my side let's just fix [this comment](https://github.com/logos-blockchain/logos-execution-zone/pull/542#discussion_r3559937964) in this PR, so that artifacts won't break

program_methods/guest/src/bin/bridge_lock.rs
moudyellaz: Right, these are stale. program_methods was dropped in your refactor and is not on dev, but somehow they were carried alongside the new-style mains under lez/programs/<name>/src/main.rs. I will delete the whole directory in this PR.

lez/cross_zone/src/lib.rs
moudyellaz: Authenticity is at the block level, not per emission: the tx comes from the peer's finalized block (validated and executed, and program_id = ping_sender means ping_sender ran), and rederive checks the block is signed by the peer's pinned key (cross_zone_verifier.rs:178). A per-emission signature would be redundant.

lez/programs/wrapped_token/core/src/lib.rs
moudyellaz: Fair, they are inconsistent. Both are fixed 32-byte domains, but this constant derives the config PDA in the guest, so changing it moves the config account id and the genesis account and regenerates the wrapped_token image id. Since it is cosmetic, I will normalize it in the deferred artifact-rebuild follow-up tracked in #611 rather than trigger a rebuild just for this.

lez/sequencer/core/src/config.rs
moudyellaz: Makes sense, Balance (lee_core, = u128) exists. I will switch the genesis amount and balance fields to Balance in the cleanup PR (#611).

program_methods/guest/src/bin/bridge_lock.rs
moudyellaz: handled [here](c9b6df2be90239219da946bb74493f4ffedce9a7)

PR #545, opened 2026-06-18
refactor(privacy_preserving_circuit): introduce helper functions to shorten long functions

jonesmarvin8:
## 🎯 Purpose

This PR introduces helper functions to shorten long functions in `privacy-preserving-circuit`; without sacrificing readability.

## ⚙️ Approach

Simple refactor.

## 🧪 How to Test

No change in functionality. All previous tests should pass.

## 🔗 Dependencies

N/A

## 🔜 Future Work

N/A

## 📋 PR Completion Checklist

- [X] Complete PR description
- [X] Implement the core functionality
- [X] Add/update tests
- [X] Add/update documentation and inline comments

Issue #135, opened 2025-10-17
Improve dummy hash mechanism for private account initialization


schouhy:
Future Work from PR #121 

The dummy hash to anchor private account initializing nullifiers to a known existing commitment root mechanism can be improved. The circuit should allow users to inject any (existing) commitment set root to make transactions that initialize private accounts look indistinguishable from private account updating transactions. Otherwise the former can be identified by those having the dummy hash as commitment set root.

moudyellaz:
Resolved by #545. The hardcoded `DUMMY_COMMITMENT_HASH` anchor was replaced with an injected `commitment_root`, so init no longer relies on the dummy-hash mechanism this issue targeted.

PR #546, opened 2026-06-18
feat: update commitment mechanism for new private account

jonesmarvin8:
## 🎯 Purpose

Addresses [Issue 135](https://github.com/logos-blockchain/logos-execution-zone/issues/135). Updates initialization of nullifier mechanism to allow for any `CommitmentSet` root paired with initial nullifiers. This removes the potential of LEZ observers to identify transactions that initializes a new private account.

## ⚙️ Approach

- [X] Update `InputAccountIdentity` (for relevant entries) to include `Option<MembershipProof>`
- [X] Update `privacy-preserving-circuit` logic for initializing nullifier. (Any valid CommitmentSetDigest works)
- [X] Update previous tests to include `commitment_root = None`
- [X] Update wallet logic to provide the current CommitmentSetDigest.

## 🧪 How to Test

All previous tests should work. Added tests: `init_with_dummy_commitment_root_produces_valid_root` and `init_nullifier_digest_is_bound_to_commitment_root`

## 🔗 Dependencies

Builds off of [PR 545](https://github.com/logos-blockchain/logos-execution-zone/pull/545)'s refactoring of privacy preserving circuit.

## 🔜 Future Work

None

## 📋 PR Completion Checklist

- [X] Complete PR description
- [X] Add/update tests


program_methods/guest/src/bin/privacy_preserving_circuit/output.rs
```diff
@@ -55,7 +63,7 @@ fn handle_private_authorized_init(
         "Found new private account with non default values"
     );
 
-    let (new_nullifier, new_nonce) = init_nullifier_and_nonce(&account_id);
+    let (new_nullifier, new_nonce) = init_nullifier_and_nonce(&account_id, membership_proof);
```
agureev: if this is specifically for init nullifiers (i.e. this is not an abstracted API over all kinds of accounts), why supply a membership proof and not explicitly a root?

seems that there is very little to gain from supplying an actual proof instead of just a root assuming the zk property for the STARK (i.e. it is not revealed whether the witness data contained just a root or a full path)

agureev: i.e. why not just replace all the proof occurences by the merkle tree root in  the init witness types?

jonesmarvin8: I considered sending just the root initially. Though, I ultimately decided from a conceptual level (to understand the circuit), it made sense to pass a Merkle proof. This ensures that the logic flow appears similar for initialized and uninitialized paths.

I'm not sure which approach would be more preferable though: (current) Merkle path across all versions vs efficient proving initialized by only passing the root. Both approaches rely on the sequencer to catch bad roots.

jonesmarvin8: @schouhy Do you have a preference in approaches?

Plausibly, providing Merkle path allows for additional abstraction. I'm not 100% sure about this.

schouhy: I agree with @agureev on this one. I think it's simpler to just pass the merkle tree root. We can modify it in the future if we need so

jonesmarvin8: Made the requested change.

program_methods/guest/src/bin/privacy_preserving_circuit/output.rs
```diff
@@ -11,11 +11,9 @@ use crate::execution_state::ExecutionState;
 
 fn init_nullifier_and_nonce(
     account_id: &AccountId,
-    membership_proof: Option<&MembershipProof>,
+    commitment_root: Option<CommitmentSetDigest>,
 ) -> ((Nullifier, CommitmentSetDigest), Nonce) {
-    let set_digest = membership_proof.map_or(DUMMY_COMMITMENT_HASH, |proof| {
-        compute_digest_for_path(&DUMMY_COMMITMENT, proof)
-    });
+    let set_digest = commitment_root.unwrap_or(DUMMY_COMMITMENT_HASH);
```
schouhy: I think it would be cleaner to have the circuit take as input a `CommitmentSetDigest` (instead of Option) and default to dummy commitment in the callers end. So that this fallback decision is a respnsibility of the caller and not baked into the circuit logic

schouhy (changes requested):
hey thanks for the PR! I'd like to request a change.

program_methods/guest/src/bin/privacy_preserving_circuit/output.rs
jonesmarvin8: Fixed. Thanks.

lez/wallet/src/lib.rs
```diff
@@ -508,6 +509,14 @@ impl WalletCore {
         }
     }
 
+    pub async fn get_commitment_root(&self) -> Result<Option<CommitmentSetDigest>> {
+        let proof = self
+            .sequencer_client
+            .get_proof_for_commitment(DUMMY_COMMITMENT)
```
agureev, line 515: This makes me think that this PR would justify setting up an RPC endpoint for getting the latest root, or maybe even attacking it to the messages a wallet listens to. This can be done in a separate PR

agureev (approved):
looks good! left a general comment that does not specifically relate to this PR

schouhy (approved):
lgtm thanks

PR #547, opened 2026-06-19
refactor(indexer)!: query IndexerCore directly in the FFI; fix #538 / #540 / #544, drop `port`

### LINKED ISSUE #538: Add header guard to `indexer_ffi.h`
Unlike `wallet_ffi.h` which has the following header guard:

https://github.com/logos-blockchain/logos-execution-zone/blob/62d9ba10f8f86db3a1f04b329a1bd9d5b893bf60/lez/wallet-ffi/wallet_ffi.h#L24-L25

the `indexer_ffi.h` does not have a header guard, which causes the consuming code to implement it manually. We can do a similar old-school header guard, or add a `#pragma once`.

### LINKED ISSUE #540: Add `init_logger` FFI functions for modules
For instance, `indexer_ffi` does not allow any way to enable the underlying `env_logger`. An exposed function to do that (either respecting `RUST_LOG` env or via a passed parameter) can help greatly for debugging purposes.

### LINKED ISSUE #544: Indexer FFI leaks the outer `Box` of every `query_*` result
The indexer FFI (`indexer_ffi.h` / the `indexer` package) returns each query result as a `PointerResult_X__OperationStatus` whose `.value` is a raw pointer created with `Box::into_raw`. The matching `free_ffi_*` functions take the payload **by value**, so they free the struct's *inner* allocations (Vecs, nested boxes) but never the *outer* box that `.value` points to. No export frees that outer box, and a C++ `free()` on it would be cross-allocator UB — so every query leaks a small fixed-size wrapper.

## Affected exports

`query_account`, `query_block`, `query_block_by_hash`, `query_transaction`, `query_block_vec`, `query_transactions_by_account`, `query_last_block` — and their `free_ffi_*` counterparts. (`start_indexer` / `stop_indexer` are fine; that handle is long-lived and explicitly stopped.)

## Impact

~8–32 bytes leaked per query (a boxed `u64` for `query_last_block`). Negligible per call, but unbounded for a long-running indexer that serves continuous queries (e.g. the explorer polls `query_last_block` every 2s and refetches blocks), so it grows without bound over time.

## Repro / evidence

Consumer side, where the leak was found and worked around: `logos-execution-zone-indexer-module/src/lez_indexer_module_impl.cpp` — every `getX()` calls `free_ffi_*(*res.value)` (by value) and deliberately leaks `res.value` (see the comment at the top of the "Indexer Queries" section). The wrapper has no safe way to free the outer box.

## Suggested fix (any one)

1. Change the `free_ffi_*` functions to take the **pointer** (`X*`) instead of the struct by value, and reconstruct + drop the whole allocation inside: `drop(Box::from_raw(ptr))` (which also drops the inner fields). The C++ then calls `free_ffi_account(res.value)` and the entire allocation is reclaimed. *(Cleanest — one call frees both levels, no extra exports.)*
2. Or add `free_ffi_*_ptr(X*)` / `free_ffi_*_result(...)` exports that free the outer box (after the existing by-value inner free), so consumers have a defined way to release it.
3. For the fixed-size, no-inner-allocation cases (e.g. `query_last_block`'s `u64`), return the value inline in `PointerResult.value` rather than boxing it, avoiding the outer allocation entirely.

> [!TIP]
>
> Option 1 is preferred — it makes ownership unambiguous (the free owns the pointer) and removes the by-value/by-pointer mismatch that caused this.

erhant:
## 🎯 Purpose

The indexer FFI (`lez/indexer/ffi`) carried several defects and unnecessary machinery. 
- It exposed no header guard (#538), no way to turn on logging (#540), and leaked the outer `PointerResult` box on every `query_*` call (#544). Now closes #538, closes #540, closes #544.
- It also spun up a full jsonrpsee RPC server and connected an in-process WebSocket client back to *itself* just to answer queries (cc. @Arjentix this is what you mentioned at https://github.com/logos-blockchain/lez-indexer-module/pull/8#pullrequestreview-4516222275 I believe) — which is what forced the meaningless `port` argument on `start_indexer`.
- Instead of waiting hardcoded 6 minutes, integration FFI tests poll for completion, drastically reducing the test time.

This PR fixes the three issues and removes the self-RPC entirely, so the FFI talks to `IndexerCore`/`IndexerStore` directly.
 
BREAKING CHANGE: downstream consumers must update (see Future Work).

## ⚙️ Approach

- [x] Fix #538 — emit `#pragma once` (`build.rs` `with_pragma_once(true)`); delete the dead, never-read `cbindgen.toml`.
- [x] Add #540 — `init_logger(string)` extern fn that takes a log-level for indexer Rust logs.
- [x] Fix #544 — the consumer-facing `free_ffi_*` (`account`, `block_opt`, `block_vec`, `transaction_opt`, `transaction_vec`) now take `*mut T` and reclaim the *outer* `Box`; `query_last_block` returns an inline `LastBlockIdResult` (no allocation → no free).
- [x] Remove the internal RPC server/client — `IndexerServiceFFI` now holds an `IndexerCore` + a background task that drains the block stream (drives ingestion); `query_*` call `store.*` directly. Deleted `client.rs` + `api/client.rs`; dropped the `indexer_service` / `indexer_service_rpc` / `jsonrpsee` / `url` / `anyhow` deps, added `indexer_core` / `futures`.
- [x] Drop the legacy `port` from `start_indexer`; stash the runtime `Handle` in the indexer so `query_*` no longer take a `runtime` argument either.
- [x] Update the `integration_tests` FFI helper and the two FFI test targets for the new signatures.
- [x] Added a `wait_for_block`-type of helper to integration FFI tests, now takes 160s instead of 460s.

`run_server` / `IndexerHandle` in `lez/indexer/service` are intentionally **unchanged** — the standalone `indexer_service` binary (port 8779) still serves RPC.

## 🧪 How to Test

Run the related tests:

```sh 
RISC0_DEV_MODE=1 cargo nextest run -p integration_tests -E 'binary(indexer_test_run_ffi) | binary(indexer_ffi_block_batching)'
```

You can also inspect the regenerated `indexer_ffi.h`: confirm `#pragma once`, `init_logger`, inline `LastBlockIdResult`, and that `query_*` / `start_indexer` carry no `runtime` / `port`.

## 🔗 Dependencies

None.

## 🔜 Future Work

- Downstream consumers must move to the new ABI: `lez_indexer_module` (drop `port`, switch frees to pointer-taking, wire up `init_logger`) and then `lez-explorer-ui`. Tracked as follow-up PRs.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


Pravdyvy (approved):
lgtm

schouhy (approved):
lgtm thanks!

lez/indexer/core/src/lib.rs
```diff
@@ -10,21 +13,30 @@ use logos_blockchain_zone_sdk::{
...
 pub mod config;
+pub mod status;
 
 #[derive(Clone)]
 pub struct IndexerCore {
     pub zone_indexer: Arc<ZoneIndexer<NodeHttpClient>>,
     pub config: IndexerConfig,
     pub store: IndexerStore,
+    /// Live ingestion status; updated by the ingest stream, read by `status`.
+    pub status: Arc<Mutex<IndexerStatus>>,
```
Arjentix: Judging by usage consider [`arc_swap`](https://crates.io/crates/arc-swap)

lez/indexer/core/src/lib.rs
```diff
@@ -37,9 +49,31 @@ impl IndexerCore {
...
+    /// store, so callers (FFI/RPC) can tell "catching up" from "failed".
+    #[must_use]
+    pub fn status(&self) -> IndexerStatus {
+        let mut snapshot = self
+            .status
+            .lock()
+            .expect("indexer status mutex poisoned")
+            .clone();
+        snapshot.indexed_block_id = self.store.get_last_block_id().ok().flatten();
+        snapshot
```
Arjentix: Looks a bit strange imo. If `indexed_block_id` in `self.status` is not tracked, then I propose to split it into different structures. So that we have status stored in self, when we add indexed_block_id to it, not inside it, and return that "big" structure with status and block_id

lez/indexer/core/src/lib.rs
```diff
@@ -37,9 +49,31 @@ impl IndexerCore {
...
+    }
+
+    /// Apply a short, non-blocking update to the shared status.
+    fn set_status(&self, update: impl FnOnce(&mut IndexerStatus)) {
+        let mut status = self.status.lock().expect("indexer status mutex poisoned");
+        update(&mut status);
+    }
```
Arjentix: A bit scary function, because user can unintentionally lock our indexer with very long `update` function. Right now I see no sense in reading current status for such an `update()` function, so I'd accept status directly and just do a quick replace here

lez/indexer/core/src/lib.rs
```diff
@@ -105,7 +159,8 @@ impl IndexerCore {
                     yield Ok(block);
                 }
 
-                // Stream ended (caught up to LIB). Sleep then poll again.
+                // Stream drained: caught up to LIB as of this cycle. Sleep then poll again.
+                self.set_status(|s| s.state = SyncState::CaughtUp);
```
Arjentix: Looks like you forgot to set error to `None` here just in case.

As I proposed above I suggest to pass a very simple type in `set_status()`. This type could just have status name and optional error and there you can provide constructors like `cautch_up()`. So that there is no way you accidentally leave error set

lez/indexer/service/src/main.rs
```diff
@@ -12,6 +12,9 @@ struct Args {
     config_path: PathBuf,
     #[clap(short, long, default_value = "8779")]
     port: u16,
+    /// Directory under which the indexer stores its `RocksDB` state.
+    #[clap(short, long, default_value = ".")]
+    data_dir: PathBuf,
```
Arjentix, line 17: Hmm, I think you need to add this flag in `Dockerfile`. So that it points to the same location as before changes, it was something like `/var/lib/indexer_service/`.
Don't bother with building images btw, they will fail because of the other reason, I'm addressing it right now.

lez/indexer/ffi/Cargo.toml
```diff
@@ -6,15 +6,17 @@ version = "0.1.0"
...
+indexer_service_protocol = { workspace = true, features = ["convert"] }
 
-url.workspace = true
+env_logger.workspace = true
 log = { workspace = true }
 tokio = { features = ["rt-multi-thread"], workspace = true }
-jsonrpsee.workspace = true
-anyhow.workspace = true
+futures.workspace = true
+# Serializes the indexer status snapshot to JSON for `query_status`.
```
Arjentix: Don't think comments here are necessary

lez/indexer/ffi/src/api/types/account.rs
```diff
@@ -111,9 +116,15 @@ impl From<&FfiAccount> for indexer_service_protocol::Account {
...
+pub unsafe extern "C" fn free_ffi_account(val: *mut FfiAccount) {
+    if val.is_null() {
+        log::error!("Trying to free a null pointer. Exiting");
+        return;
+    }
+    // Reclaim the outer box, then convert to drop the inner data buffer.
+    let boxed = unsafe { Box::from_raw(val) };
+    let orig_val: indexer_service_protocol::Account = (*boxed).into();
     drop(orig_val);
 }
```
Arjentix, line 130: Hmm, looks a bit strange for me. Do we only return boxed accounts? It wasn't a thing before as I understand. Also looks like `FfiAccount` should be passable by value.

So to me looks like old code was right, I don't get this change to be honest

Arjentix: And same for other changes like this

lez/indexer/ffi/src/indexer.rs
```diff
@@ -1,95 +1,81 @@
...
+///   store stays populated)
+/// - The [`Runtime`] they run on. It owns the underlying tokio runtime when we created it (and
+///   drops it on teardown) and merely borrows it when the caller supplied one.
 #[repr(C)]
 pub struct IndexerServiceFFI {
-    indexer_handle: *mut c_void,
-    indexer_client: *mut c_void,
+    core: *mut c_void,
+    ingest_handle: *mut c_void,
+    runtime: *mut c_void,
```
Arjentix: I think `Runtime` should be ffi safe and no need for a pointer here

lez/indexer/ffi/Cargo.toml
erhant: aggreed

lez/indexer/service/src/main.rs
erhant: Yep, added the volume path as `data_dir`

lez/indexer/ffi/src/indexer.rs
erhant: You are right, my bad! It's much better now too.

erhant: https://github.com/logos-blockchain/logos-execution-zone/pull/547/commits/ab60d93431b239f4d1f817d65341ccda03933dc9

lez/indexer/ffi/Cargo.toml
erhant: https://github.com/logos-blockchain/logos-execution-zone/pull/547/commits/674df9516352e49b0161765783d832740aca0b11

lez/indexer/service/src/main.rs
erhant: https://github.com/logos-blockchain/logos-execution-zone/pull/547/commits/674df9516352e49b0161765783d832740aca0b11

erhant: https://github.com/logos-blockchain/logos-execution-zone/pull/547/commits/674df9516352e49b0161765783d832740aca0b11

lez/indexer/ffi/src/api/types/account.rs
erhant: These changes are about #544, and I think they are correct; let me provide an example of querying & freeing an account before/after:

## Before (leaky)

```c
PointerResult result = query_account(id);
// ^ result is returned by value, lives on stack
if (result.error != Ok) { /* handle */ }

FfiAccount acct = *result.value;
// use acct like `acct.program_id`


free_ffi_account(acct); 
// inner `acct.data` is freed, but the Box that held `acct`, which was
// result.value (the Box<FfiAccount>) is now orphaned, sizeof(FfiAccount) leaks
//
// "result" is dropped (was stack-allocated)
```

## After

```c
PointerResult result = query_account(id);
// ^ result is returned by value, lives on stack
if (result.error != Ok) { /* handle */ }

FfiAccount acct = *result.value;
// use acct like `acct.program_id`
// or keep result and use like `result.value->program_id`


free_ffi_account(result.value); // drops the entire struct
// on Rust side, Box::from_raw(val) reclaims the outer Box<FfiAccount> (pointed to by PointerResult.value)
// and then (*boxed).into() reclaims the ownership & converts to `Account`,
// dropped entirely with drop(orig_val)
//
// "result" is dropped (was stack-allocated)
```

---


So by-value receipt could only ever free the inner buffer and was structurally incapable of reclaiming the outer `Box`.

The same applies to other functions like  `_block_opt` / `_block_vec` etc. also returned boxed via `PointerResult`, but note that `free_ffi_block` / `free_ffi_transaction`  deliberately stay by-value as those are for values you already hold inline (e.g. an element out of an `FfiVec`) where no outer box exists.

lez/indexer/core/src/lib.rs
erhant: Indeed!

lez/indexer/core/src/lib.rs
erhant: Yeah this was overengineered, fixed.

lez/indexer/core/src/lib.rs
erhant: fixed 🐛

lez/indexer/core/src/lib.rs
erhant: done, i didn't know about that one!

lez/indexer/core/src/lib.rs
erhant: https://github.com/logos-blockchain/logos-execution-zone/pull/547/commits/edb4a4bf7d2ec9617ed4525b79286bf0bce7b284

lez/indexer/core/src/lib.rs
erhant: https://github.com/logos-blockchain/logos-execution-zone/pull/547/commits/edb4a4bf7d2ec9617ed4525b79286bf0bce7b284

lez/indexer/core/src/lib.rs
erhant: https://github.com/logos-blockchain/logos-execution-zone/pull/547/commits/edb4a4bf7d2ec9617ed4525b79286bf0bce7b284

lez/indexer/core/src/lib.rs
erhant: https://github.com/logos-blockchain/logos-execution-zone/pull/547/commits/edb4a4bf7d2ec9617ed4525b79286bf0bce7b284

lez/indexer/ffi/src/api/types/account.rs
Arjentix: Ah, I see, thanks for detailed explanation!

Arjentix (approved):
LGTM, thanks for addressing my comments!

Issue #548, opened 2026-06-19
Unbound VPK For Shared Accounts


agureev:
Each private account (including PDAs) currently fed directly the ss (as e.g.[here](https://github.com/logos-blockchain/logos-execution-zone/blob/62d9ba10f8f86db3a1f04b329a1bd9d5b893bf60/lee/state_machine/core/src/circuit_io.rs#L38)). That ss is then used for the ChaCha [encryption](https://github.com/logos-blockchain/logos-execution-zone/blob/62d9ba10f8f86db3a1f04b329a1bd9d5b893bf60/lee/state_machine/core/src/encryption/mod.rs#L85) of the account plaintext alongside the appropriate headers. The ciphertext is then written to journal.

To decrypt it, the epk needs to be provided. The epk is, similarly to the ss, is fed directly prer private account (as e.g. [here](https://github.com/logos-blockchain/logos-execution-zone/blob/62d9ba10f8f86db3a1f04b329a1bd9d5b893bf60/lee/state_machine/core/src/circuit_io.rs#L36)). It is then written [direcly](https://github.com/logos-blockchain/logos-execution-zone/blob/62d9ba10f8f86db3a1f04b329a1bd9d5b893bf60/program_methods/guest/src/bin/privacy_preserving_circuit/output.rs#L299) to the journal as-is.

This allows for two discoordination points:

1) The ss is not bound to the vpk of the owner of the account. Namely, the ss can be generated using a random public key. If that is the case, even if an account is owned by a user with the appropriate nullifier key, they would possibly not be able to discover that anything  was created for them and hence would never be able to spend the note.

2) The ss is not bound to the epk in the journal. Namely, the esk used for the generation of the ss may not be the secret key to  the public key fed in the witness. If that is the case, even if the ss is generated with the viewing key of the receiver, the receiver would never be able to decrypt the cipher if the sender puts random epk in the journal as they would lack the material for the ss.

Both points are usually acceptable in systems where each note is assumed to have a unique owner/actor. For example in the case of transfers, the sender malforming the ciphertext will drain their balance and - for all intents and purposes - burn their transfer. That is always a possibility and the responsibility of the user then to make sure that the payloads are well-formed.

However, our situation is different. We have a shared accounts design where several actors literally share the very same nullifier and viewing keypairs, which allows this scenario to become a vulnerability if a discoordination happens.

[...]

PR #550, opened 2026-06-19
feat!: viewing key and ciphertext binding

agureev:
## 🎯 Purpose

This fixes issue #548, allowing a malicious co-owner of a shared account to utilize the fact that the epk and the vpk are not bound to the account and ciphertext.

Similarly introduces some quality-of-life improvements like the view tag in-guest computation.

## ⚙️ Approach

- [x] expose ML KEM functionality in-guest
- [x] remove ss and epk as witness data
- [x] for every private account, supply the vpk and an esk in the witness instead
- [x] use the vpk as an additional argument in computing the account id, binding it
- [x] generate the ss and the epk in-guest
- [x] use the in-guest generated ss for encryption of the ciphertext
- [x] write the in-guest generated epk to journal
- [x] for PDA authorization, store the vpk alongside the npk in the appropriate  `pda_by_position` map
- [x] compute the view tag in-guest

## 🧪 How to Test

```
RISC0_DEV_MODE=1 cargo test -p lee --lib viewing_key
```
to run the two new tests

## 🔜 Future Work

Keccak speedup via risc0 circuit

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


erhant (approved):
lgtm, thanks!

Pravdyvy (approved):
lgtm

lee/state_machine/core/src/encryption/mod.rs
```diff
@@ -113,12 +115,12 @@ impl EncryptionScheme {
         commitment: &Commitment,
         output_index: u32,
     ) -> [u8; 32] {
-        let mut bytes = Vec::new();
-
-        bytes.extend_from_slice(b"LEE/v0.2/KDF-SHA256/");
-        bytes.extend_from_slice(&shared_secret.0);
-        bytes.extend_from_slice(&commitment.to_byte_array());
-        bytes.extend_from_slice(&output_index.to_le_bytes());
+        const PREFIX: &[u8; 20] = b"LEE/v0.2/KDF-SHA256/";
```
jonesmarvin8, line 144: Why change from `extend_from_slice`?

I was under the impression that we wanted to avoid hardcoding indices in case we need to change domain separator sizes or something.

lee/state_machine/core/src/encryption/shared_key_derivation.rs
```diff
@@ -94,9 +97,9 @@ impl SharedSecretKey {
     ) -> (Self, EphemeralPublicKey) {
         use risc0_zkvm::sha::{Impl, Sha256 as _};
 
-        let mut input = Vec::with_capacity(36);
-        input.extend_from_slice(message_hash);
-        input.extend_from_slice(&output_index.to_le_bytes());
+        let mut input = [0_u8; 32 + 4];
```
jonesmarvin8: Same as above. I'm unclear the change from `extend_from_slice` to `copy_from_slice`.

lee/state_machine/core/src/encryption/mod.rs
```diff
@@ -52,6 +52,18 @@ pub struct EncryptedAccountData {
     pub view_tag: ViewTag,
 }
 
+impl EncryptedAccountData {
+    #[must_use]
+    pub fn compute_view_tag(npk: &crate::NullifierPublicKey, vpk: &ViewingPublicKey) -> ViewTag {
+        const PREFIX: &[u8; 18] = b"/LEE/v0.3/ViewTag/";
+        let mut bytes = [0_u8; 18 + 32 + ViewingPublicKey::LEN];
+        bytes[0..18].copy_from_slice(PREFIX);
+        bytes[18..50].copy_from_slice(&npk.to_byte_array());
```
jonesmarvin8, line 87: In the refactoring for this function, why change from `extend_from_slice` to `copy_from_slice`? I'm happy to use either version...but I think we should be consistent.

@Arjentix Thoughts?

lee/state_machine/core/src/encryption/mod.rs
```diff
@@ -5,15 +5,15 @@ use chacha20::{
 };
 use risc0_zkvm::sha::{Impl, Sha256 as _};
 use serde::{Deserialize, Serialize};
-#[cfg(feature = "host")]
 pub use shared_key_derivation::{MlKem768EncapsulationKey, ViewingPublicKey};
```
jonesmarvin8, line 8: Originally, the feature for host was included due to some strange restriction on the MlKem crate. Has this been resolved? Just curious.

lee/state_machine/core/src/encryption/shared_key_derivation.rs
```diff
@@ -85,7 +89,6 @@ impl SharedSecretKey {
     /// avoid EPK collisions across multiple outputs in the same test.
     ///
     /// For production use [`Self::encapsulate`], which draws randomness from the OS.
-    #[cfg(any(test, feature = "test_utils"))]
```
jonesmarvin8, line 88: This was feature locked since it's only used for testing, and we want to make sure we do not accidentally use it in production code.

lee/state_machine/core/src/encryption/shared_key_derivation.rs
```diff
@@ -94,9 +97,9 @@ impl SharedSecretKey {
     ) -> (Self, EphemeralPublicKey) {
         use risc0_zkvm::sha::{Impl, Sha256 as _};
 
-        let mut input = Vec::with_capacity(36);
-        input.extend_from_slice(message_hash);
-        input.extend_from_slice(&output_index.to_le_bytes());
+        let mut input = [0_u8; 32 + 4];
+        input[0..32].copy_from_slice(message_hash);
```
moudyellaz: Now the production guest encapsulation path (output.rs:284 passes the secret esk into message_hash; wallet fills esk from OsRng). The doc says test-only and message_hash "public, unsuitable for real encryption," which is inverted. What do you think of renaming message_hash to esk, dropping the test-only/"public" wording, and stating that confidentiality requires a secret high-entropy esk?

lee/state_machine/core/src/encryption/shared_key_derivation.rs
moudyellaz: m = SHA256(esk || output_index) has no domain-separation prefix, unlike the KDF ("LEE/v0.2/KDF-SHA256/") and view tag ("/LEE/v0.3/ViewTag/"). No attack since esk is secret, but worth a prefix for consistency.

lee/privacy_preserving_circuit/src/output.rs
```diff
@@ -283,10 +280,14 @@ fn emit_private_output(
     post_with_updated_nonce.nonce = new_nonce;
 
     let commitment_post = Commitment::new(account_id, &post_with_updated_nonce);
+
+    let (shared_secret, epk) = SharedSecretKey::encapsulate_deterministic(vpk, esk, *output_index);
+    let view_tag = EncryptedAccountData::compute_view_tag(npk, vpk);
```
moudyellaz, line 297: view_tag is computed in-guest for every output including updates. Per #573 this becomes init-only once wallet sync changes; a one-line comment linking #573 here would mark it as intentional and tracked.

moudyellaz (approved):

lee/state_machine/core/src/encryption/mod.rs
agureev: so, the only reason i changed this to slices is simply to minimize cycles, i.e. to get away from heap-allocations in-guest

however, the wins here from this are miniscule, so if you want to, i can revert all the slice changes!

lee/state_machine/core/src/encryption/mod.rs
agureev: hm, i am sadly unsure! the only thing i can say is that everything compiles/builds, so seemingly all is fixed. the ci seems to also be ok with everything

lee/state_machine/core/src/encryption/shared_key_derivation.rs
agureev: hm but this is exactly the function we run now in-guest, no? i think we even discussed in the thread that we want to ungate it.

agureev: In particular, in-guest functionality cannot have access to OS entropy, so we need specifically this function

lee/state_machine/core/src/encryption/shared_key_derivation.rs
agureev: You're absolutely right about the documentation. Your suggestion is good, I'll add it on top.

lee/state_machine/core/src/encryption/shared_key_derivation.rs
jonesmarvin8: I may have missed the ungating this function in the thread.

After careful consideration. I agree that this function should be ungated, and we should get rid of the other `encapsulate` function since we aren't using it.

I'm not sure if OS entropy is ideal though. I wonder if we can use a PRF to generate the entropy we need; I'm exploring this is the PR for chacha20poly1305. So, this is not a blocker.

moudyellaz (approved):

schouhy (approved):
looks great thanks

lee/state_machine/core/src/encryption/mod.rs
Arjentix: Ah, agree that precomputing array length beforehand looks not so great. But maybe it's okay in such common code as keys and accounts go gain some grains of speed....

integration_tests/tests/account.rs
```diff
@@ -156,7 +156,11 @@ async fn import_private_account() -> Result<()> {
     let mut ctx = TestContext::new().await?;
 
     let key_chain = KeyChain::new_os_random();
-    let account_id = lee::AccountId::from((&key_chain.nullifier_public_key, 0));
+    let account_id = lee::AccountId::from((
+        &key_chain.nullifier_public_key,
+        &key_chain.viewing_public_key,
+        0,
+    ));
```
Arjentix, line 148: Ah, this wasn't introduced in current PR but we should finally remove this conversion from a tuple in favor of a proper constructor function

lee/privacy_preserving_circuit/src/execution_state.rs
```diff
@@ -43,12 +44,13 @@ pub struct ExecutionState {
...
+    /// Map from a private-PDA `pre_state`'s position in `account_identities` to the (npk, vpk,
     /// identifier) supplied for that position. Built once in `derive_from_outputs` by walking
-    /// `account_identities` and consulting `npk_if_private_pda`. Used later by the claim and
+    /// `account_identities` and consulting `npk_vpk_if_private_pda`. Used later by the claim and
     /// caller-seeds authorization paths to verify
-    /// `AccountId::for_private_pda(program_id, seed, npk, identifier) == pre_state.account_id`.
-    private_pda_npk_by_position: HashMap<usize, (NullifierPublicKey, Identifier)>,
+    /// `AccountId::for_private_pda(program_id, seed, npk, vpk, identifier) ==
+    /// pre_state.account_id`.
+    private_pda_by_position: HashMap<usize, (NullifierPublicKey, ViewingPublicKey, Identifier)>,
```
Arjentix, line 53: I suggest putting this tuple into a separate structure and use it there as hashmap value

Arjentix (approved):
LGTM

integration_tests/tests/account.rs
agureev: We can continue the discussion on #584 and related threads

lee/privacy_preserving_circuit/src/execution_state.rs
agureev: We can continue the discussion on #584 and related threads

agureev:
Note: proof generation test on CI now takes >90 minutes given the changes

moudyellaz (approved):
lgtm, thanks for addressing my comments

jonesmarvin8 (approved):
lgtm

Issue #560, opened 2026-06-22
Remove `encrypted_private_post_states` and `new_commitments` position linking


agureev:
Currently the ciphertext data gets [pushed](https://github.com/logos-blockchain/logos-execution-zone/blob/2c122440ceec413b314dd6dab9fb818833376f75/program_methods/guest/src/bin/privacy_preserving_circuit/output.rs#L294-L301) alongside the commitments vector. While this does not reveal any plaintext information under usual assumptions, this does provide the linking of commitment to the ciphertext via the position, i.e. the i-th position of the commitment vector is the commitment to the plaintext of the i-th position of the encrypted post-states vector.

This currently can leak some information. E.g. the unbound data-length size allows to possibly differentiate between the guest programs run (some guest programs can allow only a limited-size data blobs) and hence decrease the anonymity pool for certain applications.

As a possible mediation, we can consider introducing either a user-supplied or deterministic in-guest shuffling of the encrypted private post-states vector.

In particular, either the user provides a separate witness data that tells exactly in what order do the encrypted data go, introducing explicit entropy, or we provide in-guest ordering based e.g. on the bytes-ordering of the ciphertexts.

Issue #561, opened 2026-06-22
Remove `new_nullifiers` and `new_commitments` position linking


agureev:
Currently any new nullifiers get [pushed](https://github.com/logos-blockchain/logos-execution-zone/blob/2c122440ceec413b314dd6dab9fb818833376f75/program_methods/guest/src/bin/privacy_preserving_circuit/output.rs#L280-L294) alongside the commitments. While this does not reveal any plaintext information under usual assumptions, this does provide the linking of commitment to the nullifier, not to the same nullifier as the plaintext, but does reveal that the particular commitment and a particular nullifier is linked by the same account id and e.g. leaks some information about the nonce.

As a possible mediation, we can consider introducing either a user-supplied or deterministic in-guest shuffling of the commitments vector.

In particular, either the user provides a separate witness data that tells exactly in what order do the commitments go into the journal, introducing explicit entropy, or we provide in-guest ordering based e.g. on the bytes-ordering of the commitments/nullifiers.

PR #564, opened 2026-06-22
feat!: obfuscate note and cipher order

agureev:
## 🎯 Purpose

This closes issues #560 and #561 by sorting the produced vectors via a standard sorting methods, for nullifier outputs comparing by the nullifiers and for encryption data by comparing the ciphers.

A particular attack this partially closes is a possibility of draining the anonymity pool (in the classical sense of trying to link a nullifier to a commitment) being decreased for a specific user getting to know a specific nullifier was used with program P. For example, if N uses `auth_transfer` then the anonymity pool of N decreases to only those commitments that appear in the first and second positions.

This, however, does not evidently help for this specific case in situations where there is only one private account or where all the accounts have the same viewing tags.

Generally, this approach does contain strictly less metadata than the current one.

## ⚙️ Approach

- [x] Expose byte accessors
- [x] Remove index from encryption
- [x] Replace commitment by nullifier for ChaCha KDF
- [x] Sort the note and encryption output vectors for PPC output
- [x] Change the wallet functionality with new decryption mechanism

## 🧪 How to Test

Check the changed `cicrcuit.rs` test, ensuring that the order is scrambled based on sorting.

## 🔗 Dependencies

#592, #598, #591, #585, #550

## 🔜 Future Work

- Reform `decode_insert`, currently very little is gained by it

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


lee/state_machine/core/src/encryption/mod.rs
```diff
@@ -114,39 +114,32 @@ impl EncryptionScheme {
...
         cipher.apply_keystream(buffer);
     }
 
-    fn kdf(
-        shared_secret: &SharedSecretKey,
-        commitment: &Commitment,
-        output_index: u32,
-    ) -> [u8; 32] {
+    fn kdf(shared_secret: &SharedSecretKey, nullifier: &Nullifier) -> [u8; 32] {
         const PREFIX: &[u8; 20] = b"LEE/v0.2/KDF-SHA256/";
```
moudyellaz: Since the KDF input changed incompatibly, could the domain prefix bump to `LEE/v0.3/KDF-SHA256/`? Old and new ciphertexts currently share a derivation domain, no collision today since the input lengths differ, but the explicit version bump matches the other v0.3 domains and keeps the separation intentional.

lez/wallet/src/storage/key_chain.rs
```diff
@@ -455,6 +451,21 @@ impl UserKeyChain {
         Some(new_nullifier)
     }
 
+    #[must_use]
+    pub fn locate_spend(&self, account_id: AccountId, message: &Message) -> Option<usize> {
```
moudyellaz, line 474: `locate_spend` derives the update nullifier from current local state, and the caller `continue`s silently on a miss. If sync advances the account between submission and decode (possible with the #585 nullifier watch), the self-note decode is skipped with no trace. Could a miss log a warning, or the nullifier be taken from the pre-state the tx was built from?

moudyellaz (approved):
lgtm, thanks. Left some comments. To confirm the compatibility assumption: pre-fork notes under the old commitment+index KDF won't decrypt on wallet restore, this assumes a chain reset like the rest of the stack, right?

lee/state_machine/core/src/encryption/mod.rs
agureev: Addressed in [d69071d](https://github.com/logos-blockchain/logos-execution-zone/pull/564/commits/d69071d1327411e3dfd408db1b79bd67a55e350f)

lez/wallet/src/storage/key_chain.rs
agureev: Should be addressed in [131ed6d](https://github.com/logos-blockchain/logos-execution-zone/pull/564/commits/131ed6d0255d3f9b2bfbab09efa18dd70fbab572)!

jonesmarvin8 (approved):
lgtm.

schouhy (approved):
lgtm thanks!

PR #569, opened 2026-06-23
fix(encryption): strengthen esk generation logic

jonesmarvin8:
## 🎯 Purpose

The post state for account states, resulting from a privacy transaction, is encrypted using Chacha20. The encryption key (shared secret) is the encapsulation of some 32-byte string using the account's `vpk`; this computation is performed in-guest (a change introduced in [PR 550](https://github.com/logos-blockchain/logos-execution-zone/pull/550)).

Current workflow:
- `esk` (`osRng`) is generated in host, and provided to the guest (`privacy_preserving_circuit`)
- `esk` is provided `encapsulate_deterministic` (with constant 0; was originally index)
- `encapsulate_deterministic` generates the `shared_secret` (used for as the encryption key).

Chacha20 is a one-time pad encryption scheme. Thus, it is crucial to ensure that `shared_secret` is not repeated.

Potential issues:
- `encapsulate_deterministic` relies on similarly named `RustCrypto` [function](https://docs.rs/ml-kem/latest/ml_kem/struct.EncapsulationKey.html#method.encapsulate_deterministic) that warns to only use if you know what you're doing. E.g., values provided should be random. A companion function `encapsulate` generates the randomness. However, in-guest cannot use this function due to reliance of `osRng`. Instead, we obtain rng values from host.
- `esk` is the only source of rng. This results in `shared_secret` collisions for the same `vpk`. E.g., a private transaction from two private accounts that use the same private account keys (`npk`, `vpk`) will be encrypted using the same `shared_secret`.
- Randomess (from `esk`) is entirely dependent on the user's system. If compromised, then encryption can be compromised (for other user's accounts). Unclear whether this yields any downstream ramifications.

Mitigation:
- Introduce a per account `esk` (`EphemeralSecretKey: [u8; 32]`) which is dependent on system's rng and account.
  - `esk = Sha256(PREFIX || account_id || os_random || nonce)`
  - `PREFIX = b"/LEE/v0.3/esk"`
  - `account_id`: ensures uniqueness between separate private accounts (even if the accounts share keys).
  - `os_random`: rng from client machine
  - `nonce`: provides rng from account's current state.

Remarks:
- Collision resistant among unique private accounts (`account_id`) and different states of the same account (`nonce`).
- This approach is account index (in the transaction) independent.
- `os_random` provides rng for encrypting an initialized private account.
- `nonce` strengthens rng for encrypting an updated private account. Note: secure encryption is likely the least of worries for a compromised host machine that proves a private account update; e.g., `nsk` may be leaked.

## ⚙️ Approach

- Convert `EphemeralSecretKey` from alias to struct
- Simplify `encapsulate_deterministic` to be based solely on a given `esk`
- Strengthen logic for deriving esk (`new`) as described in Propose.

## 🧪 How to Test

All previous tests work; added tests `esk_is_deterministic`, `esk_differs_for_different_acount_id`, `esk_differs_for_different_os_random` and `esk_differs_for_different_nonce`.

## 🔗 Dependencies

Based on [PR 550](https://github.com/logos-blockchain/logos-execution-zone/pull/550)

## 🔜 Future Work

N/A

## 📋 PR Completion Checklist

- [X] Complete PR description
- [X] Add/update tests


integration_tests/tests/tps.rs
```diff
@@ -293,14 +293,14 @@ fn build_privacy_transaction() -> PrivacyPreservingTransaction {
         vec![
             InputAccountIdentity::PrivateAuthorizedUpdate {
                 vpk: sender_vpk,
-                esk: [0; 32],
+                os_random: [0; 32],
```
Pravdyvy: It is mainly about naming conventions, but `os_random` sounds like a function name for me. Maybe rename into `os_random_seed` or just `random_seed`?

Pravdyvy (approved):
lgtm, minor comment, feel free to ignore.

integration_tests/tests/tps.rs
jonesmarvin8: Renamed `os_random` to `random_seed`.

agureev (approved):
looks great! i think this is a direct improvement over the current design!

could you by change rebase this on top of [2a0595f](https://github.com/logos-blockchain/logos-execution-zone/pull/550/commits/2a0595fbd62fb10d0f7d94979a475ba6c1ba8ad4) with the regenrated elfs? let me know when you do and i will merge it to 550 straight away

PR #571, opened 2026-06-24
fix(nix): build risc0 Metal kernels on macOS by fixing xcrun env

0x-r4bbit:
risc0 compiles its Metal (GPU) prover kernels by invoking `xcrun metal` / `xcrun metallib`. Under nix, the darwin stdenv exports DEVELOPER_DIR and SDKROOT pointing at its own SDK, which makes xcrun look for the `metal` tool in the wrong place and fail with:

    error: cannot execute tool 'metal' due to missing Metal Toolchain;
           use: xcodebuild -downloadComponent MetalToolchain

...even when a working Metal Toolchain is installed (the same call succeeds in a clean environment where those vars are unset).

Wrap xcrun with a small shim, placed first in PATH, that clears DEVELOPER_DIR/SDKROOT for `metal`/`metallib` invocations only — so they resolve the system Xcode Metal Toolchain — while every other xcrun call passes through with the nix environment intact. This makes wallet-ffi build with real GPU prover kernels on macOS, with no --override-input workaround needed downstream.

Note: on recent macOS the Metal Toolchain is a per-user component, so `xcodebuild -downloadComponent MetalToolchain` must have been run by the building user (and builds still require `--option sandbox false`).

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -99,6 +99,28 @@
             sha256 = recursionZkrHash;
           };
 
+          # risc0 compiles its Metal (GPU) prover kernels by invoking
+          # `xcrun metal` / `xcrun metallib`. Under nix, the darwin stdenv sets
+          # DEVELOPER_DIR/SDKROOT to its own SDK, which makes `xcrun` look for
+          # the `metal` tool in the wrong place and fail with
+          #   error: cannot execute tool 'metal' due to missing Metal Toolchain
+          # even when a working Metal Toolchain is installed. This wrapper, put
+          # first in PATH, clears those two vars for metal/metallib invocations
+          # only — so they resolve the real system Xcode Metal Toolchain — while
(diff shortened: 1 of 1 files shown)


Pravdyvy (approved):
lgtm

moudyellaz (approved):
lgtm, thanks. Just one favor please could you pull the PR template when you open a PR against this repo? check [here](https://github.com/logos-blockchain/logos-execution-zone/blob/main/.github/pull_request_template.md)

0x-r4bbit:
@moudyellaz I've actually removed all of the template because I thought the provided information in the message is enough.

moudyellaz:
> @moudyellaz I've actually removed all of the template because I thought the provided information in the message is enough.

no worries but we use this template so it is more organized :)

PR #581, opened 2026-06-26
feat(indexer): recoverable invalid blocks

### LINKED ISSUE #579: Recovering from invalidated L2 blocks within the Indexer
The indexer replays the L2 blocks the sequencer inscribes on Bedrock (the L1/DA layer) to reconstruct LEZ state.

We had  introduced a hot-fix in #520 so the indexer would stop rejecting sequencer-generated deposit transactions that legitimately modify system accounts. We had also recently decided the indexer should **not** check whether the sequencer behaved well at the transaction level. Bedrock already validates block signatures, and the sequencer validates transaction contents. The indexer should therefore trust the inscriptions.

In the ingest loop any `put_block` failure is only logged, and the L1 cursor is then advanced **unconditionally** and the block is `yield`ed as `Ok`. We do not propagate what kind of error happened as well (i.e. did RocksDB fail, or was it an invalid state transition).

https://github.com/logos-blockchain/logos-execution-zone/blob/64f8444a67e5cc93fb755f9c9292034bed2a657d/lez/indexer/core/src/lib.rs#L143-L151

With the coming of decentralized sequencers; I think the handling here should be reconsidered. The two other alternatives to silently ignoring are:

- Stopping the node
- Recovering from a bad block

I would like to take a crack at "recovering" within this issue.

## Proposed Solution

I say we keep 2 tips (or heads)

- L1 tip, ALWAYS progresses as it does right now
- L2 tip, respects the hash-linked chain (via `BlockHeader`)

https://github.com/logos-blockchain/logos-execution-zone/blob/64f8444a67e5cc93fb755f9c9292034bed2a657d/lez/common/src/block.rs#L30-L37

The L2 shall validate the relationship of a new block w.r.t. its current tip; as long as the tip is valid, the new block shall comply with the chaining rules (`new.id => tip.id + 1`, `new.prev_block_hash == tip.hash` and such).

Now comes the state transition; as I said above, pre #520 we had some system account guards and such; let's imagine all of them removed, and we only apply the state transition function as shared among Sequencer and Indexer.

- If there are no issues on transition, all good, update the L2 `tip`.
- If there is an issue with this transition, it means that our current `tip` was the LAST valid block we have seen, and someone is being naughty and publishing invalid blocks. HOWEVER, other honest sequencers are aware of this too, and they too will be unable to progress their own state transitions; so when the honest turn comes, the new valid block proposal must have the parent block as same as our L2 `tip`.

So basically, we parked our indexer L2 tip right before the invalid block, and wait until someone else actually publishes a valid block.

### Handling the "parked" state

Our current error handling is more like `if let Err(err) = put_block(...) { error!(...) }` which simply logs no matter what went behind the scenes. We can come up with a more detailed error (something like below):

```rust
#[derive(Debug, Clone, Serialize, thiserror::Error)]
pub enum BlockIngestError {
    #[error("failed to deserialize L2 block: {0}")]
    Deserialize(String),
    #[error("unexpected block id: expected {expected}, got {got}")]
    UnexpectedBlockId { expected: u64, got: u64 },
    #[error("broken chain link: expected prev {expected_prev}, got {got_prev}")]
    BrokenChainLink { expected_prev: HashType, got_prev: HashType },
    #[error("block hash mismatch: computed {computed}, header {header}")]
    HashMismatch { computed: HashType, header: HashType },
    #[error("state transition failed: {0}")]
    StateTransition(String),   // wraps lee::error::LeeError
    #[error("storage error: {0}")]
    Storage(String),
}
```

Serializable so it can live inside the persisted breaker and the status snapshot.

---

We can also keep note of the block that broke the chain:

```rust
#[derive(Debug, Clone, Serialize, BorshSerialize, BorshDeserialize)]
pub struct ChainBreaker {
    pub block_id: u64,
    pub block_hash: HashType,
    pub prev_block_hash: HashType,
    pub l1_slot: <zone-sdk slot type>,
    pub error: BlockIngestError,
    pub first_seen: Timestamp,
    pub orphans_since: u64,
}
```

Stored in a RocksDB meta key (alongside the existing `meta_last_block_id` etc.).

This captures the **first** block that broke the chain; every subsequent non-chaining block (an orphan, since the tip is frozen) bumps `orphans_since` rather than overwriting. Survives restart. **Cleared** when the chain recovers (a block applies and advances the tip).

The deserialize-failure path (see snippet below) folds into this same breaker handling: a block we cannot even parse cannot chain, so it sets `BlockIngestError::Deserialize`, advances the L1 cursor, and freezes the tip like any other break.

https://github.com/logos-blockchain/logos-execution-zone/blob/64f8444a67e5cc93fb755f9c9292034bed2a657d/lez/indexer/core/src/lib.rs#L124-L136

---

`IndexerSyncState` (`lez/indexer/core/src/status.rs`) gains a variant:

```rust
pub enum IndexerSyncState {
    Starting,
    Syncing,
    CaughtUp,
    Error,    // transient cycle failure, e.g. L1 node unreachable
    Stalled,  // parked on a chain breaker; see the breaker details
}
```

`Stalled` is distinct from `Error`: `Error` is a recoverable cycle hiccup, `Stalled` is a frozen tip awaiting a valid continuation. The active `ChainBreaker` is embedded in the `IndexerStatus` snapshot already returned over FFI/RPC, so the existing `get_status` call answers "why aren't you progressing?".

---

I do have a bit more detailed spec in my hands, with more punctual "where to edit in code" type of details. But I did not want to clutter here with those. Regardless, I will open a branch and experiment on this idea, because I believe it will GREATLY save us trouble when it comes to decentralized sequencers.

erhant:
### Bad Block Handling

The indexer replays the L2 blocks the sequencer inscribes on Bedrock to reconstruct LEZ state. Two problems made that fragile:

1. **Silent divergence on a bad block.** Any `put_block` failure was only logged; the L1 cursor then advanced **unconditionally** and the block was yielded as `Ok`. A block that failed to apply was skipped and the indexer marched on — permanently diverging from the sequencer with no signal beyond a log line.
2. **Redundant re-validation.** The indexer re-ran transaction-level checks (the PR #520 system-account hot-fix path, per-tx stateless signature checks) that Bedrock (block signatures) and the sequencer (tx contents) already perform.

This PR makes the indexer **trust the inscriptions but verify chain structure, and park instead of silently skipping**: a bad block freezes the validated L2 tip (loud, observable, recoverable) rather than corrupting state. It also adds a **startup chain-identity check** so a stale store reused against a different/reset chain is caught at boot (and optionally auto-wiped) instead of diverging.

Closes #579.

### Stale Cache Handling

This PR also helps with the "stale RocksDB vs new chain" problem. On launch, the indexer verifies the channel still serves an **anchor block** it holds at that block's recorded L1 slot: the validated tip at the read cursor — or, when parked, the recorded parked block from `StallReason`. On positive evidence of a different chain the user can opt to exit immediately or reset the disk to the new state. This check re-uses the existing `Slot` information on disk, and is the most efficient way to work with bedrock for this purpose. 

An alternative approach would be to seek the 2nd block (the first different block, after the _deterministic_ genesis), but that requires `Slot` queries and these queries take many seconds with high CPU usage due to the greater slot difference between genesis and next block. L1 slots are wall-clock-derived, so the chain sits at slot ~14.5M immediately and inscriptions live near the tip of slot-space. A from-genesis scan to locate the block would walk millions of empty slots (~145k node block-fetches → a multi-second startup stall and a flood of Storage reads) on every boot. The tip cursor is always recent, so this reads ~one batch.

## ⚙️ Approach

Two validated positions replace the single conflated cursor: an always-advancing **L1 read cursor** and a **validated L2 tip** (`{block_id, hash}`) that advances only on a chained, atomically-applied block.

- [x] **Trust the inscriptions** — rename `execute_without_system_accounts_check_on_state` → `execute_on_state`, drop the indexer-side `transaction_stateless_check` + system-account guards from both replay paths (`block_store`, `calculate_state_for_id`). The sequencer's validation is untouched.
- [x] **Acceptance test per block** — block-id continuity + `prev_block_hash` linkage + recomputed-hash integrity (`Block::recompute_hash`). No signature/tx-content re-checks.
- [x] **Atomic apply** — `accept_block` applies to a scratch clone of state and commits in-memory only after the RocksDB write succeeds; a failed block leaves both in-memory state and the tip untouched. Removes the old non-atomic `put_block`.
- [x] **Park-not-skip** — on a non-chaining/failed block the ingest loop sets a new `Stalled` status, advances the L1 cursor only (tip frozen), and does not yield; infra errors set `Error` and re-poll without advancing. The tip unfreezes (and the stall clears) when a block whose `prev_block_hash == tip.hash` finally applies.
- [x] **Typed errors + persisted diagnostics** — `BlockIngestError` enum replaces the silent `if let Err {error!()}` swallow; a `StallReason` (first breaking block + `orphans_since` count + L1 slot + typed error) is persisted in RocksDB, survives restart, clears on recovery.
- [x] **Observability** — `IndexerSyncState::Stalled` + the `StallReason` are exposed on the status snapshot over FFI (`query_status`) and a new `getStatus` RPC method.
- [x] **Startup chain-identity check** (`chain_consistency.rs`) — `IndexerCore::new` anchors on a block the same chain must still serve at a known slot: the recorded parked block while stalled (id/hash/slot from the persisted `StallReason`), otherwise the validated tip at the read cursor (only blocks advance the cursor, so it is the tip's inscription slot). Two probes, both requiring **positive evidence**: a `channel_state` frontier check (channel missing, or channel tip slot behind the anchor slot) and a one-batch scan at the anchor slot (anchor present ⇒ same chain; same id with a different hash, or an id at/below the anchor re-served *past* the anchor slot — which catches the byte-identical re-inscribed genesis ⇒ reset). On a mismatch it refuses to start by default, or wipes + re-indexes when `allow_chain_reset` is set; anything inconclusive (unreachable L1, timeout, only newer blocks) proceeds without wiping. Wired into service + FFI startup.

## 🧪 How to Test

Unit + serialization coverage (no stack required):

```sh
RISC0_DEV_MODE=1 RISC0_SKIP_BUILD=1 cargo test -p common -p storage -p indexer_core
```

Covers: hash recompute, `BlockIngestError` round-trip, `StallReason` persist/clear/orphan-count, acceptance/park paths (bad id, hash mismatch, broken link, deserialize), **park-then-recover**, full-transfer apply + historical state, `Stalled` status serialization, `allow_chain_reset` default, and the chain-identity anchor probes (same-chain / hash-mismatch / re-inscribed-old-id / newer-id bail / frontier verdicts, plus network-failure-stays-inconclusive).

Integration (requires the full local stack):

```sh
RUST_LOG=info RISC0_DEV_MODE=1 cargo test -p integration_tests --test indexer_stall -- \
  indexer_status_rpc_reports_caught_up_with_no_stall --exact --nocapture
```

Manual (run an old `rocksdb-<channel_id>` store against a freshly reset Bedrock + sequencer on the same `channel_id`):
- **Chain reset, refuse (default):** with `allow_chain_reset` unset/false → refuses to start with a clear "holds a different chain … delete the indexer storage directory, point at a fresh one, or set `allow_chain_reset`" message.
- **Chain reset, auto-wipe:** set `"allow_chain_reset": true` → `WARN Chain reset detected (…)` naming the evidence (e.g. `channel re-serves block N <hash> at slot S past our anchor slot A`, or `channel tip slot T is behind our anchor slot A`), then re-indexes from genesis. Works even if the stale store was already parked.
- **Valid resume:** restart against the *same* chain → `Resuming indexer from cursor …` and continues with no wipe.
- **Status:** `getStatus` (RPC) / `query_status` (FFI) on a parked indexer reports `state: "stalled"` with a populated `stallReason`.

> [!TIP]
>
> You can fetch status with:
>
> ```sh
> curl -s http://localhost:8779 \
>   -H 'content-type: application/json' \
>   -d '{"jsonrpc":"2.0","id":1,"method":"getStatus","params":[]}' | jq
> ```

## 🔗 Dependencies

None.

## 🔜 Future Work

- **Integration-level park testing** needs fault-injection support in the harness; park/recover is currently unit-tested.
- **Single-sequencer caveat:** "recover" degrades to "stall forever" today (no honest 2nd sequencer to publish the continuation) — still strictly better than silent divergence; full recovery materializes under decentralized sequencers.
- **Chain-identity blind spot:** a store holding only the (deterministic) genesis block reports the identical re-inscribed genesis as consistent — harmless, since everything such a store holds is valid on the new chain too; the ingest loop simply continues on it.
- Chain-identity fetch timeout is a constant; could be config-driven.
- Minor test-assertion breadth polish noted during review.
- Sequencer state management should also take notice of valid / invalid blocks and comply with the assumption of this PR that an honest sequencer will eventually produce a block referring to the last valid block as its parent.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


lez/indexer/core/src/config.rs
```diff
@@ -20,6 +20,12 @@ pub struct IndexerConfig {
     pub consensus_info_polling_interval: Duration,
     pub bedrock_config: ClientConfig,
     pub channel_id: ChannelId,
+    /// Whether to wipe the indexer store and re-index from scratch when a genesis mismatch occurs
+    /// (i.e. the L1/sequencer was reset but the old store was reused).
+    ///
+    /// Defaults to `false`: on mismatch the indexer refuses to start.
```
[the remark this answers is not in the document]
erhant: fixed in https://github.com/logos-blockchain/logos-execution-zone/pull/581/commits/fce487f24168b3e2f29d1eb74152ca8e4bdbde36

lez/indexer/service/src/service.rs
```diff
@@ -149,6 +150,11 @@ impl indexer_service_rpc::RpcServer for IndexerService {
         Ok(tx_res)
     }
 
+    async fn get_status(&self) -> Result<serde_json::Value, ErrorObjectOwned> {
+        Ok(serde_json::to_value(self.indexer.status())
+            .expect("IndexerStatus serialization should not fail"))
+    }
```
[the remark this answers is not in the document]
erhant: fixed in https://github.com/logos-blockchain/logos-execution-zone/pull/581/commits/fce487f24168b3e2f29d1eb74152ca8e4bdbde36

lez/indexer/core/src/block_store.rs
```diff
@@ -134,78 +164,227 @@ impl IndexerStore {
...
+            return Ok(AcceptOutcome::Parked(err));
+        }
+
+        let mut stored = block.clone();
+        stored.bedrock_status = BedrockStatus::Finalized;
+        if let Err(err) = self.dbio.put_block(&stored, [0_u8; 32]) {
+            let ingest_err = BlockIngestError::Storage(err.to_string());
+            self.record_stall(Some(&block.header), l1_slot, ingest_err.clone())?;
+            return Ok(AcceptOutcome::Parked(ingest_err));
+        }
```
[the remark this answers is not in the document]
erhant: indeed, this should not progress L1 as well; I changed it from being a "parkable" error, instead this function will return error.

https://github.com/logos-blockchain/logos-execution-zone/pull/581/commits/fce487f24168b3e2f29d1eb74152ca8e4bdbde36

erhant:
Rebased to `dev`

integration_tests/tests/bridge.rs
```diff
@@ -596,6 +536,38 @@ fn create_zone_indexer_observer(
     ))
 }
 
+/// Test that the indexer status RPC reports caught-up with no stall after a clean run.
+///
+/// TODO: Integration-level park testing (publishing a bad block to force a stall) is a follow-up
+/// needing fault injection support in the test harness.
+#[ignore = "requires the full local stack (bedrock + sequencer + indexer)"]
+#[test]
+async fn indexer_status_rpc_reports_caught_up_with_no_stall() -> anyhow::Result<()> {
```
Arjentix: Looks like this needs to be addressed in this PR. We have full local stack in `TestContext`

lez/common/src/block.rs
```diff
@@ -55,6 +55,21 @@ pub struct Block {
     pub bedrock_status: BedrockStatus,
 }
 
+impl Block {
+    /// Recomputes the hash from this block's contents, for integrity verification
+    /// against the value stored in `header.hash`.
+    #[must_use]
+    pub fn recompute_hash(&self) -> BlockHash {
```
Arjentix, line 62: Indeed this means we can have an invalid `Block` with hash not corresponding to the data. Maybe we could make `Block` fields private and perform this check on construction and deserialization?

Sounds hard, but just an idea

lez/storage/src/indexer/mod.rs
```diff
@@ -202,16 +204,7 @@ impl RocksDBIO {
...
-                        .transaction_stateless_check()
-                        .map_err(|err| {
-                            DbError::db_interaction_error(format!(
-                                "transaction pre check failed with err {err:?}"
-                            ))
-                        })?
-                        // FIXME: HOT FIX (testnet v0.2): does not check for system account updates due to
-                        // sequencer-generated deposit tx'es;
-                        // CHANGE ME back to `execute_check_on_state` when the indexer can authenticate deposit transactions
-                        .execute_without_system_accounts_check_on_state(
```
Arjentix: I was thinking about this and looks like we can't trust Bedrock as I thought initially.

There is nothing stopping malicious peer to submit invalid block to Bedrock when it's his turn. Even if we introduce threshold signatures it can still publish to Bedrock

lez/storage/src/indexer/indexer_cells.rs
```diff
@@ -247,6 +247,41 @@ impl SimpleWritableCell for ZoneSdkIndexerCursorCellRef<'_> {
     }
 }
 
+/// Opaque JSON bytes for the indexer's persisted `Option<StallReason>`.
+/// Serialized via `serde_json` by the caller (mirrors the zone-sdk cursor cell).
+#[derive(BorshDeserialize)]
+pub struct StallReasonCellOwned(pub Vec<u8>);
```
Arjentix, line 250: Hmm, "reason" sounds like it should be just `String`, no?

lez/indexer/service/rpc/src/lib.rs
```diff
@@ -69,6 +69,9 @@ pub trait Rpc {
         limit: u64,
     ) -> Result<Vec<Transaction>, ErrorObjectOwned>;
 
+    #[method(name = "getStatus")]
+    async fn get_status(&self) -> Result<serde_json::Value, ErrorObjectOwned>;
```
Arjentix: Let's better return something more structured than just `serde_json::Value`

lez/indexer/core/src/lib.rs
```diff
@@ -1,26 +1,37 @@
...
+pub mod stall_reason;
 pub mod status;
 
+/// Result of comparing the indexer's stored tip against the channel.
+enum ChainIdentityOutcome {
+    /// Proceed from the cursor: channel still serves our chain, nothing to compare,
+    /// or the check was inconclusive — none of which prove a reset.
+    Consistent,
+    /// The channel serves a different block at one of our ids — a chain reset.
+    Mismatch { detail: String },
```
Arjentix: Let's have a structured type here as well and implement `Display` if needed

lez/indexer/core/src/lib.rs
```diff
@@ -51,6 +62,106 @@ impl IndexerCore {
         })
     }
 
+    /// Builds the core, then verifies the stored chain matches the channel's by
+    /// re-reading the channel at the stored tip's position. On mismatch: refuse
+    /// (error) unless `allow_reset`, in which case wipe the store and re-index from
+    /// scratch. Used at service/FFI startup in place of `new`.
```
Arjentix:
```suggestion
    /// scratch.
```

lez/indexer/core/src/lib.rs
```diff
@@ -51,6 +62,106 @@ impl IndexerCore {
...
+    pub async fn new_with_chain_check(
```
Arjentix: Btw looks like it should just be `new()` now?

lez/indexer/core/src/lib.rs
```diff
@@ -51,6 +62,106 @@ impl IndexerCore {
...
+        config: IndexerConfig,
+        storage_dir: &Path,
+        allow_reset: bool,
```
Arjentix: As I understand `allow_reset` is also passed in `config`, so why pass it here as another parameter?

lez/indexer/core/src/lib.rs
```diff
@@ -51,6 +62,106 @@ impl IndexerCore {
...
+                    "Chain reset detected ({detail}). Wiping indexer store at {} and re-indexing.",
+                    home.display()
+                );
+                drop(core); // sole owner before the ingest task is spawned → closes the DB
+                storage::indexer::RocksDBIO::destroy(&home)?;
+                Self::new(config, storage_dir)
+            }
+            ChainIdentityOutcome::Mismatch { detail } => Err(anyhow::anyhow!(
+                "Indexer store at {} holds a different chain than the channel now serves \
+                 ({detail}). Run `just clean`, point at a fresh storage dir, or set \
```
Arjentix: Let's not mention `just clean`. It's our devtool and for example our infra guys will be surprised when they'l see that error on stage

lez/indexer/core/src/lib.rs
Arjentix: Also it looks a bit strange to me that we control that from Core struct... I suggest to always fail if consistency is corrupted and decide what to do on caller-site.

I mean you literally calls `new(), drop(), new()` for this struct. This looks like this code should be in `service`

Arjentix: And then this option will become service parameter, which we pass as cmd argument for now

lez/indexer/core/src/lib.rs
Arjentix, line 172: This function now becomes very long, can we split it into more functions?

lez/indexer/core/src/block_store.rs
```diff
@@ -134,78 +164,225 @@ impl IndexerStore {
...
-                        // sequencer-generated deposit tx'es;
-                        // CHANGE ME back to `execute_check_on_state` when the indexer can authenticate deposit transactions
-                        .execute_without_system_accounts_check_on_state(
-                            &mut state_guard,
-                            block.header.block_id,
-                            block.header.timestamp,
-                        )?;
+    /// The last successfully applied block as `{block_id, hash}`, or `None` before
+    /// any block is stored (cold start). Read fresh from the store each call.
+    fn validated_tip(&self) -> Result<Option<Tip>> {
```
Arjentix: I think to many info about structure internals, this info is easily deducible from types and at the same time requires support because we need to remember to update it once we change fields of `Tip`. So let's better be concise in doc comments

lez/indexer/core/src/block_store.rs
```diff
@@ -134,78 +164,225 @@ impl IndexerStore {
...
+        };
+        Ok(Some(Tip {
+            block_id,
+            hash: block.header.hash,
+        }))
+    }
+
+    /// Returns `Some(err)` if `block` is not the valid continuation of the tip:
+    /// hash integrity, then block-id continuity, then `prev_block_hash` linkage.
+    fn acceptance_error(&self, block: &Block) -> Result<Option<BlockIngestError>> {
```
Arjentix: Ah, I don't really like this reversed approach with "we'll return Some(err)". Can it be rewritten in the style of `Result<(), BlockIngestError>`? Like `validate_block()` or something more precise?

Arjentix: So that errors are not the main path, you know

lez/indexer/core/src/block_store.rs
```diff
@@ -134,78 +164,225 @@ impl IndexerStore {
...
+    }
+
+    /// Records a stall for an inscription that could not even be parsed.
+    pub fn record_deserialize_stall(
+        &self,
+        l1_slot: serde_json::Value,
+        error: String,
+    ) -> Result<()> {
+        self.record_stall(None, l1_slot, BlockIngestError::Deserialize(error))
+    }
```
Arjentix: I don't know, something is wrong here. Public function which accepts slot as non-structured json value... And error also could be a more stronger type

lez/indexer/core/src/block_store.rs
```diff
@@ -134,78 +164,225 @@ impl IndexerStore {
...
+        }
+
+        let mut stored = block.clone();
+        stored.bedrock_status = BedrockStatus::Finalized;
+        self.dbio
+            .put_block(&stored, [0_u8; 32])
+            .context("Failed to persist accepted block")?;
+
+        // Commit in-memory state (infallible) only after the DB write succeeded.
+        *self.current_state.write().await = scratch;
```
Arjentix, line 277: Hmmmm, arcswap? Maybe, not sure

lez/indexer/core/src/ingest_error.rs
```diff
@@ -0,0 +1,57 @@
+use common::HashType;
+use serde::{Deserialize, Serialize};
+
+/// Why the indexer could not apply an L2 block from the channel. Stored inside a
+/// [`crate::stall_reason::StallReason`] and surfaced on the status snapshot.
+#[derive(Debug, Clone, Serialize, Deserialize, thiserror::Error)]
+#[serde(rename_all = "camelCase")]
+pub enum BlockIngestError {
+    #[error("failed to deserialize L2 block: {0}")]
+    Deserialize(String),
```
Arjentix: Should it be serde or borsh error here as internal?

lez/indexer/core/src/ingest_error.rs
```diff
@@ -0,0 +1,57 @@
...
+        expected_prev: HashType,
+        got_prev: HashType,
+    },
+    #[error("block hash mismatch: computed {computed}, header {header}")]
+    HashMismatch {
+        computed: HashType,
+        header: HashType,
+    },
+    #[error("state transition failed: {0}")]
+    StateTransition(String),
```
Arjentix: Let's add more structured error here as well

lez/indexer/core/src/ingest_error.rs
```diff
@@ -0,0 +1,57 @@
...
+                got: 7
+            }
+        ));
+    }
+
+    #[test]
+    fn display_is_human_readable() {
+        let err = BlockIngestError::StateTransition("nonce too low".to_owned());
+        assert_eq!(err.to_string(), "state transition failed: nonce too low");
+    }
```
Arjentix: Currently it's okay because you don't store other `Error`s inside, but when you have errors chained that won't just work. For more details see [my old comment](https://github.com/logos-blockchain/logos-execution-zone/pull/433#discussion_r3053598524)

lez/indexer/core/src/stall_reason.rs
```diff
@@ -0,0 +1,25 @@
...
+
+use crate::ingest_error::BlockIngestError;
+
+/// Diagnostic record of the first block that broke the L2 chain.
+///
+/// The block-derived fields are `None` for a deserialize break (no header was
+/// ever parsed). `l1_slot` is the zone-sdk cursor position captured as JSON.
+/// `first_seen` is the breaking block's L2 timestamp (`None` for a deserialize break).
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
```
Arjentix: Why btw?

lez/indexer/core/src/stall_reason.rs
```diff
@@ -0,0 +1,25 @@
...
+pub struct StallReason {
+    pub block_id: Option<u64>,
+    pub block_hash: Option<HashType>,
+    pub prev_block_hash: Option<HashType>,
+    pub l1_slot: serde_json::Value,
```
Arjentix: Can we have something better than json value?

lez/indexer/service/src/service.rs
```diff
@@ -149,6 +150,16 @@ impl indexer_service_rpc::RpcServer for IndexerService {
         Ok(tx_res)
     }
 
+    async fn get_status(&self) -> Result<serde_json::Value, ErrorObjectOwned> {
+        serde_json::to_value(self.indexer.status()).map_err(|err| {
```
Arjentix: See, you're doing `jsonrpsee` work here, so let's just return defined structure from this fn and that's it

lez/indexer/core/src/lib.rs
```diff
@@ -135,29 +252,113 @@ impl IndexerCore {
...
+                            error!(
+                                "Store error applying block {}: {err:#}",
+                                block.header.block_id
+                            );
+                            self.set_status(IndexerSyncStatus::error(format!(
+                                "store error: {err:#}"
+                            )));
+                            had_cycle_error = true;
+                            break;
+                        }
```
Pravdyvy: I think this approach breaks API of DB. 
`block_id` can now grow more than by 1 on consecutive blocks in DB. DB implementation very much assumes that growth is monotonic. 
More about it in the specific functions.

lez/storage/src/indexer/read_once.rs
Pravdyvy, line 45: This function now have a ambiguous failed state. `None` previously meant, that we requested block out of range.
Now there is 2 different `None`s, first one is that we requested block out of range, second one is that we requested skipped block id.

Pravdyvy (commented):
This approach breaks our DB API.
Original DB API was written with a strong assumption of monotonic `block_id` growth, i.e. consecutive `block_id`s have difference of 1.
The functions which suffer from it is `RocksDBIO::get_block` (I explained how). And `RocksDBIO::get_block_batch`.
`get_block_batch` breaks more, we stop batch at first `None`, because we assume that chain ended. This may be an immediate issue for explorer.
**There may be more examples.**
This is not only the issue with indexer DB though, sequencer will have the same problems, both DB built on the same assumption. 
We need to decide on how exactly we want to handle holes.

lez/indexer/core/src/lib.rs
Pravdyvy: This question is resolved if sequencer acts in the same fashion to indexer

lez/storage/src/indexer/read_once.rs
Pravdyvy: This question is resolved if sequencer acts in the same fashion to indexer

Pravdyvy (approved):
Comments are resolved if sequencer acts in the same fashion to indexer. lgtm

lez/indexer/core/src/stall_reason.rs
erhant: will use `Slot`

lez/indexer/core/src/block_store.rs
erhant: Indeed! Will fix 🙏

lez/indexer/service/rpc/src/lib.rs
erhant: will return `IndexerStatus` with conversions impl'd

lez/indexer/service/src/service.rs
erhant: Indeed, no need for `::Value` juggling at all

lez/indexer/core/src/lib.rs
erhant: Yeah indeed, looks like these were forgotten when I decided to add `allow_chain_reset` as config instead of parameter 😓

lez/indexer/core/src/lib.rs
erhant: https://github.com/logos-blockchain/logos-execution-zone/pull/581#discussion_r3515298819 :/

lez/indexer/core/src/stall_reason.rs
erhant: Removing

lez/indexer/service/src/service.rs
erhant: https://github.com/logos-blockchain/logos-execution-zone/pull/581/commits/24469d91f07a3f2a788cafbaa53a6ca26f14b6a7

lez/indexer/core/src/stall_reason.rs
erhant: https://github.com/logos-blockchain/logos-execution-zone/pull/581/commits/24469d91f07a3f2a788cafbaa53a6ca26f14b6a7

lez/indexer/core/src/stall_reason.rs
erhant: https://github.com/logos-blockchain/logos-execution-zone/pull/581/commits/24469d91f07a3f2a788cafbaa53a6ca26f14b6a7

lez/indexer/core/src/block_store.rs
erhant: https://github.com/logos-blockchain/logos-execution-zone/pull/581/commits/24469d91f07a3f2a788cafbaa53a6ca26f14b6a7

lez/indexer/core/src/lib.rs
erhant: https://github.com/logos-blockchain/logos-execution-zone/pull/581/commits/24469d91f07a3f2a788cafbaa53a6ca26f14b6a7

lez/indexer/core/src/lib.rs
erhant: https://github.com/logos-blockchain/logos-execution-zone/pull/581/commits/24469d91f07a3f2a788cafbaa53a6ca26f14b6a7

lez/indexer/core/src/lib.rs
erhant: https://github.com/logos-blockchain/logos-execution-zone/pull/581/commits/24469d91f07a3f2a788cafbaa53a6ca26f14b6a7

lez/indexer/core/src/lib.rs
erhant: https://github.com/logos-blockchain/logos-execution-zone/pull/581/commits/24469d91f07a3f2a788cafbaa53a6ca26f14b6a7

lez/indexer/service/rpc/src/lib.rs
erhant: https://github.com/logos-blockchain/logos-execution-zone/pull/581/commits/24469d91f07a3f2a788cafbaa53a6ca26f14b6a7

lez/indexer/core/src/lib.rs
erhant: grouped some of the repetitive parts under `advance_cursor`, I think it's borderline right now.

erhant: https://github.com/logos-blockchain/logos-execution-zone/pull/581/commits/24469d91f07a3f2a788cafbaa53a6ca26f14b6a7

lez/indexer/core/src/lib.rs
```diff
@@ -51,6 +90,75 @@ impl IndexerCore {
...
+    /// Detects when the local store belongs to a different chain than the connected L1
+    /// (e.g. a wiped/restarted bedrock) so startup can reset instead of silently
+    /// diverging. Mostly a dev convenience; won't trigger during normal live indexing.
+    ///
+    /// Compares the first block the channel serves at/after the cursor (the tip's
+    /// recent L1 slot, so ~one batch — not a from-genesis scan) against our stored
+    /// block of the *same id*. Comparing by the channel's id, not our tip id, is what
+    /// catches a *shorter* reset chain: it has no block at our tip id, but its low-id
+    /// block here differs from ours.
+    async fn chain_identity_outcome(&self) -> Result<ChainIdentityOutcome> {
```
schouhy: I'd rename this method to `verify_chain_consistency` or something along those lines

lez/indexer/core/src/lib.rs
```diff
@@ -1,26 +1,37 @@
...
 pub mod block_store;
 pub mod config;
+pub mod ingest_error;
+pub mod stall_reason;
 pub mod status;
 
+/// Result of comparing the indexer's stored tip against the channel.
+enum ChainIdentityOutcome {
+    /// Proceed from the cursor: channel still serves our chain, nothing to compare,
+    /// or the check was inconclusive — none of which prove a reset.
```
schouhy: why not explicitely add a `Inconclusive` variant

lez/indexer/core/src/lib.rs
```diff
@@ -51,6 +90,75 @@ impl IndexerCore {
...
+    /// block of the *same id*. Comparing by the channel's id, not our tip id, is what
+    /// catches a *shorter* reset chain: it has no block at our tip id, but its low-id
+    /// block here differs from ours.
+    async fn chain_identity_outcome(&self) -> Result<ChainIdentityOutcome> {
+        // Don't skip parked stores: the stall is persisted and only clears on a
+        // successful apply, so skipping would re-park forever and never catch a reset.
+        // Safe anyway — a same-chain park sits at an id we never applied, so the lookup
+        // below misses → Consistent.
+        let Some(cursor) = self.store.get_zone_cursor()? else {
+            return Ok(ChainIdentityOutcome::Consistent); // empty / cold store
```
schouhy: I think it's confusing to return Consistent variant in this case. I'd add a `Inconclusive` and be explicit in the caller that it treats the inconclusive case the same as the consistent one.

lez/indexer/core/src/lib.rs
```diff
@@ -51,6 +90,75 @@ impl IndexerCore {
...
+        let Some(ours) = self.store.get_block_at_id(channel_block.header.block_id)? else {
+            return Ok(ChainIdentityOutcome::Consistent);
+        };
+        Ok(compare_block(&ours, &channel_block))
+    }
+
+    /// Reads the first block the channel serves at/after the tip's slot. `next_messages`
+    /// is exclusive, so `cursor - 1` includes the tip's own slot. `None` = inconclusive
+    /// (timeout/error, or bedrock's LIB behind our tip → empty stream).
+    async fn fetch_channel_block_from(&self, cursor: Slot) -> Result<Option<Block>> {
```
schouhy: this function returns the same `Ok(None)` value in three different cases: nothing to compare, network error, network timeout. The caller can't tell them apart and treats them all as a consistent variant. Shouldn't we treat the network cases differently?

lez/indexer/core/src/block_store.rs
```diff
@@ -134,78 +166,234 @@ impl IndexerStore {
...
+        // Clear a recorded stall now that a block has chained and applied. Only write
+        // when one is actually present, so the steady state does no extra per-block
+        // write (and can't turn a transient clear-write error into a spurious park).
+        if self.get_stall_reason()?.is_some() {
+            self.set_stall_reason(&None)?;
+        }
```
schouhy: if any of these `?` is triggered the function returns `Err`. The caller treats all Err uniformly but at this point the block is already stored in the db.

lez/indexer/core/src/lib.rs
```diff
@@ -59,9 +167,11 @@ impl IndexerCore {
     pub fn status(&self) -> IndexerStatus {
         let sync = IndexerSyncStatus::clone(&self.status.load());
         let indexed_block_id = self.store.get_last_block_id().ok().flatten();
+        let stall_reason = self.store.get_stall_reason().ok().flatten();
```
schouhy: the rocksdb read error or json parsing error coming from `get_stall_reason` are collapsed to `None` after `get_stall_reason().ok().flatten()`, indistinguishible from the case where no stall is actually recorded.

the same `get_stall_reason().ok().flatten()` bites in other parts of the code.

lez/indexer/core/src/block_store.rs
```diff
@@ -134,78 +166,234 @@ impl IndexerStore {
...
+        None => {
+            if block.header.block_id != GENESIS_BLOCK_ID {
+                return Err(BlockIngestError::UnexpectedBlockId {
+                    expected: GENESIS_BLOCK_ID,
+                    got: block.header.block_id,
+                });
+            }
+        }
+        Some(tip) => {
+            let expected = tip.block_id.checked_add(1).expect("block id overflow");
```
schouhy: is adding a `BlockIdOverflow` variant to the error enum and replacing the `expect` here too much?

if we keep the expect, I'd rephrase the message to "block id should not overflow"

lez/indexer/core/src/block_store.rs
```diff
@@ -134,78 +166,234 @@ impl IndexerStore {
...
+
+    /// Records a stall for an inscription that could not even be parsed.
+    pub fn record_deserialize_stall(&self, l1_slot: Slot, error: String) -> Result<()> {
+        self.record_stall(None, l1_slot, BlockIngestError::Deserialize(error))
+    }
+
+    /// Validates `block` against the tip and, if it chains, applies it atomically
+    /// (scratch clone, commit only on full success) and advances the tip. On any
+    /// failure records the stall and returns `Parked` without touching state.
+    pub async fn accept_block(&self, block: &Block, l1_slot: Slot) -> Result<AcceptOutcome> {
```
schouhy, line 243: if for some reason this function receives a block the indexer has already processed, it will treat it as invalid. I think we need to treat different the case where the block with unexpected id has already been seen and was valid?

schouhy (commented):
Hey thanks for the PR! left some comments

lez/storage/src/indexer/mod.rs
erhant: Yes that is the case due to our decentralized sequencer plans, but I would rather leave this in another PR where we handle `ZoneMessage::Deposit` and `ZoneMessage::Withdraw` as well, one of which will require an additional patch from `logos-blockchain`.

lez/indexer/core/src/lib.rs
erhant: Removed `new_with_chain_check` too, as config handles that option

lez/indexer/core/src/ingest_error.rs
erhant: Fixed with https://github.com/logos-blockchain/logos-execution-zone/pull/581/commits/6c65195f0c400dab8efba82537efb92d567dc844

Now, `StateTransition` has a `reason` field that captures the exact error-chain of underlying `lee::error::LeeError`

It has to be a String here because the `BlockIngestError` itself must be ser/de. I could impl ser/de to lee Error as well, if that is applicable.

lez/indexer/core/src/ingest_error.rs
erhant: Fixed (see https://github.com/logos-blockchain/logos-execution-zone/pull/581#discussion_r3521616431)

lez/indexer/core/src/ingest_error.rs
erhant: The underlying error is a Borsh error from `borsh::from_slice`, but it is a good-ol `std::io::Error` which is kind of naked. I need ser/de in particular for `BlockIngestError` as I persist it within RocksDB; so instead I go with `anyhow` approach to get the entire error chain as a string.

Fixed in https://github.com/logos-blockchain/logos-execution-zone/commit/6c65195f0c400dab8efba82537efb92d567dc844

lez/indexer/core/src/block_store.rs
erhant: Changed `l1_slot` to `Slot`, but the real issue here is about the error type.

This descends from a `borsh::from_slice` returned `Err`, which is of type `std::io::Error`. It has nothing derived for us (Clone etc.) but `BlockIngestError` must be serializable (as we store it in RocksDB, to remember why we stalled across restarts).

So, I instead use the `anyhow` approach of converting the generic `Error` into an `anyhow` error, and then get its full chain with `{:#}`. That is what I pass into this `error` parameter here.

One thing that I can do is actually just use `record_stall`, it's a one-place called function anyways. Will do that instead!

Fixed in https://github.com/logos-blockchain/logos-execution-zone/commit/6c65195f0c400dab8efba82537efb92d567dc844

lez/indexer/core/src/ingest_error.rs
erhant: Also relevant comment at https://github.com/logos-blockchain/logos-execution-zone/pull/581#discussion_r3521624565

lez/indexer/core/src/lib.rs
erhant: Added a `ChainMismatch` that is both detailed and implements `Display` in https://github.com/logos-blockchain/logos-execution-zone/commit/6c65195f0c400dab8efba82537efb92d567dc844

lez/indexer/core/src/lib.rs
erhant: The idea was that as long it is not a Mismatch, the indexer will proceed; but I aggree its better to be explicit here.
Added an `Inconclusive` variant!

lez/indexer/core/src/lib.rs
erhant: Along with https://github.com/logos-blockchain/logos-execution-zone/pull/581#discussion_r3516393725, added an Inconclusive variant 🙏

lez/indexer/core/src/lib.rs
erhant: Will now use `Inconclusive` enum variant for these.

lez/storage/src/indexer/indexer_cells.rs
erhant: It's a JSON object that we have to parse with `serde_json` anyways, so I think `serde_json::from_slice` is enough over these bytes.
While there is not much perf difference between `from_str` and `from_slice` for `serde_json`, I think working with bytes in particular for RocksDB would be more suitable.
What do you think?

lez/indexer/core/src/block_store.rs
erhant: We used that on `IndexerSyncStatus` before which was very read-heavy (basically polled every few sec) but the scratch-state here is not so much. I think `RwLock` works nicely here. Also `ArcSwap` I believe is a bit more perf-heavy, and if I recall corectly it would clone the state in the background (so that readers see the old one and writer writes on the to-be-swapped one)

lez/indexer/core/src/lib.rs
erhant: https://github.com/logos-blockchain/logos-execution-zone/pull/581/commits/e9a30c400819862a8eea2d513907da42e820a2e1

lez/indexer/core/src/lib.rs
erhant: https://github.com/logos-blockchain/logos-execution-zone/pull/581/commits/e9a30c400819862a8eea2d513907da42e820a2e1

lez/indexer/core/src/lib.rs
erhant: https://github.com/logos-blockchain/logos-execution-zone/pull/581/commits/e9a30c400819862a8eea2d513907da42e820a2e1

lez/indexer/core/src/lib.rs
erhant: Renamed at https://github.com/logos-blockchain/logos-execution-zone/pull/581/commits/e9a30c400819862a8eea2d513907da42e820a2e1

lez/indexer/core/src/block_store.rs
erhant: Well technically we do Clone ourselves for the scratch state, but I would rather keep it explicit via RwLock rather than ArcSwap maybe, not sure too haha

lez/indexer/core/src/block_store.rs
erhant: I think reworded `expect` is enough, its a negligible edge case 🙏🏻

lez/indexer/core/src/block_store.rs
erhant: Will add a `clear_stall_if_present` that we can log on failure instead of returning.

lez/indexer/core/src/block_store.rs
erhant: Indeed I missed this case, for the record let me write below the possible scenarios of a tip (ours) & channel (theirs):

```md
# Outdated & Consistent
---(tip)--> (channel)

# Up-to-Date & Consistent
---------> (tip + channel)

# Mismatch
---------> (tip)
---------> (channel)

# Beyond-the-Date & Consistent
--(channel)--> (tip)
```

I hadn't considered the 4th case as possible, but it could very well be possible if I re-ran my bedrock over the same genesis etc.

Will add a check for that!

erhant: https://github.com/logos-blockchain/logos-execution-zone/pull/581/commits/2295bb69f23f735b9feabc958b9391b0ab7898a8

lez/indexer/core/src/block_store.rs
erhant: https://github.com/logos-blockchain/logos-execution-zone/pull/581/commits/2295bb69f23f735b9feabc958b9391b0ab7898a8

lez/indexer/core/src/lib.rs
erhant: https://github.com/logos-blockchain/logos-execution-zone/pull/581/commits/2295bb69f23f735b9feabc958b9391b0ab7898a8

lez/indexer/core/src/block_store.rs
erhant: https://github.com/logos-blockchain/logos-execution-zone/pull/581/commits/2295bb69f23f735b9feabc958b9391b0ab7898a8

lez/indexer/core/src/lib.rs
erhant: https://github.com/logos-blockchain/logos-execution-zone/pull/581/commits/24469d91f07a3f2a788cafbaa53a6ca26f14b6a7

lez/indexer/core/src/block_store.rs
erhant: https://github.com/logos-blockchain/logos-execution-zone/pull/581/changes/24469d91f07a3f2a788cafbaa53a6ca26f14b6a7 fixed with `validate_against_tip`

lez/common/src/block.rs
erhant: Hmm not sure about what this gains, maybe we can open a follow-up issue? I think we will play around with the Block structure a bit with the coming decentralized sequencer tasks too hahah

integration_tests/tests/bridge.rs
erhant: I will take a look at fault-injection now 🙏🏻

erhant: Hmm I think we can also leave this as a follow up issue, we will need some test harness modifications anyways for the decentralized sequencer part; it would be better to add them together.

We will likely introduce fault-injection, along with 2 valid-sequencer scenarios etc.

schouhy (approved):
lgtm thanks for considering my comments

integration_tests/tests/bridge.rs
Arjentix: Ah, so you want to test the invalid block case here, I see. But let's also have a test for basic successful usage?

lez/common/src/block.rs
Arjentix: So the gain is the removal of "zombie" types -- types which can be constructed but still be invalid. So ideally type should be always valid since the successful construction, so I'd like `Block` to check hash at deserialization and `new()` step or something like that. Yeah, we can move into an issue.

Does my description makes sense for you?

lez/storage/src/indexer/indexer_cells.rs
Arjentix: Yeah, it's fine

lez/indexer/core/src/block_store.rs
Arjentix: Yeah, I just suggest `ArcSwap` everywhere I see `lock.write() = new_value`, but that probably doesn't make sense in all cases

lez/indexer/core/src/lib.rs
```diff
@@ -1,26 +1,65 @@
...
+    /// A genuine divergence is still caught later when the ingest loop tries to apply and parks.
+    Inconclusive,
+    /// The channel serves a different block at one of our ids.
+    ///
+    /// Details in [`BlockMismatch`], and impl's Display trait.
+    Inconsistent(BlockMismatch),
+}
+
+/// The differing pair behind a [`ChainConsistenty::Inconsistent`]: our stored
+/// block vs. the block the channel serves at the same id, as `(block_id, hash)`.
```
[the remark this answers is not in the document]
Arjentix: Created #597 to catch this in CI

lez/indexer/core/src/ingest_error.rs
```diff
@@ -0,0 +1,67 @@
...
+    NonPublicGenesisTransaction,
+    #[error("State transition failed at transaction {tx_index}: {reason}")]
+    StateTransition {
+        /// Index of the failing transaction within the block body.
+        tx_index: u64,
+        /// Reason string from `lee::Error` to `anyhow::Error` to `{:#}`.
+        ///
+        /// This is required because `lee::Error` is not `Clone + Serialize + Deserialize`, so we
+        /// cannot store it directly.
+        reason: String,
```
Arjentix, line 39: Hmm, maybe add this derives to `lee::Error` and store it here with `#[source]` attribute?

Arjentix (approved):
LGTM, thanks

lez/indexer/core/src/lib.rs
```diff
@@ -116,48 +310,105 @@ impl IndexerCore {
 
                     let zone_block = match msg {
                         ZoneMessage::Block(b) => b,
-                        // Non-block messages don't carry a cursor position; the
-                        // next ZoneBlock advances past them implicitly.
+                        // FIXME: will be handled in prep of decentralized sequencers
                         ZoneMessage::Deposit(_) | ZoneMessage::Withdraw(_) => continue,
                     };
```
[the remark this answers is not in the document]
erhant: This is intended, will be done in a later PR that specifically handles them.

integration_tests/tests/bridge.rs
erhant: Ah yes, sorry I misunderstood. Indeed this test here is the happy-path & it works, no need for `#[ignore...]` stuff. Re-enabled & observed to pass in https://github.com/logos-blockchain/logos-execution-zone/pull/581/commits/9128bf467122908662e01b9200a53a8b70d56c71

Fault-injection ones will be implemented with the next Sequencer & Indexer related task.

lez/indexer/core/src/lib.rs
```diff
@@ -51,17 +125,113 @@ impl IndexerCore {
...
+        };
+
+        Ok(comparison_result)
+    }
+
+    /// Reads the first block the channel serves at/after the tip's slot.
+    ///
+    /// `None` means we could not read a comparable block.
+    async fn fetch_channel_block_from(&self, cursor: Slot) -> Result<Option<Block>> {
+        const TIP_FETCH_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(60);
```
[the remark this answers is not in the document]
erhant: https://github.com/logos-blockchain/logos-execution-zone/pull/581/commits/9f10c845f798a5509508b570e3ae1d0bf6ff5dc8

lez/indexer/core/src/ingest_error.rs
erhant: ~~Sure if that is ok! I didn't want to touch too much on the core types; but I guess we would make use of this elsewhere too.~~

~~If we don't like it, we can feature-gate it with some custom lez-serialize feature.~~

lez/common/src/block.rs
erhant: Yesss, I got it now.

Issue up on https://github.com/logos-blockchain/logos-execution-zone/issues/599

lez/indexer/core/src/ingest_error.rs
erhant: Hmm this looks more invasive that I thought it would be. `lee::Error` transitively holds `std::io::Error` `anyhow::Error`, and `k256::schnorr::Error`, none of which have these `derive`s neither. We should keep it as is.

lez/indexer/core/src/lib.rs
```diff
@@ -51,17 +125,113 @@ impl IndexerCore {
...
+    ///
+    /// Compares the **first** block the channel serves at/after the cursor (the tip's
+    /// recent L1 slot) against our stored block of the *same id*.
+    /// Note that we have to respect the channel's tip slot, because if our tip
+    /// is ahead of the channel even in the same chain, it may looks like we are
+    /// ahead of the channel, but we are actually on a different chain.
+    async fn verify_chain_consistency(&self) -> Result<ChainConsistency> {
+        let Some(cursor) = self.store.get_zone_cursor()? else {
+            // if we have no cursor, the store is empty or cold: nothing to compare
+            return Ok(ChainConsistency::Inconclusive);
```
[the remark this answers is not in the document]
erhant: Fixed in https://github.com/logos-blockchain/logos-execution-zone/pull/581/commits/2d9984eb95f23fa5a80b9dcdf9c3c99e54e6b087.

The check no longer anchors on the read cursor's position in the stream. It anchors on a block the same chain must still serve at a known slot: the recorded parked block (`StallReason` carries its id/hash/slot) while parked, otherwise the stored tip (only blocks advance the cursor, so when not parked the cursor is the tip's inscription slot).

lez/indexer/core/src/lib.rs
```diff
@@ -116,46 +199,74 @@ impl IndexerCore {
 
                     let zone_block = match msg {
                         ZoneMessage::Block(b) => b,
-                        // Non-block messages don't carry a cursor position; the
-                        // next ZoneBlock advances past them implicitly.
+                        // FIXME: will be handled in prep of decentralized sequencers
                         ZoneMessage::Deposit(_) | ZoneMessage::Withdraw(_) => continue,
                     };
```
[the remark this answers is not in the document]
erhant: This is intended right now, will fix in proper Deposit & WIthdraw handling

lez/indexer/core/src/ingest_error.rs
Arjentix: Ah, I see, okay

integration_tests/tests/bridge.rs
```diff
@@ -557,6 +557,37 @@ fn create_zone_indexer_observer(
     ))
 }
 
+/// Test that the indexer status RPC reports caught-up with no stall after a clean run.
+///
+/// TODO: Integration-level park testing (publishing a bad block to force a stall) is a follow-up
+/// needing fault injection support in the test harness.
+#[test]
+async fn indexer_status_rpc_reports_caught_up_with_no_stall() -> anyhow::Result<()> {
```
Arjentix: Cool, just one nit: looks like it's not related to bridge tests but it's located in `bridge.rs`. I'd better move it to some indexer related file or create a new one.

Arjentix (approved):

integration_tests/tests/bridge.rs
erhant: Yep!

erhant: https://github.com/logos-blockchain/logos-execution-zone/pull/581/commits/08609c36234c64e42b865c019daff5006aa42531

There was also an error that apparently failed in CI but passed locally; also fixed.

lez/indexer/core/src/status.rs
```diff
@@ -1,9 +1,10 @@
 use serde::Serialize;
 
+use crate::stall_reason::StallReason;
+
 /// Coarse lifecycle state of the indexer's ingestion loop, so a client can tell
 /// "still catching up" apart from "something went wrong".
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize)]
-#[serde(rename_all = "snake_case")]
 pub enum IndexerSyncState {
```
moudyellaz, line 8: Removing the rename_all attrs changes the existing query_status JSON (caught_up -> CaughtUp, lastError -> last_error). lez-indexer-module parses the old names (lez_indexer_module_impl.h), so this needs either the attrs restored on these two types, or a lockstep module update plus a ! / BREAKING CHANGE marker on the commit.

lez/indexer/core/src/chain_consistency.rs
```diff
@@ -0,0 +1,500 @@
...
+    /// cursor (only blocks advance the cursor, so when not parked it is the
+    /// tip's inscription slot). See [`Self::verify_chain_at_anchor`].
+    pub(crate) async fn verify_chain_consistency(&self) -> Result<ChainConsistency> {
+        let anchor = if let Some(stall) = self.store.get_stall_reason()? {
+            Anchor {
+                slot: stall.l1_slot,
+                block: stall.block_id.zip(stall.block_hash),
+            }
+        } else {
+            let Some(cursor) = self.store.get_zone_cursor()? else {
```
moudyellaz: If the persisted cursor lags the tip by 2+ blocks (set_zone_cursor failures are warn-only), this anchor pairs the tip with a too-old slot and the scan reports ReinscribedBlock for the same chain. accept_block already receives l1_slot; persisting the tip's own inscription slot with the tip and anchoring on that would remove the dependence on the cursor invariant.

lez/common/src/transaction.rs
```diff
@@ -129,16 +128,12 @@ impl LeeTransaction {
...
-    /// accepted, including sequencer-generated deposit transactions that
-    /// legitimately modify the bridge account. The `TransactionOrigin::Sequencer`
-    /// tag that lets the sequencer bypass the guard is not carried in the block,
-    /// so the indexer cannot yet distinguish deposit txs from user txs.
-    ///
-    /// REMOVE ME when the indexer can authenticate deposit transactions.
-    pub fn execute_without_system_accounts_check_on_state(
+    /// The indexer replays blocks the sequencer already validated and inscribed on Bedrock,
+    /// so it trusts those inscriptions and re-derives state without re-validating them.
+    pub fn execute_on_state(
```
moudyellaz, line 136: Can we get a tracking issue for block-signature verification in the indexer (the trust-model thread deferred to the ZoneMessage PR)? With the tx stateless checks gone it is the only missing authentication on the ingest path.

moudyellaz (approved):

lez/indexer/core/src/chain_consistency.rs
erhant: very nice find! will fix as you suggest, makes sense 🙏🏻

lez/indexer/core/src/status.rs
erhant: I will keep this removed, and open an issue on indexer + explorer for TODOs, including this.
We already need to implement some stuff there related to this PR; along with UI additions for "stall" displays.

lez/common/src/transaction.rs
erhant: On it! 🫡

lez/indexer/core/src/status.rs
erhant: https://github.com/logos-blockchain/logos-execution-zone/pull/581/changes/1547a51fa66ace494af72337a0c0f015ddb6fb65 commit renamed

lez/common/src/transaction.rs
erhant: https://github.com/logos-blockchain/logos-execution-zone/issues/602

lez/indexer/core/src/chain_consistency.rs
erhant: I've edited `put_block` to also accept the L1 Slot so we know where the last L2 block is coming from in L1, overwriting a `TipSlot` in RocksDB (very small cost). While changing the `put_block` signature has some multi-file edits, its the cleanest & explicit approach we can take.

https://github.com/logos-blockchain/logos-execution-zone/pull/581/commits/9b2b4a3d8ca714fa089df634a230d1e0764dcdd7

lez/indexer/core/src/status.rs
erhant: issues https://github.com/logos-blockchain/lez-explorer-ui/issues/13 and https://github.com/logos-blockchain/lez-indexer-module/issues/17 opened

erhant:
There seems to be an edge issue on graceful shutting down & mistakenly stalling; see https://github.com/logos-blockchain/logos-execution-zone/issues/605

Will fix

EDIT: fixed in #610

lez/indexer/core/src/status.rs
```diff
@@ -1,9 +1,10 @@
...
     /// Booted; no ingestion cycle has run yet.
     Starting,
```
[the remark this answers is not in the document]
erhant: This is intended.

lez/indexer/core/src/lib.rs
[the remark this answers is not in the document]
erhant: This is intended for now.

lez/indexer/core/src/lib.rs
```diff
@@ -116,46 +199,74 @@ impl IndexerCore {
...
+                // Stream drained. Stay Stalled if parked; otherwise we are caught up.
+                // A store error here must not be collapsed to "no stall recorded":
+                // that would wrongly flip us to caught-up, so we log and hold state.
+                match self.store.get_stall_reason() {
+                    Ok(None) => self.set_status(IndexerSyncStatus::caught_up()),
+                    Ok(Some(_)) => {}
+                    Err(err) => {
+                        warn!("Failed to read stall reason after draining stream; not marking caught up: {err:#}");
+                    }
+                }
```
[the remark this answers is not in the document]
erhant: This is correct (even if its an edge case), fixed in https://github.com/logos-blockchain/logos-execution-zone/pull/581/commits/3656b7cbd0875e81b25d569b8ba352bded1a927c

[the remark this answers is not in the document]
erhant: Instead of an in-memory check, we simply ensure the `park_undeserializable` passes

PR #583, opened 2026-06-30
fix(sequencer): don't accept zero deposits & remove unreachable

Arjentix:
## 🎯 Purpose

- Fixes https://github.com/logos-co/audit-reports/issues/104

## ⚙️ Approach

- Forbid bridge transactions which doesn't do anything (we can't forbid exactly Deposits because private execution exists)
- Remove `unreachable!` just in case and use `None` instead
- Add test

## 🧪 How to Test

```bash
cargo test --package integration_tests --test bridge --all-features -- public_bridge_deposit_with_zero_amount_is_rejected
```

## 🔗 Dependencies

None

## 🔜 Future Work

None

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


lez/sequencer/core/src/lib.rs
```diff
@@ -729,26 +729,23 @@ fn extract_bridge_withdraw_data(tx: &LeeTransaction) -> Option<WithdrawArg> {
...
-            let recipient_pk =
-                logos_blockchain_key_management_system_service::keys::ZkPublicKey::from(
-                    BigUint::from_bytes_le(&bedrock_account_pk),
-                );
+    let bridge_core::Instruction::Withdraw {
+        amount,
+        bedrock_account_pk,
+    } = instruction
+    else {
+        return None;
```
erhant, line 737: should we add a TODO comment here about handling other instructions around here later?

integration_tests/tests/bridge.rs
```diff
@@ -98,6 +98,66 @@ async fn public_bridge_deposit_invocation_is_dropped() -> anyhow::Result<()> {
...
+        .sequencer_client()
+        .get_account_balance(bridge_account_id)
+        .await?;
+    let vault_balance_after = ctx
+        .sequencer_client()
+        .get_account_balance(recipient_vault_id)
+        .await?;
+    let tx_on_chain = ctx.sequencer_client().get_transaction(tx_hash).await?;
+
+    assert_eq!(bridge_balance_after, bridge_balance_before);
```
erhant, line 151: do we need this assertion (and the one below)? if a 0-amount tx were to succeed, this would still be true I think

erhant (approved):
LGTM, two comments

Pravdyvy (approved):
lgtm

lez/sequencer/core/src/lib.rs
Arjentix: No, I don't think so, the function is called `extract_bridge_withdraw_data()`

integration_tests/tests/bridge.rs
Arjentix: Well, yes, but just in case. This is how other tests are implemented in this module

moudyellaz (approved):

PR #584, opened 2026-06-30
refactor: abstract address plaintext into a struct

agureev:
## 🎯 Purpose

This addresses @Arjentix comment regarding the fact that we have been overusing the `(npk, vpk, identifier)` tuple both in account computation and in state.

This PR aims to refactor this and generally clean up some circuit computations given the inclusion of a fairly large `vpk` in-guest

## ⚙️ Approach

The approach is to make a `PrivateAddressPlaintext` struct that holds the aforementioned data. Connected to that, we make some modifications to minimize `vpk` copying and improve readability, such as introducing a `vpk` lifetime so that the struct does not need any copies to be created, while also removing the `private_pda_by_position` to make the lifetime self-contained and not proliferate to the execution state.

Note that since `sha2` is fairly fast in-guest (hence `npk` computation does not cost a lot of cycles), while the `vpk` is large, removing the position map does not introduce much caching deficiency.

However, this can be instead replaced by keeping a map and making the `ExecutionState` also parametrizable by `vpk` lifetimes

## 🧪 How to Test

## 🔗 Dependencies

#550

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


lee/state_machine/core/src/nullifier.rs
```diff
@@ -12,22 +16,18 @@ pub type Identifier = u128;
...
+impl PrivateAddressPlaintext<'_> {
     /// Derives an [`AccountId`] for a regular (non-PDA) private account from the nullifier public
     /// key and identifier.
     #[must_use]
-    pub fn for_regular_private_account(
-        npk: &NullifierPublicKey,
-        vpk: &ViewingPublicKey,
-        identifier: Identifier,
-    ) -> Self {
+    pub fn account_id(&self) -> AccountId {
```
Arjentix: I suggest to keep `PrivateAddressPlaintext` as a simple struct and have `AccountId` constructor from it. Like this:

```rust
let _ = AccountId::for_regular_private_account(private_address_plaintext);
let _ = AccountId::for_private_pda(private_address_plaintext, program_id, pda_seed);
```

program_methods/guest/src/bin/privacy_preserving_circuit/execution_state.rs
```diff
@@ -44,13 +43,6 @@ pub struct ExecutionState {
     /// `AccountId` entry or as an equality check against the existing one, making the rule: one
     /// `(program, seed)` → one account per tx.
     pda_family_binding: HashMap<(ProgramId, PdaSeed), AccountId>,
-    /// Map from a private-PDA `pre_state`'s position in `account_identities` to the (npk, vpk,
-    /// identifier) supplied for that position. Built once in `derive_from_outputs` by walking
-    /// `account_identities` and consulting `npk_vpk_if_private_pda`. Used later by the claim and
-    /// caller-seeds authorization paths to verify
-    /// `AccountId::for_private_pda(program_id, seed, npk, vpk, identifier) ==
-    /// pre_state.account_id`.
-    private_pda_by_position: HashMap<usize, (NullifierPublicKey, ViewingPublicKey, Identifier)>,
```
Arjentix, line 53: Cool that this was removed

lee/state_machine/core/src/nullifier.rs
```diff
@@ -36,12 +36,6 @@ impl AccountId {
     }
 }
 
-impl From<(&NullifierPublicKey, &ViewingPublicKey, Identifier)> for AccountId {
-    fn from((npk, vpk, identifier): (&NullifierPublicKey, &ViewingPublicKey, Identifier)) -> Self {
-        Self::for_regular_private_account(npk, vpk, identifier)
-    }
-}
```
Arjentix, line 43: This is also cool, finally no `From` left for `AccountId` is it became too hard to read in caller code

Issue #573, opened 2026-06-24
Change Shielded Wallet Sync given View Tag Semantics Change


agureev:
Given issue #560, in conversation with the team, it was decided that it would be best to *only* compute the `view_tag` in-guest for initialized accounts. This would improve privacy as the view tag now only would reveal partial information about initialized accounts owners.

This means that we would not be able to use the usual view tag optimization for wallet syncing. However, given that we know an updated accounts plaintext X with the knowledge of the nsk for owned and shared accounts, we know transactions in which an account is updated via computing the nullifier for X and waiting for a transaction in which it appears.

For accounts generated by sole-owned accounts, we don't even need to sync as we already know which commitments we computed and using which plaintexts. We simply need confirmations that the transaction we submitted goes through.

Issue #559, opened 2026-06-22
lez/wallet: restore doesn't reset last_synced_block (empty balances after restore) + sync panics on counterparty-chosen identifier / malformed tx


davidrusu:
# lez/wallet: restore doesn't reset `last_synced_block` (empty balances after restore) + sync panics on counterparty-chosen identifier / malformed tx

**Repo:** `logos-blockchain/logos-execution-zone`

## Summary

Three wallet-local robustness bugs in the L2 wallet: (1) restoring from mnemonic doesn't reset the
sync cursor, so a restored wallet never rescans history and shows empty/wrong balances; (2) a
counterparty-chosen recipient identifier can panic the wallet's block sync (`.expect`) for imported
private keys; (3) a `.unwrap()` on note decryption can panic on a malformed polled tx. All are
client-side (no chain-accepted incorrect artifact), but (1) silently shows wrong balances and (2) is a
remotely-triggerable wallet-sync crash.

## Severity

Medium (wallet-local). (1) is the clearest real correctness bug; (2) is a wallet-sync DoS.

### 1. Restore leaves `last_synced_block` stale → restored accounts never rescan history
- `lez/wallet/src/storage.rs:88-99` `Storage::restore` resets `key_chain` and `labels` but **not**
  `last_synced_block`.
- `sync_to_block` (`lez/wallet/src/lib.rs:692-695`) early-returns when `last_synced_block >= block_id`
  and starts the stream at `last_synced_block + 1` (`:707`).
- After `RestoreKeys` (`cli/mod.rs:274-281,354-383`) into a home previously synced to block N, the fresh
  key_chain only scans blocks `> N`, so all private-account state committed in blocks `0..=N` is never
  re-derived → restored wallet shows zero balance / "account not found" for that history.
- Fix: set `last_synced_block = 0` (or first-relevant block) in `restore`.

### 2. `.expect("Account Id should exist")` panic on imported key receiving an unregistered identifier
- `sync_private_accounts_with_tx` (`lez/wallet/src/lib.rs:738-782`) matches outputs by view_tag, which
  is `compute_view_tag(npk, vpk)` — identifier-independent
  (`lee/state_machine/core/src/encryption/mod.rs:74`). Any output to your npk decrypts and yields a
  sender-chosen `kind`/identifier; the wallet then calls
  `insert_private_account(account_id, kind, ...).expect("Account Id should exist")` (`:778-781`).
- For `imported_private_accounts`, `insert_private_account`
[...]

PR #585, opened 2026-07-01
feat(wallet): change wallet synchronization flow

agureev:
## 🎯 Purpose

First step in solving #573 allowing to randomize viewing tags for updated accounts. Specifically, do not analyze viewing tags for accounts you already know you own, i.e. for those whose nullifiers you can compute. As the account plaintext per-ID is updated linearly, you can keep track of the state updates that way,

## ⚙️ Approach

Keep track of most recent nullifiers (i.e. the nullifier for the latest state of the account the user is holding) during sync, so that when it is encountered, the view tag is not consulted.

- [x] Add `HashMap<Nullifer, AccountID>` mutable map during sync
- [x] Add helper functions to update the map given new nullifiers
- [x] Integrate the map into the syncing process
- [x] Make the syncing skip the cipherstexts which were eagerly decrypted via nullifier spotting

## 🧪 How to Test

`cargo test -p wallet --all-features --lib nullifier_sync`

## 🔜 Future Work

Remove standard viewing tag construction for updated accounts, making them user-supplied.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


lez/wallet/src/storage/key_chain.rs
```diff
@@ -281,6 +322,121 @@ impl UserKeyChain {
...
+    /// Decrypts the note at slot `i` for `account_id`, stores the new state, and returns the
+    /// account's next update nullifier. `None` if keys or decryption fail.
+    fn apply_nullifier_update(
+        &mut self,
+        account_id: AccountId,
+        message: &Message,
+        i: usize,
+    ) -> Option<Nullifier> {
+        let encrypted = &message.encrypted_private_post_states[i];
+        let commitment = &message.new_commitments[i];
```
moudyellaz, line 413: What do you think about using .get(i) here? Since an aligned message is only guaranteed when it comes from an honest node, this would let a malformed one fall through to the view-tag pass instead of panicking the whole sync (same spirit as the #559 fixes).

```suggestion
        let encrypted = message.encrypted_private_post_states.get(i)?;
        let commitment = message.new_commitments.get(i)?;
```

lez/wallet/src/storage/key_chain.rs
```diff
@@ -281,6 +322,121 @@ impl UserKeyChain {
...
+                continue;
+            };
+            let nsk = keys.nullifier_secret_key;
+            index.track(account_id, &entry.account, &nsk);
+        }
+
+        index
+    }
+
+    /// Applies every watched nullifier the `message` publishes: decrypts the position-aligned
```
moudyellaz, line 377: The nullifier-to-slot alignment this relies on is what #564 is about to shuffle. When order is obfuscated, a nullifier hit should probably trigger try-decrypt over all unhandled slots instead of slot i. Is the plan to rework this in #564, and should the alignment assumption be documented on Message until then?

lez/wallet/src/lib.rs
```diff
@@ -797,72 +797,59 @@ impl WalletCore {
...
-            for (ciph_id, encrypted_data) in tx
-                .message()
-                .encrypted_private_post_states
-                .iter()
-                .enumerate()
+            for (ciph_id, encrypted_data) in
+                message.encrypted_private_post_states.iter().enumerate()
             {
-                if encrypted_data.view_tag != view_tag {
+                // If already decrypted or the tag does nto match, skip.
```
moudyellaz: Typos: nto -> not here, and connectted -> connected in key_chain.rs line 381.

moudyellaz (commented):

moudyellaz (approved):

lez/wallet/src/storage/key_chain.rs
agureev: The plan is indeed to address this in #564. In particular, the strategy for fetching shuffled ciphers are reliant on the exact strategy of shuffling, so here it is probably better to base the fetching on the current protocol assumptions and then change it up once we agree on the shuffling explicitly. Does that sound like a good plan?

lez/wallet/src/storage/key_chain.rs
agureev: Ah, good catch! Indeed, a malicious endpoint may feed ill-formed transactions. Using `get`s is a sensible solution, asking for `is_some`. Possibly can be even better to catch this in an explicit error on top while processing the tx mismatching on length early as a misinformed tx.

However - now that I think about this - this sounds like something that should be fixed at the decoding/struct level, no? Instead of having `new_commitments`, `new_nullifiers`, `encrypted_post_states` fields, by not have a vec of `new_hashes: Vec<PrivateOutput>` with `PrivateOutput` having three fields, `nullifier`, `commitment` and `encrypted_post_state`.

That would mean that no length problems arize at all. How does that sound? Maybe I can make a separate PR for that? Or indeed just bundle this here?

lez/wallet/src/lib.rs
agureev: Handled in 1bd3960

lez/wallet/src/storage/key_chain.rs
agureev: Handled in b31aecd

lez/wallet/src/lib.rs
```diff
@@ -759,11 +760,12 @@ impl WalletCore {
         Ok(())
     }
 
-    fn sync_private_accounts_with_tx(&mut self, tx: LeeTransaction) {
-        let LeeTransaction::PrivacyPreserving(tx) = tx else {
-            return;
-        };
-
+    fn sync_private_accounts_with_tx(
```
jonesmarvin8: Function name is strange since we're receiving a message instead of tx.

lez/wallet/src/storage/key_chain.rs
jonesmarvin8: This change appears to have gotten reverted in your edits (it existed in the commit you cited, but now gone?)

lez/wallet/src/lib.rs
```diff
@@ -298,35 +299,26 @@ impl WalletCore {
...
-        if let (Some(pda_seed), Some(program_id)) = (entry.pda_seed, entry.authority_program_id) {
-            let keys = holder.derive_keys_for_pda(&program_id, &pda_seed);
+        if entry.pda_seed.is_some() {
             Some(AccountIdentity::PrivatePdaShared {
                 account_id,
-                nsk: keys.nullifier_secret_key,
-                npk: keys.generate_nullifier_public_key(),
-                vpk: keys.generate_viewing_public_key(),
-                identifier: entry.identifier,
+                nsk,
```
jonesmarvin8, line 311: Could we add a data structure to potentially bundle `npk, vpk, identifier`? I don't think it's necessary to bundle `nsk` together with these; private accounts can be initialized with or without knowledge of `nsk`.

jonesmarvin8: ^ future PR question; no worries.

jonesmarvin8 (approved):
lgtm. I left minor comments.

lez/wallet/src/storage/key_chain.rs
agureev: ah, no, they shouldn't have been reverted! one line have just probably have grown stale because of [24faedd](https://github.com/logos-blockchain/logos-execution-zone/pull/585/commits/24faedd6862f20f1981c1a6288a6f6fce170a6e1)

to make the handling explicit, in `message.rs` we introduce a method to validate lengths of all the vectors, returning the length if all the appropriate vectors match and `None` otherwise

then we insert the validation to `sync_to_block` and `decode_insert...` functions that process incoming messages from the sequencer. 

the additional detail - which was fixed in the commit on top - is that for decoding self-sent txs I assumed that the mask is always full (which it isn't based on tests). so initially i made an extra check to make sure that the note length of the message exactly matches the mask length and then relaxed it in [24faedd](https://github.com/logos-blockchain/logos-execution-zone/pull/585/commits/24faedd6862f20f1981c1a6288a6f6fce170a6e1)

lez/wallet/src/lib.rs
agureev: indeed, i have also played arounf with the idea in #584 but that did not seem to get far in terms of approval. feel free to look at it if you think it may bear fruit!

lez/wallet/src/lib.rs
agureev: good call! changed the name appropriately but minimally here: [a3593e3](https://github.com/logos-blockchain/logos-execution-zone/pull/585/commits/a3593e36b7a921462228d2be36eb4febea12b619)

jonesmarvin8 (approved):

moudyellaz (approved):
lgtm, thanks for adddressing my comments!

PR #587, opened 2026-07-01
feat!(wallet): wait for deploy tx inclusion in block

Arjentix:
Breaking:

1. Now a lot of wallet commands return `TransactionExecuted`, instead of `Empty`

2. Sequencer's `getTransaction()` RPC response changed from `tx` to `(tx, block_id)`

## 🎯 Purpose

We had a request from one of our users that deploy command exited with `0`, but actual transaction failed on chain. This is because we don't wait for it to be committed.

This PR addresses this issue and also some refactoring on the go.

## ⚙️ Approach

- Poll deploy tx
- Return `TransactionExecuted` from polling wallet functions

## 🧪 How to Test

```bash
cargo test --package integration_tests --test program_deployment --all-features -- deploy_invalid_program_fails
```

## 🔗 Dependencies

None

## 🔜 Future Work

None

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


Pravdyvy (approved):
lgtm

lez/wallet/src/lib.rs
```diff
@@ -558,10 +552,11 @@ impl WalletCore {
         tx_hash: HashType,
     ) -> Result<cli::SubcommandReturnValue> {
         println!("Transaction hash is {tx_hash}");
-        let transfer_tx = self.poll_native_token_transfer(tx_hash).await?;
-        println!("Transaction data is {transfer_tx:?}");
+        let (tx, block_id) = self.poller.poll_tx(tx_hash).await?;
+        println!("Transaction is included in block {block_id}");
+        println!("Transaction data is {tx:?}");
```
moudyellaz, line 557: Now that deploys also go through here, `{tx:?}` will debug-print the full `bytecode: Vec<u8>`, so deploying a real program dumps the whole ELF to stdout as a decimal byte list. Could we skip the data print for `LeeTransaction::ProgramDeployment` and print something like the bytecode length instead?

lez/sequencer/service/rpc/src/lib.rs
```diff
@@ -68,7 +68,7 @@ pub trait Rpc {
     async fn get_transaction(
         &self,
         tx_hash: HashType,
-    ) -> Result<Option<LeeTransaction>, ErrorObjectOwned>;
+    ) -> Result<Option<(LeeTransaction, BlockId)>, ErrorObjectOwned>;
```
moudyellaz, line 71: Small addition to the breaking note: this also changes the `getTransaction` RPC response shape from `tx` to `[tx, block_id]` on the wire, so external RPC consumers are affected too, worth a mention in the PR body.

integration_tests/tests/auth_transfer/private.rs
```diff
@@ -70,7 +70,7 @@ async fn private_transfer_to_foreign_account() -> Result<()> {
     });
 
     let result = wallet::cli::execute_subcommand(ctx.wallet_mut(), command).await?;
-    let SubcommandReturnValue::PrivacyPreservingTransfer { tx_hash } = result else {
+    let SubcommandReturnValue::TransactionExecuted { tx_hash } = result else {
         anyhow::bail!("Expected PrivacyPreservingTransfer return value");
```
moudyellaz:
```suggestion
        anyhow::bail!("Expected TransactionExecuted return value");
```
The old variant name survives in a few other error messages too (keys.rs, pinata.rs, vault.rs, setup.rs), might be worth a quick sweep.

moudyellaz (approved):
lgtm, thanks. Left few comments

lez/wallet/src/lib.rs
Arjentix: Indeed, thanks, fixed by manually implementing `Debug` for `program_deployment_transaction::Message` and truncating bytecode

lez/sequencer/service/rpc/src/lib.rs
Arjentix: Ah, true, will add to PR and commit message

integration_tests/tests/auth_transfer/private.rs
Arjentix: Indeed, thanks, fixed

PR #588, opened 2026-07-01
fix(lez/indexer): initialize testnet with the correct state

### LINKED ISSUE #586: Indexer is not aware of `Pinata`
The indexer and sequencer seed **different genesis states**. The pinata program (and the pinata system account) exist only in the sequencer's genesis. When the indexer replays a block that invokes the pinata program, its state machine rejects the block with `Unknown program`, so the block fails to index — and since the indexer can't advance past it, every subsequent block is missed by the indexer/explorer too.

## Root cause

- `sequencer_service` hardwires the testnet genesis: with `testnet` feature, `build_genesis_state` takes the testnet branch:

https://github.com/logos-blockchain/logos-execution-zone/blob/a58fbce2ff48c58b7bb5001b1a27e64b9596ee3a/lez/sequencer/core/src/lib.rs#L588-L589

This registers `programs::pinata()` **and** the pinata system account:

https://github.com/logos-blockchain/logos-execution-zone/blob/a58fbce2ff48c58b7bb5001b1a27e64b9596ee3a/lez/testnet_initial_state/src/lib.rs#L281-L296

- The **indexer** unconditionally seeds with `initial_state()`:

https://github.com/logos-blockchain/logos-execution-zone/blob/a58fbce2ff48c58b7bb5001b1a27e64b9596ee3a/lez/indexer/core/src/block_store.rs#L25-L35

This does not have Pinata program / account at all:

https://github.com/logos-blockchain/logos-execution-zone/blob/a58fbce2ff48c58b7bb5001b1a27e64b9596ee3a/lez/testnet_initial_state/src/lib.rs#L260-L279

## How to Reproduce

Local bedrock + sequencer + indexer + explorer, then run the wallet flow through `wallet pinata claim`. The claim gets sequenced fine (it's a valid block), but the indexer fails on it with an error like `Error at transition InvalidInput("Unknown program")`

The error comes from here:

https://github.com/logos-blockchain/logos-execution-zone/blob/a58fbce2ff48c58b7bb5001b1a27e64b9596ee3a/lee/state_machine/src/validated_state_diff.rs#L135-L137

The block is valid; the indexer just can't recognize a program its genesis never registered. The two genesis states must be identical, since the indexer replays exactly what the sequencer produces.

schouhy:
## 🎯 Purpose

Fixes #586. The indexer used a different genesis state than the sequencer. The sequencer's genesis has the pinata program and account, but the indexer's genesis did not. When a block claimed pinata, the indexer failed with "Unknown program" and stopped indexing all blocks after that.

## ⚙️ Approach

Made the indexer build its genesis the same way the sequencer does.

- [x] Add a `testnet` feature to `indexer_core`, same as `sequencer_core` already has
- [x] Use `initial_state_testnet()` instead of `initial_state()` in `IndexerStore::open_db` when the feature is on
- [x] Turn on the `testnet` feature for `indexer_service` and `indexer_ffi` (ffi also builds its own local genesis for the wallet)
- [x] Add a test that claims pinata and checks the indexer stays in sync

## 🧪 How to Test

Run:
```
cargo test -p integration_tests --test pinata claim_pinata_indexer_keeps_up
```

## 🔗 Dependencies

None

## 🔜 Future Work

None

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests

### DIFF
--- a/integration_tests/tests/pinata.rs
+++ b/integration_tests/tests/pinata.rs
@@ -9,7 +9,7 @@ use std::time::Duration;
 use anyhow::{Context as _, Result};
 use integration_tests::{
     TIME_TO_WAIT_FOR_BLOCK_SECONDS, TestContext, private_mention, public_mention,
-    verify_commitment_is_in_state,
+    verify_commitment_is_in_state, wait_for_indexer_to_catch_up,
 };
 use log::info;
 use sequencer_service_rpc::RpcClient as _;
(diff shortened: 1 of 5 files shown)


erhant (approved):
looks to me good 🙏🏻

moudyellaz (approved):

PR #590, opened 2026-07-02
feat(test_fixtures): use prebuilt sequencer dp to speed up integration tests

Arjentix:
## 🎯 Purpose

Integration tests became quite long after my work on moving initial state to genesis because currently each tests executes genesis and vault claims in `TestContext::new()`. This takes a lot. Running all integration tests on my machine takes 40 minutes.

## ⚙️ Approach

- Store fixture db dump with 2 blocks: genesis and vault claims
- Add a binary under `test_fixtures` to generate this dump
- Reuse this fixture dump in integration tests
- If someone wants to execute genesis and claims they can use `TestContextBuilder::from_scratch()`
- Add tests for this dump and tweak CI to catch mismatches before running integration tests

## 🧪 How to Test

See CI and it's speed time.

## 🔗 Dependencies

None

## 🔜 Future Work

None

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


Arjentix:
I've tried to remove job matrix for integration tests but a lot of indexer tests failed because indexer could not catch up. Sad that they're so flaky, but that's not the scope of this PR

lez/storage/src/sequencer/mod.rs
```diff
@@ -44,6 +46,37 @@ pub const DB_LEE_STATE_KEY: &str = "lee_state";
...
+/// Schema-agnostic single-blob snapshot of a store: every key/value pair across all column
+/// families. Lets a prebuilt store ship as one committed file instead of a rocksdb directory.
+#[derive(BorshSerialize, BorshDeserialize)]
+pub struct DbDump {
+    pub entries: Vec<DbDumpEntry>,
+}
+
+impl DbDump {
+    /// Serialize the dump to a borsh blob.
+    pub fn to_bytes(&self) -> DbResult<Vec<u8>> {
```
moudyellaz, line 66: What do you think about gzip/zstd-compressing the dump in to_bytes/from_bytes (or putting the file on LFS)? Borsh blobs like this usually compress several-fold, and since every regeneration re-adds the full 3.8MB to git history permanently, it could be worth it before the first few regenerations pile up.

lez/sequencer/core/src/lib.rs
```diff
@@ -151,15 +152,33 @@ impl<BP: BlockPublisherTrait> SequencerCore<BP> {
...
-                .await
-                .expect("Failed to publish genesis block");
+            let mut pending_blocks = store
+                .get_all_blocks()
+                .filter_ok(|block| matches!(block.bedrock_status, BedrockStatus::Pending))
+                .collect::<Result<Vec<_>, _>>()
+                .expect("Failed to read blocks from store while republishing on fresh start");
+            pending_blocks.sort_unstable_by_key(|block| block.header.block_id);
+
+            debug_assert!(
```
moudyellaz: Maybe worth promoting this debug_assert to a hard error? If a dump ever ships with genesis finalized but later blocks pending, a release build would happily publish a channel without its genesis block and the indexer would just park. Failing loudly here seems cheap insurance.

moudyellaz (approved):

lez/storage/src/sequencer/mod.rs
Arjentix: This is a cool idea. After implementing this with zstd compression level of 19 (1-22 are possible values, but 19 should be enough compaction while not spending too much time on decoding) the binary sized dropped to 448KB, just like ordinary program artifact

lez/sequencer/core/src/lib.rs
Arjentix: Makes sense, changes do `assert!`

Arjentix:
On my machine running just integration tests now takes 12 minutes. Much better than 40 minutes before this PR.

Unfortunately all tests involving indexer are flaky and often fail for me locally. We need to address this in future.

Pravdyvy (approved):
lgtm

PR #591, opened 2026-07-03
feat(circuit): view tag randomization for updated accounts

agureev:
## 🎯 Purpose

This PR, on top of both #550 and #585, fixes issue #573

Namely, the user feeds the view tag directly as the input for the updated accounts. This allows to increase the anonymity pool for a given user with `(npk, vpk)` account, as now the attacker does not necessarily know whether the view tag honestly may point in the user's direction as it may be supplied with a hiding intent.

Due to #585 changes to the wallet syncing mechanism, this does not impact discovery.

## ⚙️ Approach

- [x] Introduce `view_tag` field for all updated accounts
- [x] Remove `view_tag` generation in PPC for updated accounts, replacing it with direct writes
- [x] Change account management to generate new accounts with 

## 🧪 How to Test

## 🔗 Dependencies

#550, #585

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


lez/wallet/src/account_manager.rs
```diff
@@ -442,6 +443,7 @@ impl AccountManager {
                         InputAccountIdentity::PrivateAuthorizedUpdate {
                             vpk: pre.vpk.clone(),
                             random_seed: pre.random_seed,
+                            view_tag: random_view_tag(),
```
moudyellaz, line 434: Shared-account import: if the entry's state is stale relative to blocks already synced (share happens, co-owner updates, import lands after the wallet passed that block), the nullifier watch starts from a spent nullifier and with randomized tags there is no view-tag fallback anymore, so the account desyncs permanently. Should importing a shared entry trigger a rescan from the shared state's block, or should the entry carry its block height?

lee/privacy_preserving_circuit/src/output.rs
```diff
@@ -271,7 +275,7 @@ fn emit_private_output(
     post_state: Account,
     account_id: &AccountId,
     kind: &PrivateAccountKind,
-    npk: &NullifierPublicKey,
+    view_tag: ViewTag,
```
moudyellaz, line 278: Worth a small test asserting the journal's update note carries exactly the supplied view_tag (and an init note still carries the derived one), since that is the feature.

moudyellaz (changes requested):
thanks, requested some changes

lee/privacy_preserving_circuit/src/output.rs
agureev: [7f35c9e](https://github.com/logos-blockchain/logos-execution-zone/pull/591/commits/7f35c9e960ac5cf56842b60f19bd0a7028b10a13) done!

lez/wallet/src/account_manager.rs
agureev: Indeed, this was something I overlooked. Handled in [5ad5263](https://github.com/logos-blockchain/logos-execution-zone/pull/591/commits/5ad5263ac03fb8e9ef1dddbb8ff5aec396783821), launching a catch-up for newly added shared accounts, syncing via nullifiers starting at the initial one deterministically derived from the ID. [c10b912](https://github.com/logos-blockchain/logos-execution-zone/pull/591/commits/c10b9120fac048c5052e1dfd8bed208bc2e3f37c) also adds an explicit command to add a regular shared account with a given identifier which we lacked before

moudyellaz (approved):

lez/wallet/src/lib.rs
```diff
@@ -352,8 +352,60 @@ impl WalletCore {
...
+            return Ok(());
+        }
+
+        info!("Scanning shared account {account_id:#?} from genesis to block {cursor}");
+
+        let mut index = NullifierIndex::default();
+        index.track_initialization(account_id);
+
+        let poller = self.poller.clone();
+        let mut blocks = std::pin::pin!(poller.poll_block_range(1..=cursor));
```
schouhy, line 443: do we need this to scan from genesis in all cases?

schouhy (approved):
lgtm thanks. left one minor comment.

lez/wallet/src/lib.rs
agureev: Not necessarily, there are quite a few optimization that can happen, as Moudy metnioned. This may include being sent a specific block from which to sync as well as a specific state of the account to wait for (e.g. instead of syncing the state of the account from the beginning just sync from a specific state represented by a specific nullifier).

I left out such optimizations from this PR as I think this may require a general conversation about shared-secret-sharing API but I certainly should write an issue on this.

agureev: #636 filed

schouhy: cool thanks!

Issue #562, opened 2026-06-22
Consider Padding Ciphertext Data


agureev:
Currently, all fields of the ciphertext plaintext other than the `Data` field are bound in size, with it being appended at the end of the cipher. What that means is that the ciphertext length is dependent on the size of the account data.

This information may leak some data about the application used. E.g. some applications may use only very small data while others always carry large vectors.  Similarly, if an actor knows an application used, they can use the length of the cipher to possibly discover what kind of data was encoded.

To mitigate this, it may be possible to consider either deterministic (e.g. pad to the next power of 2) or user-provided padding.

jonesmarvin8:
This is an issue that @schouhy and I have discussed before. I would be interested to know how padding affects (1) encryption sizes and (2) privacy proof generation.

PR #592, opened 2026-07-03
feat: dummy note padding

agureev:
## 🎯 Purpose

This addresses the issue of function leakage for private transactions. Namely, previously there could be a link between a program being executed and how many notes are used. This PR allows the users to generate dummy notes to obfuscate how many actual accounts were utilized in a transaction.

Pads explicitly to 7 transactions in-wallet based on the currently supported programs. Seems that AMM is the largest in terms of accounts, supporting 7. Generally, users can decide to pad as much as they want.

## ⚙️ Approach

- [x] Introduce `DummyInput` struct and appropriate nullifier/creation/ciphertext interface
- [x] Make the circuit io accept a vector of dummy inputs
- [x] Make the circuit compute the nullifiers and commitments based on the inputs and write to journal

## 🧪 How to Test

## 🔗 Dependencies

#598, #591, #585, #550

## 🔜 Future Work

Randomize note order, pad ciphertexts to hide data

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


integration_tests/tests/auth_transfer/private.rs
```diff
@@ -84,7 +83,6 @@ async fn private_transfer_to_foreign_account() -> Result<()> {
     let tx = fetch_privacy_preserving_tx(ctx.sequencer_client(), tx_hash).await;
     assert_eq!(tx.message.new_commitments[0], new_commitment1);
 
-    assert_eq!(tx.message.new_commitments.len(), 2);
     for commitment in tx.message.new_commitments {
```
moudyellaz, line 86: Could these become `assert_eq!(tx.message.new_commitments.len(), 7)` instead of being dropped? That would pin the pad-to-max behavior end-to-end, right now nothing outside the wallet unit test checks that the padding actually reaches the chain.

lee/state_machine/core/src/nullifier.rs
```diff
@@ -109,6 +109,14 @@ impl Nullifier {
         bytes.extend_from_slice(account_id.value());
         Self(Impl::hash_bytes(&bytes).as_bytes().try_into().unwrap())
     }
+
+    #[must_use]
+    pub fn for_dummy(nullifier_seed: &[u8; 32]) -> Self {
```
moudyellaz, line 114: Could we add pinned-value tests for `Nullifier::for_dummy` and `Commitment::for_dummy` like the `for_private_pda_matches_pinned_value` ones? These prefixes are consensus-critical once real proving is live, a silent change here should fail a test.

lee/privacy_preserving_circuit/src/output.rs
```diff
@@ -262,7 +263,29 @@ pub fn compute_circuit_output(
...
     }
 
+    for dummy in dummy_inputs {
+        emit_dummy_output(&mut output, dummy);
+    }
+
+    output
+}
+
+fn emit_dummy_output(output: &mut PrivacyPreservingCircuitOutput, dummy: DummyInput) {
```
moudyellaz, line 273: Nothing bounds `dummy_inputs`: the wallet pads to 7, but the circuit and node accept any count, and each dummy permanently grows the commitment tree and nullifier set at near-zero prover cost. Block size caps the tx bytes but not the cumulative state. Should the node enforce a notes-per-tx cap now that note count is decoupled from program semantics, or is this deferred to the fee model?

lez/wallet/src/account_manager.rs
```diff
@@ -627,6 +660,34 @@ fn random_view_tag() -> ViewTag {
...
+    let mut bytes = vec![0; len];
+    OsRng.fill_bytes(&mut bytes);
+    bytes
+}
+
+/// Generates a dummy note: random bytes sized to a default-account ciphertext, a real
+/// ML-KEM ciphertext epk toward a throwaway key, and a random view tag.
+fn random_dummy_note() -> EncryptedAccountData {
+    // Sized to a default-account ciphertext; matching real data sizes is a separate issue.
+    let ciphertext_len = PrivateAccountKind::HEADER_LEN
```
moudyellaz, line 702: Since dummy ciphertexts are default-account-sized and AMM/token notes carry data, ciphertext length still separates real notes from padding for exactly the programs motivating the pad-to-7. Is the ciphertext-padding follow-up planned to land before this padding is relied on for function privacy? Asking about ordering, I see it's acknowledged as future work.

moudyellaz (commented):
Left some comments to consider.

integration_tests/tests/auth_transfer/private.rs
agureev: I agree that we should check that padding reaches the chain, but I am unsure we should replace all the test length assertions with that as padding is not - based on this PR - a protocol-level primitive. It is something that is a wallet-level behavior that should be malleable and not touch other tests based on it.

lee/privacy_preserving_circuit/src/output.rs
agureev: I think this should be deferred to the fee model. This is not much different from launching a program with enormous amounts of accounts used, there does not seem to be an inherent reason to limit this.

What do you think? I can also add a comment/issue/TODO explicitly addressing this.

lez/wallet/src/account_manager.rs
agureev: Ciphertext padding is not something that is planned to lang before this PR is merged, but it is recognized in issue #562 and is independent of this issue is particular.

I was planning to address this in a separate parallel PR this week. Does that sound good?

lee/privacy_preserving_circuit/src/output.rs
moudyellaz: linking this to an issue would be great, thanks!

integration_tests/tests/auth_transfer/private.rs
moudyellaz: Then how about one dedicated test instead, e.g. a single transfer asserting `tx.message.new_commitments.len() == 7 (or MAX_PRIVATE_ACCOUNTS)`? That way the padding-reaches-chain property is pinned in exactly one place, and if the convention changes only that test moves.

lee/state_machine/core/src/nullifier.rs
agureev: Addressed in: [af4797f](https://github.com/logos-blockchain/logos-execution-zone/pull/592/commits/af4797fcd48b4f00bbfa69f083659b20fab68b3d)

integration_tests/tests/auth_transfer/private.rs
agureev: This for now - sadly - has to be an integration test even if the behavior is regarding wallet tx generation, rather than sending. I added the appropriate integration test here: [db40bf7](https://github.com/logos-blockchain/logos-execution-zone/pull/592/commits/db40bf727bb219e8f3113e361eb081ace2ce7b8d), but after #614 this can become a unit test for the wallet.

lee/privacy_preserving_circuit/src/output.rs
agureev: Filed as #622

moudyellaz (approved):
lgtm, thanks.

integration_tests/tests/auth_transfer/private.rs
```diff
@@ -172,7 +171,6 @@ async fn private_transfer_to_owned_account_using_claiming_path() -> Result<()> {
         .context("Failed to get private account commitment for sender")?;
     assert_eq!(tx.message.new_commitments[0], sender_commitment);
 
-    assert_eq!(tx.message.new_commitments.len(), 2);
```
jonesmarvin8, line 175: Should we maintain the assert on commitments size to make sure nothing is silently dropped (including the dummy notes)?

jonesmarvin8: I think this was actually addressed with Moudy's comment; it didn't show due to resolved.

lez/wallet/src/account_manager.rs
```diff
@@ -398,6 +406,30 @@ impl AccountManager {
...
+
+    /// Generate the dummy inputs that pad this transaction's private-account count up to
+    /// `MAX_PRIVATE_ACCOUNTS`.
+    pub fn dummy_inputs_default(&self) -> Vec<DummyInput> {
+        let private_count = self
+            .states
+            .iter()
+            .filter(|state| matches!(state, State::Private(_)))
+            .count();
+        self.dummy_inputs(Self::MAX_PRIVATE_ACCOUNTS.saturating_sub(private_count))
```
schouhy, line 430: let's log a warning when the sub saturates

lee/state_machine/src/privacy_preserving_transaction/circuit/mod.rs
```diff
@@ -69,6 +70,22 @@ pub fn execute_and_prove(
...
+    execute_and_prove_with_dummy_inputs(
+        pre_states,
+        instruction_data,
+        account_identities,
+        vec![],
+        program_with_dependencies,
+    )
+}
+
+pub fn execute_and_prove_with_dummy_inputs(
```
schouhy: can we rename this to `execute_and_prove_with_padded_inputs`? I got confused by the name, I read it at first as "with dummy inputs" as in contrast to "with real inputs".

schouhy (approved):
LGTM thanks!

lee/state_machine/src/privacy_preserving_transaction/circuit/mod.rs
agureev: addressed in 2e8a01a

lez/wallet/src/account_manager.rs
agureev: addressed in [2e8a01a](https://github.com/logos-blockchain/logos-execution-zone/pull/592/commits/2e8a01ac95eab639ef4b0bf698f60aa8de35cacf)

PR #594, opened 2026-07-03
fix(lez/wallet): drop signing with recipient keys

schouhy:
## 🎯 Purpose

Wallet signs deshielded and public transfers with the recipient's key too, leaking that sender and recipient are the same wallet.

## ⚙️ Approach

- Use `AccountIdentity::PublicNoSign` instead of `Public` for deshielded transfers.
- Modify `into_public_identity` to take a `sign` flag argument.

## 🧪 How to Test

Run a deshielded and public transfers to a recipient owned by the same wallet, confirm in explorer only the sender signs.

## 🔗 Dependencies

None

## 🔜 Future Work

None

## 📋 PR Completion Checklist

- [x] Complete PR description
- [ ] Add/update tests

### DIFF
--- a/lez/wallet/src/cli/mod.rs
+++ b/lez/wallet/src/cli/mod.rs
@@ -159,14 +159,22 @@ impl CliAccountMention {
         }
     }
 
+    /// Convert to an [`crate::AccountIdentity`] for use in a public transaction.
+    ///
+    /// The `sign` flag indicates whether to sign or not with the account keys.
     #[must_use]
-    pub fn into_public_identity(self, account_id: lee::AccountId) -> crate::AccountIdentity {
+    pub fn into_public_identity(
+        self,
+        account_id: lee::AccountId,
+        sign: bool,
+    ) -> crate::AccountIdentity {
         match self {
             Self::KeyPath(key_path) => crate::AccountIdentity::PublicKeycard {
                 account_id,
                 key_path,
             },
-            Self::Id(_) | Self::Label(_) => crate::AccountIdentity::Public(account_id),
+            Self::Id(_) | Self::Label(_) if sign => crate::AccountIdentity::Public(account_id),
+            Self::Id(_) | Self::Label(_) => crate::AccountIdentity::PublicNoSign(account_id),
         }
     }
 }
--- a/lez/wallet/src/cli/programs/amm.rs
+++ b/lez/wallet/src/cli/programs/amm.rs
@@ -138,9 +138,9 @@ impl AmmProgramAgnosticSubcommand {
             ) => {
                 let tx_hash = Amm(wallet_core)
                     .send_new_definition(
-                        user_holding_a.into_public_identity(a),
-                        user_holding_b.into_public_identity(b),
-                        user_holding_lp.into_public_identity(lp),
+                        user_holding_a.into_public_identity(a, true),
+                        user_holding_b.into_public_identity(b, true),
+                        user_holding_lp.into_public_identity(lp, true),
                         balance_a,
                         balance_b,
                     )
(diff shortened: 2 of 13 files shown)


lez/wallet/src/cli/programs/amm.rs
```diff
@@ -170,8 +170,8 @@ impl AmmProgramAgnosticSubcommand {
             (AccountIdWithPrivacy::Public(a), AccountIdWithPrivacy::Public(b)) => {
                 let tx_hash = Amm(wallet_core)
                     .send_swap_exact_input(
-                        user_holding_a.into_public_identity(a),
-                        user_holding_b.into_public_identity(b),
+                        user_holding_a.into_public_identity(a, true),
+                        user_holding_b.into_public_identity(b, true),
```
Arjentix, line 174: Hmm, can't you swap on other user account?

lez/wallet/src/cli/mod.rs
```diff
@@ -162,14 +162,22 @@ impl CliAccountMention {
         }
     }
 
+    /// Convert to an [`crate::AccountIdentity`] for use in a public transaction.
+    ///
+    /// The `sign` flag indicates whether to sign or not with the account keys.
     #[must_use]
-    pub fn into_public_identity(self, account_id: lee::AccountId) -> crate::AccountIdentity {
+    pub fn into_public_identity(
```
Arjentix, line 166: Ah, I hoped we've already removed this method, because it makes no sense to me. But probably not the scope of this PR

Arjentix (approved):
LGTM

Pravdyvy (approved):
lgtm

lez/wallet/src/cli/programs/native_token_transfer.rs
```diff
@@ -123,8 +123,8 @@ impl AuthTransferSubcommand {
                 (AccountIdWithPrivacy::Public(from), AccountIdWithPrivacy::Public(to)) => {
                     let to_mention = to_account.expect("matched Some branch");
                     NativeTokenTransferProgramSubcommand::Public {
-                        from: Some(from_account.into_public_identity(from)),
-                        to: Some(to_mention.into_public_identity(to)),
+                        from: Some(from_account.into_public_identity(from, true)),
+                        to: Some(to_mention.into_public_identity(to, false)),
```
moudyellaz, line 127: The red CI is caused by a protocol rule: the first time a program takes ownership of an account (first transfer to a fresh account, mint to a new holder), that account must sign the tx. All 4 failing suites hit this as `ClaimedUnauthorizedAccount`.

So the recipient's signature can't be dropped unconditionally: it's required for new accounts, and only redundant (the leak this PR fixes) for existing ones. Suggestion: decide per recipient, `Public` while the account is still unclaimed, `PublicNoSign` once it exists. The wallet already fetches the recipient account, so the owner check is available.

I verified the existing-recipient side locally on this branch: `successful_transfer_to_existing_account` and `deshielded_transfer_to_public_account` pass, and the transfer tx's witness set contains exactly one signature, the sender's.

lez/wallet/src/cli/mod.rs
moudyellaz, line 166: Could `successful_transfer_to_existing_account` (integration_tests/tests/auth_transfer/public.rs) also assert the privacy property directly? After the balance checks: fetch the blocks, find the public tx whose `account_ids` contain both sender and receiver, and assert `witness_set().signatures_and_public_keys()` maps to exactly `[sender]`. I ran exactly that locally on this branch and it passes, so it's a drop-in guard against recipient signing regressing.

moudyellaz (changes requested):
The existing-recipient flows check out (ran them locally, details inline), but the unconditional no-sign breaks claim-on-first-use flows, which is what the red CI is.

moudyellaz (approved):

lez/wallet/src/cli/programs/amm.rs
schouhy: ah what do you mean?

lez/wallet/src/cli/programs/native_token_transfer.rs
schouhy: I kept the "sign only with the sender" on the wallet behavior. To maintain the spirit of the tests I avoided using the wallet in those and created txs with a method specific for those tests where the transaction contains the signature of the recipient too.

lez/wallet/src/cli/mod.rs
schouhy: done

PR #596, opened 2026-07-04
feat(lez/wallet): add change-network command

schouhy:
## 🎯 Purpose

Adds a way to switch the wallet's network from the CLI.

## ⚙️ Approach

- [x] Add `wallet change-network` command
- [x] Prompt user to pick testnet, local, or custom
- [x] Set sequencer_addr based on the choice
- [x] Make testnet the default network

## 🧪 How to Test

Run `wallet change-network`, pick each option (1, 2, 3), and check the config file has the right `sequencer_addr`.

## 🔗 Dependencies

None

## 🔜 Future Work

None

## 📋 PR Completion Checklist

- [x] Complete PR description
- [ ] Add/update tests

### DIFF
--- a/lez/wallet/src/cli/network.rs
+++ b/lez/wallet/src/cli/network.rs
@@ -1,0 +1,38 @@
+use std::{convert::Infallible, str::FromStr};
+
+use url::Url;
+
+const TESTNET_SEQUENCER_ADDR: &str = "https://testnet.lez.logos.co";
+const LOCAL_SEQUENCER_ADDR: &str = "http://127.0.0.1:3040";
+
+/// A `change-network` argument: `testnet`, `local`, or a custom sequencer URL.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum NetworkAlias {
+    Testnet,
+    Local,
+    Other(String),
+}
+
+impl FromStr for NetworkAlias {
+    type Err = Infallible;
+
+    fn from_str(network: &str) -> Result<Self, Self::Err> {
+        Ok(match network {
+            "testnet" => Self::Testnet,
+            "local" => Self::Local,
+            other => Self::Other(other.to_owned()),
+        })
+    }
+}
+
+impl TryFrom<NetworkAlias> for Url {
+    type Error = url::ParseError;
+
+    fn try_from(alias: NetworkAlias) -> Result<Self, Self::Error> {
+        match alias {
+            NetworkAlias::Testnet => TESTNET_SEQUENCER_ADDR.parse(),
+            NetworkAlias::Local => LOCAL_SEQUENCER_ADDR.parse(),
+            NetworkAlias::Other(url) => url.parse(),
+        }
+    }
+}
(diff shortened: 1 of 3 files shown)


moudyellaz (approved):
lgtm, thanks

lez/wallet/src/cli/network.rs
```diff
@@ -0,0 +1,52 @@
...
+        "3" => prompt("Enter sequencer URL: ")?
+            .parse::<Url>()
+            .context("Invalid sequencer URL")?,
+        other => anyhow::bail!("Invalid choice: {other}"),
+    };
+
+    let mut config = wallet_core.config().clone();
+    config.sequencer_addr = sequencer_addr;
+    wallet_core.set_config(config);
+    wallet_core.store_config_changes().await?;
```
Arjentix: Hm, so this is a config change. And we already have setters for config fields. So these two commands will intersect... Maybe we can integrate this into `wallet config set sequencer_addr`?

erhant: adding to Daniil's comment, I wonder if we could get rid of interactivity all together

- `wallet config set sequencer_addr` is a single command, non-interactive, and similarly:
- `wallet change-network [ testnet | local | <anything-else> ]` would also be non-interactive, and serve the purpose (+ easier to test)

erhant (changes requested):
LGTM with one exception: im not sure about the interactivity requirement of this command, commented at https://github.com/logos-blockchain/logos-execution-zone/pull/596/changes#r3594420664

lez/wallet/src/cli/network.rs
schouhy: I'm fine with either option (interactive or non-interactive). If I remember correctly this was an upstream request to @moudyellaz. Should I change the approach to a non-interactive shorthand of the current config setter?

moudyellaz: @schouhy interactivity wasn't part of the original ask, the goal was automatic default storage/config paths plus making it easy to see and pick which sequencer. The named aliases are the useful bit, so non-interactive `change-network [ testnet | local | <url> ]` works well i think.

schouhy: cool I'll make the change

lez/wallet/src/cli/network.rs
```diff
@@ -0,0 +1,14 @@
+use url::Url;
+
+const TESTNET_SEQUENCER_ADDR: &str = "https://testnet.lez.logos.co";
+const LOCAL_SEQUENCER_ADDR: &str = "http://127.0.0.1:3040";
+
+/// Resolve a `change-network` argument (`testnet`, `local`, or a custom sequencer URL)
+/// to a sequencer URL.
+pub fn resolve_network_alias(network: &str) -> Result<Url, url::ParseError> {
```
erhant: could be a nit; but we can have a enum here with `Testnet, Local, Other(String)` variant that can implement `FromStr` and `TryInto Url`; up to you 🙏🏻

erhant (approved):
LGTM, one nit comment only

lez/wallet/src/cli/network.rs
schouhy: sounds good, added in [b8bb687](https://github.com/logos-blockchain/logos-execution-zone/pull/596/commits/b8bb6876ddcac62831843acec1141109a555949a)

PR #598, opened 2026-07-06
feat!: batched merkle proof fetching

agureev:
## 🎯 Purpose

Previously, the lifecycle of fetching the proofs to consume private accounts involved individual calls to the sequencer. This means that there is a chance - even for an honest sequencer - that the commitments fetched are fetched against different roots. Such requests reveal that a specific actor calling the rpc point most probably: 

1) owns the commitment 
2) is planning to spend it 
3) will spend them in a transactions whose roots will be those that got fetched against

This means that a sequencer appearing to be honest would be able to drastically decrease the anonymity pool of such transactions.

While this PR does not necessarily resolve this leak, it allows to at the very least mask the nullifiers from explicit linking by allowing to fetch against all of the accounts one is planning to consume in one go. This means that even if you were the only one who fetched at that specific root, the anonymity pool for each of the commitments fetched would be at least the amount of private accounts used in the transaction.

This evidently does not work for transactions with one private account, yet this will be resolved by padding notes. Moreover, this helps hide padding notes under a uniform root. Without this, roots used for padding may reveal information.

## ⚙️ Approach

- [x] Introduce a unified endpoint to fetch a vector of merkle proofs and a root against a vector of commitments
- [x] Integrate into wallet functionality 

## 🧪 How to Test

Run integration tests

## 🔗 Dependencies

#585

## 🔜 Future Work

Add padding notes to increase the anonymity pool and add function privacy.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests

### DIFF
--- a/lez/sequencer/service/rpc/src/lib.rs
+++ b/lez/sequencer/service/rpc/src/lib.rs
@@ -6,8 +6,8 @@ use jsonrpsee::types::ErrorObjectOwned;
 #[cfg(feature = "client")]
 pub use jsonrpsee::{core::ClientError, http_client::HttpClientBuilder as SequencerClientBuilder};
 use sequencer_service_protocol::{
-    Account, AccountId, Block, BlockId, ChannelId, Commitment, HashType, LeeTransaction,
-    MembershipProof, Nonce, ProgramId,
+    Account, AccountId, Block, BlockId, ChannelId, Commitment, CommitmentSetDigest, HashType,
+    LeeTransaction, MembershipProof, Nonce, ProgramId,
 };
 
 #[cfg(all(not(feature = "server"), not(feature = "client")))]
@@ -76,11 +76,11 @@ pub trait Rpc {
         account_ids: Vec<AccountId>,
     ) -> Result<Vec<Nonce>, ErrorObjectOwned>;
 
-    #[method(name = "getProofForCommitment")]
-    async fn get_proof_for_commitment(
+    #[method(name = "getProofsAndRoot")]
+    async fn get_proofs_and_root(
         &self,
-        commitment: Commitment,
-    ) -> Result<Option<MembershipProof>, ErrorObjectOwned>;
+        commitments: Vec<Commitment>,
+    ) -> Result<(Vec<Option<MembershipProof>>, CommitmentSetDigest), ErrorObjectOwned>;
 
     #[method(name = "getAccount")]
     async fn get_account(&self, account_id: AccountId) -> Result<Account, ErrorObjectOwned>;
--- a/lez/wallet/src/account_manager.rs
+++ b/lez/wallet/src/account_manager.rs
@@ -7,6 +7,7 @@ use lee_core::{
     Commitment, CommitmentSetDigest, Identifier, InputAccountIdentity, MembershipProof,
     NullifierPublicKey, NullifierSecretKey, SharedSecretKey,
     account::{AccountWithMetadata, Nonce},
+    compute_digest_for_path,
     encryption::ViewingPublicKey,
 };
 use rand::{RngCore as _, rngs::OsRng};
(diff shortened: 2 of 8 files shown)


lez/sequencer/service/rpc/src/lib.rs
```diff
@@ -76,11 +76,11 @@ pub trait Rpc {
         account_ids: Vec<AccountId>,
     ) -> Result<Vec<Nonce>, ErrorObjectOwned>;
 
-    #[method(name = "getProofForCommitment")]
-    async fn get_proof_for_commitment(
+    #[method(name = "getProofsAndRoot")]
```
moudyellaz, line 79: This removes `getProofForCommitment` from the RPC surface and two public `WalletCore` methods, could the commit get the `!` and a BREAKING CHANGE footer, and the PR body a breaking note? No in-org consumers left (checked module/indexer/UI repos), but it is external RPC surface.

lez/wallet/src/account_manager.rs
```diff
@@ -505,7 +488,34 @@ async fn prepare_public_keycard_state(
...
+            State::Public { .. } | State::PublicKeycard { .. } => None,
+        })
+        .unzip();
+
+    let (proofs, root) = wallet
+        .get_proofs_and_root(commitments)
+        .await
+        .map_err(ExecutionFailureKind::SequencerError)?;
+
+    for (pre, proof) in private.iter_mut().zip(proofs) {
```
moudyellaz: Could we add a `proofs.len() == commitments.len()` check before the zip? A short server response currently degrades silently into Init variants and fails much later with an opaque on-chain rejection.

moudyellaz (changes requested):
lgtm, thanks. Left some comments to consider.

lez/sequencer/service/rpc/src/lib.rs
agureev: [1920ca9](https://github.com/logos-blockchain/logos-execution-zone/pull/598/commits/1920ca9427b488badf5211a8c650706363585dd7) now has a bang with explanation, alongside the PR also having a bang.

lez/wallet/src/account_manager.rs
agureev: Good call, generally implemented a validation function making sure that  the return values are trustworthy per the specified semantics; [79b3d62](https://github.com/logos-blockchain/logos-execution-zone/pull/598/commits/79b3d6226ecc91d2deeff51d81956822fb252ce5)

lez/sequencer/service/rpc/src/lib.rs
jonesmarvin8: Could these changes be made note of in [LEZ specs](https://app.notion.com/p/Specs-LEZ-Specs-3928f96fb65c80f1a7a2f452fa4e83a4). This way nothing is overlooked when updating the specs.

lez/wallet/src/account_manager.rs
```diff
@@ -327,8 +327,7 @@ impl AccountManager {
                 } => {
                     let pre = private_shared_acc_preparation(
                         wallet, account_id, nsk, npk, vpk, identifier, true,
-                    )
-                    .await?;
+                    );
```
jonesmarvin8, line 330: Why was `?` dropped?

lez/sequencer/service/rpc/src/lib.rs
agureev: > Could these changes be made note of in [LEZ specs](https://app.notion.com/p/Specs-LEZ-Specs-3928f96fb65c80f1a7a2f452fa4e83a4). This way nothing is overlooked when updating the specs.

Done!

lez/wallet/src/account_manager.rs
agureev: just because `private_shared_acc_preparation`'s type changed. it now does not return `Result` as we fetch proofs separately

jonesmarvin8 (approved):
lgtm. Left a minor comment/question.

moudyellaz (approved):

PR #600, opened 2026-07-07
feat(wallet)!: Multi-sequencer client 

Pravdyvy:
## 🎯 Purpose

Changes into wallet client to accomodate multiple sequencers.

## ⚙️ Approach

- [x] Replaced one URL/Auth pair with list in configuraton.
- [x] Clients now are callibrated if their metrics are absent. 
- [x] Client now handles metric collection. Every sequencer actualized at the start of session, metering happens for every request. Considering the fact, that different requests take different amount of time, metering `get_last_block_id` request.
- [x] Optimal sequencer choice implemented based on a list of characteristics, namely: latest block id, error number, latency average and variance. 
- [x] Leading sequencer is chosen for the duration of one wallet session. 
- [x] To emulate this behaviour, modified FFi to update and store metrics at `wallet_ffi_save` call.

## 🧪 How to Test

So far only unit tests available, at `lez/wallet/src/multi_client.rs`, mocked integration tests will be added later. Current tests are not broken, callibration is almost non-present(1 request) for them. 

## 🔗 Dependencies

None

## Breaking changes

- `WalletCore` initialization is now asynchronous, minor change, does not affect much, mostly needed to support `tokio::fs::File`.
- Wallet config JSON schema changed.
- Wallet FFI functions `wallet_ffi_open` and `wallet_ffi_create_new` now take additional argument `metrics_path` for metrics location.

## 🔜 Future Work

- Create mocked tests to test leader choice closer to reality.
- Add ability to distribute request between multiple sequencers, I think `send_transaction` is a valid candidate for distribution.
- Add CLI and FFI method to add sequencers and check/clean metrics.
- Add parallel sequencers actualization.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


integration_tests/tests/bridge.rs
```diff
@@ -440,8 +440,8 @@ async fn bedrock_deposit_claim_and_withdraw_round_trip_succeeds() -> anyhow::Res
 
     // Now claim funds from vault back to recipient
     let nonces = ctx
-        .wallet()
-        .get_accounts_nonces(vec![recipient_id])
+        .wallet_mut()
+        .get_accounts_nonces(&[recipient_id])
```
Arjentix: I wonder why it requires mutable wallet now

lez/wallet/configs/debug/wallet_config.json
```diff
@@ -1,5 +1,7 @@
 {
-    "sequencer_addr": "http://127.0.0.1:3040",
+    "sequencers_conn_data": [{
```
Arjentix: Hm, let's have more user friendly name, maybe `sequencers`?

lez/wallet/src/config.rs
```diff
@@ -39,20 +48,22 @@ pub struct WalletConfig {
     pub seq_poll_max_retries: u64,
     /// Max amount of blocks to poll in one request.
     pub seq_block_poll_max_amount: u64,
-    /// Basic authentication credentials
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub basic_auth: Option<BasicAuth>,
+    /// Limit number of sequencer polls during callibration, should not be zero
+    pub callibration_limit: usize,
```
Arjentix: I haven't seen this in updated config.json...

Also let's maybe assign default value with serde here?

lez/wallet/src/lib.rs
```diff
@@ -93,45 +99,69 @@ pub struct WalletCore {
...
     storage_path: PathBuf,
 
-    poller: TxPoller,
-    pub sequencer_client: SequencerClient,
+    metrics_path: PathBuf,
+    metrics: HashMap<Url, Metrics>,
+    /// Wrapping metric updates in Arc<RwLock> to not break interfaces too much.
+    ///  
+    /// It is assumed, that wallet methods can be accesed via immutable reference.
+    metric_updates: Arc<RwLock<Vec<MetricsUpdate>>>,
```
Arjentix: Currently I'm not sure about that metrics naming. Soon I plan to start working on real metrics with prometheus and etc. Maybe we can call this `Stats`?

lez/wallet/src/lib.rs
```diff
@@ -187,6 +208,21 @@ impl WalletCore {
...
     }
 
+    #[must_use]
+    pub fn optimal_poller(&self) -> TxPoller {
+        TxPoller::new(self.config(), self.leader_owned())
+    }
+
+    #[must_use]
+    pub fn leader_owned(&self) -> SequencerClient {
+        self.multi_sequencer_client.leader.clone()
```
Arjentix: You access leader by field. Does it mean it's assumed that leader is always fixed here? Maybe better to extract it to a `leader()` method where in will be silently recalculated or something?

lez/wallet/src/lib.rs
```diff
@@ -223,6 +259,43 @@ impl WalletCore {
...
+            .context("Failed to create file")?;
+        file.write_all(&metrics_serialized)
+            .await
+            .context("Failed to write to file")?;
+        file.sync_all().await.context("Failed to sync file")?;
+
+        println!("Stored metrics at {}", self.metrics_path.display());
+
+        Ok(())
+    }
```
Arjentix: Why we need to store them in a file?

Arjentix: Also let's rename to `dump_metrics()` or something like that, because we also have `update_metrics()` which from first point of view have the same meaning as `store_metrics()`

lez/wallet/src/lib.rs
```diff
@@ -414,14 +487,38 @@ impl WalletCore {
...
     pub async fn get_account_balance(&self, acc: AccountId) -> Result<u128> {
-        Ok(self.sequencer_client.get_account_balance(acc).await?)
+        let call_f = async |client: &SequencerClient| client.get_account_balance(acc).await;
+
+        let (call_res, metrics_update) = self.multi_sequencer_client.metered_call(call_f).await;
+
+        {
+            let mut metric_updates_guard = self.metric_updates.write().await;
+            metric_updates_guard.push(metrics_update);
+        }
```
Arjentix: Maybe we can encapsulate metrics inside multiclient? So that we don't have to do that on every external call

lez/wallet/src/main.rs
```diff
@@ -21,7 +20,7 @@ use wallet::{
 async fn main() -> Result<()> {
     let Args {
         continuous_run,
-        auth,
+        auth: _auth,
```
Arjentix: Hm, probably can be removed as argument at all now

lez/wallet/src/cli/mod.rs
```diff
@@ -285,21 +283,25 @@ pub async fn execute_subcommand(
...
                 .context("Transaction finalization error")?
         }
     };
 
+    // Kind of a sledgehammer solution, but it is not clear if there is the case to not store
+    // metrics
+    wallet_core
+        .store_metrics()
+        .await
+        .context("Failed to store metrics")?;
```
Arjentix: Let's move to `main()`?

lez/wallet/src/cli/config.rs
```diff
@@ -76,9 +69,6 @@ impl ConfigSubcommand {
     ) -> Result<SubcommandReturnValue> {
         let mut config = wallet_core.config().clone();
         match key.as_str() {
-            "sequencer_addr" => {
```
Arjentix, line 79: Maybe we need something like `wallet config sequencer add` and `wallet config sequencer remove` then

lez/wallet/src/multi_client.rs
```diff
@@ -0,0 +1,951 @@
...
+    pub latency_var: f32,
+    pub sample_size: usize,
+    pub latest_block_id: u64,
+    pub errors: u64,
+}
+
+#[derive(Debug, Clone)]
+pub struct MetricsUpdate {
+    pub latency: f32,
+    pub new_latest_block_id: Option<u64>,
```
Arjentix:
```suggestion
    pub new_latest_block_id: Option<BlockId>,
```

lez/wallet/src/multi_client.rs
```diff
@@ -0,0 +1,951 @@
...
+use url::Url;
+
+use crate::config::SequencerConnectionData;
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct Metrics {
+    pub latency_avg: f32,
+    pub latency_var: f32,
+    pub sample_size: usize,
+    pub latest_block_id: u64,
```
Arjentix:
```suggestion
    pub latest_block_id: BlockId,
```

lez/wallet/src/multi_client.rs
```diff
@@ -0,0 +1,951 @@
...
+use sequencer_service_rpc::{RpcClient as _, SequencerClient, SequencerClientBuilder};
+use serde::{Deserialize, Serialize};
+use url::Url;
+
+use crate::config::SequencerConnectionData;
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct Metrics {
+    pub latency_avg: f32,
+    pub latency_var: f32,
```
Arjentix, line 24:
```suggestion
    pub latency_avg: Duration,
    pub latency_var: Duration,
```

I think it should be duration, no?

lez/wallet/src/multi_client.rs
```diff
@@ -0,0 +1,951 @@
...
+}
+
+#[derive(Clone)]
+pub struct MultiSequencerClient {
+    // For now we store only leader, it is possible, that
+    // in future for important sends(for example for transactions)
+    // we would want to distribute call between known sequencers
+    pub leader: SequencerClient,
+    pub leader_url: Url,
+}
```
Arjentix, line 88: Agreed, but let's encapsulate that and don't make these pub

lez/wallet/src/multi_client.rs
```diff
@@ -0,0 +1,951 @@
...
+
+                let metric_mut = metrics.get_mut(sequencer_addr).unwrap();
+                metric_mut.apply_updates(&[metric_updates]);
+            // Otherwise actualize client data
+            } else {
+                metrics.insert(
+                    sequencer_addr.clone(),
+                    callibrate_client(&sequencer_client, callibration_limit).await,
+                );
+            }
```
Arjentix: Let's try `metrics.entry()` to not to search twice

lez/wallet/src/multi_client.rs
```diff
@@ -0,0 +1,951 @@
...
+
+    #[must_use]
+    pub const fn leader_ref(&self) -> &SequencerClient {
+        &self.leader
+    }
+
+    #[must_use]
+    pub fn leader_clone(&self) -> SequencerClient {
+        self.leader.clone()
+    }
```
Arjentix, line 192: Let's just have `pub fn leader(&self) -> &SequencerClient` and then caller can call `clone()` if they want

lez/wallet/src/multi_client.rs
```diff
@@ -0,0 +1,951 @@
...
+}
+
+pub async fn actualize_client(client: &SequencerClient) -> MetricsUpdate {
+    let now = tokio::time::Instant::now();
+
+    let block_id = client.get_last_block_id().await.ok();
+
+    #[expect(clippy::as_conversions, reason = "int to float conversion is safe")]
+    let latency = tokio::time::Instant::now().duration_since(now).as_millis() as f32;
```
Arjentix: This code is repeated with `calibrate_client()`. Let's extract it into a separate function and reuse

lez/wallet/src/multi_client.rs
```diff
@@ -0,0 +1,951 @@
...
+    let min_lat_var_addr = client_vec.iter().fold(client_vec[0], |acc, x| {
+        let old = metrics.get(acc).unwrap();
+        let (old_lat, old_var) = (old.latency_avg, old.latency_var);
+        let new = metrics.get(*x).unwrap();
+        let (new_lat, new_var) = (new.latency_avg, new.latency_var);
+
+        let new_std = new_var.sqrt();
+        let old_std = old_var.sqrt();
+
+        // Client is better if its averabe is better and variance does not make it worse
```
Arjentix:
```suggestion
        // Client is better if its average is better and variance does not make it worse
```

Arjentix:
Suddenly it felt so nice to review non AI-generated code

lez/wallet/src/config.rs
erhant: ++

lez/wallet/src/multi_client.rs
```diff
@@ -0,0 +1,951 @@
...
+        let Ok(block_id) = block_id else {
+            errors = errors.saturating_add(1);
+            continue;
+        };
+
+        latest_block_id = block_id;
+        latencies.push(latency);
+    }
+
+    // Precision loss if fine there
```
erhant:
```diff
- if
+ is
```

lez/wallet/src/multi_client.rs
```diff
@@ -0,0 +1,951 @@
...
+    let sample_size = latencies.len();
```
erhant, line 330: if all attempts in `0..callibration_limit` fail, we will get `0` for sample size, and the fields below will be `NaN` and passed into Metrics

a serialized Metrics will write NaN's as `null`, and on re-open a `null` will NOT be deserialized to `f32`, and therefore fail.

a check that `sample_size != 0` could save the trouble here (even if its edge case)

lez/wallet/src/main.rs
erhant: ++

lez/wallet/src/multi_client.rs
```diff
@@ -0,0 +1,951 @@
...
+
+            // If there is no metrics for client, callibrate it
+            if metrics.contains_key(sequencer_addr) {
+                let metric_updates = actualize_client(&sequencer_client).await;
+
+                log::debug!(
+                    "Metered call for {sequencer_addr:?}, metric updates is {metric_updates:?}"
+                );
+
+                let metric_mut = metrics.get_mut(sequencer_addr).unwrap();
```
erhant: maybe `if let Some(m) = metrics.get_mut` would look nice with the `if` above?

lez/wallet/src/multi_client.rs
```diff
@@ -0,0 +1,951 @@
...
+                metric_mut.apply_updates(&[metric_updates]);
+            // Otherwise actualize client data
```
erhant: I think this comment belongs to the `true` branch, and the comment above belongs here (the `false` branch)

lez/wallet/src/config.rs
erhant: the correct spelling is `calibration` without the double-l, probably occurs in other places too in this PR

lez/wallet/src/cli/config.rs
```diff
@@ -37,7 +37,7 @@ impl ConfigSubcommand {
         } else if let Some(key) = key {
             match key.as_str() {
                 "sequencer_addr" => {
```
erhant: should we use `sequencers_conn_data` here as well?

lez/wallet/src/multi_client.rs
```diff
@@ -0,0 +1,951 @@
...
+
+    let max_block_id = metrics.get(max_block_id_addr).unwrap().latest_block_id;
+
+    // Sort out all clients running late
+    client_vec = client_vec
+        .iter()
+        .filter_map(|x| {
+            let latest_block_id = metrics.get(*x).unwrap().latest_block_id;
+
+            (latest_block_id == max_block_id).then_some(*x)
```
erhant, line 394: i wonder if this is too strict, could it be possible that a client is just 1 block behind?
in particular, the client list is constructed at different times instead of parallel, so we could mistakenly remove the "first polled" client because the "last polled" one ended up in the new block maybe

integration_tests/tests/wallet_ffi.rs
```diff
@@ -388,10 +396,18 @@ fn new_wallet_ffi_with_default_config(password: &str) -> Result<FfiCreateWalletO
 fn load_existing_ffi_wallet(home: &Path) -> Result<*mut WalletHandle> {
     let config_path = home.join("wallet_config.json");
     let storage_path = home.join("storage.json");
+    let metrics_path = home.join("metrics.json");
```
erhant, line 376: we should add this to `just clean`

lez/wallet/src/multi_client.rs
```diff
@@ -0,0 +1,951 @@
...
+        })
+        .collect();
+
+    // Get the clients with lesser or equal to average error count
+    #[expect(clippy::as_conversions, reason = "int to float conversion is safe")]
+    let avg_err_count = (client_vec.iter().fold(0_u64, |acc, x| {
+        acc.saturating_add(metrics.get(*x).unwrap().errors)
```
erhant: we are using `errors` here, but it is a lifetime metric that is never normalized / refreshed

after a long run without any resets, this may be erroneous; at least we could normalize it w.r.t sample size (ie. num of sequencers)?

lez/wallet/configs/debug/wallet_config.json
Pravdyvy: Renamed

lez/wallet/src/config.rs
Pravdyvy: Updated config with this value as well.

Pravdyvy: If this one should receive default value, then others must as well? What they should be, then? Let's leave default values for later.

lez/wallet/src/lib.rs
Pravdyvy: Renamed into `Statistics`.

lez/wallet/src/lib.rs
Pravdyvy: Added function.

lez/wallet/src/lib.rs
Pravdyvy: It is basically a part of a persistent data. Don't want to store it in `storage.json`, I don't think it fit there.

Pravdyvy: Renamed into `dump_statistics()`

lez/wallet/src/lib.rs
Pravdyvy: Moved updates into client, main statistics fit better into `WalletCore`.

lez/wallet/src/main.rs
Pravdyvy: Removed

lez/wallet/src/cli/mod.rs
Pravdyvy: One can use wallet as a library, for example we act like this in integration tests. So it is the best place.

Pravdyvy: Also renamed this function twice, `store_metrics` -> `dump_statistics` -> `client_rotation`, now additionally modifies leader at the end of a command.

lez/wallet/src/cli/config.rs
Pravdyvy: Yes, it is in plans to extend cli in one of the next PRs to represent changes.

lez/wallet/src/multi_client.rs
Pravdyvy: Applied

lez/wallet/src/multi_client.rs
Pravdyvy: Applied

lez/wallet/src/multi_client.rs
Pravdyvy: Encapsulated

lez/wallet/src/multi_client.rs
Pravdyvy: No, it would affect calculations. Duration is an integer, statistics must be floats.

lez/wallet/src/multi_client.rs
Pravdyvy: I would like to do it, but actualization and callibration are asyncronous, i can not simply use them in closures.

integration_tests/tests/bridge.rs
Pravdyvy: Removed, artifact of previous variants.

lez/wallet/src/multi_client.rs
Pravdyvy: Modified.

lez/wallet/src/multi_client.rs
Pravdyvy: Extracted, reused.

lez/wallet/src/multi_client.rs
Pravdyvy: Applied.

lez/wallet/src/multi_client.rs
Pravdyvy: Applied

lez/wallet/src/multi_client.rs
Pravdyvy: True. Now return `None` in this cases and drop this sequencer from leaderboard.

lez/wallet/src/multi_client.rs
Pravdyvy: Yes, added

lez/wallet/src/multi_client.rs
Pravdyvy: Swapped

lez/wallet/src/config.rs
Pravdyvy: Fixed

lez/wallet/src/cli/config.rs
Pravdyvy: Yes, in one of previous threads renamed into `sequencers`, renamed here as well.

integration_tests/tests/wallet_ffi.rs
Pravdyvy: Added

lez/wallet/src/multi_client.rs
Pravdyvy: Now use error ratios.

lez/wallet/src/multi_client.rs
Pravdyvy: I would argue that 1 block is enough to fail all your transactions. Imagine you interact(privately) with some popular program, like AMM.
Also in future I would like to distribute "send" requests, like transaction between multiple best sequencers, this check gives me little hope that they are aligned in state.

But there is possible solution, we can actualize all sequencers in parallel. I would leave it as the future task, becasue it is not trivial, considering that we want to collect latencies with some measure of honesty.

erhant: > I would argue that 1 block is enough to fail all your transactions. Imagine you interact(privately) with some popular program, like AMM.

Yes of course, but my point was that perhaps the client is honest but we had polled it long ago so its seen as stale, so maybe we be optimistic about it etc.

I agree that the fitting solution to this is parallel actualization, can be left for later (with maybe a TODO comment / tracker issue)?

Pravdyvy: Added as a `Future work` item. Will create tracking issue after merging.

lez/wallet/src/config.rs
moudyellaz: `callibration_limit` has no serde default and the debug `wallet_config.json` doesn't include it, so loading the shipped config fails to deserialize. Could we add `#[serde(default = "default_callibration_limit")]` (non-zero) and add the key to the JSON? That keeps existing configs working.

lez/wallet/src/multi_client.rs
moudyellaz, line 330: If every calibration poll errors (or `callibration_limit` is 0), `sample_size` is 0, so `latency_avg`/`latency_var` are NaN, get persisted, and make all `choose_leader` comparisons silently false. Guard the empty-sample case so a dead sequencer sorts last.

lez/wallet/src/multi_client.rs
```diff
@@ -0,0 +1,951 @@
...
+        &self.leader
+    }
+
+    #[must_use]
+    pub fn leader_clone(&self) -> SequencerClient {
+        self.leader.clone()
+    }
+
+    // Keeping this call abstract, in case if we need to do more than one request
+    pub async fn metered_call<R, E, I: AsyncFn(&SequencerClient) -> Result<R, E>>(
```
moudyellaz, line 195: `choose_leader` runs once in `new` and `metered_call` always targets `self.leader`, so a leader that fails mid-session bricks every call until restart, and accumulated metrics only take effect on the next open. Intentional for v1 (re-selection is #623), or should repeated failures trigger re-selection here?

lez/wallet/src/multi_client.rs
```diff
@@ -0,0 +1,951 @@
...
+        &self,
+        call: I,
+    ) -> (Result<R, E>, MetricsUpdate) {
+        let resp = tokio::join!(call(self.leader_ref()), actualize_client(self.leader_ref()));
```
moudyellaz: This fires a second `get_last_block_id` alongside every real call, so each wallet op costs two requests to the leader and the recorded latency is the probe's, not the actual call's (send_transaction latency is never measured). Could we time the real call instead of firing a separate probe?

lez/wallet/src/main.rs
moudyellaz: Agreed. Right now `--auth` is still parsed but dropped, so it silently does nothing; auth only comes from per-sequencer config. Let's either remove the flag or wire it to a sequencer entry, so it's not a silent no-op.

lez/wallet/src/multi_client.rs
```diff
@@ -0,0 +1,951 @@
...
+
+    #[expect(clippy::as_conversions, reason = "int to float conversion is safe")]
+    let client_vec = client_vec[..(client_vec
+        .iter()
+        .position(|item| (metrics.get(*item).unwrap().errors as f32) > avg_err_count)
+        .unwrap_or(client_vec.len()))]
+        .to_vec();
+
+    // Choose clients with least latency and variance
+    let min_lat_var_addr = client_vec.iter().fold(client_vec[0], |acc, x| {
```
moudyellaz, line 434: The final tie-break folds a non-transitive comparator over HashMap iteration order, so among sequencers with equal block/errors the chosen leader can vary run to run. Sort by a single key (e.g. `latency_avg + std`) for a stable pick.

lez/wallet/src/lib.rs
```diff
@@ -223,6 +259,43 @@ impl WalletCore {
         Ok(())
     }
 
+    async fn update_metrics(&mut self) -> Result<()> {
```
moudyellaz: Updates are applied under a read lock then cleared under a separate write lock; a `metered_call` pushing between the two is dropped unapplied. Drain under one write lock (`mem::take`) instead.

lez/wallet-ffi/wallet_ffi.h
```diff
@@ -1631,17 +1633,21 @@ struct FfiCreateWalletOutput wallet_ffi_create_new(const char *config_path,
...
  * # Returns
  * - Opaque wallet handle on success
  * - Null pointer on error (call `wallet_ffi_get_last_error()` for details)
  *
  * # Safety
  * All string parameters must be valid null-terminated UTF-8 strings.
+ * `handle` must be a valid wallet handle from `wallet_ffi_create_new` or `wallet_ffi_open`.
  */
-struct WalletHandle *wallet_ffi_open(const char *config_path, const char *storage_path);
+struct WalletHandle *wallet_ffi_open(const char *config_path,
```
moudyellaz, line 1647: Doc block lists a `handle` param this fn doesn't take and drops `storage_path`, which is still in the signature. Regenerate the header.

lez/wallet/src/multi_client.rs
```diff
@@ -0,0 +1,951 @@
...
+                        ))
+                        .collect(),
+                    );
+                }
+                builder
+                    .build(sequencer_addr)
+                    .context("Failed to create sequencer client")?
+            };
+
+            // If there is no metrics for client, callibrate it
```
moudyellaz: Comment here describes the opposite branch (`contains_key` actualizes, the `else` calibrates). Also `let mut client_vec = vec![]` at :299 is immediately reassigned, and `optimal_poller().clone()` at lib.rs:931 clones an already-owned value.

moudyellaz (approved):
lgtm, thanks. Left some comments and questions.

lez/wallet/src/cli/config.rs
Arjentix: Ok, cool

lez/wallet/src/lib.rs
```diff
@@ -414,14 +477,27 @@ impl WalletCore {
...
+        self.statistics.get(sequencer_url)
+    }
+
     /// Get account balance.
     pub async fn get_account_balance(&self, acc: AccountId) -> Result<u128> {
-        Ok(self.sequencer_client.get_account_balance(acc).await?)
+        Ok(self
+            .multi_sequencer_client
+            .metered_call(async |client: &SequencerClient| client.get_account_balance(acc).await)
+            .await?)
```
Arjentix, line 490: Looks nice now, thanks

Arjentix (approved):
LGTM, thanks

lez/wallet/src/config.rs
Pravdyvy: Added serde default

lez/wallet/src/multi_client.rs
Pravdyvy: Now this function return `Option`, with `None` on this case. If client failed all callibration attempts, it just get dropped from a leaderboard.

lez/wallet/src/multi_client.rs
Pravdyvy: Intentional.

lez/wallet/src/multi_client.rs
Pravdyvy: Real calls can take different amounts of time, which makes statistic unstable. We can be comparing client from which we requested 1 block to one from which we requested 100.
Also it is very useful method to call, because it, in addition to latency, can give us block id updates. 

It would be better to meter call proper, but I don't know how to normalize statistic, open to ideas.

lez/wallet/src/main.rs
Pravdyvy: Removed

lez/wallet/src/multi_client.rs
Pravdyvy: I am not sure I understand, relation inside this fold is basically relation on pairs of floats (a,b) such that
```
(a_1, b_1) > (a_2, b_2) <=> (a_1 <= a_2) AND ((a_1 + sqrt(b_1)) < (a_2 + sqrt(b_2)))
```
which is a transitive relation for finite floats.

lez/wallet/src/lib.rs
Pravdyvy: Fixed.

lez/wallet-ffi/wallet_ffi.h
Pravdyvy: Fixed docstring, regenerated the header.

lez/wallet/src/multi_client.rs
Pravdyvy: Fixed all

moudyellaz (approved):
lgtm, thanks for addressing my comments and answering my questions. It is much clear now.

PR #603, opened 2026-07-08
feat!(indexer, sequencer): common two-tip chain state for decentralized sequencing

erhant:
## 🎯 Purpose

Foundation for **decentralized sequencing**: move LEZ from a single sequencer toward a committee of accredited sequencers sharing one channel, using the L1's round-robin turn model. Each sequencer produces a block only on its own turn and on the channel's latest block, applies the blocks its peers produce, and rebuilds its own block when it loses a competing-write race — so every honest node converges on the same chain and state.

This PR delivers the shared foundation plus the sequencer side of that model.

The multi-admin roster-change flow are to-do.

> [!NOTE]
> Boot behavior change: on restart the sequencer replays its stored chain and aborts startup if the result doesn't reproduce the persisted state (e.g. genesis actions were edited in the config) — instead of silently running on a state that diverges from the channel. Reset with `just clean` or restore the original config.

## ⚙️ Approach

Extracted the block validate-then-apply path into a shared crate, generalized it into a two-tier chain state, and wired the sequencer to produce on turn, follow peers, and recover from orphans.

- [x] Add `lez/chain_state` crate: storage-free `apply_block(tip, block, state)` + shared `BlockIngestError` / `StallReason` / `Tip` / `AcceptOutcome`, lifted from the indexer's apply path
- [x] Add two-tier `ChainState`: irreversible `final` tier + reorg-able `head` tier (fed by `adopted` / `orphaned` / `finalized`), with a single `final_stall` on the final tier (an invalid *adopted* block just freezes the head tip and self-heals via reorg/finalization)
- [x] Migrate the indexer onto the shared `apply_block` — removes its duplicate `ingest_error.rs` / `stall_reason.rs`; crate-root re-exports keep the indexer service/protocol unchanged; existing indexer tests pass as-is (parity). #607's retry path (`BlockIngestError::is_retryable`, `AcceptOutcome::RetryableFailure`) was folded into `chain_state` during the rebase so both consumers share it. `validate_against_tip` / `apply_block_to_state` are exported separately so the indexer keeps validating before paying for the scratch-state clone
- [x] Gate sequencer production on turn (`is_our_turn`, wired into the main loop) — produce only when authorized
- [x] Build blocks on the `ChainState` head and make it authoritative — **remove `self.state`** (breaking). `publish_block` now returns the assigned `MsgId`, so our block is applied to the head immediately and deduped when it comes back as `adopted`
- [x] Follow peers: `on_follow` reads `channel_update` (adopted/orphaned) + finalized, applies them to the head via one batched `apply_channel_update` (single head re-derivation for all orphans), and mirrors them to the store via `RocksDBIO::store_followed_blocks` (blocks, tip meta, head state and the final-tier snapshot in one atomic write batch). Both the follow and produce paths persist **under the chain lock**, so disk write order always matches apply order
- [x] Orphan rebuild: return a reverted block's user txs to the mempool so the next on-turn production re-includes them on the new head — via non-blocking `try_push` (a blocking push from the publisher's drive task can deadlock against production, which is the only mempool consumer)
- [x] Restart-safe tier reconstruction: persist a final-tier snapshot (`final_lee_state` + `final_block_meta`) atomically with the follow update that advanced it; on boot, anchor `final` on it (pre-genesis state when absent) and replay stored blocks above it as `head`, so a post-restart orphan of a not-yet-finalized block still reverts. Restored entries carry hash-derived sentinel `MsgId`s (the real ones aren't persisted); orphan/finalize events correlate by block hash
- [x] Finalization hardening: a re-delivered finalized block at/below the final tip is idempotent (`AlreadyApplied`, no spurious stall), and a finalized block chaining on an unfinalized head entry finalizes that prefix first (finality is prefix-monotone). A same-height `adopted` competitor arriving without its orphan sibling is warned about and ignored
- [x] Replace the cloning `state()` with a scoped `with_state(|s| …)` read (no whole-state copy)

Post-review additions:

- [x] Reorg-safe store: `store_followed_blocks` pins the tip meta to the post-update head tip and deletes stale blocks above it in the same batch, so a net-shortening reorg can't leave orphaned payloads that brick restart replay
- [x] Fail-stop persistence: a failed follow persist panics the publisher's drive task, whose termination now trips `SequencerHandle::failed()` via a watch channel; the node halts instead of running with memory ahead of disk (or as a follow-death zombie). 
- [x] `adminConfigureChannel` removed from the RPC entirely: roster changes are a local one-shot (`configure_channel` binary → `post_channel_config`) that signs with `<home>/bedrock_signing_key` and posts the config op straight to bedrock. Authorization is holding the admin key file, matching the L1's trust model (it rejects non-admin signers). Structural validation (non-empty keys, threshold bounds, nonzero `posting_timeframe` / `posting_timeout`, timeout ≥ timeframe) moved into `post_channel_config`. New `--listen-address` flag (default `0.0.0.0` unchanged) to bind the RPC to loopback.
- [x] `V03State` now derives `PartialEq`/`Eq` for production, so the restart consistency check can use `==` (previously it was dervied in `test` only)
- [x] Cleanup pass: single-owner `Tip`/`BlockMeta`/`Block` conversions (`From` impls), shared `StallReason::new`/`escalate` used by both the sequencer and indexer stall paths, deduplicated integration-test client builders

### Overall Tests

```sh
cargo +nightly fmt --all
RISC0_DEV_MODE=1 cargo test -p chain_state -p indexer_core -p sequencer_core -p storage -p mempool
RISC0_DEV_MODE=1 cargo test -p chain_state -p indexer_core -p sequencer_core -p storage -p mempool --all-features
cargo machete
```

- **Indexer parity** — the existing `indexer_core` `block_store` tests (accept / park / recover / idempotent re-delivery), including #607's retry-gate tests, pass unchanged after the migration.
- **`chain_state` unit tests** — `apply_block` (each validation + apply path) and `ChainState` (adopt/advance, bad-block freeze without stall, orphan revert, channel-update reorg, finalize-through, backfill, invalid-finalized stall, finalized re-delivery dedup, prefix finalization, hash-correlated restore/revert).
- **`sequencer_core` tests** — production builds on the head; balance/nonce reads go through `with_state`; restart restores unfinalized blocks as head and recovers from a post-restart orphan; restart re-anchors on the persisted final snapshot.
- **`storage` tests** — followed-block batches land blocks, tip meta, state and the final snapshot atomically; re-deliveries don't demote finalized blocks or clobber state.

### Integration Test

We have an integration test for the multi-sequencer setting:

```sh
RISC0_DEV_MODE=1 cargo test -p integration_tests --release --test multi_sequencer -- --nocapture
```

### Live Demo

Do a fresh start with `just clean` (and remove Sequencer-2 tmpdir if you ran this test before, I'm using `"$TMPDIR/lez-sequencer2"` for it here)

1. **Terminal 1**: `just run-bedrock`
2. **Terminal 2**: `just run-sequencer`: starts solo as the **first sequencer**; its first inscription creates the channel, making it admin. It logs Bedrock signing public key: `<hex1>` at startup.
3. Accredit both keys (once sequencer 1 is producing). From `lez/sequencer/service/`, print each node's pubkey (sequencer 2's key will be created here in this command, before its first boot):

```sh
# <hex1>
cargo run --release -q -p sequencer_service --bin bedrock_pubkey -- configs/debug/sequencer_config.json --home .

# <hex2>
cargo run --release -q -p sequencer_service --bin bedrock_pubkey -- configs/debug/sequencer_config.json --home "$TMPDIR/lez-sequencer2"
```

4. Rotate the committee via the admin RPC (~20 s round-robin turns, 30-slot stall timeout):

```sh
cargo run --release -q -p sequencer_service --bin configure_channel -- configs/debug/sequencer_config.json --home . --keys <hex1>,<hex2> --posting-timeframe 20 --posting-timeout 30
```

5. `Ok` = the bedrock node accepted the signed config op; channel acceptance is still asynchronous — watch for `Turn update: our_turn=…` log lines in the sequencer logs.

6. **Terminal 3**: `just run-sequencer "" "$TMPDIR/lez-sequencer2" 3041` to start a **second sequencer** off the same config; it syncs the existing chain and produces on its turns.

7. **Terminal 4**: `just run-indexer` which serves only finalized blocks, regardless of which sequencer published them.

8. **Terminal 5**: (optional) `just run-explorer` to see the blocks, do some TXes etc.

9. Verify that:
  - both sequencers alternate `our_turn` and produce in their windows
  - a tx submitted to the off-turn sequencer waits in its mempool and lands when its turn arrives
  - heads on ports 3040/3041 track the same chain

```sh
$ curl -s localhost:3041 -H 'content-type: application/json' \
  -d '{"jsonrpc":"2.0","id":1,"method":"getLastBlockId","params":[]}'
# {"jsonrpc":"2.0","id":1,"result":207}
$ curl -s localhost:3040 -H 'content-type: application/json' \
  -d '{"jsonrpc":"2.0","id":1,"method":"getLastBlockId","params":[]}'
# {"jsonrpc":"2.0","id":1,"result":207}
```

## 🔗 Dependencies

None

## 🔜 Future Work

- **Multi-admin roster change** — committee config via `prepare_tx` → collect signatures → `submit_signed_tx`.
- **Sequencer follow-path retry parity** — the indexer retries `is_retryable` (transient) apply failures via `ApplyRetryGate`; the sequencer's `ChainState` currently parks on all failures (documented in `on_follow` and on `AcceptOutcome::RetryableFailure`). Add retry to the follow path.
- **Persistence follow-ups** — the final-tier snapshot + head replay landed in this PR; still open: 
  a. full `BlocksProcessed` atomicity — checkpoint, blocks, state and orphan reverts in one write batch (#633; an orphan-only update currently persists nothing, so a crash there resurrects the orphaned block until its replacement arrives)
  b. durable resubmission of orphaned user txs (#634) — `try_push` can drop on a full mempool, and any halt loses the in-memory mempool
  c. a sidecar `block_id → MsgId` cell so restored head entries stop needing hash-derived sentinel `MsgId`s 
  d. thread the finalized inscription's L1 slot for the invalid-finalized stall (https://github.com/logos-blockchain/logos-blockchain/pull/3147)
  e. the sequencer's in-memory tx-hash cache isn't updated for followed peer blocks (rebuilt from all stored blocks on restart, so live peer-tx lookups may miss; the indexer's persistent index is the query source of truth).

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


lez/storage/src/sequencer/mod.rs
```diff
@@ -612,6 +612,34 @@ impl RocksDBIO {
...
+            .is_some_and(|stored| stored.header.hash == block.header.hash);
+
+        if !already_stored {
+            self.atomic_update(block, &[], vec![], state)?;
+        }
+        if finalized {
+            self.mark_block_as_finalized(block_id)?;
+        }
+        Ok(())
+    }
```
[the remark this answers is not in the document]
erhant: Fixed in https://github.com/logos-blockchain/logos-execution-zone/pull/603/commits/72f148e6210d6cf9a15b98bfcc2bde1c23497e66

`put_block` now updates `latest_block_meta` on `block_id >= last` so a same-height overwrite refreshes the tip hash; the restart path seeds from the reorg winner.

lez/storage/src/sequencer/tests.rs
```diff
@@ -0,0 +1,108 @@
...
+    // A reorg replaces block 2: the competing block wins the slot.
+    let block2b = produce_dummy_block(2, Some(HashType([9; 32])), vec![]);
+    dbio.store_followed_block(&block2b, &state_with_balance(300), false)
+        .unwrap();
+
+    let stored = dbio.get_block(2).unwrap().expect("block 2 is stored");
+    assert_eq!(stored.header.hash, block2b.header.hash);
+    assert!(matches!(stored.bedrock_status, BedrockStatus::Pending));
+    assert_eq!(stored_balance(&dbio), 300);
+}
```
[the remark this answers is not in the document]
erhant: Added assertion for `latest_block_meta.{id, hash}` as well (https://github.com/logos-blockchain/logos-execution-zone/pull/603/commits/72f148e6210d6cf9a15b98bfcc2bde1c23497e66)

lez/chain_state/DESIGN.md
```diff
@@ -0,0 +1,302 @@
...
+`StallReason`, and `Tip`.
+
+```rust
+struct Tip { block_id: u64, hash: HashType, l1_slot: Slot }
+
+enum AcceptOutcome { Applied, AlreadyApplied, Parked(BlockIngestError) }
+```
+
+`Tip` carries `l1_slot` (recorded atomically with the tip) because the anchor /
+chain-consistency logic keys on the inscription slot, not just `(id, hash)`.
```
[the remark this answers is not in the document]
erhant: Updated in https://github.com/logos-blockchain/logos-execution-zone/pull/603/commits/72f148e6210d6cf9a15b98bfcc2bde1c23497e66

lez/sequencer/core/src/lib.rs
```diff
@@ -351,26 +384,35 @@ impl<BP: BlockPublisherTrait> SequencerCore<BP> {
...
             .await
             .context("Failed to publish block to Bedrock")?;
 
+        // Apply our own block to the head with the MsgId the publish assigned it,
+        // so the head advances and the later adopted redelivery dedups.
+        let head_state = {
+            let mut chain = self.chain.lock().expect("chain state mutex poisoned");
+            chain.apply_adopted(this_msg, &block);
+            chain.head_state().clone()
+        };
```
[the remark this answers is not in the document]
erhant: Fixed in https://github.com/logos-blockchain/logos-execution-zone/pull/603/commits/e19afd660cbe0631d247d4d41ca4fb779676b238

Now, persistence only happens when `apply_adopted` returns `Applied`; on a lost race (i.e. `AlreadyApplied`, or `Parked` after a same-height reorg) we warn and skip, letting the follow path persist the canonical block.

The re-queue logic is already implemented: Zone SDK surfaces our invalidated inscription via `channel_update.orphaned`, and `apply_follow_update` re-queues its user txs unconditionally, so we don't need to requeue here.

lez/sequencer/core/src/lib.rs
```diff
@@ -578,6 +638,81 @@ struct BlockWithMeta {
...
+async fn apply_follow_update(
+    dbio: Arc<RocksDBIO>,
+    chain: Arc<Mutex<ChainState>>,
+    mempool_handle: MemPoolHandle<(TransactionOrigin, LeeTransaction)>,
+    update: block_publisher::FollowUpdate,
+) {
+    // Apply under the lock and collect what to persist; take a single
+    // head snapshot. Release the lock before touching disk so the
+    // producer is never blocked on the follow path's I/O.
+    let (adopted, finalized, resubmit_txs, head_snapshot) = {
```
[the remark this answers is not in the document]
erhant: Going with option 2: `store_followed_blocks` now persists the whole update (blocks, tip meta, state after the last block) in a single RocksDB `WriteBatch`, so a crash can no longer leave the stored state ahead of the stored blocks.

Per-block state snapshots were rejected on cost (a full `V03State` clone per block during backfill batches). 

NOTE: while fixing this a sibling crash window was found: the `zone-sdk` checkpoint is persisted before the follow write, so a crash in between resumes past blocks that never landed. That needs checkpoint+blocks+state in one batch (the sdk's documented event contract) and is deliberately left as a follow-up (TODO in code).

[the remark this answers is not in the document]
erhant: Fixed in https://github.com/logos-blockchain/logos-execution-zone/pull/603/commits/d927d3591b607cb0e416a6bb171d6e29f2a4fc91

lez/sequencer/core/src/lib.rs
```diff
@@ -578,6 +753,111 @@ struct BlockWithMeta {
...
+            .collect();
+
+        let mut final_advanced = false;
+        for (this_msg, block) in &finalized {
+            // FIXME: thread the finalized inscription's L1 slot once the
+            // sdk surfaces it; only used for the invalid-finalized stall.
+            if matches!(
+                chain.apply_finalized(*this_msg, block, Slot::from(0)),
+                AcceptOutcome::Applied
+            ) {
```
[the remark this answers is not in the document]
erhant: Opened https://github.com/logos-blockchain/logos-blockchain/issues/3139 about this

lez/storage/src/sequencer/mod.rs
```diff
@@ -612,6 +643,92 @@ impl RocksDBIO {
...
+                batch_tip = Some(BlockMeta {
+                    id,
+                    hash: to_write.header.hash,
+                });
+            }
+        }
+
+        if batch.is_empty() && final_snapshot.is_none() {
+            return Ok(());
+        }
```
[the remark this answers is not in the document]
erhant: Known limitation: this is future-work item (3.a) in the PR description ("an orphan-only update currently persists nothing, so a crash there resurrects the orphaned block until its replacement arrives").

Note that convergence is not lost permanently: the resurrected head is warn-ignored by the same-height adopted guard, but the replacement's finalization rebases the head onto the finalized chain (`apply_finalized_direct`, covered by `finalized_unknown_block_rebases_head`), so the divergence is bounded by finality.

lez/storage/src/sequencer/mod.rs
```diff
@@ -42,6 +43,10 @@ pub const DB_META_UNSEEN_WITHDRAW_COUNT_KEY: &str = "unseen_withdraw_count";
 
 /// Key base for storing the LEE state.
 pub const DB_LEE_STATE_KEY: &str = "lee_state";
+/// Key base for storing the LEE state at the last L1-finalized block.
+pub const DB_FINAL_LEE_STATE_KEY: &str = "final_lee_state";
```
Arjentix, line 52: We already store state in db. And I think we do that only on checkpoints from zone sdk. So I wonder why this is also needed.

If it's really needed I'd think how can we unify it with existing state stored in db so that we don't store the state twice.

lez/storage/src/sequencer/mod.rs
```diff
@@ -516,6 +527,26 @@ impl RocksDBIO {
             .map(|opt| opt.map(|val| val.0))
     }
 
+    /// `(state, meta)` at the last L1-finalized block; `None` until the first
+    /// finalization is observed.
+    pub fn get_final_snapshot(&self) -> DbResult<Option<(V03State, BlockMeta)>> {
+        let Some(meta) = self.get_opt::<FinalBlockMetaCellOwned>(())? else {
+            return Ok(None);
+        };
+        let state = self.get::<FinalLeeStateCellOwned>(())?;
```
Arjentix, line 584: Hmmm, theoretically we can get a race here, isn't it? You do two reads and a write can happen in between?

Maybe we need [`snapshot`](https://docs.rs/rocksdb/latest/rocksdb/type.Snapshot.html) here.

Probably we have this problem not only here

lez/storage/src/sequencer/mod.rs
```diff
@@ -612,6 +643,92 @@ impl RocksDBIO {
...
+    /// TODO: the zone-sdk checkpoint is persisted by `on_checkpoint` *before*
+    /// this write. Full `BlocksProcessed` atomicity (checkpoint, blocks, state
+    /// and orphan reverts in one batch) is a follow-up.
+    pub fn store_followed_blocks(
+        &self,
+        blocks: &[(&Block, bool)],
+        state: &V03State,
+        final_snapshot: Option<(&V03State, &BlockMeta)>,
+    ) -> DbResult<()> {
+        let last_block_in_db = self.get_meta_last_block_in_db()?;
```
Arjentix, line 733: Hmm, so in this function you first read and then write. Also possible to have a race here as I understand. We probably need to revisit the whole db interaction. Maybe in another PR

lez/storage/src/sequencer/mod.rs
```diff
@@ -612,6 +643,92 @@ impl RocksDBIO {
...
+            let already_stored = stored
+                .as_ref()
+                .is_some_and(|stored| stored.header.hash == block.header.hash);
+
+            if already_stored && !finalized {
+                continue;
+            }
+
+            let mut to_write = if already_stored {
+                stored.expect("`already_stored` implies a stored block")
```
Arjentix: You first check for is_some() and then call expect(), better to have one `if let Some()` instead

lez/storage/src/sequencer/mod.rs
```diff
@@ -612,6 +643,92 @@ impl RocksDBIO {
...
+        if batch.is_empty() && final_snapshot.is_none() {
+            return Ok(());
+        }
+        // Meta is staged once for the batch's highest block.
+        // (`>=` so a same-height overwrite refreshes the tip hash, matching `put_block`)
+        if let Some(tip) = batch_tip
+            && tip.id >= last_block_in_db
+        {
+            self.put_meta_last_block_in_db_batch(tip.id, &mut batch)?;
+            self.put_meta_latest_block_meta_batch(&tip, &mut batch)?;
```
Arjentix: Hmmm, so this happens for every tx, but it's done in a batch. Meaning that intermediate writes like latest block meta will be overwritten by the future writes. Maybe we can do it once in the end?

lez/chain_state/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+
+anyhow.workspace = true
+log.workspace = true
+serde.workspace = true
+thiserror.workspace = true
+
+[dev-dependencies]
+testnet_initial_state.workspace = true
+# we use borsh to compare byte-to-byte matching within tests
+# (it sorts hashmap's before serialization, so we can compare two states for equality)
```
Arjentix:
```suggestion
```

lez/chain_state/src/chain.rs
```diff
@@ -0,0 +1,872 @@
...
+///
+/// Only the final tier stalls: an invalid `adopted` block just freezes the
+/// head tip and self-heals via reorg or finalization.
+pub struct ChainState {
+    final_state: V03State,
+    final_tip: Option<Tip>,
+    final_stall: Option<StallReason>,
+
+    head_state: V03State,
+    head_blocks: Vec<HeadEntry>,
```
Arjentix, line 34: Having two states seems a bit strange to me, why do we need this?

lez/chain_state/src/chain.rs
```diff
@@ -0,0 +1,872 @@
...
+                    block: block.clone(),
+                });
+                AcceptOutcome::Applied
+            }
+            Err(err) => AcceptOutcome::Parked(err),
+        }
+    }
+
+    /// Reverts an orphaned head block and everything after it, then re-derives head.
+    pub fn revert_orphan(&mut self, this_msg: MsgId, block: &Block) {
```
Arjentix, line 181: When this might be needed?

lez/sequencer/core/src/lib.rs
```diff
@@ -58,11 +66,12 @@ impl DepositMetadata {
 }
 
 pub struct SequencerCore<BP: BlockPublisherTrait = ZoneSdkPublisher> {
-    state: lee::V03State,
+    /// Two-tier chain state: production builds on its head; the publisher's
+    /// `on_follow` sink feeds adopted/orphaned/finalized peer blocks into it.
+    chain: Arc<Mutex<ChainState>>,
```
Arjentix, line 68: Hmm, maybe better RwLock for reading rpc queries?

lez/sequencer/core/src/lib.rs
```diff
@@ -119,18 +128,83 @@ impl<BP: BlockPublisherTrait> SequencerCore<BP> {
...
+        // The replayed head must reproduce the persisted state byte-for-byte,
+        // else store and config disagree (e.g. edited genesis actions).
+        let replayed = borsh::to_vec(chain.head_state()).expect("state serializes");
+        let stored = borsh::to_vec(stored_head_state).expect("state serializes");
+        assert_eq!(
+            replayed, stored,
+            "Persisted state does not match the replayed chain; reset the store or restore the original config"
+        );
+
+        chain
```
Arjentix: Hm, I'd not use byte-for-byte comparison, why not to just use rust's `==`?

While byte-for-byte may sound more safe I don't think it actually saves us from something. For example we still rely on **logical** equivalence of the states, we don't operate on bytes. Also we may forget to encode something and that won't be catched

lez/sequencer/core/src/lib.rs
```diff
@@ -149,6 +223,7 @@ impl<BP: BlockPublisherTrait> SequencerCore<BP> {
             Self::on_finalized_block(store.dbio()),
             Self::on_deposit_event(store.dbio(), mempool_handle.clone()),
             Self::on_withdraw_event(store.dbio()),
+            Self::on_follow(store.dbio(), Arc::clone(&chain), mempool_handle.clone()),
         )
```
Arjentix, line 203: Hm, I'm looking just at this and I'm not sure about passing chain here...

I'd prefer if we continue structuring things like producer-consumer. It works well with mempool. Maybe we can just have the same thing but for tx from bedrock / other sequencers. But the difference would be that they will be applied immediately and not by timeout like we do with the mempool

lez/sequencer/core/src/lib.rs
```diff
@@ -335,6 +409,18 @@ impl<BP: BlockPublisherTrait> SequencerCore<BP> {
...
 
+    /// Publisher sink adapter over [`apply_follow_update`].
+    fn on_follow(
+        dbio: Arc<RocksDBIO>,
+        chain: Arc<Mutex<ChainState>>,
+        mempool_handle: MemPoolHandle<(TransactionOrigin, LeeTransaction)>,
+    ) -> block_publisher::OnFollowSink {
+        Box::new(move |update: block_publisher::FollowUpdate| {
+            apply_follow_update(&dbio, &chain, &mempool_handle, update);
+            Box::pin(std::future::ready(()))
```
Arjentix: If it's sync you don't have to make `OnFollowSink` async

lez/sequencer/core/src/lib.rs
```diff
@@ -565,10 +713,37 @@ impl<BP: BlockPublisherTrait> SequencerCore<BP> {
...
+                withdraw_threshold,
+            )
+            .await
+    }
+
+    /// Shared handle to the two-tier follow state.
+    #[must_use]
+    pub fn chain(&self) -> Arc<Mutex<ChainState>> {
+        Arc::clone(&self.chain)
     }
```
Arjentix: Not sure why would anyone need this. This would allow callers to lock our chainstate which will prevent updates, sounds scary

lez/sequencer/core/src/block_publisher.rs
```diff
@@ -41,6 +50,40 @@ pub type OnDepositEventSink =
...
+/// Commands the drive task executes with `&mut sequencer`.
+enum Command {
+    /// Publish an inscription (+ atomic withdrawals); responds with the assigned `MsgId`.
+    Publish {
+        inscription: Inscription,
+        withdrawals: Vec<WithdrawArg>,
+        resp: oneshot::Sender<Result<MsgId>>,
+    },
+    /// Post a `ChannelConfig` op replacing the accredited keys / rotation params.
+    ConfigureChannel {
```
Arjentix: Everyday we become closer to Actor architecture

lez/sequencer/service/src/service.rs
```diff
@@ -201,8 +205,68 @@ impl<BC: BlockPublisherTrait + Send + 'static> sequencer_service_rpc::RpcServer
         let channel_id = self.sequencer.lock().await.block_publisher().channel_id();
         Ok(ChannelId(*channel_id.as_ref()))
     }
+
+    async fn admin_configure_channel(
+        &self,
+        request: ConfigureChannelRequest,
+    ) -> Result<(), ErrorObjectOwned> {
+        request.validate().map_err(invalid_params)?;
```
Arjentix: Let's perform validation on deserialization step

lez/sequencer/service/Cargo.toml
```diff
@@ -1,6 +1,7 @@
 [package]
 name = "sequencer_service"
 version = "0.1.0"
+default-run = "sequencer_service"
```
Arjentix, line 4: TIL about default-run. Why do you need this btw?

lez/storage/src/sequencer/mod.rs
```diff
@@ -612,6 +643,92 @@ impl RocksDBIO {
...
+        }
+        self.put_lee_state_in_db_batch(state, &mut batch)?;
```
moudyellaz, line 772: This `lee_state` write is unconditional, but the meta write just above is height-guarded (`tip.id >= last_block_in_db`). Since produce (`atomic_update`) and follow (`store_followed_blocks`) race on this same key with only the ChainState lock (released before the write), a late follow write for a lower block can leave `lee_state` behind the stored meta, and restart's `restore_chain_state` assert then panics in a loop. Could we gate the state write on tip height too, or hold the lock across the write?

lez/sequencer/core/src/lib.rs
```diff
@@ -578,6 +753,111 @@ struct BlockWithMeta {
...
+
+        // User txs of orphaned blocks, returned to the mempool below.
+        let resubmit_txs: Vec<LeeTransaction> = orphaned
+            .iter()
+            .flat_map(|(_, block)| resubmittable_txs(block))
+            .collect();
+
+        // Outcomes align with `adopted`.
+        let outcomes = chain.apply_channel_update(&orphaned, &adopted);
+        let mut to_persist: Vec<(&Block, bool)> = adopted
```
moudyellaz, line 978: `to_persist` is adopted-only, so a net-shortening reorg (orphaned 8,9,10, shorter adopted 8') overwrites id 8 and writes `lee_state@8'` but leaves orphaned payloads 9,10 on disk. Restart then replays 8',9,... and panics on 9's broken link every boot, a permanent brick. Could we delete ids in `(new_head_tip, old_max]` in the same batch? No test covers this today.

lez/sequencer/core/src/lib.rs
```diff
@@ -578,6 +753,111 @@ struct BlockWithMeta {
...
+    // One atomic write for the whole update: blocks, tip meta and the state
+    // after the last block land together, so a crash can never leave the
+    // stored state ahead of the stored blocks.
+    //
+    // TODO: the zone-sdk checkpoint is persisted by `on_checkpoint` before
+    // this write; a crash in between resumes past these blocks without them
+    // ever landing in the store. Full `BlocksProcessed` atomicity (checkpoint
+    // + blocks + state in one batch, per the sdk's event contract) is a
+    // follow-up.
+    if let Err(err) = dbio.store_followed_blocks(
```
moudyellaz: The in-memory ChainState was already advanced under the lock above, so if this persist fails we've only logged it, with no rollback or halt, and memory is now ahead of disk. The next update builds on the advanced chain and can skip blocks that never landed, so restart comes up behind and forks. Could we halt or retry here instead of logging through?

lez/sequencer/core/src/lib.rs
```diff
@@ -578,6 +753,111 @@ struct BlockWithMeta {
...
+            chain.head_state().clone(),
+            final_snapshot,
+        )
+    };
+
+    // One atomic write for the whole update: blocks, tip meta and the state
+    // after the last block land together, so a crash can never leave the
+    // stored state ahead of the stored blocks.
+    //
+    // TODO: the zone-sdk checkpoint is persisted by `on_checkpoint` before
```
moudyellaz: This TODO's window is a real data-loss path: on_checkpoint persists the zone-sdk cursor before this write, and is_fresh_start derives from checkpoint presence, so a crash in between makes restart resume past blocks that never landed, needing an operator resync. Could we get a tracking issue for the BlocksProcessed atomicity if it's not folded in here?

lez/sequencer/service/src/service.rs
```diff
@@ -201,8 +205,68 @@ impl<BC: BlockPublisherTrait + Send + 'static> sequencer_service_rpc::RpcServer
         let channel_id = self.sequencer.lock().await.block_publisher().channel_id();
         Ok(ChannelId(*channel_id.as_ref()))
     }
+
+    async fn admin_configure_channel(
```
moudyellaz: This has no caller auth and the RPC server binds `0.0.0.0` (lib.rs:156), yet it signs a committee-roster change with this node's admin key. Since a first/solo sequencer is the channel admin by default, anyone who can reach the port can swap the committee keys and evict the honest node. Could we bind to loopback / gate it behind auth, and note the trust assumption on the method?

lez/sequencer/core/src/lib.rs
```diff
@@ -578,6 +753,111 @@ struct BlockWithMeta {
...
+        error!("Failed to persist followed blocks: {err:#}");
+    }
+
+    // Rebuild orphaned work: return its user txs to the mempool so the
+    // next on-turn production re-includes them on the new head.
+    //
+    // We use [`try_push`] here because this is called from the
+    // publisher's drive task, and only the block production drains the mempool.
+    // A blocking push on a full mempool would deadlock here.
+    for tx in resubmit_txs {
```
moudyellaz, line 1034: Two loss paths for orphaned user txs here: `try_push` warns and drops on a full mempool, and txs popped into a block whose `publish_block` then fails are never put back (the node exits instead). A mined-then-orphaned user tx can vanish while the client thinks it landed. A durable resubmit path would close it.

lez/sequencer/core/src/lib.rs
```diff
@@ -771,6 +1060,19 @@ fn build_bridge_deposit_tx_from_event(event: &PendingDepositEventRecord) -> Resu
     )))
 }
 
+/// User transactions of an orphaned block to return to the mempool: everything
+/// except the trailing clock tx and sequencer-generated bridge deposits (those are
+/// replayed from their own bedrock events, not the mempool).
+fn resubmittable_txs(block: &Block) -> Vec<LeeTransaction> {
```
moudyellaz, line 1248: This excludes only the clock tx and bridge deposits, so an orphaned cross-zone inbox tx (Sequencer origin, sequencer-only program) gets re-injected as User origin, hitting the path the ingress guard rejects and duplicating the watcher's L1 replay. Could we also exclude `is_sequencer_only_program` / all non-User txs here?

lez/chain_state/src/chain.rs
```diff
@@ -0,0 +1,872 @@
...
+            return AcceptOutcome::AlreadyApplied;
+        }
+
+        let mut scratch = self.final_state.clone();
+        match apply_block(self.final_tip.as_ref(), block, &mut scratch) {
+            Ok(()) => {
+                self.final_state = scratch;
+                self.final_tip = Some(tip_of(block));
+                self.final_stall = None;
+                self.head_blocks.clear();
```
moudyellaz, line 283: This rebase drops every head block above the finalized one, but `apply_follow_update` only resubmits from the `orphaned` list, not from a suffix dropped here. A finalized competitor arriving without a preceding orphaned event for our suffix (backfill path) silently loses our own blocks' user txs. Safe within one well-formed BlocksProcessed (orphaned processed first), but no safety net if ordering differs. Worth a guard or a comment pinning the assumption?

moudyellaz (changes requested):
thanks for this PR, left some inline comments (if i understood the approach), they are all related to durability. I will leave here also some non blocking comments.
- Orphan resubmit runs over peers' orphaned blocks too (lib.rs:784), committee-wide mempool bloat.
- apply_finalized_direct below-tip idempotency arm is hash-blind (chain.rs:242), asymmetric with the equal-height arm.
- ConfigureChannelRequest::validate doesn't bound posting_timeframe/timeout (protocol/src/lib.rs:44); 0 or timeout < timeframe passes and can stall turns.
- Turn gating is untested (mock hardcodes true) and block_create_timeout > turn window can starve a node's own turn.
- Indexer dropped the ingest_error/stall_reason module paths; types are re-exported so no wire break, but module-path imports break.

I also verified these and they work pretty fine: getStatus and on-disk StallReason/BlockIngestError formats are preserved, the RPC surface is purely additive (no JSON wire break), the apply.rs migration is behavior-identical, the spin bump is a legit yanked-crate fix, and multi_sequencer is a meaningful test.

lez/chain_state/src/chain.rs
erhant: This is the single entire point of this PR actually:
- we need one state to keep track of the final state, that we keep building over as we receive finalized blocks
- we need the other head state to keep the actual latest state built from `finalState + [...unfinalizedBlocks]`

when we get orphaned / adopted tx'es, we are able to rollback / build-over the existing final state to reconstruct the correct head state, until we actually receive the finalized ones to update our finalized state on top

lez/storage/src/sequencer/mod.rs
erhant: Similar to https://github.com/logos-blockchain/logos-execution-zone/pull/603#discussion_r3613298019

We do need two states, one final and one head state. Storing them on disk besides memory allows us to continue on restart from where we have left off, e.g. `final_state -> ...txs -> head_state`.

lez/sequencer/service/Cargo.toml
erhant: I needed a separate binary `bedrock_pubkey` which is a pubkey helper for committee accreditation, needed for the live-demo locally (see PR description)

when you have multiple binaries, `cargo run -p sequencer_service` needs a specific binary name to run, but if you have a default that you always run you can just specify that in `default-run` and it will select it for you, so this way `just run-sequencer` etc. works as before with no modifications.

lez/sequencer/core/src/lib.rs
erhant: Yeah, will drop `pub` 👌🏻

lez/storage/src/sequencer/mod.rs
erhant: Will do 🙏🏻

lez/chain_state/src/chain.rs
erhant: Basic orphaning, consider the scenario:

```ml
final --> tx1 --> tx2 --> tx3 --> head
```

Then, consider getting `tx2` to be orphaned; meaning that it is no longer part of the head state and you should drop it. Since this is a "chain" we must truncate the state upto that point, and keep the rest:

```ml
                   cut here
final --> tx1 -/- tx2 --> tx3 --> head

(* new head *)
final --> tx1 --> head
```

lez/sequencer/core/src/lib.rs
erhant: Could be, but the critical sections are short & `follow*` path writes frequently. RwLock could cause writer starvation under heavy RPC load. I suggest we keep it, and measure under heavy load with a tracker issue if we want to.

lez/sequencer/core/src/lib.rs
erhant: `V03State` does not have a derived `PartialEq` (it has only when `test`ing), so `==` is not there. If its ok we can enable that for all, but I thought there was a reason to keep it testing only.

> Also we may forget to encode something and that won't be catched

I didn't get this one, you mean like if we have a field that we override its borsh traits to not actually encode/decode?

lez/sequencer/core/src/lib.rs
erhant: Agreed, will fix

lez/sequencer/service/src/service.rs
erhant: Will do!

erhant: https://github.com/logos-blockchain/logos-execution-zone/pull/603/commits/6eb14350625b28f73911304ce12bd26bd58eef83

erhant: I did the usual `try_from` for this, but I still kept `validate` as a function on its own; easier tests & could be used on builder-pattern if we need to construct that as well if needed.

lez/sequencer/core/src/lib.rs
erhant: https://github.com/logos-blockchain/logos-execution-zone/pull/603/commits/6eb14350625b28f73911304ce12bd26bd58eef83

lez/sequencer/core/src/lib.rs
erhant: https://github.com/logos-blockchain/logos-execution-zone/pull/603/commits/6eb14350625b28f73911304ce12bd26bd58eef83

lez/storage/src/sequencer/mod.rs
erhant: https://github.com/logos-blockchain/logos-execution-zone/pull/603/commits/6eb14350625b28f73911304ce12bd26bd58eef83

lez/storage/src/sequencer/mod.rs
erhant: Fixed in https://github.com/logos-blockchain/logos-execution-zone/pull/603/commits/c8dd76a94b9b554a9bb77faca70ee2dd792bb66f, now holds the lock across the write. Both paths now persist under the `chain` lock, so disk writes land in order.

Comes with a bonus of saving us two per-update `V03State` clones too.

A small trade off: RPC reads via `with_state` can briefly block behind a RocksDB batch write (win for correctness at the cost of a small latency)

lez/sequencer/core/src/lib.rs
erhant: Good catch...

Now we take the post-update head tip as well, and drop those above it. Added a new `net_shortening_reorg_drops_stale_blocks` test for it too.

Added in https://github.com/logos-blockchain/logos-execution-zone/pull/603/commits/c8dd76a94b9b554a9bb77faca70ee2dd792bb66f

lez/sequencer/core/src/lib.rs
erhant: I've changed it to `panic` for now, but will change to a better behavior.

lez/sequencer/service/src/service.rs
erhant: Added a `--listen-address` arg to Sequencer so we have a way out of 0.0.0.0 by default. (https://github.com/logos-blockchain/logos-execution-zone/pull/603/commits/c8dd76a94b9b554a9bb77faca70ee2dd792bb66f)

Added comment https://github.com/logos-blockchain/logos-execution-zone/pull/603/commits/d417a4e828aee5423b2ddaf6da65dea6b38ee787

lez/sequencer/core/src/lib.rs
erhant: `try_push` here is deliberate, otherwise we get a deadlock. Will open a issue to come up with a better handling here (goes a bit beyond this PR).

lez/chain_state/src/chain.rs
erhant: Added comment in https://github.com/logos-blockchain/logos-execution-zone/commit/c8dd76a94b9b554a9bb77faca70ee2dd792bb66f

lez/sequencer/core/src/lib.rs
erhant: Ah nice catch, added a `is_sequencer_only_tx` function and now checks for it here. https://github.com/logos-blockchain/logos-execution-zone/commit/c8dd76a94b9b554a9bb77faca70ee2dd792bb66f

lez/storage/src/sequencer/mod.rs
erhant: Fixed in https://github.com/logos-blockchain/logos-execution-zone/pull/603/commits/c8dd76a94b9b554a9bb77faca70ee2dd792bb66f, now both store-writes run under the same chian lock.

I agree that a broader PR is more fitting the revisit for the whole DB thing.

lez/storage/src/sequencer/mod.rs
erhant: Theoretically yes, but this is only called on start once; so there wont be any race.

Regardless, this should be revisited (as also mentioned at https://github.com/logos-blockchain/logos-execution-zone/pull/603#discussion_r3604027356) with a "db revisit" PR.

lez/storage/src/sequencer/mod.rs
erhant: It's done per block, and the meta keys are staged once outside the loop.

Also in https://github.com/logos-blockchain/logos-execution-zone/commit/c8dd76a94b9b554a9bb77faca70ee2dd792bb66f the meta is pinned to the head tip passed in by the caller, so there is no per-iteration tracking at all now too.

lez/sequencer/core/src/lib.rs
erhant: I see you are leaning towards actors here too heheh

Follow-updates have to apply synchronously in event order before the next head-tick here, and a queue to do that here would complicate things a bit too I think; if done for this part alone*

I would rather keep this as is right now, perhaps we can revisit the actor-based approach to do a general overhaul on the sequencer?

lez/sequencer/core/src/lib.rs
erhant: Created the issue https://github.com/logos-blockchain/logos-execution-zone/issues/633

lez/sequencer/core/src/lib.rs
erhant: Improved in https://github.com/logos-blockchain/logos-execution-zone/pull/603/commits/b34ab22b64460317ab6bfefa9a75aa0017f2ddfa: the persist failure still `panic`s, but the publisher's drive task now signals its termination through a watch channel that `SequencerHandle`'s `failed` method selects on, so the panic reliably halts the whole node through the normal shutdown path.

erhant: Still needs a proper follow-up (opened on https://github.com/logos-blockchain/logos-execution-zone/issues/634)

lez/sequencer/core/src/lib.rs
Arjentix: > you mean like if we have a field that we override its borsh traits to not actually encode/decode?

Yes, something like that

> If its ok we can enable that for all, but I thought there was a reason to keep it testing only.

I think it was like that just because it was needed in tests but not in production code, let's make it in production and use here

lez/sequencer/core/src/lib.rs
Arjentix: > I see you are leaning towards actors here too heheh

image

lez/sequencer/core/src/lib.rs
erhant: Added `PartialEq, Eq` in https://github.com/logos-blockchain/logos-execution-zone/pull/603/commits/4ce1bc7086e9a1252fb455406c0c2ecd9a2227e9

test_fixtures/src/setup.rs
```diff
@@ -169,6 +193,31 @@ pub async fn setup_sequencer_from_prebuilt(
...
+    )
+    .await
+}
+
+/// Like [`setup_sequencer`], but with a pre-generated bedrock (Ed25519, 32-byte
+/// seed) signing key.
+///
+/// This allows the tests to know the sequencer's public key before it boots which
+/// is required to accredit a committee member that has not started yet.
+pub async fn setup_sequencer_with_bedrock_key(
```
Arjentix: Btw just in case my PR #606 will be merged first: I've introduced a builder pattern for sequencer setup, so in case when you'll need to rebase I suggest to replace it following builder pattern

lez/chain_state/src/apply.rs
```diff
@@ -0,0 +1,308 @@
...
+            }
+        }
+    }
+    Ok(())
+}
+
+/// Applies a block's transactions to `state`, mapping every failure to a
+/// [`BlockIngestError`] so the caller can park rather than crash. Operates in
+/// place; the caller commits only on `Ok`.
+pub fn apply_block_to_state(block: &Block, state: &mut V03State) -> Result<(), BlockIngestError> {
```
Arjentix, line 125: In future I'd like to have something like `pub struct LezState(V03State)` which will encapsulate all of this validation and application logic inside methods

lez/chain_state/src/chain.rs
```diff
@@ -0,0 +1,848 @@
...
+}
+
+/// The head tier (reorg-able, from `adopted`/`orphaned`) over the final tier
+/// (irreversible, from `finalized`).
+///
+/// `head_state` is given by `final_state` replayed through `head_blocks`.
+///
+/// Only the final tier stalls: an invalid `adopted` block just freezes the
+/// head tip and self-heals via reorg or finalization.
+pub struct ChainState {
```
Arjentix, line 28: Oh, so you kind of already introduced what I had in mind. Ok, then maybe later let's only work with it as a top-level state structure and won't expose `V03State` so much?

So I mean that for `V03State` we have state transition functions like `transition_from_public_transaction()` and etc. I'd like to have the same logic for `ChainState` but with something like `apply_block()`. Does it make sense to you?

Arjentix: Ah, I see, you already have functions like that, cool. Then I'd consider removing `V03State` at all from caller-site.

lez/sequencer/core/src/block_publisher.rs
```diff
@@ -41,6 +50,39 @@ pub type OnDepositEventSink =
 pub type OnWithdrawEventSink =
     Box<dyn Fn(WithdrawInfo) -> Pin<Box<dyn Future<Output = ()> + Send>> + Send + 'static>;
 
+/// The channel delta the follow path consumes from one `Event::BlocksProcessed`,
+/// with inscription payloads decoded into `(MsgId, Block)` pairs.
+pub struct FollowUpdate {
+    pub adopted: Vec<(MsgId, Block)>,
+    pub orphaned: Vec<(MsgId, Block)>,
+    pub finalized: Vec<(MsgId, Block)>,
```
Arjentix, line 78: Let's add docs for each field so that it will be clear what these names mean

lez/sequencer/core/src/block_publisher.rs
```diff
@@ -56,20 +98,51 @@ pub trait BlockPublisherTrait: Clone {
...
 
     fn channel_id(&self) -> ChannelId;
+
+    /// Whether this sequencer is currently authorized to write to the channel.
+    fn is_our_turn(&self) -> bool;
+
+    /// A [`watch::Receiver`] that closes when the publisher's background driver terminates
+    /// (a panicked sink, an ended event stream). No channel events are
+    /// processed past that point, so the node must halt.
+    fn driver_watch(&self) -> watch::Receiver<()>;
```
Arjentix: Hmm, can't we use cancellation token technique we discussed before? So that block publisher cancels the token and that propagates to the whole sequencer.

Ahh, actor frameworks would solve this more beautifully as well...

lez/sequencer/service/src/bin/bedrock_pubkey.rs
```diff
@@ -0,0 +1,35 @@
+//! Prints the sequencer's bedrock signing public key (hex) without booting it.
+//!
+//! Loads `<home>/bedrock_signing_key` from the given sequencer config, creating
+//! the key if it doesn't exist yet, so that a node can be accredited into the
+//! channel committee before its first boot.
```
Arjentix, line 5: Instead of this whole new binary maybe we can make a flag for sequencer_service to do that?

Arjentix (approved):
LGTM, left some minor comments

lez/storage/src/sequencer/mod.rs
moudyellaz: thanks for the lock. What worries me here is that the produce path stays asymmetric though: atomic_update writes lee_state unconditionally (:773) but put_block only guards meta with >= (:495) and doesn't prune higher payloads. So an orphan-only revert below disk head + a produce on that head tears (state, blocks) and crash-loops restart. Can a channel update arrive orphaned-without-adopted? If so atomic_update needs the same pin-meta + prune as the follow path.

lez/sequencer/service/src/service.rs
moudyellaz: --listen-address helps, thanks. But the RPC is still unauthenticated, so anyone who can reach the bound address can sign a roster change with the node's admin key. Is auth on adminConfigureChannel in scope for this PR, or should it be a tracked follow-up before this runs anywhere non-local?

lez/chain_state/src/chain.rs
```diff
@@ -0,0 +1,848 @@
...
+
+        self.head_blocks
+            .iter()
+            .find(|entry| entry.block.header.block_id == block_id)
+            .map(|entry| entry.block.header.hash)
+    }
+
+    /// Applies an adopted head block. On failure the head tip stays at the last
+    /// valid block and no stall is recorded; the head self-heals.
+    pub fn apply_adopted(&mut self, this_msg: MsgId, block: &Block) -> AcceptOutcome {
```
schouhy, line 111: I'd call this method apply_block

schouhy, line 111: this method can do the self-healing if the new adopted block has an id older than the head tip. I think it's better to not depend on orphaned blocks if the adopted list has all we need to rollback changes. And handle orphaned blocks by just adding their transactions back to the mempool when they come

schouhy (commented):
hey thanks for the PR! left some comments

lez/chain_state/src/chain.rs
erhant: hmm I think we should keep it as is, because its' nice to separate `apply_adopted` and `apply_finalized`, and they both use an existing `apply_block` helper that takes a `tip`, `block` and mutable `state`, i.e. applying a block to head / final state, respectively.

Renaming this is still possible because we will have one `apply_block` and another `ChainState::apply_block`, but I think we should keep it as is

lez/sequencer/service/src/bin/bedrock_pubkey.rs
erhant: I was about to type "I think we should keep it a separate binary" but after revisiting, it does make sense lol
will add it and see how ergonomic it becomes 🙏🏻

lez/sequencer/core/src/block_publisher.rs
erhant: The driver watch here is practically equivalent to a cancellation, but I also didn't like `panic` at all.
I actually created 3 different sessions, asking Claude each time can we come up with a better handling here, it always suggested panic & recover 🤔 

I will try cancellation here too, will let you know

lez/chain_state/src/chain.rs
erhant: good idea! its much cleaner this way indeed

fixed in https://github.com/logos-blockchain/logos-execution-zone/pull/603/commits/3d65101b194961f9f9c74cb5c87f6963c35a2618

lez/sequencer/core/src/block_publisher.rs
erhant: added in https://github.com/logos-blockchain/logos-execution-zone/pull/603/commits/3d65101b194961f9f9c74cb5c87f6963c35a2618

test_fixtures/src/setup.rs
erhant: Noted ✍🏻

lez/chain_state/src/chain.rs
erhant: some details to rephrase about the current implementation:

- a plain tip-extension still builds on the `head`, no cost paid extra
- a previous post-final pre-head extension (e.g. our tip is at 100 but adoptions start from 97, finalization at 95) now rederives the head (95->96) and then applies the adoptions on top of it; basically what you said in this comment!
- if for some wicked reason we get adopted block that precede our final block, I just log-and-ignore them
- orphan-driven reverts are still in the code, because we do need to handle them explicitly (for mempool) and it may be possible to have no adoptions but have orphans etc.

lez/storage/src/sequencer/mod.rs
erhant: ah yes my bad; fixed in https://github.com/logos-blockchain/logos-execution-zone/pull/603/commits/3d65101b194961f9f9c74cb5c87f6963c35a2618

- when we are `follow`ing: if the head tip is below the disk head, we rewind explicitly:

https://github.com/logos-blockchain/logos-execution-zone/pull/603/changes/3d65101b194961f9f9c74cb5c87f6963c35a2618#diff-d5bc892dca9a0df72b3e0a3071f9bd2eb9356641fbce33491d10ac938a2e4925R714

- when we are `put`ting blocks, if the head tip is below the disk head they are removed:

https://github.com/logos-blockchain/logos-execution-zone/pull/603/changes/3d65101b194961f9f9c74cb5c87f6963c35a2618#diff-d5bc892dca9a0df72b3e0a3071f9bd2eb9356641fbce33491d10ac938a2e4925R495-R496

added unit tests for them too

lez/sequencer/service/src/bin/bedrock_pubkey.rs
erhant: I did fold it into `sequencer_service` as a flag locally, but it doesn't really seem to add anything and in the end I would prefer keeping the dedicated binary: it's a one-shot key-file utility with no reason to carry the service's full arg surface.

lez/sequencer/core/src/block_publisher.rs
erhant: while it did not change "how things look" overall, it is more amenable to our general plains in the future of refactoring the sequencer logic (actors etc.?) and proper cancellations; added in https://github.com/logos-blockchain/logos-execution-zone/pull/603/commits/66d090d0c92a149e6bb46ecbe371eb4b7b5e0b5b

we still do require a `panic`, but now instead of a "failing channel seen as termination", we have a drop-guarded cancellation that cancels itself on failure, achieving the same affect but more directly

otherwise we were using `channel.changed()` event that would only be active on failure, because the channel never writes; so it was kind of an ugly side-effect trick, drop-cancellation is much cleaner

lez/sequencer/service/src/service.rs
erhant: Went further than auth: the endpoint is gone. There was no need for this to be an RPC call in the first place apparently; roster changes are now a local one-shot (configure_channel binary) that signs with `<home>/bedrock_signing_key` and posts the config op straight to bedrock.
  
Authorization is now about holding the admin key file, which matches the L1's actual trust model (it rejects non-admin signers). The sequencer RPC no longer carries any admin surface, so there's nothing left to authenticate.

Done in https://github.com/logos-blockchain/logos-execution-zone/pull/603/commits/66743751e6493b91904e4a40d3e4fe0461a9ada6

lez/sequencer/service/src/bin/bedrock_pubkey.rs
erhant: + now we have another binary `lez/sequencer/service/src/bin/configure_channel.rs` as well :D (following https://github.com/logos-blockchain/logos-execution-zone/pull/603#discussion_r3630561719)

lez/storage/src/sequencer/mod.rs
```diff
@@ -633,6 +691,92 @@ impl RocksDBIO {
...
+            if *finalized {
+                to_write.bedrock_status = BedrockStatus::Finalized;
+            }
+            self.put_block_payload(&to_write, &mut batch)?;
+        }
+
+        // A shrink-only update (orphans without adopted replacements) has no
+        // payloads to write but must still rewind the tip meta, or the stored
+        // state tears against the stale disk head on the next produce.
+        let tip_rewound = head_tip.is_some_and(|tip| tip.id < last_block_in_db);
```
schouhy: dont we need `tip_rewound` to be `true` if `head_tip` is `None`?

maybe `let tip_rewound = last_block_in_db > head_tip.map_or(0, |tip| tip.id);`?

lez/sequencer/core/src/tests.rs
```diff
@@ -1247,3 +1275,501 @@ fn pda_mechanism_with_pinata_token_program() {
...
+    // A peer block wins height 2 while "our" block is in flight.
+    let peer_block = common::test_utils::produce_dummy_block(2, Some(genesis_meta.hash), vec![]);
+    sequencer
+        .chain()
+        .lock()
+        .expect("chain mutex poisoned")
+        .apply_adopted(MsgId::from([9; 32]), &peer_block);
+
+    // Our competing block at the same height: same parent, different content.
+    let tx = common::test_utils::produce_dummy_empty_transaction();
```
schouhy: I think this dummy transaction is invalid. The tests passes because the whole block fails. With a valid transaction the test fails

```diff
-    let tx = common::test_utils::produce_dummy_empty_transaction();
+    let acc1 = initial_public_user_accounts()[0].account_id;
+    let acc2 = initial_public_user_accounts()[1].account_id;
+    let tx = common::test_utils::create_transaction_native_token_transfer(
+        acc1,
+        0,
+        acc2,
+        10,
+        &create_signing_key_for_account1(),
+    );
```

schouhy (commented):
hey thanks for considering my sugggestions!

moudyellaz (approved):
lgtm, thanks for re-addressing my comments!

schouhy (approved):
Left some comments. Otherwise LGTM great work!

lez/sequencer/core/src/tests.rs
erhant: Great finding actually, it caught a deeper bug that I introduced in the latest adopt self-heal & rebase,

the dummy tx is right, it has zero nonces but one signature, so the block failed to apply and the test passed via `Parked`. With your valid tx it exposed that `record_produced_block` was inheriting the new adopted-driven reorg: our own block would have displaced the peer's channel-adopted block. Added `ChainState::apply_produced` (extend-only, never reorgs as our block isn't on the channel yet) and took your diff.

Fixed in https://github.com/logos-blockchain/logos-execution-zone/pull/603/commits/abe62495fde7ab91bcb287fa8156aff310ab9a6e

lez/storage/src/sequencer/mod.rs
erhant: `head_tip: None` means the chain holds no blocks at all, which implies nothing was applied (no payloads, and no final snapshot either) so there is nothing to rewind to as well. Note that this is the "store" function, so a `None` tip means the expected head updates were not there too. We can add an early return to `head: None` case. 

Restructured in https://github.com/logos-blockchain/logos-execution-zone/commit/abe62495fde7ab91bcb287fa8156aff310ab9a6e

PR #604, opened 2026-07-08
refactor!: bundle private/public i/o into action structs

agureev:
## 🎯 Purpose

This refactors the message and the PPC output structs, enforcing the bijection between pre and post-states on the struct level. Before, e.g. having separate vectors for `new_commitments` and `new_nullifiers` would require being careful about their length mismatch.

This refactor reduces the mismatch possibilities by making sure that all pre and post state fields for public and private accounts come paired in an appropriate struct. E.g. the privacy-preserving circuit output struct becomes:

```
pub struct PrivacyPreservingCircuitOutput {
    pub public_actions: Vec<PublicAction>,
    pub private_actions: Vec<PrivateAction>,
    pub block_validity_window: BlockValidityWindow,
    pub timestamp_validity_window: TimestampValidityWindow,
}
```
The name `Actions` is taken from ZCash as a stand-in, feel free to suggest better naming.

## ⚙️ Approach

- [x] Define private and public action structs
- [x] Change the structs utilizing any paired vectors, i.e. Message and Output
- [x] Change usepoints of said structs
- [x] Refactor other parts of the codebase that could benefit from the change

## 🧪 How to Test

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


lee/state_machine/src/privacy_preserving_transaction/message.rs
```diff
@@ -1,24 +1,27 @@
...
-use crate::{AccountId, error::LeeError};
+use crate::AccountId;
 
 const PREFIX: &[u8; 32] = b"/LEE/v0.3/Message/Privacy/\x00\x00\x00\x00\x00\x00";
 
+#[derive(Debug, Clone, PartialEq, Eq, BorshSerialize, BorshDeserialize)]
+pub struct PublicActionWithOwner {
+    pub account_id: AccountId,
+    pub post_state: Account,
+}
```
Arjentix: Name of the struct does not reflect what's inside I think

lez/indexer/service/protocol/src/convert.rs
```diff
@@ -281,27 +281,24 @@ impl From<PublicMessage> for lee::public_transaction::Message {
...
-                .map(Into::into)
+            public_post_states: public_actions.into_iter().map(|a| a.post_state.into()).collect(),
+            encrypted_private_post_states: private_actions
+                .iter()
+                .map(|a| a.encrypted_post_state.clone().into())
                 .collect(),
-            new_commitments: new_commitments.into_iter().map(Into::into).collect(),
-            new_nullifiers: new_nullifiers
+            new_commitments: private_actions.iter().map(|a| a.commitment.clone().into()).collect(),
+            new_nullifiers: private_actions
```
Arjentix: I think It's worth changing this whole struct the same way you changed the core struct.

In case of any problems on Explorer side you can always use Claude, haha

Arjentix (commented):
Like the approach, much cleaner now.

Left some minor comments, will wait for undrafting and then will approve

lez/indexer/service/protocol/src/convert.rs
agureev: should be better now! the message format has been changed accordingly!

lee/state_machine/src/privacy_preserving_transaction/message.rs
agureev: Renamed to `PublicActionWIthID`, agreed that `Owner` is a vague term with program ownership alongside authorization

lee/state_machine/core/src/circuit_io.rs
```diff
@@ -147,18 +148,53 @@ impl InputAccountIdentity {
...
 
+#[derive(Serialize, Deserialize, BorshSerialize, BorshDeserialize)]
+#[cfg_attr(
+    any(feature = "host", test),
+    derive(Debug, Clone, Default, PartialEq, Eq)
+)]
+pub struct PrivateAction {
+    pub nullifier: Nullifier,
+    pub root: CommitmentSetDigest,
+    pub commitment: Commitment,
```
moudyellaz, line 162: One thing worth guarding here: after `obfuscate_output_ordering`, this `commitment` no longer belongs to this action's `nullifier`/`encrypted_post_state` (that's #561's unlinking doing its job), but the struct shape suggests they're paired, and the indexer protocol mirrors it downstream. Could commitments move out to a separate `Vec<Commitment>`, since nullifier/root/note are truly paired and commitments are an independent set? If you'd rather keep the bundling, a doc note on this field at both definition sites would save a future consumer from re-linking them.

lee/state_machine/src/privacy_preserving_transaction/message.rs
```diff
@@ -1,24 +1,27 @@
...
+pub struct PublicActionWithID {
+    pub account_id: AccountId,
+    pub post_state: Account,
+}
+
 #[derive(Clone, Default, PartialEq, Eq, BorshSerialize, BorshDeserialize)]
 pub struct Message {
-    pub public_account_ids: Vec<AccountId>,
+    pub public_actions: Vec<PublicActionWithID>,
     pub nonces: Vec<Nonce>,
```
moudyellaz, line 23: `nonces` is the one parallel vector left after this refactor, positionally paired with `public_actions`. Could `nonce` live inside `PublicActionWithID`, or does the signing/serialization format constrain it?

moudyellaz (commented):
Left two comments. The three failing CI jobs look like infra (archive extraction + GitHub rate limit fetching risc0), probably just need a rerun.

lee/state_machine/src/privacy_preserving_transaction/message.rs
agureev: I agree that it would be nice to pair these! Indeed, the signing format currently seems to narrow it as the nonces for signature-checking are gotten by position from the nonces vector. There is probably some nice way to refactor this as well, but it seemed to be something to be undertaken in parallel, I concentrated only on the pre-post state pairing here.

lee/state_machine/core/src/circuit_io.rs
agureev: I think I would rather keep the binding, otherwise this reintroduces the issue of length-matching.

But I agree that a doc addition would be nice!

Pravdyvy (approved):
lgtm

lee/state_machine/core/src/circuit_io.rs
agureev: Should be addressed in [0e957b5](https://github.com/logos-blockchain/logos-execution-zone/pull/604/commits/0e957b5c08367b712da9b77e725715de8afe56f8)

Let me know if there are any other places you want me to document this

lee/state_machine/src/privacy_preserving_transaction/message.rs
agureev: Filed #655 to keep track of this

moudyellaz (approved):
lgtm, thanks for addressing my comments.

Arjentix (approved):
LGTM, thanks

PR #606, opened 2026-07-08
feat(sequencer): bootstrap state from Bedrock

Arjentix:
## 🎯 Purpose

This PR implements sequencer bootstrapping from Logos Blockchain Node (Bedrock).

## ⚙️ Approach

- Basically sequencer just compares local store and bedrock channel on startup
- Different combinations result in different behavior, all of them tested
- Indexer bootstrap logic previously implemented in #581 moved to a separate crate to unify validation between services

## 🧪 How to Test

Added integrations and unit tests.

Unit tests:

```bash
RISC0_DEV_MODE=1 cargo nextest run --package sequencer_core --lib --all-features -- tests::reconstruction
```

Integration tests:

```bash
RISC0_DEV_MODE=1 cargo nextest run --package integration_tests --test sequencer_bootstrap --all-features
```

## 🔗 Dependencies

- https://github.com/logos-blockchain/logos-execution-zone/pull/581

## 🔜 Future Work

None

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


lez/chain_consistency/Cargo.toml
erhant, line 1: hmm perhaps we are both doing a similar thing here with https://github.com/logos-blockchain/logos-execution-zone/pull/603/changes#diff-92d263813ee4b250bf1c03b72dd3713966d8b97760a8af32e71f7a903e08e25c

Arjentix: Ah, yes, probably. Well, however merge first wins :trollface:

lez/sequencer/core/src/lib.rs
```diff
@@ -66,13 +70,20 @@ pub struct SequencerCore<BP: BlockPublisherTrait = ZoneSdkPublisher> {
     block_publisher: BP,
 }
 
+/// Outcome of the startup verify-and-reconstruct pass.
+struct ReconstructionSummary {
+    /// Whether the channel served any blocks. Drives the fresh-start decision
+    /// between publishing our own genesis and adopting the channel's history.
+    channel_had_blocks: bool,
+}
```
Pravdyvy: Kind of a strange struct, I suppose it will contain additional fields in the future?

Pravdyvy (approved):
lgtm, one question.

lez/indexer/core/src/lib.rs
```diff
@@ -125,6 +123,49 @@ impl IndexerCore {
...
+    fn get_startup_anchor(&self) -> Result<Option<Anchor>> {
+        if let Some(stall) = self.store.get_stall_reason()? {
+            return Ok(Some(Anchor::new(
+                stall.l1_slot,
+                stall.block_id.zip(stall.block_hash),
+            )));
+        }
+
+        // not stalled, so anchor on the tip at its own inscription slot
+        let Some(slot) = self.store.get_tip_slot()?.or(self.store.get_zone_cursor()?) else {
```
erhant: can use `or_else(|| ...)` instead to avoid redundant call to `get_zone_cursor`

lez/sequencer/core/src/tests/reconstruction.rs
```diff
@@ -0,0 +1,401 @@
...
+    let witness_set = lee::public_transaction::WitnessSet::for_message(&message, &[signing_key]);
+    LeeTransaction::Public(lee::PublicTransaction::new(message, witness_set))
+}
+
+/// The channel is ahead of the local store and its extra blocks carry
+/// bridge deposit and withdraw transactions. Reconstruction must replay them,
+/// reproducing both the minted/burned balances and the withdraw reconciliation
+/// bookkeeping the production path records.
+#[tokio::test]
+async fn reconstructs_channel_with_deposit_and_withdraw() {
```
erhant: This test passes because `MockBlockPublisher` drops `_on_deposit_event` (`mock.rs:54`) and never re-delivers the L1 deposit. In production, the real `zone-sdk` cold-start backfill actually re-delivers already-finalized deposits, and that path results in a **double-mint** issue.

---

A fresh sequencer has no checkpoint, so zone-sdk cold-starts and backfills the whole channel from genesis to LIB. `extract_finalized_items` re-emits a `FinalizedOp::Deposit` for every channel deposit in that range and `block_publisher` routes it straight into `on_deposit_event`. So every deposit already on an adopted channel is guaranteed to come back as a live event.

Imagine the scenario below, for a fresh **Sequencer B** adopting **Sequencer A**'s channel where A already produced+finalized a bridge deposit for `op_id` equal to `X`:

- **B boots fresh** and has empty store, `initial_checkpoint = None` so `is_fresh_start = true`.

- `replay_unfulfilled_deposit_events` is no-op, store is empty

- `BP::new` starts zone-sdk with `initial_checkpoint = None`, so a cold-start backfill re-scans the channel from genesis on the drive task.

- Backfill re-observes A's finalized deposit for op ID `X` via `on_deposit_event`:
  - `add_pending_deposit_event(X)` succeeds, (fresh store, `X` unknown and therefore not deduped),
  - builds the mint tx and **pushes it to the mempool**.

- `verify_and_reconstruct` replays A's blocks. At the deposit block, `apply_reconstructed_block`:
  - `apply_block` executes the bridge `Deposit`; this is **first mint tx** into the recipient vault,
  - `store.update` → `mark_pending_deposit_events_submitted([X])` flips the pending row to *submitted*, but the mint tx is **already in the mempool** and marking the RocksDB row doesn't remove it.

- Later `produce_new_block` pops the queued mint tx for the same op ID `X`: this is the **second mint**. The bridge program doesn't dedup — it destructures `l1_deposit_op_id: _` instead!

As a result the recipient vault is credited twice for one L1 deposit.

---

Reordering doesn't save it: if the backfill fires *after* reconstruction, reconstruction had no pending row to mark (it never *inserts* one), so `add_pending_deposit_event(X)` still returns `Ok(true)` and enqueues the mint → same double mint.

Root cause: dedup lives in the `pending_deposit_events` table, but the mint is driven by the mempool; reconstruction only touches the table, so it can neither cancel a queued mint nor pre-register `X` so the backfill dedups it.

This test here pre-seeds the pending row and calls `verify_and_reconstruct` directly, only getting the "mark an existing row submitted" path, so no event / mempool stuff.

erhant (changes requested):
LGTM overall but one important finding (commented at the test function)

lez/sequencer/core/src/lib.rs
```diff
@@ -205,6 +221,185 @@ impl<BP: BlockPublisherTrait> SequencerCore<BP> {
...
+    ) -> Result<ReconstructionSummary> {
+        let anchor_record = store
+            .get_zone_anchor()
+            .context("Failed to read zone anchor")?;
+
+        let after_slot = anchor_record
+            .and_then(|record| record.slot.checked_sub(1))
+            .map(Slot::from);
+        let messages = publisher
+            .read_channel_after(after_slot)
```
moudyellaz: This read is unbounded (`stream.collect()`, no timeout) and any error flows through `?` to the `.expect` at line 175, so a transient Bedrock hiccup at boot can crash-loop or hang the sequencer. The indexer's check uses a 60s timeout and downgrades errors to Inconclusive to proceed. Could we match that fail-open posture here?

lez/sequencer/core/src/lib.rs
```diff
@@ -205,6 +221,185 @@ impl<BP: BlockPublisherTrait> SequencerCore<BP> {
...
+        // Derive bridge bookkeeping from the block's transactions, matching the
+        // production path so the reconciliation counters stay consistent.
+        let mut deposit_event_ids = Vec::new();
+        let mut withdrawals = Vec::new();
+        for tx in &block.body.transactions {
+            if let Some(deposit_id) = extract_bridge_deposit_id(tx) {
+                deposit_event_ids.push(deposit_id);
+            }
+            if let Some(withdraw) = extract_bridge_withdraw_data(tx) {
+                withdrawals.push(withdraw_event_reconciliation_key(&withdraw.outputs)?);
```
moudyellaz: Reconstructed withdrawals increment `unseen_withdraw_count`, but the decrement only fires when Bedrock re-emits the Withdraw event, which won't happen for a peer's block whose event was already consumed. Each one leaves a permanent phantom count (reconciliation state, not ledger). Could we skip the increment on the replay path?

lez/chain_consistency/src/consistency.rs
```diff
@@ -168,116 +176,104 @@ enum AnchorProbe {
...
-    ///
-    /// Only positive evidence of a different chain yields `Inconsistent`.
-    /// Absence of data stays `Inconclusive`.
-    async fn verify_chain_at_anchor(&self, anchor: &Anchor) -> Result<ChainConsistency> {
-        match self.node.channel_state(self.config.channel_id).await {
-            Ok(state) => {
-                if let Some(mismatch) = frontier_verdict(anchor.slot, state.map(|s| s.tip_slot)) {
-                    return Ok(ChainConsistency::Inconsistent(mismatch));
+    // `next_messages` is exclusive, so `slot - 1` includes the anchor slot.
+    let Some(from_slot) = anchor.slot.into_inner().checked_sub(1) else {
```
moudyellaz, line 270: At anchor slot 0, this path returns Inconclusive without scanning, but `classify_channel` has no such guard and goes on to probe slot 0, so the two give different verdicts on the same content. Could `classify_channel` get the same slot-0 guard so they stay in sync?

lez/sequencer/core/src/lib.rs
```diff
@@ -205,6 +221,185 @@ impl<BP: BlockPublisherTrait> SequencerCore<BP> {
...
+        // ran (`!is_fresh_start`), because its cold-start backfill re-persists a
+        // checkpoint by the time we reach here — reading the store now would always
+        // see one. Together they mean we produced blocks and zone-sdk processed
+        // channel activity in a prior run.
+        let local_tip = store
+            .latest_block_meta()
+            .context("Failed to read latest block meta")?
+            .id;
+        let had_checkpoint_before_start = !is_fresh_start;
+        let has_committed_to_channel = local_tip > GENESIS_BLOCK_ID && had_checkpoint_before_start;
```
moudyellaz: A node that produced only genesis (tip == GENESIS) has this false, so if its channel was wiped the refuse-to-resume check is skipped and it silently re-bootstraps. Only the window between producing genesis and block 1, but the heuristic is technically wrong there.

lez/sequencer/core/src/lib.rs
```diff
@@ -205,6 +221,185 @@ impl<BP: BlockPublisherTrait> SequencerCore<BP> {
...
+            hash: tip.hash,
+        };
+        chain_consistency::validate_against_tip(Some(&cc_tip), block).map_err(|err| {
+            anyhow!(
+                "Channel block {block_id} does not extend local tip {}: {err}",
+                tip.id
+            )
+        })?;
+
+        let mut scratch = state.clone();
```
moudyellaz, line 394: Full V03State clone per replayed block, so the first reconstruct of a long chain is O(chain x state) in time and allocation. Maybe it is worth it to amortize if chains get long?

lez/storage/src/sequencer/mod.rs
```diff
@@ -359,6 +363,14 @@ impl RocksDBIO {
         self.del::<ZoneSdkCheckpointCellOwned>(())
     }
 
+    pub fn get_zone_cursor(&self) -> DbResult<Option<ZoneAnchorRecord>> {
```
moudyellaz: This cell/method is `zone_cursor` but holds a `ZoneAnchorRecord`, and the sequencer calls it `anchor`. Maybe rename the cell to match?

lez/sequencer/core/src/lib.rs
```diff
@@ -205,6 +221,185 @@ impl<BP: BlockPublisherTrait> SequencerCore<BP> {
...
+            .latest_block_meta()
+            .context("Failed to read latest block meta")?
+            .id;
+        let had_checkpoint_before_start = !is_fresh_start;
+        let has_committed_to_channel = local_tip > GENESIS_BLOCK_ID && had_checkpoint_before_start;
+        if has_committed_to_channel && channel_tip_slot.is_none() {
+            return Err(anyhow!(
+                "Sequencer holds committed blocks (tip {local_tip}) but the Bedrock channel \
+                 no longer exists on the connected chain — the channel was wiped or the node \
+                 points at a different chain. Refusing to resume onto a foreign channel."
```
moudyellaz: The prebuilt-fixture path can hit this and panic (is_fresh_start=false + tip>genesis + empty channel) if the committed dump ever captures a checkpoint, which the new Justfile hook re-rolls. Maybe worth guarding?

lez/sequencer/core/src/tests/reconstruction.rs
```diff
@@ -0,0 +1,401 @@
...
+        .enumerate()
+        .map(|(index, id)| {
+            let block = store.get_block_at_id(id).expect("read").expect("present");
+            block_to_channel_message(&block, (index as u64 + 1) * slot_step)
+        })
+        .collect()
+}
+
+#[tokio::test]
+async fn reconstructs_missing_channel_blocks_into_fresh_store() {
```
moudyellaz, line 41: These cover the happy paths well, but the failure branches of `apply_reconstructed_block` (hash-mismatch, missing-locally, reinscribed-genesis) go through `verify_and_reconstruct` untested. Maybe add a couple of divergence cases?

moudyellaz (approved):
lgtm, thanks. Left some comments and questions.

lez/sequencer/core/src/lib.rs
Arjentix: Ah, yes, I agree it's strange. Probably I've let AI do too much in this PR. I was also considering to remove this structure but decided that maybe it's more readable. But after your comment I can see that it confuses not only me, so I'll just return bool from function and reverse it's value, so that the meaning will be "was channel empty?"

lez/indexer/core/src/lib.rs
Arjentix: It turned out to be a bit uglier, but I agree that it's better to avoid such things, good catch!

lez/sequencer/core/src/tests/reconstruction.rs
Arjentix: Good catch, thanks! I fixed it with an additional check if deposit was already submitted

lez/sequencer/core/src/lib.rs
Arjentix: Agree, collecting all blocks is not nice, there could be a lot of them. Changed to reading block on by one, validate it and apply it. I think crashing is okay here, sequencer makes no sense if it can't catch up to the latest state when it starts

lez/sequencer/core/src/lib.rs
Arjentix: Thanks, fixed and added test

lez/chain_consistency/src/consistency.rs
Arjentix: Fixed by removing `classify_channel()` at all when addressing your [other comment](https://github.com/logos-blockchain/logos-execution-zone/pull/606#discussion_r3605371415)

lez/sequencer/core/src/lib.rs
Arjentix: Fixed by changing to `if let Some(local_tip) = local_tip`, so genesis id is no longer used here

lez/sequencer/core/src/lib.rs
Arjentix: This is currently implemented in the same way we did this on Indexer side. Currently I see no easy way to improve it. The only option I can imagine is transaction rollback, but that's not currently implemented.

 Let's consider this in further PR when we really stuck in this performance bottle neck?

Arjentix: - Opened #635

lez/storage/src/sequencer/mod.rs
Arjentix: Renamed

lez/sequencer/core/src/lib.rs
Arjentix: Indeed something like that happened to me at the beginning, but I fixed it by deleting zone checkpoint when generating this fixture. So no real problem exists here

lez/sequencer/core/src/tests/reconstruction.rs
Arjentix: Added more tests

erhant (approved):
LGTM! 💐

PR #607, opened 2026-07-09
fix!(indexer): initial prep for breakpoint fix

### LINKED ISSUE #605: indexer: Ctrl-C during indexing corrupts the store (missing breakpoint bricks startup; false park + cursor advance strands valid blocks)
## Summary

A `Ctrl-C` during the indexing (catch-up) phase can permanently corrupt the indexer's RocksDB store in **two independent ways**:

1. **Missing breakpoint → unrecoverable init failure.** The periodic state snapshot ("breakpoint") is written *outside* the block's atomic `WriteBatch` and is never retried after a failure. A store can end up with `last_block_in_db` past a 100-block boundary but without the breakpoint that boundary should have produced. Every subsequent start then fails in `IndexerStore::open_db` with:

   ```
   Error: Failed to initialize indexer service
   Caused by:
       Logic Error: "breakpoint" not found
   ```

   This happens *before* the chain-consistency check, so `allow_chain_reset` never gets a chance to run. The only recovery is deleting the store.

2. **False park + cursor advance → permanent stall.** Shutdown tears down the tokio runtime / risc0 host executor while the ingest loop is still running. The resulting transient infra error (`Failed to execute program: rx len failed`) is classified by `apply_block_to_scratch` as `BlockIngestError::StateTransition` → the block is **parked** and, critically, the L1 cursor **advances past it**. On restart the perfectly valid parked blocks are behind the cursor and are never re-delivered: the tip stays frozen forever and every future block orphans.

Both were produced by a single `Ctrl-C` while testing #581 (park-not-skip ingest, branch `erhant/indexer-recoverable-invalid-blocks`).

## Observed log

```
[2026-07-08T15:30:36Z INFO  indexer_core] Indexed L2 block 162
^C[2026-07-08T15:30:36Z INFO  indexer_service] Received Ctrl-C signal
[2026-07-08T15:30:36Z INFO  indexer_service] Shutting down server...
[2026-07-08T15:30:36Z INFO  indexer_service] Server shutdown complete
[2026-07-08T15:30:36Z ERROR indexer_core] Parked at block 163: State transition failed at transaction 0: Failed to execute program: rx len failed
[2026-07-08T15:30:36Z ERROR indexer_core] Parked at block 164: Unexpected block id: expected 163, got 164
[2026-07-08T15:30:36Z ERROR indexer_core] Parked at block 165: Unexpected block id: expected 163, got 165
error: interrupted by SIGINT
```

Note the parks happen *after* "Server shutdown complete": the ingest loop is not stopped by the Ctrl-C handler, it just keeps running against a dying runtime.

Restart:

```
[2026-07-08T15:30:56Z INFO  indexer_service] Starting Indexer Service RPC server on 0.0.0.0:8779
Error: Failed to initialize indexer service
Caused by:
    Logic Error: "breakpoint" not found
```

## Evidence (direct dump of the corrupted store)

Read-only dump of `lez/indexer/service/rocksdb-0101…01` after the failure:

| key | value |
|---|---|
| `cf_block` | blocks **1..=162** present |
| `cf_meta/last_block_in_db` | **162** |
| `cf_breakpoint` | **only breakpoint 0** — breakpoint 1 (state @ block 100) missing |
| `cf_meta/last_breakpoint_id` | 0 |
| `cf_meta/zone_sdk_indexer_cursor` | **15487340** |
| `cf_meta/tip_slot` | 15487295 (block 162) |
| `cf_meta/stall_reason` | `{"block_id":163, …, "error":{"StateTransition":{"tx_index":0,"reason":"Failed to execute program: rx len failed"}}, "l1_slot":15487309, "orphans_since":2}` |

- Init failure chain: `IndexerStore::open_db` (`lez/indexer/core/src/block_store.rs:52`) → `final_state()` → `calculate_state_for_id(162)` → `closest_breakpoint_id(162) = 1` (`lez/storage/src/indexer/mod.rs`) → `get_breakpoint(1)` → `Logic Error: "breakpoint" not found`.
- Stall permanence: stored cursor `15487340` > parked block 163's inscription slot `15487309`, so blocks 163–165 (which are valid — they only "failed" because the executor was being torn down) can never be re-delivered.

### 1. Breakpoint write is non-atomic, expensive, and never retried (pre-existing, `lez/storage/src/indexer/write_atomic.rs`)

`put_block` commits the block + `last_block` meta in one `WriteBatch`, then — only afterwards, non-atomically — calls `put_next_breakpoint()` when `block_id % BREAKPOINT_INTERVAL == 0`. `put_next_breakpoint` recomputes the snapshot via `calculate_state_for_id(N)`: a **full replay of all blocks since the previous breakpoint through the risc0 executor** (~0.3–1 s during catch-up — by far the longest synchronous window in the accept path, so an untimely exit is disproportionately likely to land inside it).

If the process dies (or the executor is torn down) in that window: the block and `last_block` are durable, the breakpoint is not. On the next run the boundary block is re-delivered and hits the `AlreadyApplied` short-circuit in `accept_block` (`block_store.rs`), which never re-attempts the breakpoint. Nothing else ever writes it. `closest_breakpoint_id` assumes breakpoints exist by pure arithmetic, so the first restart with a tip past the boundary dies at init, permanently.

### 2. Shutdown doesn't stop ingestion; transient executor failure is treated as an invalid block (new risk from #581's park-not-skip)

The Ctrl-C handler shuts down the RPC server but not the ingest stream. Runtime teardown kills the risc0 host channel; the in-flight `accept_block` fails with `rx len failed`, which `apply_block_to_scratch` maps — like *every* apply error — to `BlockIngestError::StateTransition`. The ingest loop then takes the `Parked` path: persists a `StallReason` **and advances the L1 cursor** past the block (and past its orphaned successors). Parking is only sound for *deterministically* invalid blocks; for transient infra failures it converts a clean shutdown into permanent divergence.

### 3. `open_or_create` clobbers breakpoint meta on every open (pre-existing, `lez/storage/src/indexer/mod.rs`)

Every open unconditionally rewrites breakpoint 0 and resets `last_breakpoint_id = 0`. After a restart the breakpoint counter restarts from 0, so subsequent `put_next_breakpoint` calls write *lagging* snapshots (at block 200 it writes breakpoint 1 = state@100 instead of breakpoint 2). Enough restarts alone — no kill-timing race required — leave `closest_breakpoint_id(tip)` pointing at a breakpoint that was never written, producing the same init death.

## Proposed fixes

1. **Write the breakpoint inside the block's `WriteBatch`.** When `block_id % BREAKPOINT_INTERVAL == 0`, the scratch state `accept_block` just validated *is* the breakpoint state — serialize it into the same batch as the block. Atomic, no replay, no executor involvement; the `put_next_breakpoint` replay path can be deleted.
2. **Degrade gracefully on missing breakpoints.** `calculate_state_for_id` should fall back to the nearest *existing* breakpoint ≤ the computed id (worst case: breakpoint 0) instead of hard-erroring — a missing snapshot then costs a longer replay instead of bricking the store.
3. **Separate transient failures from deterministic invalidity.** Executor/infra failures during apply must take the existing `Err` path (no park, no cursor advance, re-poll next cycle), not `Parked`. Additionally, wire the Ctrl-C signal to cancel the ingest stream *before* the runtime teardown kills the executor, so shutdown can't produce spurious apply failures at all.
4. **Seed breakpoint 0 / `last_breakpoint_id` only when absent** in `open_or_create`.

(2) also wants a regression test: build a store past block 100, delete breakpoint 1, reopen — must start and serve correct state.

## Recovery (workaround)

The store cannot be repaired by config (`allow_chain_reset` is unreachable — init fails first). Delete the store directory (`just clean` or `rm -rf lez/indexer/service/rocksdb-<channel_id>`) and re-index; everything is re-derivable from L1. Until (1)/(3) land, any Ctrl-C during catch-up near a multiple of 100 blocks can reproduce the corruption.

## Related

- #579 / #581 — park-not-skip ingest (root causes 1 and 3 pre-date it; cause 2 is new risk introduced by parking)

erhant:
## 🎯 Purpose

Fixes two ways a `Ctrl-C` during the indexing (catch-up) phase could permanently corrupt the indexer's RocksDB store:

1. **Missing breakpoint → bricked startup.** The periodic state snapshot ("breakpoint") was written *outside* the block's atomic `WriteBatch`, recomputed via a ~100-block risc0 replay, and never retried after a failure. A store whose tip passed a 100-block boundary without its snapshot failed every subsequent boot with `Logic Error: "breakpoint" not found` — before the chain-consistency check, so `allow_chain_reset` couldn't help.
2. **False park → permanent stall.** Shutdown tore down the runtime while the ingest loop was still processing; the dying risc0 executor's transient error (`rx len failed`) was classified as an invalid block, parking the indexer and advancing the L1 cursor past perfectly valid blocks — unrecoverable divergence.

Closes #605.

BREAKING CHANGE: The indexer now requires `CancellationToken` for the graceful shutdown.

## ⚙️ Approach

- [x] **Atomic breakpoint write** — `accept_block` passes its validated scratch state (which *is* the post-block state) into `put_block`; the snapshot and `last_breakpoint_id` land in the same `WriteBatch` as the block at every `BREAKPOINT_INTERVAL` boundary. The replay-based `put_next_breakpoint` is deleted: no recomputation, no executor dependency, no torn-write window.
- [x] **Fallback replay** — `calculate_state_for_id` walks down to the nearest breakpoint that *exists* (breakpoint 0 is the floor) instead of hard-erroring, so a store with a missing snapshot costs a longer replay instead of being bricked. Already-corrupted stores become bootable again.
- [x] **Non-clobbering seeding** — `open_or_create` seeds breakpoint 0 / `last_breakpoint_id` only on a fresh store; reopens no longer reset the breakpoint counter (pre-existing bug that made snapshots lag one interval per restart).
- [x] **Retryable vs deterministic failures** — new `AcceptOutcome::ApplyFailed`: a `StateTransition` failure no longer parks immediately or persists a stall (a `FIXME` documents that this error should eventually carry a structured cause separating genuine rejection from executor/storage infra failure). Structural failures (bad id, broken link, hash mismatch, bad clock tx) still park immediately.
- [x] **Retry-before-park** — `ApplyRetryGate` in the ingest loop: the same block must fail `APPLY_RETRY_LIMIT = 3` consecutive delivery cycles (cursor untouched in between) before the stall is recorded and the cursor advances. Transient executor failures self-heal; genuinely invalid blocks still park, ~3 poll-intervals later. The gate resets only on `Applied` — the re-delivered `AlreadyApplied` prefix after a retry cycle must not clear the failing block's streak.
- [x] **Cooperative shutdown** — the service's existing `CancellationToken` is threaded into the subscription/ingest loop, which now exits cleanly at a message boundary on `Ctrl-C` instead of being killed mid-work by runtime teardown (the source of the spurious `rx len failed` parks).

## 🧪 How to Test

```sh
RISC0_DEV_MODE=1 RISC0_SKIP_BUILD=1 cargo test -p storage -p indexer_core
```

Manual (the original incident): run the full local stack, `Ctrl-C` the indexer mid catch-up (ideally just past block 100), restart. Expect `Shutdown requested; stopping block ingestion` on exit, `Resuming indexer from cursor …` on restart, no `"breakpoint" not found`, and `getStatus` not reporting `stalled`.

## 🔗 Dependencies

None, but built for #581 

## 🔜 Future Work

- Structure the `StateTransition` error (tracked by the in-code `FIXME`s) so the retry gate can narrow to genuine infra failures instead of retrying all state-transition errors. (#610)

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests

### DIFF
--- a/lez/indexer/core/src/ingest_error.rs
+++ b/lez/indexer/core/src/ingest_error.rs
@@ -40,6 +40,19 @@ pub enum BlockIngestError {
     },
 }
 
+impl BlockIngestError {
+    /// Whether the failure may be transient rather than a property of the block.
+    ///
+    /// FIXME: `StateTransition` is too coarse — its `reason` string mixes genuine
+    /// state-transition rejections with infra failures (risc0 executor teardown,
+    /// storage errors). Once it carries a structured cause, narrow this so only
+    /// infra failures retry.
+    #[must_use]
+    pub const fn is_retryable(&self) -> bool {
+        matches!(self, Self::StateTransition { .. })
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
--- a/lez/indexer/core/src/lib.rs
+++ b/lez/indexer/core/src/lib.rs
@@ -10,6 +10,7 @@ use log::{error, info, warn};
 use logos_blockchain_zone_sdk::{
     CommonHttpClient, Slot, ZoneMessage, adapter::NodeHttpClient, indexer::ZoneIndexer,
 };
+use retry::ApplyRetryGate;
 pub use stall_reason::StallReason;
 
 use crate::{
@@ -22,8 +23,12 @@ pub mod block_store;
 pub mod chain_consistency;
 pub mod config;
 pub mod ingest_error;
+mod retry;
 pub mod stall_reason;
 pub mod status;
+
+/// Consecutive failed apply attempts of the same block before parking.
+const APPLY_RETRY_LIMIT: u32 = 3;
 
 #[derive(Clone)]
 pub struct IndexerCore {
@@ -167,6 +172,7 @@ impl IndexerCore {
 
         async_stream::stream! {
             let mut cursor = initial_cursor;
+            let mut retry_gate = ApplyRetryGate::new();
 
             if cursor.is_some() {
                 info!("Resuming indexer from cursor {cursor:?}");
@@ -215,6 +221,7 @@ impl IndexerCore {
 
                     match self.store.accept_block(&block, slot).await {
                         Ok(AcceptOutcome::Applied) => {
+                            retry_gate.reset();
                             info!("Indexed L2 block {}", block.header.block_id);
                             self.set_status(IndexerSyncStatus::syncing());
                             self.advance_cursor(&mut cursor, slot);
@@ -237,13 +244,33 @@ impl IndexerCore {
                             self.advance_cursor(&mut cursor, slot);
                         }
                         Ok(AcceptOutcome::ApplyFailed(ingest_err)) => {
-                            error!(
-                                "Apply failed at block {}: {ingest_err}",
-                                block.header.block_id
-                            );
-                            self.set_status(IndexerSyncStatus::error(ingest_err.to_string()));
-                            had_cycle_error = true;
-                            break;
+                            let attempts = retry_gate.register_failure(block.header.block_id);
+                            if attempts >= APPLY_RETRY_LIMIT {
+                                error!(
+                                    "Parked at block {} after {attempts} failed apply attempts: {ingest_err}",
+                                    block.header.block_id
+                                );
+                                if let Err(err) = self.store.record_stall(
+                                    Some(&block.header),
+                                    slot,
+                                    ingest_err.clone(),
+                                ) {
+                                    warn!("Failed to record stall reason: {err:#}");
+                                }
+                                self.set_status(IndexerSyncStatus::stalled(ingest_err.to_string()));
+                                self.advance_cursor(&mut cursor, slot);
+                                retry_gate.reset();
+                            } else {
+                                error!(
+                                    "Failed to apply block {} (attempt {attempts}/{APPLY_RETRY_LIMIT}), will retry: {ingest_err}",
+                                    block.header.block_id
+                                );
+                                self.set_status(IndexerSyncStatus::error(format!(
+                                    "apply failed, retrying: {ingest_err}"
+                                )));
+                                had_cycle_error = true;
+                                break;
+                            }
                         }
                         Err(err) => {
                             // Infrastructure error (DB read/write), not a bad block.
--- a/lez/indexer/core/src/lib.rs
+++ b/lez/indexer/core/src/lib.rs
@@ -250,12 +250,21 @@ impl IndexerCore {
                                     "Parked at block {} after {attempts} failed apply attempts: {ingest_err}",
                                     block.header.block_id
                                 );
+                                // The stall must be durable before the cursor moves.
                                 if let Err(err) = self.store.record_stall(
                                     Some(&block.header),
                                     slot,
                                     ingest_err.clone(),
                                 ) {
-                                    warn!("Failed to record stall reason: {err:#}");
+                                    error!(
+                                        "Failed to record stall reason for block {}: {err:#}",
+                                        block.header.block_id
+                                    );
+                                    self.set_status(IndexerSyncStatus::error(format!(
+                                        "store error: {err:#}"
+                                    )));
+                                    had_cycle_error = true;
+                                    break;
                                 }
                                 self.set_status(IndexerSyncStatus::stalled(ingest_err.to_string()));
                                 self.advance_cursor(&mut cursor, slot);
--- a/lez/indexer/core/src/lib.rs
+++ b/lez/indexer/core/src/lib.rs
@@ -243,7 +243,7 @@ impl IndexerCore {
                             // L1 proceeds regardless
                             self.advance_cursor(&mut cursor, slot);
                         }
-                        Ok(AcceptOutcome::ApplyFailed(ingest_err)) => {
+                        Ok(AcceptOutcome::RetryableFailure(ingest_err)) => {
                             let attempts = retry_gate.register_failure(block.header.block_id);
                             if attempts >= APPLY_RETRY_LIMIT {
                                 error!(
--- a/lez/indexer/core/src/lib.rs
+++ b/lez/indexer/core/src/lib.rs
@@ -236,6 +236,15 @@ impl IndexerCore {
                             // L1 proceeds regardless
                             self.advance_cursor(&mut cursor, slot);
                         }
+                        Ok(AcceptOutcome::ApplyFailed(ingest_err)) => {
+                            error!(
+                                "Apply failed at block {}: {ingest_err}",
+                                block.header.block_id
+                            );
+                            self.set_status(IndexerSyncStatus::error(ingest_err.to_string()));
+                            had_cycle_error = true;
+                            break;
+                        }
                         Err(err) => {
                             // Infrastructure error (DB read/write), not a bad block.
                             // will re-poll from the same cursor next cycle.
--- a/lez/indexer/service/src/main.rs
+++ b/lez/indexer/service/src/main.rs
@@ -34,7 +34,9 @@ async fn main() -> Result<()> {
     let cancellation_token = listen_for_shutdown_signal();
 
     let config = indexer_service::IndexerConfig::from_path(&config_path)?;
-    let indexer_handle = indexer_service::run_server(config, data_dir.as_path(), port).await?;
+    let indexer_handle =
+        indexer_service::run_server(config, data_dir.as_path(), port, cancellation_token.clone())
+            .await?;
 
     tokio::select! {
         () = cancellation_token.cancelled() => {
--- a/lez/indexer/service/src/service.rs
+++ b/lez/indexer/service/src/service.rs
@@ -2,7 +2,7 @@ use std::{path::Path, pin::pin, sync::Arc};
 
 use anyhow::{Context as _, Result, bail};
 use arc_swap::ArcSwap;
-use futures::{StreamExt as _, never::Never};
+use futures::StreamExt as _;
 use indexer_core::{IndexerCore, config::IndexerConfig};
 use indexer_service_protocol::{
     Account, AccountId, Block, BlockId, HashType, IndexerStatus, Transaction,
@@ -14,6 +14,7 @@ use jsonrpsee::{
 };
 use log::{debug, error, info, warn};
 use tokio::sync::mpsc::UnboundedSender;
+use tokio_util::sync::CancellationToken;
 
 pub struct IndexerService {
     subscription_service: SubscriptionService,
@@ -21,9 +22,13 @@ pub struct IndexerService {
 }
 
 impl IndexerService {
-    pub async fn new(config: IndexerConfig, storage_dir: &Path) -> Result<Self> {
+    pub async fn new(
+        config: IndexerConfig,
+        storage_dir: &Path,
+        shutdown: CancellationToken,
+    ) -> Result<Self> {
         let indexer = IndexerCore::new(config, storage_dir).await?;
-        let subscription_service = SubscriptionService::spawn_new(indexer.clone());
+        let subscription_service = SubscriptionService::spawn_new(indexer.clone(), shutdown);
 
         Ok(Self {
             subscription_service,
@@ -170,15 +175,17 @@ impl indexer_service_rpc::RpcServer for IndexerService {
 struct SubscriptionService {
     parts: ArcSwap<SubscriptionLoopParts>,
     indexer: IndexerCore,
+    shutdown: CancellationToken,
 }
 
 impl SubscriptionService {
-    pub fn spawn_new(indexer: IndexerCore) -> Self {
-        let parts = Self::spawn_respond_subscribers_loop(indexer.clone());
+    pub fn spawn_new(indexer: IndexerCore, shutdown: CancellationToken) -> Self {
+        let parts = Self::spawn_respond_subscribers_loop(indexer.clone(), shutdown.clone());
 
         Self {
             parts: ArcSwap::new(Arc::new(parts)),
             indexer,
+            shutdown,
         }
     }
 
@@ -192,12 +199,16 @@ impl SubscriptionService {
             // Respawn the subscription service loop if it has finished (either with error or panic)
             if guard.handle.is_finished() {
                 drop(guard);
-                let new_parts = Self::spawn_respond_subscribers_loop(self.indexer.clone());
+                let new_parts = Self::spawn_respond_subscribers_loop(
+                    self.indexer.clone(),
+                    self.shutdown.clone(),
+                );
                 let old_handle_and_sender = self.parts.swap(Arc::new(new_parts));
                 let old_parts = Arc::into_inner(old_handle_and_sender)
                     .expect("There should be no other references to the old handle and sender");
 
                 match old_parts.handle.await {
+                    Ok(Ok(())) => {}
                     Ok(Err(err)) => {
                         error!(
                             "Subscription service loop has unexpectedly finished with error: {err:#}"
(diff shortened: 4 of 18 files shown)


lez/storage/src/indexer/write_atomic.rs
```diff
@@ -216,18 +223,27 @@ impl RocksDBIO {
...
         }
 
+        if let Some(state) = breakpoint {
+            debug_assert!(
+                block
+                    .header
+                    .block_id
+                    .is_multiple_of(BREAKPOINT_INTERVAL.into()),
+                "breakpoint snapshot must accompany an interval-boundary block"
+            );
```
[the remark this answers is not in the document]
erhant: Fixed to return an error in https://github.com/logos-blockchain/logos-execution-zone/pull/607/commits/4c79ce8d9ce49b43246a0103e81607dc215a0ca3

lez/indexer/service/src/service.rs
```diff
@@ -192,12 +199,16 @@ impl SubscriptionService {
             // Respawn the subscription service loop if it has finished (either with error or panic)
             if guard.handle.is_finished() {
```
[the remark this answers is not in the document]
erhant: Added guard in https://github.com/logos-blockchain/logos-execution-zone/pull/607/commits/4c79ce8d9ce49b43246a0103e81607dc215a0ca3

lez/storage/src/indexer/mod.rs
```diff
@@ -156,11 +158,26 @@ impl RocksDBIO {
             ));
         }
 
-        let br_id = closest_breakpoint_id(block_id);
-        let mut state = self.get_breakpoint(br_id)?;
+        // walk down to the nearest snapshot that exists
+        let mut breakpoint_id = closest_breakpoint_id(block_id);
+        let mut state = loop {
```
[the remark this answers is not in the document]
erhant: We do not use `get_meta_last_breakpoint_id` here because it may be stale if the last breakpoint was deleted, and simply computing it from `block_id` is enough

lez/storage/src/indexer/write_atomic.rs
```diff
@@ -216,18 +223,29 @@ impl RocksDBIO {
...
+        if let Some(state) = breakpoint {
+            if !block
+                .header
+                .block_id
+                .is_multiple_of(BREAKPOINT_INTERVAL.into())
+            {
+                return Err(DbError::db_interaction_error(
+                    "Breakpoint snapshot must accompany an interval-boundary block".to_owned(),
+                ));
+            }
```
Pravdyvy: Isn't inverse also true? 
If someone provided `None` on boundary block, then it should fail as well.

Pravdyvy (commented):
Thank you for the PR. One comment on the `put_block` logic.

test_fixtures/src/setup.rs
```diff
@@ -101,10 +101,15 @@ pub async fn setup_indexer(bedrock_addr: SocketAddr) -> Result<(IndexerHandle, T
...
 
-    indexer_service::run_server(indexer_config, temp_indexer_dir.path(), 0)
-        .await
-        .context("Failed to run Indexer Service")
-        .map(|handle| (handle, temp_indexer_dir))
+    indexer_service::run_server(
+        indexer_config,
+        temp_indexer_dir.path(),
+        0,
+        tokio_util::sync::CancellationToken::new(),
```
Arjentix, line 108: So as I understand in future there should be one token that we'll pass in Sequencer and Indexer to also fix our Ctrl-C problem in tests when docker containers are leaked

lez/storage/src/indexer/mod.rs
```diff
@@ -156,8 +159,33 @@ impl RocksDBIO {
             ));
         }
 
-        let br_id = closest_breakpoint_id(block_id);
-        let mut state = self.get_breakpoint(br_id)?;
+        // walk down to the nearest snapshot that exists
+        //
+        // NOTE: we do not use `get_meta_last_breakpoint_id` here because
+        // it may be stale if the last breakpoint was deleted, and simply
+        // computing it from `block_id` is enough
```
Arjentix: Maybe remove last breakpoint at all then?

lez/indexer/core/src/ingest_error.rs
```diff
@@ -40,6 +40,19 @@ pub enum BlockIngestError {
     },
 }
 
+impl BlockIngestError {
+    /// Whether the failure may be transient rather than a property of the block.
+    ///
+    /// FIXME: `StateTransition` is too coarse — its `reason` string mixes genuine
+    /// state-transition rejections with infra failures (risc0 executor teardown,
+    /// storage errors). Once it carries a structured cause, narrow this so only
+    /// infra failures retry.
```
Arjentix, line 49: What's the blocker for that?

lez/indexer/core/src/lib.rs
```diff
@@ -236,6 +243,44 @@ impl IndexerCore {
                             // L1 proceeds regardless
                             self.advance_cursor(&mut cursor, slot);
                         }
+                        Ok(AcceptOutcome::ApplyFailed(ingest_err)) => {
```
Arjentix: Shouldn't we check that error is retryable here?

lez/indexer/service/src/main.rs
Arjentix, line 52: So what does cancellation_token passing here? Wasn't this code enough?

Arjentix: I thought old code would just print "Shutting down server..." and then drop indexer service which should drop gracefully, no?

Arjentix: Ah, I see, subscription loop should be stopped gracefully as well

Arjentix (approved):
LGTM, thanks

test_fixtures/src/setup.rs
erhant: Exactly 🙏🏻

lez/indexer/core/src/ingest_error.rs
erhant: Details within #610 , we need to know the reason behind `ProgramExecutionError` variant within `LeeError`. I think the fix will have to do with something below:

https://github.com/logos-blockchain/logos-execution-zone/blob/3bfa210eac8942a39393d79574b7e2e587443a56/lee/state_machine/src/program/mod.rs#L55-L86

Instead of mapping errors as `ProgramExecutionError` as-is, we can use `exit_code` as our actual error marker, and return a `ProgramExecutionError(ExitCode)` as our actual error code. If the execution itself returns an error (an `anyhow::Error`) then we know whatever went wrong is not about the tx, but about the host machine itself, e.g.
- out-of-memory causing r0vm to be killed
- a SIGINT killing r0vm before its caller

lez/indexer/service/src/main.rs
erhant: Yep, that future was a `Never`-returning one, now it should cancel and return `()` properly.

lez/storage/src/indexer/mod.rs
erhant: Yeah there is no need for this; I had kept it just in case but especially with a fresh reset following this PR; there will no longer be a need on this.

erhant: removed in https://github.com/logos-blockchain/logos-execution-zone/pull/607/commits/f668f3de17140b9d6bcbde0eb0c02b2d3300e893

lez/indexer/core/src/lib.rs
erhant: We actually check it before, and then construct `ApplyFailed` **if and only if** the error is retriable (stated in docstring as well); maybe I can use a more explicit name instead of `ApplyFailed`, like `RetryableFailure`?

EDIT: yeah actually I think I will rename it to `RetryableFailure` hahah

erhant: Done in https://github.com/logos-blockchain/logos-execution-zone/pull/607/commits/313ca7959e4b1c1032b42f3f517bdb99889dee8f

lez/storage/src/indexer/write_atomic.rs
erhant: I think it's okay to let it allow `None` on a boundary block, letting the caller have the choice of persisting or not persisting a breakpoint.

Pravdyvy: This function doc comment says:
```
/// Put a block atomically (via [`WriteBatch`]) along with its L1 header, `Slot`,
    /// and — for interval-boundary blocks — its post-state snapshot.
```
We should either update doc comment or logic. Right now it is inconsistent. 
Also I highly doubt, that actively choosing to skip some breakpoints will have wide usage.

lez/indexer/service/src/main.rs
Arjentix: Hmm, wait, but... Isn't it strange that dropping a struct would result in a leak of task? Maybe it's better to incapsulate this behavior inside? So that `impl Drop` somehow stops the task?

lez/storage/src/indexer/write_atomic.rs
erhant: On a second thought; I think you are right. We should ensure the post-state to be provided. In https://github.com/logos-blockchain/logos-execution-zone/pull/607/commits/15512a4c97073f5a6dbda2ea15a8cc020ebd0bf5 I changed the logic as follows:

- `put_block` now always takes a reference to the post-state, which is always there in the main code path, and a reference to it is cheap.
- only when the `block.id` is a boundary, we store it's breakpoint in DB
- the caller does not have to manage "am i a boundary block?" check at all now that it is handled where it is needed

Note that this required me to add the `initial_state` as that state for some test dummy blocks here and there, but I think it's worth the trade off

What do you think about this approach?

lez/indexer/service/src/main.rs
erhant: `SubscriptionService` does still have `impl Drop` that `abort()`s the task but it does it a bit too "ungracefully", which is why we needed the token.

https://github.com/logos-blockchain/logos-execution-zone/blob/15512a4c97073f5a6dbda2ea15a8cc020ebd0bf5/lez/indexer/service/src/service.rs#L293-L297

That drop fires when ownership unwinds, which on Ctrl-C happens at an _arbitrary_ point during runtime teardown. This is why the old loop kept processing and persisted the false parks. And `abort()` is abrupt (kills at whatever await point it lands on, no clean return, no way to gate the respawn path), while `Drop` being sync can't await or coordinate anything. 

So the token simply allows us a bit more graceful shutdown.

erhant: One extra "security" would be to also cancel the token within `drop` impl above, but not the cancellation token itself; instead a `child_token` of it's parent that should explicitly be passed on construction (so that dropping the service does not cancel every other thing out there).

But I think thats too overkill.

lez/storage/src/indexer/write_atomic.rs
Pravdyvy: Fine by me

Pravdyvy (approved):
lgtm

lez/indexer/service/src/main.rs
erhant: @copilot do you think we could do this with `Drop` alone?

erhant: Spoke with @Arjentix and we've aggreed on a middle-ground; we can auto-cancel the token within `Drop` impl, the token is still passed from outside but the user is free to pass a token as-is or via `child_token`.

Done in https://github.com/logos-blockchain/logos-execution-zone/pull/607/commits/0f15c60cd06736e82ff5bc8d6580014d478949c2

Issue #611, opened 2026-07-09
cross-zone: cleanup follow-up (genesis/config extraction, verifier replay semantics, code comments)


moudyellaz:
Tracking the follow-up cleanups agreed during review of #542. The genesis and config extraction, the verifier replay fix, and the code comments below are the committed host-only cleanup pass and land as one PR. The deferred and general-layer sections are documented here for reference and are handled separately.

Genesis and config extraction (host-only):
- Move the six cross-zone builtin programs out of testnet_initial_state::initial_programs into a cross-zone-specific setup, gated behind the cross-zone config or via a GenesisAction::DeployProgram, so the production builtin list is unchanged for a zone with no cross-zone config. (testnet_initial_state/src/lib.rs:296)
- Move the wrapped_token config account out of initial_public_accounts into the same cross-zone genesis setup. (testnet_initial_state/src/lib.rs:279)
- Use the Balance alias (lee_core, = u128) for the GenesisAction amount and balance fields instead of hardcoded u128. (sequencer/core/src/config.rs:32)

Verifier replay semantics (host-only):
- The verifier bails on an already-seen message key, but the inbox treats a replay as an idempotent no-op on chain, so the verifier can halt an honest indexer on a legitimate replay. Change verify_block to continue on an already-seen key rather than bail, and add a test. (cross_zone_verifier.rs:125, cross_zone_inbox/src/main.rs:76)

Code comments and demo safety (host-only):
- Note the trust boundary at the verifier call site: we trust the peer's finalized block and do not re-execute source transactions. (cross_zone glue)
- Note the single-key limitation and its tie to decentralized sequencing at the pinned-key field. (cross_zone_verifier.rs:75)
- Mark the bridge demo not production-safe with a doc comment, pending source verification (see general layer below), on cross_zone_bridge.rs and the bridge_lock/wrapped_token cores.

Deferred (needs guest changes and artifact rebuild):
- Replace V03State::insert_genesis_account seeding with a genesis Configure transaction, so per-zone genesis accounts (inbox config, bridge-lock holdings) are produced by a transaction rather than by direct state insertion. This requires adding genesis-only Configure instructions to the cross_zone_inbox and bridge_lock guests, which changes their ELFs and forces an artifact rebuild, so it is done as its own PR, not in the host-only cleanup above. (state/mod.rs:203)
[...]

Issue #272, opened 2026-01-07
explicit panic on `wallet token new`


fryorcraken:
See commands below:

```
▶ wallet account new public 
Generated new account with account_id Public/6gFwLyoyV6bcFDgyL5rVbzo6H3fPmT9s5uZJLiYhXyFN at path /1
Stored persistent accounts at "./wallet-alice/storage.json"

▶ wallet account new public 
Generated new account with account_id Public/5Qfw3S73dJdmgXQStRG7ynhQv2AP8sZ29Qrm6CMJHrCF at path /0/0
Stored persistent accounts at "./wallet-alice/storage.json"

▶ RUST_BACKTRACE=1 wallet token new --name MOONCOIN --total-supply 21000000 --definition-account-id Public/6gFwLyoyV6bcFDgyL5rVbzo6H3fPmT9s5uZJLiYhXyFN --supply-account-id Public/5Qfw3S73dJdmgXQStRG7ynhQv2AP8sZ29Qrm6CMJHrCF 

thread 'main' (896637) panicked at /home/fryorcraken/src/logos-blockchain/lssa/wallet/src/cli/programs/token.rs:1437:21:
explicit panic
stack backtrace:
   0: __rustc::rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::panicking::panic
   3: <wallet::cli::programs::token::CreateNewTokenProgramSubcommand as wallet::cli::WalletSubcommand>::handle_subcommand::{{closure}}
   4: <wallet::cli::programs::token::TokenProgramSubcommand as wallet::cli::WalletSubcommand>::handle_subcommand::{{closure}}
   5: wallet::cli::execute_subcommand_with_auth::{{closure}}
   6: tokio::runtime::park::CachedParkThread::block_on
   7: tokio::runtime::context::runtime::enter_runtime
   8: tokio::runtime::runtime::Runtime::block_on
   9: wallet::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.

▶ RUST_BACKTRACE=full wallet token new --name MOONCOIN --total-supply 21000000 --definition-account-id Public/6gFwLyoyV6bcFDgyL5rVbzo6H3fPmT9s5uZJLiYhXyFN --supply-account-id Public/5Qfw3S73dJdmgXQStRG7ynhQv2AP8sZ29Qrm6CMJHrCF  

[...]

PR #615, opened 2026-07-13
refactor!(cross-zone): deploy programs at genesis instead of builtins

moudyellaz:
## 🎯 Purpose

Review cleanups for the cross-zone messaging feature (follows #542, now merged), addressing the follow-ups tracked in #611. Moves the cross-zone builtin programs out of the production genesis, fixes the indexer's replay handling, adds explanatory and safety docs, and removes stale duplicate crates.

## ⚙️ Approach

Cross-zone programs are no longer registered in the production genesis; a zone that participates declares the ones it uses via a new GenesisAction::DeployProgram (sequencer) and a matching deploy_programs list (indexer). The cross-zone genesis accounts are seeded through the state constructor rather than an ad-hoc mutator.

- [x] Remove the stale top-level programs/ core crates, duplicates of lez/programs/*/core
- [x] Deploy the cross-zone builtins at genesis via GenesisAction::DeployProgram and the indexer deploy_programs list, so production genesis carries none of them
- [x] Seed the cross-zone genesis accounts (inbox config, wrapped-token config, bridge-lock holdings) via with_programs / with_public_accounts, dropping V03State::insert_genesis_account
- [x] Accept a replayed cross-zone dispatch as an on-chain no-op instead of halting the indexer
- [x] Use the Balance alias for the GenesisAction amounts
- [x] Add trust-boundary, single-key, reference-adapter, and bridge-demo not-production-safe docs

## 🧪 How to Test

Dev mode, needs Docker for Bedrock. On macOS set DYLD_FALLBACK_FRAMEWORK_PATH.
- RISC0_DEV_MODE=1 cargo nextest run --workspace --exclude integration_tests --exclude test_fixtures --all-features
- RISC0_DEV_MODE=1 cargo nextest run -p integration_tests --test cross_zone_verified --test cross_zone_bridge --test cross_zone_ingress_guard

## 🔗 Dependencies

None

## 🔜 Future Work

None

## ⚠️ Breaking Changes

Genesis state root changes: cross-zone builtins are no longer in the production genesis. The sequencer and indexer configs gain a cross-zone program deploy list (GenesisAction::DeployProgram and deploy_programs) that cross-zone-participating zones must set.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


lez/indexer/core/src/cross_zone_verifier.rs
agureev: Based on line #272 in indexer's `lib.rs`, it seems that this `seen` ends up being populated even if the `accept_block` does not succeed (e.g. when `Parked`) as the state is still mutated and not dropped.

So this means that if I am a dishonest sequencer at state S and block N, I can first submit a proper tx T with e.g. a `ping` with a key K but mangle something (e.g. the block height, set it to N + 2) so that `verify_block` succeeds but `accept_block` parks. At this point, the indexer have recorded the key, so the next time I encounter the key I skip it.

Now, the (malicious) sequencer generates the next block again starting at S and N again, but now uses the same key K but faking the payload, instead pretending it is e.g. `Mint`. This should be caught on the indexer by rederivation, looking at the origin of the transaction, but the indexer verifier includes K in `seen` already, so the rederivation is skipped, recognizing the mint.

Why not actually make the guest inbox just crash alongside the indexer like it did before? Or I guess maybe it introduces its own timing issues. Or maybe make `verify_block` return a vector of keys to be inserted after the block has been accepted?

agureev (changes requested):
Thank you for the PR! Left a comment regarding skipping already-seen keys and not dropping them on non-accepted blocks.

lez/indexer/core/src/cross_zone_verifier.rs
moudyellaz: Good catch! thanks. I took your second option: verify_block returns the keys, recorded via record_seen only on AcceptOutcome::Applied, so seen mirrors the inbox's on-chain seen-shard (a parked or retryable-failed block records nothing). Added a regression test for the park-then-forge case.

agureev (approved):
LGTM, thank you for addressing the comment!

lee/state_machine/src/state/mod.rs
```diff
@@ -287,6 +279,50 @@ impl V03State {
...
 
+    /// Order-independent fingerprint of the genesis-relevant state: the public
+    /// account set, the deployed program set, and the commitment-set digest.
+    ///
+    /// The sequencer and the indexer build genesis separately from their own
+    /// configs (the cross-zone deploy set is seeded directly, not carried by a
+    /// genesis transaction), so a mismatched `DeployProgram` / `deploy_programs`
+    /// list would otherwise diverge silently. Both nodes log this at startup;
+    /// equal values mean the two genesis states agree. Entries are sorted by id
+    /// before hashing, so the value does not depend on `HashMap` iteration order.
```
Arjentix:
```suggestion
```

lee/state_machine/src/state/mod.rs
```diff
@@ -287,6 +279,50 @@ impl V03State {
...
+    #[must_use]
+    pub fn genesis_fingerprint(&self) -> [u8; 32] {
```
Arjentix, line 297: Let's apply destructuring pattern in the body of this function: `let Self { public_state, programs, ..... } = self;`

So that we don't forget to update this function when we add a new field into the struct

lez/cross_zone/src/lib.rs
erhant: `amount` can use `Balance` instead of `u128` as per #611

lez/indexer/core/src/cross_zone_verifier.rs
```diff
@@ -112,20 +115,25 @@ impl CrossZoneVerifier {
...
+    pub async fn verify_block(&self, block: &Block) -> Result<Vec<MessageKey>> {
+        let mut verified = Vec::new();
         for tx in &block.body.transactions {
             let Some(msg) = Self::decode_dispatch(tx) else {
                 continue;
             };
 
             let key = message_key(&msg.src_zone, msg.src_block_id, msg.src_tx_index);
             if self.seen.read().await.contains(&key) {
-                bail!(
```
erhant, line 125: would a `debug!` log be useful here?

lez/sequencer/core/src/lib.rs
```diff
@@ -662,24 +685,35 @@ fn build_genesis_state(config: &SequencerConfig) -> (lee::V03State, Vec<LeeTrans
...
-    // Seed this zone's cross-zone inbox config so the inbox guest can authorize
-    // inbound peer messages (zone-specific config, not produced by any tx).
-    if let Some(cross_zone) = &config.cross_zone {
-        let self_zone = *config.bedrock_config.channel_id.as_ref();
-        let (config_id, config_account) =
-            cross_zone::build_inbox_config_account(self_zone, cross_zone);
-        state.insert_genesis_account(config_id, config_account);
-    }
+/// Bridge-lock holder balances configured for this zone's genesis.
+fn bridge_lock_holdings(genesis: &[GenesisAction]) -> Vec<(lee::AccountId, u128)> {
```
erhant: this `u128` can be a `Balance`?

erhant (approved):
LGTM 🙏🏻 
few small comments

lez/cross_zone/src/lib.rs
moudyellaz: done

lez/indexer/core/src/cross_zone_verifier.rs
moudyellaz: done

lez/sequencer/core/src/lib.rs
moudyellaz: done

lez/testnet_initial_state/src/lib.rs
```diff
@@ -292,12 +271,6 @@ fn initial_programs() -> Vec<Program> {
         programs::vault(),
         programs::faucet(),
         programs::bridge(),
-        programs::cross_zone_outbox(),
-        programs::cross_zone_inbox(),
-        programs::ping_sender(),
-        programs::ping_receiver(),
-        programs::bridge_lock(),
-        programs::wrapped_token(),
```
Arjentix, line 300: So now that we have decided that we want just hardcoded LEZ-to-LEZ communication it may make sense to leave some important generic programs here in default setup.

To me looks like inbox and outbox form the important part of this communication and probably makes sense to leave them here.

While things like ping_ programs look more like an example usage of this inbox/outbox, so they can be deleted from here, right.

lez/sequencer/core/src/config.rs
```diff
@@ -21,15 +22,22 @@ use url::Url;
...
-        amount: u128,
+        amount: Balance,
+    },
+    /// Registers a cross-zone builtin at genesis, keeping the cross-zone programs
+    /// out of the production builtin set. A zone lists the ones it uses. This set
+    /// must match the indexer's `deploy_programs`, or the two genesis states
+    /// diverge; compare the genesis fingerprint both nodes log at startup.
+    DeployProgram {
+        program: CrossZoneProgram,
     },
```
Arjentix: Let's better make it generic, so that we can deploy any program, not just cross zone program

lez/sequencer/core/src/lib.rs
```diff
@@ -650,8 +670,11 @@ fn build_genesis_state(config: &SequencerConfig) -> (lee::V03State, Vec<LeeTrans
             GenesisAction::SupplyBridgeAccount { balance } => {
                 Some(build_supply_bridge_account_genesis_transaction(*balance))
             }
-            // Force-inserted below: bridge_lock has no mint transaction.
-            GenesisAction::SupplyBridgeLockHolding { .. } => None,
+            // Seeded/registered directly (accounts via `cross_zone::genesis_accounts`,
+            // programs via `with_programs`), not by a genesis tx.
+            GenesisAction::SupplyBridgeLockHolding { .. } | GenesisAction::DeployProgram { .. } => {
+                None
+            }
```
Arjentix: Judging by our discussions in Notion we really want to have first (genesis) block to represent initial genesis configuration. And before cross zone PR that was the thing. So I suggest to fix it there. There should be account transfer and program deployment transactions right here as we do for other genesis actions

lez/indexer/core/src/config.rs
```diff
@@ -37,6 +38,12 @@ pub struct IndexerConfig {
     /// Defaults to `false`: on mismatch the indexer refuses to start.
     #[serde(default)]
     pub allow_chain_reset: bool,
+    /// Cross-zone builtins to deploy at genesis, mirroring the sequencer's
+    /// `DeployProgram` actions so the replayed genesis is identical. This set
+    /// must match the sequencer's; compare the genesis fingerprint both nodes
+    /// log at startup to confirm.
+    #[serde(default)]
+    pub deploy_programs: Vec<CrossZoneProgram>,
```
Arjentix: If you fix my suggestion about genesis then you won't need this in Indexer. It will just catch up programs from block history

Arjentix (changes requested):
Thanks for the PR, left some comments I think we need to address. After resolving them would like to re-review.

The most important IMO is this: https://github.com/logos-blockchain/logos-execution-zone/pull/615#discussion_r3596748986

lez/programs/bridge_lock/src/main.rs
Arjentix, line 1: I think you also need to add all new main.rs from `lez/programs/*` to `lez/programs/Cargo.toml` the same way all other programs are added

lee/state_machine/src/state/mod.rs
Arjentix: This is not resolved yet

lez/sequencer/core/src/lib.rs
```diff
@@ -853,27 +853,55 @@ fn replay_unfulfilled_deposit_events(
...
+            account_id,
+            balance,
+        } => Some(build_supply_account_genesis_transaction(
+            account_id, *balance,
+        )),
+        GenesisAction::SupplyBridgeAccount { balance } => {
+            Some(build_supply_bridge_account_genesis_transaction(*balance))
+        }
+        // Seeded directly (holdings via `build_holding_account`), not by a genesis tx.
+        GenesisAction::SupplyBridgeLockHolding { .. } => None,
```
Arjentix: I think this can easily be a transaction, no? Without it indexer still needs genesis configuration to know which accounts were supplied

lez/sequencer/core/src/lib.rs
```diff
@@ -883,26 +911,20 @@ fn build_genesis_state(config: &SequencerConfig) -> (lee::V03State, Vec<LeeTrans
...
-        let (config_id, config_account) =
-            cross_zone::build_inbox_config_account(self_zone, cross_zone);
-        state.insert_genesis_account(config_id, config_account);
-    }
-
     (state, genesis_txs)
 }
 
+/// Bridge-lock holder balances configured for this zone's genesis.
+fn bridge_lock_holdings(genesis: &[GenesisAction]) -> Vec<(lee::AccountId, lee::Balance)> {
```
Arjentix: If you make it a transaction this function would not be needed, but anyway I'll note that it's better to return iterator from here to remove redundant collect

Arjentix:
LGTM, left some comments, I think they can be resolved quickly

lez/programs/bridge_lock/src/main.rs
moudyellaz: All cross-zone guests are already registered as `[[bin]]` in `lez/programs/Cargo.toml` (`cross_zone_inbox`, `cross_zone_outbox`, `ping_sender`, `ping_receiver`, `bridge_lock`, `wrapped_token`), with their cores in the `programs` feature. Every `main.rs` under `lez/programs/*` has a bin, nothing missing here.

lee/state_machine/src/state/mod.rs
moudyellaz: Done in dc977db7. `genesis_fingerprint` now destructures `let Self { public_state, private_state, programs } = self;`, all fields listed with no `..`, so adding a field forces a decision here.

lez/sequencer/core/src/lib.rs
moudyellaz: Can't be a genesis tx. The holder is a bridge_lock-owned account at the holder's own pubkey, `build_holding_account` sets `program_owner = bridge_lock`, and `Lock` asserts that and debits the balance directly (the direct-move design we picked over a chained transfer to avoid the mint-from-nothing hole). Claiming ownership at a non-PDA id needs `Claim::Authorized`, which requires that account to be signer-authorized in the tx. The genesis builder has the holder's pubkey but not its key, so it can't sign; `Claim::Pda` would force a bridge_lock PDA id, not the holder's pubkey. So the holding stays direct-seeded and the indexer keeps `bridge_lock_holdings` for this one case. Making it block-reconstructable would need a bridge_lock redesign (e.g. PDA-based holdings), bigger than a genesis tx.

lez/sequencer/core/src/lib.rs
moudyellaz: Done in dc977db7. `bridge_lock_holdings` returns an iterator now, dropped the redundant `collect` (and the caller's `.into_iter()`). Keeping holdings direct-seeded per the thread above, so the function stays.

lez/sequencer/core/src/lib.rs
Arjentix: Ok, but I'd say let's keep an eye on it and maybe have a tracking issue

Arjentix (approved):
Thanks for addressing my comments

lez/sequencer/core/src/lib.rs
moudyellaz: Filed #644 to track it.

PR #618, opened 2026-07-14
feat(ci): use docker image for ci

Arjentix:
## 🎯 Purpose

Our CI uses ubuntu-latest and installs all dependencies on top of it on each job. Not only this takes some time but also makes risc0 services to deny our requests because of high load.

This PR introduces CI docker image.

## ⚙️ Approach

- Derive image tag by hashing dockerfile and rust toolchain file
- If image was already built then use it, otherwise build and push it
- Run most of the jobs from this image
- Exception: artifacts build because we can't do docker inside docker

## 🧪 How to Test

See CI

## 🔗 Dependencies

None

## 🔜 Future Work

None

## 📋 PR Completion Checklist

- [x] Complete PR description
~- [] Add/update tests~


.github/actions/run-in-ci-image/action.yml
```diff
@@ -0,0 +1,54 @@
...
+        # to mean the same thing inside and outside the container.
+        docker run --rm \
+          --network host \
+          --volume /var/run/docker.sock:/var/run/docker.sock \
+          --volume "$PWD:$PWD" \
+          --volume "$HOME/.cargo/registry:/usr/local/cargo/registry" \
+          --workdir "$PWD" \
+          "${env_args[@]}" \
+          "$INPUT_IMAGE" \
+          bash -o pipefail -c "$INPUT_RUN"
```
erhant: small nitpick: do you think `-e` here would be useful as well? it would return on first non-zero exit, if $INPUT_RUN has multiple steps etc

.github/workflows/ci-image.yml
```diff
@@ -0,0 +1,71 @@
...
+# it only when it isn't published yet.
+on:
+  workflow_call:
+    outputs:
+      image:
+        description: Image reference to pass to a job's `container:`.
+        value: ${{ jobs.ci-image.outputs.image }}
+
+jobs:
+  ci-image:
```
erhant, line 17: another nitpick: this workflow can be called both from `bench-regression.yml` and `ci.yml` if the changed paths allow it, but this will cause a double work for both workflows, instead the "first one" could build the image, and the "second one" can use that newly built image (it will wait for the first one, but wont do any compute at all)

this is done with adding:

```yml
   concurrency:
        group: ci-image-${{ github.sha }}
        cancel-in-progress: false
```

erhant (approved):
LGTM, thanks!

two small comments

Pravdyvy (approved):
lgtm

.github/actions/run-in-ci-image/action.yml
Arjentix: Cool idea, will add

.github/workflows/ci-image.yml
Arjentix: Hm, should be good, added, let's see

.github/docker/ci.Dockerfile
```diff
@@ -0,0 +1,88 @@
...
+# For `cargo +nightly fmt` in the fmt-rs job.
+RUN rustup toolchain install nightly --profile minimal --component rustfmt
+
+# The bootstrap script only ever drops rzup in $HOME/.risc0/bin, so run it
+# against the build-time HOME and keep just the binary. rzup itself then honours
+# RISC0_HOME and installs the toolchain, r0vm and cargo-risczero under /usr/local.
+RUN curl -L https://risczero.com/install | bash \
+    && mv /root/.risc0/bin/rzup /usr/local/bin/rzup \
+    && rm -rf /root/.risc0 \
+    && rzup install
```
moudyellaz: Worth pinning versions here (rzup install <version>, binstall with versioned crates)? The tag hash freezes whatever a build fetches, so any unrelated Dockerfile edit later silently rebuilds with newer r0vm/tools while the workspace still pins risc0 3.0.5. A pin makes the bump an explicit diff, and bumping it rebuilds the image by construction.

moudyellaz (approved):
lgtm, thanks. Left a minor suggestion

.github/docker/ci.Dockerfile
Arjentix: Ok, makes sense, pushed a new commit with fixed versions

PR #619, opened 2026-07-14
feat: add bulk public account reads

### LINKED ISSUE #617: Feature request: add bulk public account reads
## Summary

Add a bulk public-account query that accepts multiple account IDs and returns their full account states in one request.

Suggested API path:

```text
sequencer RPC: getAccounts(Vec<AccountId>) -> Vec<Account>
wallet core: get_accounts_public(Vec<AccountId>)
wallet FFI: wallet_ffi_get_accounts_public(...)
```

The execution-zone module can then expose the wallet FFI operation to external wallet consumers.

## Motivation

Wallet consumers commonly need to inspect every signable public account to classify account type, program owner, balance, nonce, and data. Wallet account listing currently returns only account ID and privacy kind, while full state is available only through singular account reads.

For a wallet containing `N` public accounts, discovery therefore requires `N` sequencer requests. The wallet FFI also holds its wallet-core mutex while each network request completes, so client-side asynchronous fan-out does not reduce total scan time. Latency grows linearly with wallet size and delays any feature that needs a coherent wallet snapshot.

A module-level loop alone would reduce transport overhead but retain `N` sequencer requests. Bulk support needs to reach the sequencer RPC.

## Related work

Issue #185 - Implement a wallet command to list owned accounts and their information
is a closed functional predecessor. It requests a wallet command that lists
owned accounts with their addresses and contents, but does not define a bulk
sequencer read, wallet-core or FFI API, ordered result semantics, or the
network and mutex behavior addressed here.

PR #471 - Add getAccountAtBlock RPC to indexer
mentions a future `getAccountsAtBlock(account_ids, block_id)` batch operation.
That proposal concerns historical indexer state. This request instead concerns
current public-account state fetched through the sequencer and surfaced by the
wallet FFI.

## Proposed behavior

- Accept an ordered vector of public account IDs.
- Return one full `Account` per input ID in matching order.
- Preserve the singular `getAccount` state semantics.
- Handle an empty input vector without error.
- Keep existing singular methods for compatibility.
- Provide an FFI-owned result list and matching free function.

The existing `getAccountsNonces(Vec<AccountId>)` method provides a precedent for vector account queries.

## Expected benefit

- One network round trip instead of one per account.
- One sequencer state-lock acquisition for the requested snapshot.
- Lower reconnect and account-discovery latency for larger wallets.
- Simpler consumers: one coherent account-state response instead of manual serial aggregation.

## Acceptance criteria

- Sequencer RPC exposes `getAccounts` and returns states in input order.
- Wallet core exposes a typed bulk public-account method.
- Wallet FFI exposes bulk read and memory-release functions.
- Tests cover empty, single-account, multi-account, and invalid-request behavior.
- A multi-account read performs one sequencer RPC request.

3esmit:
## 🎯 Purpose

Add bulk public-account reads across the sequencer RPC, wallet core, and wallet FFI.

Wallet consumers can now retrieve multiple full account states in one network request and one sequencer state-lock acquisition, instead of issuing one request per account.

Closes #617.

## ⚙️ Approach

- [x] Add ordered `getAccounts(Vec<AccountId>) -> Vec<Account>` sequencer RPC.
- [x] Preserve existing account lookup semantics, input order, and duplicate IDs.
- [x] Support empty, single-account, and multi-account requests.
- [x] Add `WalletCore::get_accounts_public`.
- [x] Add FFI bulk-read and matching memory-release functions.
- [x] Expose the FFI batch limit and an explicit invalid-argument error.
- [x] Limit batches to 24 accounts so worst-case JSON responses remain below the default 10 MiB transport limit.
- [x] Regenerate the C header and add unit and integration coverage.

## 🧪 How to Test

```bash
cargo test -p sequencer_service_rpc --features client,server --lib
cargo test -p wallet-ffi --lib

cargo test -p integration_tests \
  --test account \
  get_public_accounts_returns_states_in_input_order \
  -- --exact

cargo test -p integration_tests \
  --test wallet_ffi \
  test_wallet_ffi_get_accounts_public \
  -- --exact

cargo check -p wallet -p wallet-ffi -p sequencer_service
cargo fmt --all -- --check
```

The tests cover empty input, ordered multi-account results, duplicate IDs, unknown accounts, oversized requests, null FFI inputs, and nested FFI allocation cleanup.

## 🔗 Dependencies

None.

## 🔜 Future Work

Expose the new wallet FFI operation through higher-level execution-zone consumers as needed.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9097,6 +9097,7 @@ version = "0.1.0"
 dependencies = [
  "jsonrpsee",
  "sequencer_service_protocol",
+ "serde_json",
 ]
 
 [[package]]
--- a/integration_tests/tests/account.rs
+++ b/integration_tests/tests/account.rs
@@ -9,6 +9,7 @@ use key_protocol::key_management::KeyChain;
 use lee::Data;
 use lee_core::account::Nonce;
 use log::info;
+use sequencer_service_rpc::{ClientError, MAX_ACCOUNTS_PER_REQUEST, RpcClient as _};
 use tokio::test;
 use wallet::{
     account::{AccountIdWithPrivacy, HumanReadableAccount, Label},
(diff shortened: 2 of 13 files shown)


moudyellaz:
Closing this, @3esmit. The sequencer half targets `lez/sequencer/service/src/service.rs`, which no longer exists: the RPC server became an actor and that code moved to `lez/sequencer/actors/rpc_server/src/actor/service.rs`. That makes this a rewrite rather than a rebase, and the wallet-ffi half would land on a header several open PRs are actively changing. The gap stays tracked in #617.

Issue #620, opened 2026-07-15
Third party private account initialization authorization


jonesmarvin8:
Third parties can initialize private accounts through privacy transactions using the account's public keys (`npk`, `vpk`) and an `identifier`. These accounts correspond with `AccountIdentity::PrivateForeign` (in `wallet`) and `InputAccountIdentity::PrivateUnauthorized` in `lee`. The privacy preserving circuit enforces `is_authorized = false` for these accounts.

Some LEZ programs functionalities enforce `is_authorized = true` for [account initialization](https://github.com/logos-blockchain/lez-programs/blob/5b82a52c6b812a541b0b864a2fca8aa9692aaac1/programs/token/src/initialize.rs#L17), token holding for [new definition](https://github.com/logos-blockchain/lez-programs/blob/5b82a52c6b812a541b0b864a2fca8aa9692aaac1/programs/token/src/new_definition.rs#L45).

This artificially blocks privacy transactions that initializes a new private account for another entity. Only the account owner (e.g., entity that possesses the account's `nsk`) can update the account. Thus, the only interaction with the private account is its initialization.

Due to this, `is_authorized = true` for `InputAccountIdentity::PrivateUnauthorized` logic branch of privacy preserving circuit does not sacrifice the integrity or security of a user's account. This only provides authorization for uninitiallized private accounts to third parties. Any future updates or usage can only be done by the account owner.

**Changes**
`InputAccountIdentity::PrivateUnauthorized` rename to `InputAccountIdentity::PrivateForeignInit` to reflect changes in authorization.
`PrivateForeign` accounts initialized in wallet with `is_authorized = true`
Privacy preserving circuit enforces `is_authorized = true` for `PrivateForeignInit` accounts.

jonesmarvin8:
PR 621 handled this.

PR #621, opened 2026-07-15
refactor: `PrivateUnauthorized` authorization changed to true

jonesmarvin8:
## 🎯 Purpose

Address [Issue 620](https://github.com/logos-blockchain/logos-execution-zone/issues/620). `PrivateUnauthorized` accounts are accounts initialized without the `nsk`. This allows third parties to initialize the account. However, in practice LEZ programs have blocked this.

Third parties can only initialize accounts; updates **require** the `nsk`. Thus, no security or integrity compromises occur by assuming that all private account initializations are authorized.

## ⚙️ Approach

- [X] Rename `InputAccountIdentity::PrivateUnauthorized` to `InputAccountIdentity::PrivateForeignInit`
- [X] Update wallet initialization for `AccountIdentity::PrivateForeign` to include `is_authorized = true`
- [X] Update privacy preserving circuit to enforce `is_authorized = true` for `Private::ForeignInit`

## 🧪 How to Test

Simple logic change. All tests should still work.

## 🔗 Dependencies

N/A

## 🔜 Future Work

N/A

## 📋 PR Completion Checklist

- [X] Complete PR description
- [X] Add/update tests


agureev (approved):
Thank you for the PR! Generally looks good, I do not see anything that would break the semantics of current programs.

However, I want to duplicate here something mentioned in private discussions - based on my understanding, issues like these (i.e. general program semantics for private accounts) are best decided in relation to privacy agnosticism with respect to public programs which developers will be writing.

However, as discussed in conversation with @jonesmarvin8 and @schouhy, authorization is very different in public and private domain because in the public domain you can utilize and account in execution without authorizing it but you cannot in private because utilization of private account means nullification which is currently conjoined with authorization.

Based on @schouhy's suggestion, I think it would be generally best in the future to decouple nullification and authorization of programs in general. I know this is parallel to the issue of the PR in some sense, but I do think it is important to note given that both touch upon private authorization mechanisms and I think there has been no properly recorded instance of such a discussion on GH.

jonesmarvin8:
> Thank you for the PR! Generally looks good, I do not see anything that would break the semantics of current programs.
> 
> However, I want to duplicate here something mentioned in private discussions - based on my understanding, issues like these (i.e. general program semantics for private accounts) are best decided in relation to privacy agnosticism with respect to public programs which developers will be writing.
> 
> However, as discussed in conversation with @jonesmarvin8 and @schouhy, authorization is very different in public and private domain because in the public domain you can utilize and account in execution without authorizing it but you cannot in private because utilization of private account means nullification which is currently conjoined with authorization.
> 
> Based on @schouhy's suggestion, I think it would be generally best in the future to decouple nullification and authorization of programs in general. I know this is parallel to the issue of the PR in some sense, but I do think it is important to note given that both touch upon private authorization mechanisms and I think there has been no properly recorded instance of such a discussion on GH.

I think authorization has been doing two tasks as well:
- Valid initialization; for public accounts we do not use `identifiers` so a malicious account initialization would "waste that set of keys.
- Update account's data and balance: authorization for account data manipulation is dependent on program's logic.

I'm not convinced initialization should require the same authorization as updates for either public or private accounts.

agureev:
> I'm not convinced initialization should require the same authorization as updates for either public or private accounts.

Honestly probably something that the smart contract team should simply decide on. They have some assumptions they want to work with and build apps against, they can be stronger or weaker but it is probably the developers who know best what they want. Here the important point is that we just try to keep the semantics as similar as possible for the private case for interop reasons.

jonesmarvin8:
> > I'm not convinced initialization should require the same authorization as updates for either public or private accounts.
> 
> Honestly probably something that the smart contract team should simply decide on. They have some assumptions they want to work with and build apps against, they can be stronger or weaker but it is probably the developers who know best what they want. Here the important point is that we just try to keep the semantics as similar as possible for the private case for interop reasons.

@0x-r4bbit  Thoughts?

Pravdyvy (approved):
lgtm

erhant (approved):
lgtm!

PR #623, opened 2026-07-15
feat(wallet)!: Request distribution for multiple sequencers

Pravdyvy:
## 🎯 Purpose

Distribute important requests(i.e. transactions) between multiple sequencers.

## ⚙️ Approach

- [x] `MultiSequencerClient` now stores an ordered list of leaders. 
- [x] Requests is being sent and polled distributely. 
- [x] Tests update.

## 🧪 How to Test

Added unit tests which choose subset of leaders.

## 🔗 Dependencies

None

## 🔜 Future Work

- Move statistics collection into another thread, probably using actor approach.

## Breaking changes

Configuration for a wallet changed, now it takes additional subconfig for multi-sequencer client. Have a default serde flag.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


lez/wallet/src/lib.rs
```diff
@@ -810,12 +821,17 @@ impl WalletCore {
...
                     .send_transaction(LeeTransaction::Public(tx.clone()))
                     .await
             })
-            .await?)
+            .await
+            .into_iter()
+            .find(std::result::Result::is_ok)
+            .ok_or(ExecutionFailureKind::SequencerError(anyhow::anyhow!(
+                "All sequencers rejected transaction"
+            )))??)
```
erhant: This seems to repeat across few places, with a static typed String error message too

maybe can be shared with code-reuse, or a specific SequencerError variant?

lez/wallet/src/multi_client.rs
```diff
@@ -358,10 +475,11 @@ pub async fn actualize_client(client: &SequencerClient) -> StatisticsUpdate {
 }
 
 #[must_use]
-pub fn choose_leader(
+pub fn choose_leaders(
```
erhant, line 482: I wonder if using the std [BinaryHeap](https://doc.rust-lang.org/beta/std/collections/struct.BinaryHeap.html) would be nicer here; e.g. impl `Ord` for the SequencerClient (or a wrapper for it specifically) and just use a max-heap with all these clients, with statistics updated every now and then

we wont have that many clients anyways so there is probably no real complexity-wise win here, but just an idea

lez/wallet/src/multi_client.rs
```diff
@@ -34,6 +34,17 @@ pub struct StatisticsUpdate {
...
+    #[must_use]
+    pub const fn failure() -> Self {
+        Self {
+            latency: 0_f32,
+            new_latest_block_id: None,
+            is_failed: true,
+        }
+    }
+}
+
```
erhant, line 38: Would a two-variant `StatisticsUpdate::Success { latency, new_latest_block_id }` and `StatisticsUpdate::Failure` be more ergonomic?

erhant (approved):
lgtm! few comments

lez/wallet/src/lib.rs
Pravdyvy: It is quite a specific scenario. Introduced the specific ExecutionFailureKind variant.

lez/wallet/src/multi_client.rs
Pravdyvy: Good idea, refactored.

lez/wallet/src/multi_client.rs
Pravdyvy: Probably too much for this exact configuration. 
Would require merging Statistics into client where they really should not belong(I think client should be cheaply clonable).
I am not happy how this configuration is written as well, I have ideas for update, precisely move all statistic collection into actor running in parallel. It should be the best solution for wallet as CLI, as lib and as FFI.

lez/wallet/src/multi_client.rs
```diff
@@ -169,54 +179,152 @@ impl MultiSequencerClient {
...
+                    break;
+                }
+            }
+        }
 
         {
             let mut statistic_updates_guard = self.statistic_updates.write().await;
-            statistic_updates_guard.push(statistics_update);
+
+            statistic_updates_guard.extend(statistic_map.into_iter());
```
moudyellaz: HashMap::extend replaces the value for an existing key rather than appending, and this buffer only clears once per command at client_rotation. Since a single command makes several metered_get calls to the same helm url, each flush overwrites the previous one and only the last sample survives (dropped Failures included), so sample_size grows ~1 per command instead of per call. Could we merge instead, e.g. `for (url, mut v) in statistic_map { guard.entry(url).or_default().append(&mut v); }`? Same at metered_send.

lez/wallet/src/multi_client.rs
```diff
@@ -431,35 +539,44 @@ pub fn choose_leader(
...
         //
         // However one can argue that this:
         //
-        // [-old_std...................old_lat........new_lat.........+new_std.......+old_std]
+        // [-right_std < right_lat < left_lat < +left_std < +right_std]
         //
         // is still better, but it is up to discussion
-        if (new_lat <= old_lat) && ((new_lat + new_std) < (old_lat + old_std)) {
-            *x
+        if (left_lat <= right_lat) && ((left_lat + left_std) < (right_lat + right_std)) {
```
moudyellaz: This comparator never returns Equal (cmp(a,a) is Greater), so it isn't a total order, which sort_by requires. The ordering of near-equal clients ends up unspecified, and newer std sort can panic on a non-total-order comparator. Could we sort by a computed score with total_cmp and return Equal on ties instead?

lez/wallet/src/multi_client.rs
```diff
@@ -169,54 +179,152 @@ impl MultiSequencerClient {
...
+            .entry(leader_url.clone())
+            .or_default()
+            .push(statistics_update);
+
+        resp
+    }
+
+    /// Metered call for main leader(helm), to get data, necessary for send call.
+    ///
+    /// If current leader errors, we ask next one in list.
```
moudyellaz: The doc says a failing leader falls over to the next in the list, but leaders() is capped at distribution_limit, so with the default of 1 there's no failover at all, and above 1 it only covers the distribution subset rather than all healthy clients. Could we either widen failover to all calibrated clients or tweak the doc to match?

lez/wallet/src/multi_client.rs
```diff
@@ -169,54 +179,152 @@ impl MultiSequencerClient {
...
+                .entry(helm_url.clone())
+                .or_default()
+                .push(StatisticsUpdate::Failure);
+
+            for (leader, leader_url) in self.leaders().iter().skip(1) {
+                resp = self
+                    .metered_get_helper(&call, leader, leader_url, &mut statistic_map)
+                    .await;
+
+                if resp.is_err() {
```
moudyellaz: On a failed leader this records two samples for one call: the concurrent actualize probe (which can come back Success, it's an independent get_last_block_id) plus an explicit Failure. Might be cleaner to record a single outcome per call so the probe doesn't mask the failure.

moudyellaz (approved):
lgtm, thanks. Left some minor comments and couple of questions.

lez/wallet/src/multi_client.rs
Pravdyvy: True, updated with merge.

lez/wallet/src/multi_client.rs
Pravdyvy: Rewrote with `total_cmp()`

lez/wallet/src/multi_client.rs
Pravdyvy: Explicitly stated, that we failover up to a `self.config.distribution_limit`.

lez/wallet/src/multi_client.rs
Pravdyvy: Removed failure records.

lez/wallet/src/multi_client.rs
```diff
@@ -119,48 +119,58 @@ impl MultiSequencerClient {
...
+        let leader_list = choose_leaders(
+            &client_list,
+            statistics,
+            multi_sequencer_client_config.distribution_limit,
+        )
+        .ok_or_else(|| anyhow::anyhow!("Failed to find leader"))?;
 
-        let res = choose_leader(&client_list, statistics)
-            .ok_or_else(|| anyhow::anyhow!("Failed to find leader"))?;
+        assert_ne!(leader_list.len(), 0);
```
schouhy: shouldn this return `Err`?

lez/wallet/src/multi_client.rs
```diff
@@ -431,35 +547,46 @@ pub fn choose_leader(
...
-        // [-old_std...................old_lat........new_lat.........+new_std.......+old_std]
+        // [-right_std < right_lat < left_lat < +left_std < +right_std]
         //
         // is still better, but it is up to discussion
-        if (new_lat <= old_lat) && ((new_lat + new_std) < (old_lat + old_std)) {
-            *x
-        } else {
-            acc
+        let first_ordering = left_lat.total_cmp(&right_lat);
+        match first_ordering {
```
schouhy, line 575: I think the binary relation defined by this closure is not a total order. It is not antisymmetric. Suppose 
A_lat  = 10, A_std = 2
B_lat = 11, B_std = 0

if the code compares (A,B), then first_ordering is Less, which then compares with the variances and returns (10 + 2).total_cmp(11 + 0) which is Greater. If the code compares (B,A) then first_ordering is Greater and the code returns Greater too.

lez/wallet/src/multi_client.rs
schouhy, line 1: Not introduced in this PR, but the `errors` and `sample_size` include the full history. I think it would be better to have these statistics be computed over a recent window only.

lez/wallet/src/multi_client.rs
```diff
@@ -169,54 +179,160 @@ impl MultiSequencerClient {
...
+        call: I,
+    ) -> Vec<Result<R, E>> {
+        let leaders = self.leaders().iter().take(self.config().distribution_limit);
+
+        // Collecting all statistics into one map to lock updates only once
+        let mut statistic_map: HashMap<Url, Vec<StatisticsUpdate>> = HashMap::new();
+
+        let mut results = vec![];
+
+        for (leader, leader_url) in leaders {
```
schouhy, line 296: can we run these concurrently? Otherwise might not scale well with many sequencers?

lez/wallet/src/multi_client.rs
```diff
@@ -169,54 +179,160 @@ impl MultiSequencerClient {
...
+    /// Update statistics of a leader, clear statistic updates log.
+    pub async fn update_statistics(&self, statistics: &mut HashMap<Url, Statistics>) {
         let mut statistic_updates = self.statistic_updates.write().await;
-        leader_statistic.apply_updates(statistic_updates.as_ref());
+
+        #[expect(clippy::iter_over_hash_type, reason = "Ordering is unnecesary here")]
+        for (addr, statistic_updates_vec) in statistic_updates.iter() {
+            let leader_statistic = statistics
+                .get_mut(addr)
+                .expect("Leader statistic must be present after setup");
```
schouhy: shouldn't this return a result?

schouhy (approved):
hey thanks! LGTM. Left some comments

lez/wallet/src/multi_client.rs
Pravdyvy: Rewrote to return Err

lez/wallet/src/multi_client.rs
Pravdyvy: Rewrote using `f32::total_cmp()`

lez/wallet/src/multi_client.rs
Pravdyvy: Part of the PR #638

lez/wallet/src/multi_client.rs
Pravdyvy: Returning a result

lez/wallet/src/multi_client.rs
Pravdyvy: I struggle to understand, what is a recent window. We can, of course, make it configurable, but it will just put this responsibility on users.
I would say that window could probably be something akin to full circle of round-robin algorithm. What do you think?

schouhy: yep, sounds good. Not necessarily something to add to this PR though

schouhy (approved):

moudyellaz (approved):
lgtm, thanks for addressing my comments. On the failing CI test, I have fixed this in my open PR615. So feel free to merge this PR.

PR #625, opened 2026-07-15
fix!(state_machine): reject public transactions that silently drop a d…

jonesmarvin8:
## 🎯 Purpose

Address [Issue 626](https://github.com/logos-blockchain/logos-execution-zone/issues/626). Strengthen public `validate_state_diff` logic to enforce that accounts that appear in the transaction message appears in the resulting `state_diff`.

This prevents ill-formed tests for public execution to pass when the private variant fails.

**Breaking change**: may break ill-formed public tests and programs.

## ⚙️ Approach

- [X] Add logic to `validate_state_diff` to enforce `account_ids` from message included in `state_diff`.
- [X] Add `dropped_account` test program; differs from `missing_output` by having `pre_state = post_state`.
- [X] Add test verifying that `dropped_account` execution fails for public state.

## 🧪 How to Test

New test `program_should_fail_if_it_drops_a_declared_account`.

## 🔗 Dependencies

N/A

## 🔜 Future Work

N/A

## 📋 PR Completion Checklist

- [X] Complete PR description
- [X] Add/update tests


lee/state_machine/src/validated_state_diff/mod.rs
```diff
@@ -293,6 +293,17 @@ impl ValidatedStateDiff {
...
+        // Every account the caller declared as part of the transaction must appear in the final
+        // diff.
+        for account_id in &message.account_ids {
+            ensure!(
+                state_diff.contains_key(account_id),
+                InvalidProgramBehaviorError::DeclaredAccountMissingFromOutput {
+                    account_id: *account_id
+                }
+            );
+        }
```
Arjentix, line 305: So what about if-else statements in the program code where sometimes program decides not to modify an account. Do you think such account should not be in `message.account_ids`?

Arjentix: Maybe they should appear in pre and post states, but there might be no diff

jonesmarvin8: From my understanding (and doubled check). `state_diff`includes unmodified account states as well. So, all accounts appear that appear in pre and post state also appeared as message.id.

This is a bit stronger than the intended goal.
- No accounts included in a program execution is dropped
- Account ids included in message account ids must be included in pre and post states; I don't think there's a practical application that this wouldn't occur for.

Arjentix: > state_diffincludes unmodified account states as well.

Oh, ok, that's not intuitive, but ok

Arjentix (approved):
LGTM, thanks

Pravdyvy (approved):
lgtm

erhant (approved):
lgtm ty!

PR #577, opened 2026-06-25
Wallet ffi labels

Pravdyvy:
## 🎯 Purpose

Solve issue https://github.com/logos-blockchain/logos-execution-zone-module/issues/35

Handle redirection onto `dev` branch.

## ⚙️ Approach

- [x] Labels `wallet_ffi` functionality.
- [x] Labels integration tests.
- [x] Updated CI to store cache on `dev` branch and to rerun CI on merges into `dev`.
- [x] Updated `CONTRIBUTING.md` with `dev` mentions.

## 🧪 How to Test

Added 2 integration tests for labels

## 🔗 Dependencies

None

## 🔜 Future Work

Expand LEZ modle with this functionality

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Implement the core functionality
- [x] Add/update tests
- [x] Add/update documentation and inline comments

Issue #537, opened 2026-06-17
[CI] Redundant logos-blockchain-circuits dependency in bench CI step


Pravdyvy:
With the latest version of `logos-blockchain` `logos-blockchain-circuits` is now unnecessary for builds.
Benchmarks, however check the difference between main and branch, so it is was impossible to exclude circuit setup for benchmarks before. 
After merging #510 now `logos-blockchain` is updated and we can remove circuits setup from CI altogether.

PR #628, opened 2026-07-15
ci: reuse bounded integration artifacts

### LINKED ISSUE #597: Add rustdoc checks in CI
I just thought about this improvement when I saw a comment: https://github.com/logos-blockchain/logos-execution-zone/pull/581#discussion_r3529819136

Should be a good addition to our CI, however may take time to fix all broken links.

### LINKED ISSUE #627: [CI] bound repeated integration artifacts and setup work
# CI: reuse bounded per-target artifacts and deterministic tooling

## Summary

The General workflow builds one complete integration-test archive, then sends
that archive to one fresh runner per test binary. A successful 30-target run at
[`dev` commit `ea41d5a7`][tested-commit] produced this baseline:

| Measurement | Evidence |
| --- | ---: |
| Workflow jobs | 40 |
| Total runner time | 445.80 minutes |
| Integration jobs | 30 |
| Integration runner time | 309.10 minutes |
| Complete integration archive | 3.61 GiB |
| Same archive delivered to 30 workers | 108.32 GiB |
| Source-built Cargo tool installs | 37 / 122.05 minutes |
| RISC Zero installer invocations | 36 / 18.55 minutes |
| Valid-proof command | 60 minutes 12 seconds |

The dominant costs are repeated transfer, extraction, compilation, and runner
setup rather than unique test work. The archive is large enough that combining
it with extracted binaries, toolchains, and Docker layers regularly exhausts
standard hosted-runner disk.

### 1. Every integration worker receives every integration binary

The [baseline workflow] creates one archive for the complete
`integration_tests` package. It then starts one matrix job per discovered
binary. Every job downloads and extracts that complete archive, although it
runs only one binary.

The archive in the [July 15 General run][baseline-run] was 3.61 GiB. Thirty
downloads therefore move 108.32 GiB before counting the producer upload, RISC
Zero downloads, Cargo registry traffic, or Docker images. Every fresh runner
also holds the compressed archive while nextest extracts the same shared files.

GitHub-hosted jobs do not share a filesystem.[^job-isolation] Standard public
Linux runners expose 14 GB of SSD storage.[^runner-storage] Repeated archive,
toolchain, and Docker storage has already produced `No space left on device`
failures in [run 29319061408], [run 29023798490], and [run 28527704562].

The archive grew from 2.94 GiB in [July 1 artifact metadata] to 3.61 GiB on
July 15: approximately 690.25 MiB, or 22.96%, in two weeks. Deleting a few
runner files delays the next failure but retains the 30-fold transfer and
extraction multiplier.

### 2. Cargo builds a target the matrix intentionally excludes

The workflow excludes `tps` while discovering matrix entries, after the
complete package archive has already been built. Cargo therefore compiles and
archives `tps`, but no consumer executes it. Target selection must happen in
the producer before compilation and archive construction.

### 3. Thirty target jobs repeat unrelated preparation

The integration matrix consumed 309.10 runner-minutes in the baseline.
Repeated step timestamps account for:

| Repeated integration step | Runner-minutes |
| --- | ---: |
| Install system dependencies | 6.60 |
| Install RISC Zero | 15.22 |
| Install active Rust toolchain | 6.42 |
| Download complete archive | 35.03 |
| Compile and install nextest | 108.50 |
| Run tests, including full extraction | 134.45 |

Preparation before `Run tests` consumed 171.77 runner-minutes, 55.6% of the
matrix total. One job per binary provides useful independent checks and runner
parallelism; it does not require each job to repeat all provisioning or receive
unrelated binaries.

### 4. CI compiles command-line tools from source on fresh runners

The baseline compiles nextest in the unit, fixture, producer, and all 30
integration-consumer jobs. Thirty-three nextest installations consumed 115.87
runner-minutes. Taplo, cargo-machete, cargo-deny, and just add four more source
builds. All 37 Cargo tool-install steps consumed 122.05 runner-minutes, 27.4%
of total workflow runner time.

Nextest documents prebuilt CI installation and identifies
`taiki-e/install-action` as its easiest prebuilt path.[^nextest-install]
Compiling these release tools on every ephemeral runner adds dependency
downloads, target output, disk use, and mutable resolution without producing
repository code.

### 5. Artifact transport repeats compression and has no bound

The producer uploads an already compressed `.tar.zst` without an explicit
compression level, short retention period, or size ceiling. Artifact upload
therefore spends time trying to compress effectively incompressible input and
retains a large per-run transport artifact according to repository defaults.
GitHub recommends compression level `0` for large files that do not compress
well.[^artifact-compression]

Uploading the baseline archive took 1 minute 52 seconds. Consumers retain the
compressed copy until after extraction, and the workflow has no free-disk gate
before Docker-backed tests begin. Default test-profile debug information also
inflates every transported test binary.

### 6. RISC Zero installation is broad, mutable, and repeated

The original composite downloads rzup and runs bare `rzup install`. Rzup
documents that omitting a version selects the latest component version.[^rzup]
The baseline invokes this setup 36 times, consuming 18.55 runner-minutes and
recreating the same files on each runner.

The jobs need different subsets:

- integration and valid-proof consumers need `r0vm`;
- unit, producer, and generated-artifact jobs need the RISC Zero Rust
  toolchain and selected build components;
- lint and the crypto-primitives benchmark do not need RISC Zero.

The required `r0vm` 3.0.5 binary is 103.95 MiB. A measured default C++ guest
toolchain occupied approximately 1.5 GiB even though repository guests are
Rust-only. Exact build/runtime profiles and one shared runtime artifact avoid
installing unrelated components in every consumer.

### 7. Valid-proof recompiles an archived target

The original valid-proof job runs:

```console
cargo test -p integration_tests -- --exact private::private_transfer_to_owned_account
```

The filter after `--` is a test-harness filter; it does not select the
`auth_transfer` Cargo target before compilation.[^cargo-test-selection] The
command took 60 minutes 12 seconds even though the producer had already built
`auth_transfer` into the nextest archive.

The exact target archive can be reused with the conjunction
`binary(=auth_transfer) &
test(=private::private_transfer_to_owned_account)`, eliminating the duplicate
Cargo build while leaving `RISC0_DEV_MODE` unset.

### 8. The real-proof log gap hides CPU proving, not RPC waiting

In [PR #619's valid-proof job][proof-job], the log gap between
`15:38:55.702` and `16:30:40.948` lasted 51 minutes 45 seconds. The code path is:

1. `send_privacy_preserving_tx_with_pre_check` calls `execute_and_prove`.
2. `execute_and_prove` calls
   `default_prover().prove_with_opts(..., ProverOpts::succinct())`.
3. Only after proof generation returns does the wallet make the JSON-RPC
   `send_transaction` call.

The preceding `execution time: 64.930608ms` message is RISC Zero guest trace
execution, not the outer succinct proof. The silent interval is local CPU work
inside `r0vm` segment/recursion proving. Archive reuse removes duplicate
compilation but does not remove or weaken this cryptographic work.

Default nextest presentation also hides captured output for successful tests
behind a progress display. Regular target jobs should disable the progress bar
and emit each test's full stdout/stderr when it completes while retaining
parallel scheduling. The single valid-proof test can use uncaptured output and
targeted RISC Zero logging so its available logs stream live. RISC Zero may
still emit no continuous progress during an individual proving stage; the
workflow must not represent that silence as an RPC delay.

### 9. Parallel cache writers cannot share output within one run

Six baseline jobs can save the same Cargo cache key on `main` or `dev`. Jobs
cannot consume another parallel job's newly saved cache during that workflow.
Multiple writers add post-job upload contention and cache churn instead of
forming an inter-job transport channel.

GitHub distinguishes cross-run dependency caches from artifacts that pass
produced files between jobs.[^cache-purpose] Test binaries and `r0vm` are
same-run artifacts; Cargo dependencies and full RISC Zero build components are
cross-run caches. Each cache key needs one trusted writer.

The Docker publication matrix has a separate collision: all four service
images use the default BuildKit GitHub Actions cache scope. Docker documents
that a later write to the same scope overwrites the earlier cache.[^docker-cache-scope]
Each image needs a distinct scope.

### 10. Superseded workflows keep consuming runners

General and benchmark workflows originally lacked concurrency groups. A new
push to the same pull request left the obsolete multi-job run active. GitHub
supports canceling an in-progress member of the same concurrency group.[^concurrency]
Without it, rapid updates multiply the already expensive workflow.

### 11. Mutable tool and action inputs weaken reproducibility

The original workflows combine action tags, an undated nightly, unversioned
Cargo installs, and mutable RISC Zero component selection. GitHub states that
a full commit SHA is the only immutable action reference.[^secure-actions]
`cargo install --locked` locks dependencies for whichever package release is
selected; it does not select an exact package version.

Exact CLI releases, dated toolchains, exact RISC Zero components, a
checksum-pinned bootstrap, full-SHA action references, and deterministic cache
keys make setup failures attributable to committed inputs.

### 12. Existing checks spend setup time while leaving gaps

The TOML formatter runs `taplo fmt --check .`. Taplo 0.10.0 reports that input
as excluded and produces `files: []`; removing the literal `.` discovers all
80 configured TOML files. The baseline spent 2 minutes 12 seconds compiling
Taplo before checking no files. [PR #618] independently identifies and fixes
this command.

Lint installs RISC Zero even though clippy skips guest builds, but it does not
run rustdoc. [Issue #597] tracks that gap. Warnings-denied rustdoc exposes two
existing unresolved links: `ML_KEM_768_CIPHERTEXT_LEN` and `Program`.

The [baseline benchmark workflow] listens to `lez/key_protocol/**` and
`lee/core/**`, while the repository paths are `lee/key_protocol/**` and
`lee/state_machine/core/**`. Relevant edits therefore do not trigger it. The
benchmark also installs RISC Zero although `crypto_primitives_bench` compiles
without it.

[Issue #537] overlaps only historically. External-circuits setup became
obsolete after [PR #510] and was removed by [commit 808cea7]; [PR #507] was the
earlier temporary pin. The remaining work is unrelated RISC Zero benchmark
setup and stale path filters, not a revival of the completed external-circuits
task.

### 13. A shared CI image does not remove full-archive fan-out

[PR #618] now passes its latest [General run][pr-618-general] and
[benchmark run][pr-618-benchmark]. It usefully prebuilds system dependencies
and tools, removes repeated source installs, and fixes Taplo discovery.

Its integration topology still stores one 3.61 GiB complete archive and sends
it to all 30 target jobs. Each job then pulls the CI image and extracts the
full archive to run one binary. A representative target spent approximately
47 seconds downloading the archive, 57 seconds pulling the image, and 31
seconds extracting it. The approach addresses tool/system provisioning, but
retains the dominant archive transfer, extraction, disk multiplier, and
nested-container boundary.

The reusable parts are complementary: deterministic prebuilt tooling should
be retained, while the artifact boundary must become target-scoped. A future
trusted-branch-built immutable image can be evaluated separately for jobs that
are safe inside that boundary.

### Target-scoped integration transport

1. Discover all non-`tps` integration targets from `cargo metadata`.
2. Build each target exactly once in one producer with exact
   `--test <target>` and `binary(=<target>)` selectors. Reuse the producer's
   Cargo target directory across all builds.
3. Upload one already-compressed archive per target through the official
   artifact client in raw-file mode. Each archive has a 128 MiB ceiling, the
   complete archive set has a 4 GiB aggregate ceiling, and the producer
   requires at least 6 GiB free after each upload. Retain each archive for one
   day and delete each producer copy immediately after upload.
4. Compare the uploaded-target list with discovered Cargo metadata. Missing,
   extra, or duplicate targets fail the producer.
5. Generate a 30-target matrix with immutable artifact IDs from the validated
   list. Preserve one independently reported job per target, allow all 30 to
   run in parallel, and require no CI mapping edit for a new Cargo target.
6. Download and extract only the named target archive in each job, delete the
   compressed input, and require at least 6 GiB free before Docker-backed
   execution.
7. Disable the nextest progress bar, report every status, and emit full
   stdout/stderr when each regular test completes. Keep parallel scheduling and
   default fail-fast behavior within the target; other target jobs continue.
8. Reuse `auth_transfer` for the exact real-proof test. Stream this single
   test's output uncaptured, enable available RISC Zero phase diagnostics, and
   keep development mode disabled.

### Deterministic shared setup

- Install exact prebuilt nextest, Taplo, cargo-machete, cargo-deny, and just
  releases with source fallback disabled.
- Install exact RISC Zero build/runtime profiles through checksum-pinned rzup.
- Upload the producer's version-checked `r0vm` as a small one-day artifact and
  move it into place in each consumer.
- Keep one trusted writer per Cargo/RISC Zero cache. Pull requests may restore
  non-secret caches but cannot save them.
- Disable Cargo incremental compilation and test binary debug information in
  CI to reduce transient output and align cache keys.
- Pin third-party actions to full SHAs, use Node 24-compatible first-party
  artifact/cache actions, pin rustfmt to a dated nightly, and use exact tool
  versions.
- Give each Docker publication image a separate BuildKit cache scope and
  cancel superseded workflow runs.

### Coverage and workflow correctness

- Keep fixture gating, unit, integration, valid-proof, generated-artifact,
  lint, dependency, formatting, and benchmark checks.
- Remove Taplo's ineffective `.` argument.
- Add warnings-denied rustdoc for the workspace and two guest packages; repair
  only the two broken intra-doc links it exposes.
- Regenerate the affected checked-in guest artifact and verify its ImageID is
  unchanged.
- Correct both benchmark path filters and remove unrelated RISC Zero setup.
- Keep General pull-request permissions read-only and require no image
  publication from pull-request code.

### Final per-target run

[General run 29457865053][per-target-run] discovered, built, uploaded, and
executed all 30 non-`tps` integration targets as independently reported jobs.
[Benchmark run 29457864998][per-target-benchmark] completed successfully in
8 minutes 51 seconds.

| Measurement | Baseline | Per-target run |
| --- | ---: | ---: |
| Integration targets | 30 | 30 |
| Individual integration archive | 3.61 GiB, all targets | 48.65--70.80 MiB, one target |
| Retained integration archives | 3.61 GiB | 2.03 GiB |
| Target/proof archive fan-out | 108.32 GiB | 2.10 GiB |
| Integration consumer time | 309 minutes 6 seconds | 122 minutes 26 seconds |
| Test result | -- | 114 passed, 1 skipped |
| General runner time | 445 minutes 48 seconds | 239 minutes 12 seconds |
| General wall time | 74 minutes 44 seconds | 67 minutes 50 seconds |

The producer created and uploaded all 30 target archives in 13 minutes 21
seconds. Integration consumers used 60.4% less runner time than the baseline,
and total General runner time fell by 46.3%. Every archive remained below its
128 MiB ceiling; the largest used 55.3% of that limit. The complete 2.03 GiB
set remained below the 4 GiB aggregate ceiling.

Thirty target downloads plus the repeated `auth_transfer` proof archive move
2.10 GiB, 98.1% below the baseline archive fan-out. Including a conservative
31 downloads of the 66.18 MiB runtime artifact gives 4.10 GiB of combined
integration/proof payload, 96.2% below the baseline archive traffic alone.
The retained target archives are 43.8% smaller than the baseline's single
complete archive.

All consumer extraction footprints were approximately 375--477 MiB. A
representative `config` worker downloaded only its approximately 71 MiB
archive, extracted approximately 474 MiB in 0.87 seconds, retained
approximately 88 GiB free, and printed completed-test status plus stdout and
stderr without a progress bar.

The valid-proof job reused the `auth_transfer` archive without a Cargo build,
kept `RISC0_DEV_MODE` unset, and completed its exact test command in 53 minutes
58 seconds. After 64 milliseconds of guest execution, RISC Zero recursion
stage logs and nextest minute heartbeats continued until proving completed;
the subsequent RPC connection followed immediately. The full General run
completed successfully with all 40 jobs green.

## Relationship to existing CI work

- [PR #498] introduced build-once nextest archives and binary-scoped jobs.
  Per-target archives preserve that matrix while removing full-archive fan-out.
- [PR #590] added the fixture gate. All target jobs remain conditional on it.
- [PR #577] restricted cache writes to trusted branches. Cargo and RISC Zero
  caches retain that policy with one writer per key.
- [PR #417] and [issue #278] improved fork support. General CI remains
  read-only and performs no pull-request image publication.
- [PR #311] locked nextest installation. Exact prebuilt installation preserves
  the locked version without compiling it on every runner.
- [Issue #597] is covered by warnings-denied rustdoc and the two link fixes.
- [Issue #537], [PR #507], [PR #510], and [commit 808cea7] concern completed
  external-circuits cleanup. Only the remaining benchmark setup/path issues
  overlap.
- [Issue #321] remains separate. `cargo deny check` stays enabled and advisory
  policy is unchanged.
- [PR #618] validates prebuilt environment benefits and the Taplo correction.
  Per-target artifacts address the full-archive/image fan-out it retains.
- Existing machete, toolchain, script, quality, image, license, performance,
  benchmark, and indexer work from [issue #132], [issue #178], [issue #181],
  [issue #182], [issue #222], [PR #195], [PR #248], [PR #317], [PR #431],
  [PR #439], and [PR #494] remains intact.

## Acceptance criteria

- Every non-`tps` Cargo integration target has exactly one uploaded archive;
  missing, extra, or duplicate targets fail CI.
- General CI exposes one independently reported job per target and allows all
  target jobs to run concurrently.
- No target worker downloads or extracts another target's archive.
- Every target archive is below 128 MiB, retained for one day, uploaded
  without redundant compression, and deleted locally after upload/extraction.
- The complete target archive set remains below 4 GiB, and the producer
  retains at least 6 GiB free throughout archive creation.
- Every target worker retains at least 6 GiB free after extraction and before
  Docker-backed execution.
- Regular target jobs show all test statuses and full stdout/stderr after each
  test completes, with no progress bar. Tests remain parallel within a target
  and use default fail-fast behavior; independent target jobs continue.
- Valid-proof reuses `auth_transfer`, performs no Cargo build, streams its
  single test's output live, and keeps `RISC0_DEV_MODE` unset.
- Direct General workflow provisioning contains no `cargo install`; exact
  prebuilt tool downloads use checksum verification and no source fallback.
- RISC Zero components, rustfmt, tools, and third-party actions are exact and
  immutable inputs. First-party artifact/cache actions use Node 24-compatible
  releases.
- Cache writes occur only on trusted `main` or `dev` pushes, with one writer
  per cache key/profile. Docker publication caches use distinct scopes.
- General CI remains read-only and pull-request CI requires no package/image
  publication permission.
- Taplo checks the configured TOML set; rustdoc, formatting, lint, dependency,
  unit, fixture, integration, real-proof, generated-artifact, and benchmark
  checks remain present.
- The final hosted run records all 30 archive sizes, disk headroom, output
  behavior, runner time, and wall time before merge.

[^artifact-compression]: [`actions/upload-artifact` compression guidance][artifact-compression]
[^cache-purpose]: [GitHub dependency-cache documentation][cache-purpose]
[^cargo-test-selection]: [Cargo test target-selection documentation][cargo-test-selection]
[^concurrency]: [GitHub workflow concurrency documentation][concurrency]
[^docker-cache-scope]: [Docker GitHub Actions cache-scope documentation][docker-cache-scope]
[^job-isolation]: [Understanding GitHub Actions][job-isolation]
[^nextest-install]: [Nextest prebuilt installation documentation][nextest-install]
[^runner-storage]: [GitHub-hosted runner specifications][runner-storage]
[^rzup]: [RISC Zero rzup documentation][rzup]
[^secure-actions]: [GitHub secure-use documentation][secure-actions]

[artifact-compression]: https://github.com/actions/upload-artifact#altering-compressions-level-speed-v-size
[baseline benchmark workflow]: https://github.com/logos-blockchain/logos-execution-zone/blob/ea41d5a7385689e7e9683d2f882572387d617847/.github/workflows/bench-regression.yml
[baseline workflow]: https://github.com/logos-blockchain/logos-execution-zone/blob/ea41d5a7385689e7e9683d2f882572387d617847/.github/workflows/ci.yml
[baseline-run]: https://github.com/logos-blockchain/logos-execution-zone/actions/runs/29418633146
[cache-purpose]: https://docs.github.com/en/actions/concepts/workflows-and-actions/dependency-caching
[cargo-test-selection]: https://doc.rust-lang.org/cargo/commands/cargo-test.html#target-selection
[commit 808cea7]: https://github.com/logos-blockchain/logos-execution-zone/commit/808cea737edffa04031fe8e053ea5bbe7171fdd7
[concurrency]: https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency
[docker-cache-scope]: https://docs.docker.com/build/cache/backends/gha/#scope
[per-target-benchmark]: https://github.com/logos-blockchain/logos-execution-zone/actions/runs/29457864998
[per-target-run]: https://github.com/logos-blockchain/logos-execution-zone/actions/runs/29457865053
[issue #132]: https://github.com/logos-blockchain/logos-execution-zone/issues/132
[issue #178]: https://github.com/logos-blockchain/logos-execution-zone/issues/178
[issue #181]: https://github.com/logos-blockchain/logos-execution-zone/issues/181
[issue #182]: https://github.com/logos-blockchain/logos-execution-zone/issues/182
[issue #222]: https://github.com/logos-blockchain/logos-execution-zone/issues/222
[issue #278]: https://github.com/logos-blockchain/logos-execution-zone/issues/278
[issue #321]: https://github.com/logos-blockchain/logos-execution-zone/issues/321
[issue #537]: https://github.com/logos-blockchain/logos-execution-zone/issues/537
[issue #597]: https://github.com/logos-blockchain/logos-execution-zone/issues/597
[job-isolation]: https://docs.github.com/en/actions/get-started/understand-github-actions
[july 1 artifact metadata]: https://api.github.com/repos/logos-blockchain/logos-execution-zone/actions/runs/28531560278/artifacts
[nextest-install]: https://nexte.st/docs/installation/pre-built-binaries/
[pr #195]: https://github.com/logos-blockchain/logos-execution-zone/pull/195
[pr #248]: https://github.com/logos-blockchain/logos-execution-zone/pull/248
[pr #311]: https://github.com/logos-blockchain/logos-execution-zone/pull/311
[pr #317]: https://github.com/logos-blockchain/logos-execution-zone/pull/317
[pr #417]: https://github.com/logos-blockchain/logos-execution-zone/pull/417
[pr #431]: https://github.com/logos-blockchain/logos-execution-zone/pull/431
[pr #439]: https://github.com/logos-blockchain/logos-execution-zone/pull/439
[pr #494]: https://github.com/logos-blockchain/logos-execution-zone/pull/494
[pr #498]: https://github.com/logos-blockchain/logos-execution-zone/pull/498
[pr #507]: https://github.com/logos-blockchain/logos-execution-zone/pull/507
[pr #510]: https://github.com/logos-blockchain/logos-execution-zone/pull/510
[pr #577]: https://github.com/logos-blockchain/logos-execution-zone/pull/577
[pr #590]: https://github.com/logos-blockchain/logos-execution-zone/pull/590
[pr #618]: https://github.com/logos-blockchain/logos-execution-zone/pull/618
[pr-618-benchmark]: https://github.com/logos-blockchain/logos-execution-zone/actions/runs/29446434412
[pr-618-general]: https://github.com/logos-blockchain/logos-execution-zone/actions/runs/29446434518
[proof-job]: https://github.com/logos-blockchain/logos-execution-zone/actions/runs/29428039747/job/87419099459?pr=619
[run 28527704562]: https://github.com/logos-blockchain/logos-execution-zone/actions/runs/28527704562
[run 29023798490]: https://github.com/logos-blockchain/logos-execution-zone/actions/runs/29023798490
[run 29319061408]: https://github.com/logos-blockchain/logos-execution-zone/actions/runs/29319061408/job/87041311986
[runner-storage]: https://docs.github.com/en/actions/reference/runners/github-hosted-runners
[rzup]: https://github.com/risc0/risc0/blob/release-3.0/rzup/README.md
[secure-actions]: https://docs.github.com/en/actions/reference/security/secure-use
[tested-commit]: https://github.com/logos-blockchain/logos-execution-zone/commit/ea41d5a7385689e7e9683d2f882572387d617847

3esmit:
## 🎯 Purpose

Reduce CI disk use and repeated setup while preserving one independently
reported job per integration target. Integration binaries are built in one
producer, and each of the 30 jobs receives only its own bounded nextest archive
plus a shared, versioned RISC Zero runtime. Existing fixture, Docker-backed
integration, real-proof, generated-artifact, and benchmark checks remain.

Problem statement and baseline evidence: #627.

Closes #627.
Closes #597.

<details><summary>Details</summary>
<p>

### Preserve target-level parallelism with target-scoped artifacts

- Discover every non-`tps` integration target from `cargo metadata`.
- Build each target once in a shared producer target directory. Exact
  `--test <target>` and `binary(=<target>)` selectors avoid compiling or
  archiving unrelated integration binaries.
- Upload one already-compressed nextest archive per target through the official
  artifact client in raw-file mode. Each archive has a 128 MiB ceiling, the
  complete set has a 4 GiB aggregate ceiling, and the producer requires at
  least 6 GiB free after each upload. Archives are retained for one day and
  deleted from producer disk immediately after upload.
- Compare the complete uploaded-target list with the discovered Cargo target
  list. Missing, extra, or duplicate archives fail the producer.
- Generate the 30-job matrix, including immutable artifact IDs, from that
  validated list. New Cargo test targets require no workflow mapping edit.
  Each worker downloads, extracts, and runs only its own target archive, then
  deletes the compressed input before Docker-backed execution.
- Require at least 6 GiB of free runner disk after extraction.
- Reuse the `auth_transfer` archive for the real-proof check instead of
  compiling that integration package again.

This retains the binary-scoped check names and runner concurrency introduced
by #498 without sending the complete integration package to every runner.
`test-fixtures-tests`, added by #590, still gates every integration job.

### Restore useful test output

Regular integration jobs disable nextest's progress bar and report every test
status. Captured stdout and stderr from successful tests are emitted as soon as
that test completes; failed-test output remains immediate. Tests within one
target stay parallel and use nextest's default fail-fast behavior. Independent
target jobs continue to report concurrently.

Nextest cannot stream uncaptured, non-interleaved output while running tests in
parallel. The real-proof job selects exactly one test, so it uses
`--no-capture` to stream that test live without sacrificing useful
parallelism. `RISC0_INFO=1` and targeted host logging stream every diagnostic
RISC Zero emits. An individual proving stage can still be silent; the explicit
proof-start message distinguishes that wait from compilation or RPC activity.

### Reuse deterministic tools and runtimes

- Install exact prebuilt releases of nextest 0.9.140, Taplo 0.10.0,
  cargo-machete 0.9.2, cargo-deny 0.20.2, and just 1.56.0 with source fallback
  disabled.
- Install exact RISC Zero components through checksum-pinned rzup: `r0vm`
  3.0.5, RISC Zero Rust 1.94.1, and `cargo-risczero` 3.0.5.
- Split RISC Zero into build and runtime profiles. The producer uploads its
  version-checked 103.95 MiB `r0vm`; consumers move that binary into place
  rather than installing RISC Zero or retaining a second staged copy.
- Keep one trusted writer per Cargo/RISC Zero cache. Pull requests may restore
  non-secret caches but cannot save them.
- Upgrade first-party cache, upload, and download actions to Node 24-compatible
  releases. All action references remain full-SHA pinned.
- Pin rustfmt to `nightly-2026-04-14`, use minimal stable toolchains, isolate
  Docker BuildKit cache scopes, and cancel superseded runs.

### Preserve and repair quality checks

- Remove Taplo's literal `.` input so all 80 configured TOML files are
  discovered.
- Add warnings-denied rustdoc for the workspace and both guest packages.
- Keep formatting, clippy, machete, cargo-deny, unit, fixture, integration,
  real-proof, generated-artifact, and benchmark checks.
- Correct benchmark triggers to `lee/key_protocol/**` and
  `lee/state_machine/core/**`, and remove unrelated RISC Zero setup from that
  benchmark.
- Keep pull-request General CI read-only and avoid image publication from
  untrusted code.

The two Rust source edits only repair pre-existing intra-doc links exposed by
the new warnings-denied rustdoc check:

- `lee/state_machine/core/src/encryption/shared_key_derivation.rs` resolves
  `ML_KEM_768_CIPHERTEXT_LEN` through `crate`.
- `lez/programs/src/lib.rs` resolves `Program` through
  `lee::program::Program`.

They do not change runtime behavior or public API. Because the guest source
metadata changed, the checked-in privacy circuit binary is regenerated. Its
RISC Zero ImageID remains
`36924f1232186568e6c06e7eacdcfbfc4b788c1af49b42848e872d635afd23d9`.

### Baseline

The successful [`dev` baseline][baseline-run] used 40 jobs and 445.80
runner-minutes:

| Measurement | Baseline |
| --- | ---: |
| Integration jobs | 30 |
| Integration runner time | 309.10 minutes |
| Complete integration archive | 3.61 GiB |
| Same archive delivered to 30 workers | 108.32 GiB |
| Source-built Cargo tool installs | 37 / 122.05 minutes |
| Nextest source builds | 33 / 115.87 minutes |
| RISC Zero installer invocations | 36 / 18.55 minutes |
| Valid-proof command | 60 minutes 12 seconds |

The archive grew from 2.94 GiB to 3.61 GiB in two weeks. Standard public Linux
runners expose 14 GB of SSD storage, and the old topology has produced
`No space left on device` failures in [run 29319061408], [run 29023798490],
and [run 28527704562]. Full measurements and source links are recorded in
#627.

### Final per-target hosted run

[General run 29457865053][per-target-run] discovered, built, uploaded, and
executed all 30 non-`tps` integration targets as independently reported jobs.
[Benchmark run 29457864998][per-target-benchmark] completed successfully in
8 minutes 51 seconds.

| Measurement | Baseline | Per-target run |
| --- | ---: | ---: |
| Integration targets | 30 | 30 |
| Individual integration archive | 3.61 GiB, all targets | 48.65--70.80 MiB, one target |
| Retained integration archives | 3.61 GiB | 2.03 GiB |
| Target/proof archive fan-out | 108.32 GiB | 2.10 GiB |
| Integration consumer time | 309 minutes 6 seconds | 122 minutes 26 seconds |
| Test result | -- | 114 passed, 1 skipped |
| General runner time | 445 minutes 48 seconds | 239 minutes 12 seconds |
| General wall time | 74 minutes 44 seconds | 67 minutes 50 seconds |

The producer created and uploaded all 30 target archives in 13 minutes 21
seconds. Integration consumers used 60.4% less runner time than the baseline,
and total General runner time fell by 46.3%. Every archive remained below its
128 MiB ceiling; the largest used 55.3% of that limit. The complete 2.03 GiB
set remained below the 4 GiB aggregate ceiling.

Thirty target downloads plus the repeated `auth_transfer` proof archive move
2.10 GiB, 98.1% below the baseline archive fan-out. Including a conservative
31 downloads of the 66.18 MiB runtime artifact gives 4.10 GiB of combined
integration/proof payload, 96.2% below the baseline archive traffic alone.
The retained target archives are 43.8% smaller than the baseline's single
complete archive.

All consumer extraction footprints were approximately 375--477 MiB. A
representative `config` worker downloaded only its approximately 71 MiB
archive, extracted approximately 474 MiB in 0.87 seconds, retained
approximately 88 GiB free, and printed completed-test status plus stdout and
stderr without a progress bar.

The valid-proof job reused the `auth_transfer` archive without a Cargo build,
kept `RISC0_DEV_MODE` unset, and completed its exact test command in 53 minutes
58 seconds. After 64 milliseconds of guest execution, RISC Zero recursion
stage logs and nextest minute heartbeats continued until proving completed;
the subsequent RPC connection followed immediately. The full General run
completed successfully with all 40 jobs green.

### Real-proof delay diagnosis

The approximately one-hour real-proof step is not an archive or RPC delay.
In [PR #619's valid-proof job][proof-job], the log gap from
`15:38:55.702` to `16:30:40.948` is 51 minutes 45 seconds. Code flow enters
`default_prover().prove_with_opts(..., ProverOpts::succinct())` before the
JSON-RPC `send_transaction` call. The preceding RISC Zero
`execution time: 64.930608ms` line measures guest trace execution, not outer
succinct proof generation. The silence is local CPU proving through `r0vm`.

Archive reuse removes the duplicate Cargo build before this work. It does not
weaken or skip cryptographic proving: `RISC0_DEV_MODE` remains unset. The new
live-output and RISC Zero logging settings make the phase visible, but do not
claim to shorten the proof itself.

## Relationship to other CI work

- [PR #498] introduced the build-once nextest archive and one job per binary.
  This PR preserves its check granularity while replacing full-archive fan-out
  with one target-scoped archive per worker.
- [PR #590] added the fixture gate. Every target worker still depends on it.
- [PR #577] restricted cache writes to trusted branches. This PR preserves
  that policy and reduces each cache key to one writer.
- [PR #417] and [issue #278] improved fork support. General CI remains
  read-only and requires no package publication.
- [PR #311] locked nextest installation. This PR keeps an exact version and
  replaces repeated source compilation with a checksummed prebuilt release.
- [Issue #597] is covered by warnings-denied rustdoc and the two link repairs.
- [Issue #537] overlaps only with historical external-circuits/benchmark
  setup. [PR #510] made that setup obsolete and [commit 808cea7] removed it;
  this PR only removes the remaining unrelated RISC Zero benchmark setup and
  repairs the benchmark triggers.
- [Issue #321] remains separate. `cargo deny check` stays enabled; advisory
  policy is unchanged.
- Existing CI work from [issue #132], [issue #178], [issue #181],
  [issue #182], [issue #222], [PR #195], [PR #248], [PR #317], [PR #431],
  [PR #439], and [PR #494] remains intact.

### Difference from PR #618

[PR #618] now passes its latest [General run][pr-618-general] and
[benchmark run][pr-618-benchmark]. It prebuilds system dependencies and tools
in a reusable CI image and also fixes Taplo discovery. Those changes remove
per-job tool compilation.

Its integration boundary remains one complete 3.61 GiB archive delivered to
30 target jobs. Each target job then pulls the CI image and extracts the full
archive to run one binary; a representative job spent approximately 47
seconds downloading the archive, 57 seconds pulling the image, and 31 seconds
extracting it. It therefore retains the full-archive disk and transfer
multiplier and a nested container boundary for Docker-dependent tests.

This PR keeps the same 30 target-level checks but changes the shared boundary:
each job receives only its 48.65--70.80 MiB target archive and the 66.18 MiB
runtime artifact. Small tools use exact prebuilt releases, larger RISC Zero
components use exact-key caches, and Docker-backed tests remain native host
jobs. Pull-request CI does not need to build or publish an image. The two
approaches are composable if an immutable, trusted-branch-built image is later
shown to improve container-safe jobs without restoring the artifact
multiplier.

## Validation

Completed locally:

- [x] `git diff --check`
- [x] Parse every changed workflow and composite-action YAML file.
- [x] actionlint 1.7.12 and ShellCheck 0.11.0.
- [x] Confirm every third-party `uses:` reference is a 40-character SHA.
- [x] Confirm no General workflow provisioning step runs `cargo install`.
- [x] Confirm Cargo metadata automatically drives target uploads and matrix
  entries without a maintained target list.
- [x] Build and list an exact target archive; verify the 128 MiB ceiling.
- [x] Extract `auth_transfer` and verify the exact real-proof selection.
- [x] `cargo +nightly-2026-04-14 fmt --check`.
- [x] Taplo formatting check over 80 TOML files.
- [x] Regenerate the privacy circuit binary and verify its unchanged ImageID.

Hosted validation:

- [x] Complete all General and benchmark jobs on GitHub-hosted runners.
- [x] Confirm the producer reports all 30 archives exactly once and every
  archive remains below 128 MiB.
- [x] Confirm every target job downloads only its named archive, retains at
  least 6 GiB free after extraction, and reports full completed-test output
  without a progress bar.
- [x] Record final wall time, runner time, archive metadata, and disk
  measurements against the baseline.
- [x] Confirm valid-proof streams output and still performs a non-development
  succinct proof.

## Dependencies

None. PR #618 is related work, not a prerequisite.

</p>
</details>

[baseline-run]: https://github.com/logos-blockchain/logos-execution-zone/actions/runs/29418633146
[commit 808cea7]: https://github.com/logos-blockchain/logos-execution-zone/commit/808cea737edffa04031fe8e053ea5bbe7171fdd7
[per-target-benchmark]: https://github.com/logos-blockchain/logos-execution-zone/actions/runs/29457864998
[per-target-run]: https://github.com/logos-blockchain/logos-execution-zone/actions/runs/29457865053
[issue #132]: https://github.com/logos-blockchain/logos-execution-zone/issues/132
[issue #178]: https://github.com/logos-blockchain/logos-execution-zone/issues/178
[issue #181]: https://github.com/logos-blockchain/logos-execution-zone/issues/181
[issue #182]: https://github.com/logos-blockchain/logos-execution-zone/issues/182
[issue #222]: https://github.com/logos-blockchain/logos-execution-zone/issues/222
[issue #278]: https://github.com/logos-blockchain/logos-execution-zone/issues/278
[issue #321]: https://github.com/logos-blockchain/logos-execution-zone/issues/321
[issue #537]: https://github.com/logos-blockchain/logos-execution-zone/issues/537
[issue #597]: https://github.com/logos-blockchain/logos-execution-zone/issues/597
[pr #195]: https://github.com/logos-blockchain/logos-execution-zone/pull/195
[pr #248]: https://github.com/logos-blockchain/logos-execution-zone/pull/248
[pr #311]: https://github.com/logos-blockchain/logos-execution-zone/pull/311
[pr #317]: https://github.com/logos-blockchain/logos-execution-zone/pull/317
[pr #417]: https://github.com/logos-blockchain/logos-execution-zone/pull/417
[pr #431]: https://github.com/logos-blockchain/logos-execution-zone/pull/431
[pr #439]: https://github.com/logos-blockchain/logos-execution-zone/pull/439
[pr #494]: https://github.com/logos-blockchain/logos-execution-zone/pull/494
[pr #498]: https://github.com/logos-blockchain/logos-execution-zone/pull/498
[pr #510]: https://github.com/logos-blockchain/logos-execution-zone/pull/510
[pr #577]: https://github.com/logos-blockchain/logos-execution-zone/pull/577
[pr #590]: https://github.com/logos-blockchain/logos-execution-zone/pull/590
[pr #618]: https://github.com/logos-blockchain/logos-execution-zone/pull/618
[pr-618-benchmark]: https://github.com/logos-blockchain/logos-execution-zone/actions/runs/29446434412
[pr-618-general]: https://github.com/logos-blockchain/logos-execution-zone/actions/runs/29446434518
[proof-job]: https://github.com/logos-blockchain/logos-execution-zone/actions/runs/29428039747/job/87419099459?pr=619
[run 28527704562]: https://github.com/logos-blockchain/logos-execution-zone/actions/runs/28527704562
[run 29023798490]: https://github.com/logos-blockchain/logos-execution-zone/actions/runs/29023798490
[run 29319061408]: https://github.com/logos-blockchain/logos-execution-zone/actions/runs/29319061408/job/87041311986


Arjentix:
Please, keep description simple and more like a high overview.

For details we can look at the diff and for issue description we can go to the issue.

.github/workflows/publish_images.yml
```diff
@@ -15,20 +15,20 @@ jobs:
   risc0_base:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
```
Arjentix, line 18: Why to pin like this?

.github/dependabot.yml
Arjentix, line 1: This change seems out of scope for this PR. At least I'd expect it in a separate commit

3esmit:
> Please, keep description simple and more like a high overview.
> 
> For details we can look at the diff and for issue description we can go to the issue.

Fixed. Removed the Issue details, and kept the other details hidden in the PR body.

.github/workflows/ci.yml
```diff
@@ -14,88 +16,90 @@ on:
...
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v5
-        with:
-          ref: ${{ github.event.pull_request.head.sha || github.head_ref }}
+      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
 
       - name: Install taplo-cli
-        run: cargo install --locked taplo-cli
+        uses: taiki-e/install-action@43aecc8d72668fbcfe75c31400bc4f890f1c5853 # v2.83.2
```
Arjentix, line 47: What's wrong with cargo install?

.github/workflows/ci.yml
```diff
@@ -107,202 +111,344 @@ jobs:
...
+            --exclude test_program_guests \
+            --exclude test_methods_guests
+
+          for package in test_program_guests test_methods_guests; do
+            cargo doc \
+              -p "$package" \
+              --all-features \
+              --no-deps \
+              --target-dir "$RUNNER_TEMP/rustdoc-$package"
+          done
```
Arjentix, line 134: Why exclude them?

.github/workflows/ci.yml
```diff
@@ -107,202 +111,344 @@ jobs:
...
+          tool: nextest@${{ env.NEXTEST_VERSION }}
+          fallback: none
 
       - name: Run tests
         env:
           RISC0_DEV_MODE: "1"
-          RUST_LOG: "info"
-        run: cargo nextest run --workspace --exclude integration_tests --exclude test_fixtures --all-features
+          RUST_LOG: info
+        run: cargo nextest run --workspace --exclude integration_tests --exclude test_fixtures --all-features --no-fail-fast
```
Arjentix: I don't think this is desirable. Typically we want to see CI results ASAP. If we spot a error, we can run tests locally to find other failures.

.github/workflows/ci.yml
```diff
@@ -14,88 +16,90 @@ on:
 
 permissions:
   contents: read
```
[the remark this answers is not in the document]
3esmit: No permission change is needed. The completed [PR run](https://github.com/logos-blockchain/logos-execution-zone/actions/runs/29449181322) restored caches and uploaded/downloaded all five same-run artifacts with `permissions: contents: read`; the producer and every artifact consumer passed. These actions use the Actions runtime credentials for the current workflow. An `actions: read` token is needed when `download-artifact` is explicitly used to fetch from another workflow run or repository. Keeping the workflow at least privilege.

.github/workflows/ci.yml
```diff
@@ -107,202 +111,344 @@ jobs:
...
+            | sort
+            | unique[]
+          ' "$shard_dir/metadata.json" > "$shard_dir/discovered.targets"
+
+          shard_targets=(
+            "indexer_ffi_state_consistency keys token indexer_block_batching indexer_test_run_ffi shared_accounts account"
+            "amm cross_zone_bridge pinata indexer_state_consistency_with_labels indexer_test_run vault program_deployment"
+            "indexer_ffi_state_consistency_with_labels ata auth_transfer indexer_stall indexer_ffi_block_batching private_pda block_size_limit cross_zone_state_machine"
+            "cross_zone_verified two_zone bridge indexer_state_consistency cross_zone_ping wallet_ffi config cross_zone_ingress_guard"
+          )
```
Arjentix: Oh no no no, how are we supposed to add new tests then? Should we always remember to add them to CI?

.github/workflows/ci.yml
```diff
@@ -107,202 +111,344 @@ jobs:
...
-      - uses: ./.github/actions/install-risc0
+      - uses: ./.github/actions/restore-risc0-runtime
 
-      - name: Install active toolchain
-        run: rustup install
-
-      - name: Download integration test archive
-        uses: actions/download-artifact@v4
+      - name: Install nextest
+        uses: taiki-e/install-action@43aecc8d72668fbcfe75c31400bc4f890f1c5853 # v2.83.2
```
Arjentix, line 371: So I think in description I saw that you says that on dev we install nextest multiple times, but in your PR this is cached, am I right?

So how does it work then?

.github/workflows/ci.yml
```diff
@@ -107,202 +111,344 @@ jobs:
...
+          profile: build
 
       - name: Restore Rust cache
-        uses: Swatinem/rust-cache@v2
+        uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # v2.9.1
         with:
           shared-key: ci-rust-cache
-          save-if: ${{ github.ref == 'refs/heads/main' || github.ref == 'refs/heads/dev' }}
+          cache-bin: false
+          save-if: false
```
Arjentix, line 511: Why btw? Why saving on main and dev isn't a good thing?

.github/workflows/bench-regression.yml
```diff
@@ -2,8 +2,8 @@ on:
   pull_request:
     paths:
       - "tools/crypto_primitives_bench/**"
-      - "lez/key_protocol/**"
-      - "lee/core/**"
+      - "lee/key_protocol/**"
+      - "lee/state_machine/core/**"
```
Arjentix, line 6: Good catch

.github/actions/install-risc0/action.yml
```diff
@@ -1,10 +1,72 @@
...
+        key: risc0-${{ runner.os }}-${{ runner.arch }}-${{ inputs.profile }}-rzup-0.5.0-r0vm-3.0.5-rust-1.94.1-cargo-risczero-3.0.5
+
+    - name: Install pinned RISC Zero components
+      env:
+        GITHUB_TOKEN: ${{ github.token }}
+        RISC0_PROFILE: ${{ inputs.profile }}
       run: |
-        curl -L https://risczero.com/install | bash
-        /home/runner/.risc0/bin/rzup install
+        risc0_home="${RISC0_HOME:-$HOME/.risc0}"
```
Arjentix, line 29: Why so complex code instead of script execution? Is it just to not to rely on their script?

.github/actions/build-integration-shard/action.yml
```diff
@@ -0,0 +1,69 @@
...
+    required: true
+runs:
+  using: composite
+  steps:
+    - name: Build integration test shard ${{ inputs.shard }}
+      env:
+        RISC0_DEV_MODE: "1"
+        SHARD: ${{ inputs.shard }}
+      run: |
+        if [[ ! "$SHARD" =~ ^[0-3]$ ]]; then
```
Arjentix: I've noticed that changing the amount of shards is not just changing one constant. We have to check each usage of this 0-3 thing

.github/workflows/ci.yml
Arjentix: Ok, I see, this thing installs binaries directly, this may save some time

.github/actions/build-integration-shard/action.yml
3esmit: I will reput them individually as it was before.

.github/workflows/ci.yml
3esmit: I agree, will undo this.

.github/workflows/publish_images.yml
3esmit: This pins the action to immutable code. Tags such as `v5` can move, while the full commit SHA always resolves to the reviewed revision. The trailing `# v5.0.0` comment keeps the release readable, do you prefer to keep it pinned to tag?

Arjentix:
Thanks for the PR, I like some changes.

But I'd like to raise some questions. Especially considering my PR #618.

- I feel like my approach with universal Docker image kind of encapsulates all dependencies for our job in one place and this looks good. Jobs easily pull existing image and don't redownload/rebuild the dependencies. Do you think your approach with cached artifacts is better? To me it looks more cherry-picked in terms of dependencies (e.g. in your case just risc0) and looks like it requires more code.
- Pining all versions of everything doesn't seem to be so reasonable for us IMO. I don't think we faced any problems with that, while it helps us to not to forget to update local tools when needed (e.g. cargo fmt).
- I kind of like sharding approach, 40 jobs we have right now is a bit overwhelming. From the other side it's hard to decide which tests should go at which shard logically. E.g. if I want to check if integration tests from module `authenticated_transfer` have passed I need to manually look at every shard job and look for it. And if some test is flaky and I want to rerun it I need to rerun the whole shard and wait for 30 mins, while on dev and in #618 it's a separate tiny job

.github/dependabot.yml
3esmit: Agreed. Full-SHA pins are part of making this CI revision deterministic, but repository-wide update automation is separate policy. I removed the Dependabot configuration from this PR and will leave it for a follow-up.

.github/workflows/ci.yml
3esmit: Exactly. Each fresh runner downloads the exact prebuilt nextest release instead of compiling it and its dependencies. In the baseline, 33 nextest source builds consumed 115.87 runner-minutes; the prebuilt install takes only a few seconds per runner. `fallback: none` also prevents an unnoticed source build.

.github/workflows/ci.yml
3esmit: They are excluded only from the shared workspace rustdoc invocation. The loop immediately below runs rustdoc for both guest packages in separate target directories. Their generated guest artifacts use colliding output names when built into one target directory, so running them separately preserves full coverage without the collision.

.github/workflows/ci.yml
3esmit: Agreed. I restored nextest's default fail-fast behavior for unit and fixture jobs. Integration targets remain independent jobs, so unrelated targets can still report concurrently without making one job continue after its first failure.

.github/workflows/ci.yml
3esmit: Nextest itself is not restored from the Cargo cache. Each fresh runner downloads the exact prebuilt nextest binary, which takes seconds instead of compiling it from source for several minutes. The Rust cache is for repository dependencies across runs, while the producer's compiled integration tests are passed to same-run workers as workflow artifacts. These are separate reuse paths.

.github/workflows/ci.yml
3esmit: Saving on `main` and `dev` is still enabled, but only in the integration producer for this shared key. Other jobs restore the same cache read-only. Parallel jobs cannot use a cache saved by another job in the current run, so multiple writers only upload competing copies of the same key. One trusted writer keeps the useful branch cache without that contention.

.github/actions/install-risc0/action.yml
3esmit: Yes, mainly to make the bootstrap and installed components deterministic. The previous installer path resolved mutable defaults and installed more than these jobs need. This checks the rzup binary checksum, asserts rzup 0.5.0, then installs exact `r0vm`, Rust, and `cargo-risczero` versions for the chosen profile. On an exact cache hit, the install commands are skipped. I can add a short comment in the action so that intent is visible next to the code.

3esmit:
@Arjentix I started working on this fix at the same as you. I came across this because I submitted some PRs related to wallet, and the CI was failing again and again due out of space, and not just that, the CI was so long, that I decided to contribute to fix it. 
I am unsure exactly what is the best approach, but I am sure that the CI needs to be fixed somehow. As this is not my repository home (my tasks live in `lez-programs` repository), I don't have any strong preference for which solution for this problem is taken, just that a solution is merged.

3esmit:
Follow-up proposal (intentionally out of scope for this PR): gate the real succinct-proof job by change impact so unrelated PRs do not spend about an hour proving an unchanged circuit.

This should be implemented and validated in a separate PR. That PR can test the policy against representative changed-file cases before enabling skips:

- always run on `main`, `dev`, scheduled runs, and manual runs;
- run on PRs that change the proof integration test, guest/circuit code or artifacts, verifier/state-machine paths, relevant wallet/sequencer code, RISC Zero/toolchain configuration, workspace dependency resolution, or the proof workflow itself;
- use dependency metadata where practical instead of a permanently hand-maintained package list;
- treat renamed/deleted files, incomplete API results, parse errors, and unknown paths as proof-required (fail closed);
- leave a visible successful job/summary explaining why proof ran or was skipped.

PR #628 stays focused on artifact reuse, bounded disk use, automatic per-target integration jobs, and readable logs. It also changes proof-sensitive files, so retaining the real proof run here gives the required validation signal.

3esmit:
@copilot resolva os conflitos de merge nesta branch.

PR #630, opened 2026-07-17
feat!: ciphertext padding

### LINKED ISSUE #562: Consider Padding Ciphertext Data

agureev:
## 🎯 Purpose

Closes #562. Note ciphertexts are variable length, so the size of the account data behind each note is visible on chain. Every note the wallet emits is now padded to a uniform 512 bytes, which together with #592 action padding and #591 view-tag randomization makes a private transaction look like 7 identical-shaped actions.

## ⚙️ Approach

- `PrivacyPreservingCircuitInput` carries `ciphertext_padding: Option<u32>`, threaded to `EncryptionScheme::encrypt`, which zero-extends the plaintext before ChaCha.
- The pad is a floor, not an exact size: a plaintext longer than it is encrypted at its own length instead of panicking the proof.
- The wallet pads to `CIPHERTEXT_PAD_SIZE` (512) and sizes its dummy notes to match.

## 🧪 How to Test

`cargo test -p lee_core --features host encrypt`, `cargo nextest run -p lee -E 'test(padded)'`, and the `private_transaction_padding` integration test.

## 💥 Breaking

`PrivacyPreservingCircuitInput` gains a field, so the circuit image id changes. Nodes and wallets pinned to the previous image id reject proofs from this one.


lee/state_machine/core/src/encryption/mod.rs
```diff
@@ -120,11 +120,20 @@ impl EncryptionScheme {
...
         nullifier: &Nullifier,
+        pad_to_len: Option<u32>,
     ) -> Ciphertext {
         // Plaintext: PrivateAccountKind::HEADER_LEN bytes header || account bytes.
         // Both variants produce the same header length — see PrivateAccountKind::to_header_bytes.
         let mut buffer = kind.to_header_bytes().to_vec();
         buffer.extend_from_slice(&account.to_bytes());
+        if let Some(pad_to_len) = pad_to_len {
+            let pad_to_len = usize::try_from(pad_to_len).expect("pad length fits in usize");
+            assert!(
```
moudyellaz, line 131: Padding closes the length leak nicely. One robustness note: the `pad_to_len >= buffer.len()` assert runs in-guest, so any account whose plaintext exceeds CIPHERTEXT_PAD_SIZE panics proving and the wallet can't send any private tx for it. At pad 512, HEADER_LEN (81) + 68 overhead leaves ~363 bytes for data, vs DATA_MAX_LENGTH (100 KiB) allowed, so it's latent but could bite on larger payloads. Could we add a host-side check before env write returning a typed error naming the account and overflow, instead of an opaque in-guest panic?

lez/wallet/src/account_manager.rs
```diff
@@ -5,8 +5,8 @@ use keycard_wallet::{KeycardWallet, python_path};
 use lee::{AccountId, PrivateKey, PublicKey, Signature};
 use lee_core::{
     Commitment, CommitmentSetDigest, DummyInput, Identifier, InputAccountIdentity, MembershipProof,
-    NullifierPublicKey, NullifierSecretKey, PrivateAccountKind, SharedSecretKey,
-    account::{Account, AccountWithMetadata, Nonce},
+    NullifierPublicKey, NullifierSecretKey, SharedSecretKey,
+    account::{AccountWithMetadata, Nonce},
     compute_digest_for_path,
     encryption::{
```
moudyellaz, line 11: Padding is wired always-on here (Some(CIPHERTEXT_PAD_SIZE) at the one call site), so the issue's opt-out for users who'd rather not pay the storage cost isn't reachable from the wallet, every private tx now writes 7×512 bytes. Reasonable as a default, could you add a line noting the always-on is deliberate, or leave a knob for later?

moudyellaz (commented):
Padding closes the #562 length leak cleanly, but i think that the added tests only assert padded ciphertext length. A padded encrypt then decrypt roundtrip would be worth adding, since the trailing-zero tolerance in Account::from_cursor is what keeps padded notes decryptable and nothing currently pins it. Encrypt with Some(512), decrypt, assert the original kind and account come back.

lee/state_machine/core/src/encryption/mod.rs
```diff
@@ -114,17 +114,26 @@ impl EncryptedAccountData {
...
 impl EncryptionScheme {
+    /// `pad_to_len` is a floor: shorter plaintexts are zero-extended to it, longer ones are left
+    /// alone. Decryption needs no counterpart, `Account` bytes are length-prefixed.
     #[must_use]
     pub fn encrypt(
         account: &Account,
         kind: &PrivateAccountKind,
         shared_secret: &SharedSecretKey,
         nullifier: &Nullifier,
+        pad_to_len: Option<u32>,
```
jonesmarvin8, line 125: Shouldn't the docstring for `pad_to_len` be shifted to the entry? Also, might be good to mention when `None` is desirable.

jonesmarvin8 (approved):
Lgtm. I added one minor comment.

It is worth noting that this PR was originally designed prior to the support of program updates. As such, the privacy transactions did not reveal the specific exected programs. Now, we do; each program called during a privacy execution is emitted.

Padding (with dummy accounts) was intended to obfuscate the programs executed privately. This can be partially recovered with [PR 824](https://github.com/logos-blockchain/logos-execution-zone/pull/824) (functional privacy for immutable programs).

Perhaps, it makes sense to make ciphertext padding and dummy accounts optional or dependent on program immutability (to save on fees for privacy executions)?

lee/state_machine/core/src/encryption/mod.rs
moudyellaz: Done in 731b921b5.

moudyellaz:
@jonesmarvin8  Re the program visibility, that part of the original motivation is gone. Note length still leaks the size of the account data behind each note, which is #562, and that stays true whether or not the program is visible.

On the fee side, padding is free today: PRIVATE_GAS_STOR is flat (PROOF_BYTES + PRIVATE_PAD_BYTES) and size is not an input to the assessment, so making the pad optional saves nothing in gas, only block bytes, while shrinking the anonymity set for whoever opts out. That is why the wallet keeps it always on.

Immutability-conditional padding makes more sense once #824 lands and function privacy actually recovers something; happy to revisit it there rather than in this PR.

Two things changed since your review: the pad is now capped at MAX_CIPHERTEXT_PADDING (8 KiB) so a prover cannot inflate a transaction into a whole block at the flat fee, and the guest now rejects dummy notes shorter than the requested pad instead of trusting the host to size them.

schouhy (approved):
lgtm thanks

PR #638, opened 2026-07-22
feat(wallet): multi-sequencer client parallelizm

Pravdyvy:
## 🎯 Purpose

Parallelization of various multi-sequecner client operations.

## ⚙️ Approach

- [x] Parallel actualization function.
- [x] Parallel callibration.
- [x] Parallel transaction distribution.

## 🧪 How to Test

Usual tests must pass

## 🔗 Dependencies

None

## 🔜 Future Work

None

## 📋 PR Completion Checklist

- [x] Complete PR description
- [ ] Add/update tests


lez/wallet/src/multi_client.rs
```diff
@@ -478,85 +521,160 @@ pub async fn actualize_client(client: &SequencerClient) -> StatisticsUpdate {
...
             ) > avg_err_ratio
         })
-        .unwrap_or(client_vec.len()))]
+        .unwrap_or(res_vec.len()))]
         .to_vec();
 
     // Choose clients with least latency and variance
-    client_vec.sort_by(|a, b| {
-        let left = statistics.get(*a).unwrap();
+    res_vec.sort_by(|a, b| {
```
moudyellaz, line 685: this is something that existed i think since PR623. the comparator isn't a total order: for a = (latency_avg 1, latency_var 100) and b = (latency_avg 2, latency_var 1), cmp(a,b) and cmp(b,a) both return Greater. sort_by then settles on whatever input order gives it, which is now config order rather than latency, and it panics with "comparison function does not correctly implement a total order" once the candidate set gets large enough. maybe you should compare a single score: (a.latency_avg + a.latency_var.sqrt()).total_cmp(&(b.latency_avg + b.latency_var.sqrt()))?

lez/wallet/src/multi_client.rs
```diff
@@ -118,35 +124,40 @@ impl MultiSequencerClient {
...
-                    "Metered call for {sequencer_addr:?}, statistic updates is {statistic_updates:?}"
-                );
+            if statistics.contains_key(sequencer_addr) {
+                actualization_list.push((sequencer_addr.clone(), sequencer_client.clone()));
+            } else {
+                calibration_list.push((sequencer_addr.clone(), sequencer_client.clone()));
+            }
 
-                statistic_mut.apply_updates(&[statistic_updates]);
+            client_list.push((sequencer_addr.clone(), sequencer_client));
```
moudyellaz, line 142: switching client_list from HashMap<Url, _> to Vec drops the implicit dedup, so two config entries with the same sequencer_addr no longer collapse. same url twice means it takes two leader_list slots, gets the tx sent to it twice, has its statistics counted twice, and metered_get failover can retry the node that just failed instead of a different one. in multi_actualize_clients/multi_calibrate_clients the handle_map insert also replaces one duplicate's JoinHandle, and dropping a JoinHandle detaches rather than cancels, so that task keeps running and its result is thrown away. could you dedup conn_data by sequencer_addr before building the lists?

lez/wallet/src/multi_client.rs
```diff
@@ -478,85 +521,160 @@ pub async fn actualize_client(client: &SequencerClient) -> StatisticsUpdate {
     })
 }
 
+// Next 2 functions can be done in uniform way right now, but it will be incredibly cursed.
+// ToDo: Return to it when "return-type notation" is stable
+
+pub async fn multi_actualize_clients(
```
moudyellaz: the three new paths (multi_actualize_clients, multi_calibrate_clients, metered_send_transaction) have no coverage; could you add one test that parallel setup yields the same statistics as the serial path for a fixed client set?

lez/wallet/src/multi_client.rs
```diff
@@ -280,31 +298,55 @@ impl MultiSequencerClient {
...
+                        actualize_client(curr_leader.clone())
+                    ),
+                    curr_url,
+                )
+            });
+        }
 
-        for (leader, leader_url) in leaders {
-            let (resp, statistics_update) = tokio::join!(call(leader), actualize_client(leader));
+        while let Some(resp) = join_set.join_next().await {
```
moudyellaz, line 342: a panicking send task turns into ExecutionFailureKind::JoinError, and find(Result::is_ok) in lib.rs drops every Err, so all the caller ever sees is MultiSequencerTransactionSendError with no cause. could you log the per-leader errors at warn before returning?

lez/wallet/src/multi_client.rs
```diff
@@ -118,35 +124,40 @@ impl MultiSequencerClient {
...
-                statistic_mut.apply_updates(&[statistic_updates]);
+            client_list.push((sequencer_addr.clone(), sequencer_client));
+        }
 
-                client_list.insert(sequencer_addr.clone(), sequencer_client);
-            // Otherwise calibrate client data
-            } else if let Some(client_statistics) = calibrate_client(
-                &sequencer_client,
-                multi_sequencer_client_config.calibration_limit,
+        // This actually runs in one thread, but eah exact future is parallelized.
```
moudyellaz: typo, "eah" -> "each". also the comment isn't quite right anymore: both helpers call tokio::task::spawn and both binaries run a multi-thread runtime, so the probes do spread across worker threads. and line 303, "Less abstract that" -> "than".

lez/wallet/src/multi_client.rs
```diff
@@ -478,85 +521,160 @@ pub async fn actualize_client(client: &SequencerClient) -> StatisticsUpdate {
...
+    )]
+    for (url, task) in handle_map {
+        let task_opt = task.await.ok();
+        statistic_updates.push((url, task_opt));
+    }
+
+    statistic_updates
+}
+
+pub async fn multi_calibrate_clients(
```
moudyellaz: these two have no callers outside the module, and this PR makes calibrate_client/actualize_client/choose_leaders private, so could they be private too?

lez/wallet/src/multi_client.rs
```diff
@@ -581,16 +699,10 @@ pub fn choose_leaders(
     });
 
     Some(
-        client_vec
-            .iter()
+        res_vec
+            .into_iter()
+            .map(|(a, b)| (b.clone(), a.clone()))
```
moudyellaz: res_vec holds references, so this clones every client and url on the way out. since client_vec is owned now, retain + into_iter would drop the clones entirely.

moudyellaz (approved):
lgtm, thanks! left some minor comments

lez/wallet/src/multi_client.rs
Pravdyvy: True. Also `(a.latency_avg + a.latency_var.sqrt()).total_cmp(&(b.latency_avg + b.latency_var.sqrt()))?` gives everything I need. 
Updated.

lez/wallet/src/multi_client.rs
Pravdyvy: I even think it should be an error. Added uniqueness check which `bail!` on failure.

lez/wallet/src/multi_client.rs
Pravdyvy: It is kind of complex. I can implement mock client to test this exact case, which will be implemented as regular client with trapping middleware. It will take some time.
Alternatively I can include testing of this in the integration test PR #643 on which I work right now, this also will take time.
In place testing will work faster, but integration tests would not be mocked, what do you think is preferable?

lez/wallet/src/multi_client.rs
Pravdyvy: Added logging in `inspect_err`.

lez/wallet/src/multi_client.rs
Pravdyvy: Fixed typos and comment.

lez/wallet/src/multi_client.rs
Pravdyvy: Removed `pub`

lez/wallet/src/multi_client.rs
Pravdyvy: Updated

lez/wallet/src/multi_client.rs
```diff
@@ -10,13 +10,19 @@
 use std::{collections::HashMap, path::Path, sync::Arc};
 
 use anyhow::{Context as _, Result};
+use common::{HashType, transaction::LeeTransaction};
+use itertools::Itertools as _;
 use lee_core::BlockId;
+use log::warn;
```
erhant: There are some parts where we use `log::warn!` OR `warn!` ; and I honestly think the fully qualified path `log::warn!` is much readable, and we should eventually use that alone.

lez/wallet/src/multi_client.rs
```diff
@@ -478,119 +532,175 @@ pub async fn actualize_client(client: &SequencerClient) -> StatisticsUpdate {
...
+    }
+
+    statistic_updates
+}
+
+async fn multi_calibrate_clients(
+    clients: Vec<(Url, SequencerClient)>,
+    calibration_limit: usize,
+) -> Vec<(Url, Option<Statistics>)> {
+    let mut handle_map = HashMap::new();
```
erhant, line 567: Would a [JoinSet](https://docs.rs/tokio/latest/tokio/task/join_set/struct.JoinSet.html) do better here? It is used elsewhere too

lez/wallet/src/multi_client.rs
```diff
@@ -478,119 +532,175 @@ pub async fn actualize_client(client: &SequencerClient) -> StatisticsUpdate {
...
+                }
+            })
+            .flatten();
+        statistics.push((url, task_opt));
+    }
+
+    statistics
+}
+
 #[must_use]
```
erhant: should be below the comment block

erhant (approved):
LGTM, few small comments

lez/wallet/src/multi_client.rs
Pravdyvy: Replaced all entries of `warn!` with `log::warn!` in this file.

lez/wallet/src/multi_client.rs
Pravdyvy: Moved

lez/wallet/src/multi_client.rs
Pravdyvy: I want to know url of client in a task, which failed to join. JoinSet would not allow something like that.

lez/wallet/src/multi_client.rs
moudyellaz: #643 works fine

lez/wallet/src/multi_client.rs
moudyellaz: typo "adresses" -> "addresses".

lez/wallet/src/multi_client.rs
moudyellaz: nice, that covers the join error. the send errors at 356 are still dropped by find(is_ok) though, so a total failure has no cause attached. leader_url is right there, another inspect_err would do it. not from your PR, so fine to leave if you prefer.

lez/wallet/src/multi_client.rs
moudyellaz: could 672 use truncate(pos) instead of res_vec[..pos].to_vec()? same result without cloning every survivor.

lez/wallet/src/multi_client.rs
moudyellaz: just a note, i think JoinSet can actually handle this: join_next_with_id hands back the task Id with the result, and JoinError::id() survives a panic, so an Id -> Url map keeps attribution even when a join fails. metered_send_transaction already does the tuple trick for the non-panic case. what you have works fine, so no need to change it, just so the constraint doesn't get taken as given later.

lez/wallet/src/multi_client.rs
Pravdyvy: Yes, updated.

lez/wallet/src/multi_client.rs
Pravdyvy: Fixed

Issue #634, opened 2026-07-20
fix(sequencer): durable resubmission of orphaned user transactions


erhant:
> Follow-up from https://github.com/logos-blockchain/logos-execution-zone/pull/603#discussion_r3604568578

When a sequencer's block is orphaned, `apply_follow_update` returns its user transactions to the mempool so the next on-turn production re-includes them. Three loss paths remain:

1. **Full mempool drops the tx.** The resubmit uses `mempool_handle.try_push`, which warns and drops when the mempool is full. `try_push` is deliberate — a blocking push from the publisher's drive task would deadlock, since only block production drains the mempool — but the dropped tx is gone.

2. **Publish failure exits without re-queue.** Transactions popped from the mempool into a block whose `publish_block` then fails are never put back; the error propagates through the main loop and the node exits.

3. **Any halt loses the whole in-memory mempool.** #603 made halts prompt and deliberate (a persist failure panics the drive task, whose termination now trips `SequencerHandle::failed()`), which is correct for state consistency, but every such exit also drops all queued and reclaimed txs, since the mempool has no persistence.

In all cases a submitted (or mined-then-orphaned) user transaction can vanish while the client believes it landed.

---

Possible direction: a persisted resubmit queue in the sequencer store (drained on turn, ahead of the mempool), so orphan reclaims and popped-but-unpublished txs survive both backpressure and restarts. Note that fixing (2) alone by re-queuing in memory before exiting would be pointless (the exit wipes the mempool anyway) so persistence is the actual requirement. Needs to compose with the orphan-resubmit filters added in #603 (clock tx, bridge deposits, and sequencer-only cross-zone txs are excluded).

erhant:
The three loss paths split cleanly on one question: *did the client ever receive an inclusion signal?*

**Paths (2) and (3) — never published.** The tx never reached the channel, so the client got no inclusion signal, and `sendTransaction` never promised one: it returns a bare `HashType` the moment the tx lands in the local mempool channel (`service.rs:50,91-95`). The wallet's `poll_tx` then fails **loudly** with `"Transaction not found in preconfigured amount of blocks"` (`poller.rs:34-64`). That is ordinary volatile-mempool behaviour; the client resubmits. Persisting these would have the sequencer guarantee durability for txs it merely accepted into a bounded in-memory channel — over-promising relative to its own API, on the one surface that is client-driven and therefore DoS-exposed.

[...]

PR #642, opened 2026-07-23
fix(sequencer): follow-up fixes for #633 and #639

### LINKED ISSUE #633: fix(sequencer): persist the zone-sdk checkpoint atomically with followed blocks and state
Follow-up from https://github.com/logos-blockchain/logos-execution-zone/pull/603#discussion_r3604568569; TODOs already in code on `RocksDBIO::store_followed_blocks` and `apply_follow_update`.

The `zone-sdk` checkpoint is persisted by `on_checkpoint` *before* the follow path's atomic write of blocks + tip meta + state. A crash in that window makes restart resume the sdk stream **past** blocks that never landed in the store (`is_fresh_start` is derived from checkpoint presence), so the node comes up with a gap it can never backfill — an operator resync is required.

As a proposed fix: fold the checkpoint into the same RocksDB `WriteBatch` as the blocks, tip meta, state and orphan reverts of the `BlocksProcessed` event that produced it, per the sdk's documented event contract.

This would also naturally cover the remaining orphan-only-update gap from #603: an orphan-only update currently persists nothing, so a crash there resurrects the orphaned tip until its replacement arrives (bounded today by the finalization rebase, but full-event atomicity closes it properly).

### LINKED ISSUE #639: bridge: L1 to L2 deposits lack chain-level exactly-once dedupe by `l1_deposit_op_id`
## Problem

Bridge deposit minting has no replay protection in chain state. The bridge program ignores its dedup key (as in, `lez/programs/bridge/src/main.rs` literally destructures `l1_deposit_op_id: _`) and just chains a vault transfer, and the
sequencer-built deposit tx carries an empty witness set (see `build_bridge_deposit_tx_from_event`), so LEE's nonce check never fires
(nonces are validated per-signature only). The same deposit instruction can therefore be applied to state any number of times, transferring the amount again on each application.

Exactly-once currently rests entirely on **per-node** bookkeeping (`pending_deposit_events` records + the production-time `is_deposit_event_submitted` guard). That bookkeeping neither survives reorgs nor coordinates across sequencers.

### Known manifestations (same root cause)

1. **Multi-sequencer double-credit.** Every accredited sequencer receives the finalized `Deposit` event and enqueues its own mint tx. A includes it in a block; B's follow path does no deposit bookkeeping, so B's local guard passes and B includes its copy on its turn → recipient credited twice. (flagged by @schouhy on #603)
2. **Deposit loss on orphan.** The including block is orphaned: the record
   stays `submitted_in_block_id = Some(N)` (never un-marked),
   `resubmittable_txs` deliberately skips deposit txs, and startup replay only
   re-queues records with `submitted_in_block_id = None` → the mint is never
   re-attempted. Worse, `remove_fulfilled_pending_deposit_events_up_to_block`
   deletes records by block-id ordering at finality, so a replacement block
   *without* the tx finalizing at that height permanently erases the evidence.
3. **Bootstrap double-mint** (single node): cold-start backfill re-delivers an already-finalized deposit into the mempool while reconstruction replays the block that already minted it. Mitigated node-locally in #606 (`is_deposit_event_submitted` production guard + reconstruction marking), but that guard is per-store and does nothing for **problem (1)** above.

**Prior art**: #555 (closed) framed this as "transient queue vs. durable processed-ledger"; the #606 guard extends the queue bookkeeping. Both are point patches on node-local state. The class persists because dedup state (RocksDB records) and mint-driving state (mempool) are separate stores that can disagree, and no store is shared between sequencers.

## Proposed fix

Move exactly-once into chain state, the only thing all sequencers agree on:

- [ ] **Bridge program enforces dedup on `l1_deposit_op_id`**: e.g. create a "deposit receipt" PDA seeded by the op id (creation fails if it exists), or track consumed ids in bridge PDA state. A duplicate then fails validation everywhere: dropped at block production, and a block containing one is rejected by every follower via the shared `apply_block`.
- [ ] **Follow-path bookkeeping**: extract deposit op ids from adopted/finalized blocks in `apply_follow_update` and mark local records submitted during follow-path persistence, so a sequencer doesn't attempt a duplicate its peer already included.
- [ ] **Orphan un-marking**: reset records from orphaned blocks to unsubmitted so the existing replay machinery re-queues them — safe once the program dedups.
- [ ] **Fix finality cleanup** to remove records by op-id-observed-in-finalized-block, not by block-id ordering.
- [ ] Rebuild committed `artifacts/` + prebuilt test fixture; the program change alters the risc0 guest image (`just build-artifacts`).

With program-level dedup in place, the node-local guards from #555/#606 collapse into optimizations rather than correctness mechanisms.

## Acceptance tests

- **Duplicates**: A block containing a duplicate `l1_deposit_op_id` fails `apply_block`; a duplicate mint is dropped at block build.
- **Multi-sequencer integration**: a bridge deposit with a two-sequencer committee credits the recipient vault exactly once on both chains.
- **Reorg**: orphan a deposit-carrying block; the mint lands in the replacement branch exactly once.
- The #606 reconstruction tests keep passing.

## Notes

- Depends on #603 (the follow path this builds on); this is a follow-up issue.
- Test-infra gap worth closing alongside: `MockBlockPublisher` drops `on_deposit_event`, so re-delivery semantics are invisible to unit tests. (this also caused a bug on #606, noticed during review)
- Folding the follow-path deposit marking into one atomic batch overlaps with #633 (full `BlocksProcessed` atomicity) so we should coordinate the fix of these two together.

erhant:
### Closes #633

The zone-sdk checkpoint — the stream resume cursor — was persisted **before** the blocks, state, and bookkeeping it covers. `Event::BlocksProcessed` fanned out into four independent RocksDB writes (checkpoint first, then finalization cleanup, then deposit/withdraw records, then the block batch), so a crash in that window resumed the sdk stream *past* blocks that never landed in the store — a gap the node cannot backfill, requiring an operator resync.

This folds everything one sequencer event produces into a single atomic `WriteBatch`, so the resume cursor can never advance ahead of the effects it covers.

### Closes #639

Bridge L1 to L2 deposits had no chain-level replay protection: the bridge program destructured `l1_deposit_op_id: _` and the sequencer-built mint tx carries an empty witness set, so LEE's nonce check (which is per-signature) never fired. The same deposit instruction could be applied to state any number of times. Exactly-once rested entirely on per-node RocksDB bookkeeping, which neither survived reorgs nor coordinated across sequencers — giving multi-sequencer double-credit, deposit loss on orphan, and bootstrap double-mint.

> [!NOTE]
>
> I also took a look at #634 but that may require some invasive mempool refactoring; will defer that on a later PR.

> [!NOTE]
> **Why enforce deposit exactly-once in the bridge program rather than with a node-level tracker?**
>
> A node-local tracker is actually what we had before (`submitted_in_block_id`), but it can't deliver the guarantee under decentralized sequencing. Duplicates don't only originate locally, they arrive inside *peer* blocks we adopt or replay during reconstruction, and those paths are validated by the guest, never by our block-building filter.
>
> So, the check must live where peer blocks are actually validated. A private tracker is also per-node: a second sequencer that independently observed the same L1 deposit has no record that we already minted it, so the only shared source of truth all nodes agree on is the L2 chain state itself. Anchoring dedup in a state-resident receipt PDA means it reverts automatically on a reorg (no hand-written rollback), and it stays a replayable ledger invariant that an indexer or a reconstructing node can verify from chain data alone rather than trusting private storage.
>
> The node still keeps a `deposit_already_minted` check during production, but purely as an optimization that reads the authoritative receipt to avoid emitting a wasted no-op tx, the correctness boundary is the program's no-op-on-existing-receipt. The mechanism mirrors the existing `cross_zone_inbox` seen-shard pattern (cc @moudyellaz).

### #633 fixes:

- [x] **Add** `StoreUpdate` — a struct describing every effect of one event (checkpoint, blocks, tip meta, head state, final snapshot, deposit/withdraw bookkeeping, channel anchor) — and `RocksDBIO::store_update`, which stages all of it into one `WriteBatch`.
- [x] **Change** the publisher to deliver one `FollowUpdate` (checkpoint + adopted/orphaned/finalized + deposits + withdrawals) through a single `OnFollowSink`, replacing the four separate sinks (`CheckpointSink`/`FinalizedBlockSink`/`OnDepositEventSink`/`OnWithdrawEventSink`). `BlockPublisherTrait::new` drops from 9 args to 5.
- [x] **Fix** the produce path: `publish_block` now returns its checkpoint, persisted inside `atomic_update`'s batch. The checkpoint carries the sdk's `pending_txs`, so storing a produced block without it would restore a pending set missing the just-published inscription — which the sdk would never resubmit.
- [x] **Change** bridge-deposit mints to drain from the store at block production instead of the mempool. The persisted deposit record is the queue; the old async `on_deposit_event` sink did a blocking `mempool.push().await` from the drive task, which a non-turn sequencer never drains — a hard **deadlock**. This also subsumes the startup `replay_unfulfilled_deposit_events` and removes a double-mint race.
- [x] **Fix** deposit/withdraw bookkeeping to fold in memory across the whole event before writing once — the records live in single whole-vector cells, so per-effect disk re-reads would silently drop earlier staged writes (one event can carry several deposits during backfill).
- [x] **Change** `apply_reconstructed_block` from three separate writes to one `store_update` call.

### #639 fixes:

- [x] Add `bridge_core::deposit_receipt_{seed,account_id}` :a per-op-id PDA (domain-separated hash), following the same house pattern as `cross_zone_inbox`'s seen-shard.
- [x] Fix the bridge program to take a 3rd receipt account: if it already exists the deposit is a **no-op**, otherwise it claims the receipt via `Claim::Pda` and chains the vault transfer. The receipt's existence in chain state is the exactly-once ledger, so every follower reaches the same verdict through `apply_block`. Much better for decentralized sequencing.
- [x] Change the production drain to skip deposits whose receipt already exists in the state being built on. This is all multi sequencer correct and reorg correct with no bookkeeping (an orphan reverts the receipt with the block, so the next turn re-mints).
- [x] Remove the node-local dedup that this supersedes: the `is_deposit_event_submitted` guard and the entire `submitted_in_block_id` field. `PendingDepositEventRecord` is now a pure liveness queue.
- [x] Fix finality cleanup to remove records by op-id observed in a finalized block (`remove_deposit_records`) instead of by block-id ordering, which erased evidence when a replacement block finalized at the same height.
- [x] Add a borrowing `V03State::get_public_account` so the drain's receipt check doesn't `clone` an `Account` per record.
 
> [!NOTE]
>
> While #639  acceptance criteria say a duplicate should fail `apply_block`; we implement a no-op instead.
>
> A duplicate here is the natural multi-sequencer race (**B** builds a mint before adopting **A**'s block that already had it), not a byzantine act — and LEZ blocks are all-or-nothing at apply, so fail-hard would park the block on every follower, while on the producer it propagates through ? and kills the node. Acceptance therefore becomes "recipient credited exactly once" rather than "block rejected."
  
## 🧪 How to Test

```sh
RISC0_DEV_MODE=1 cargo test -p storage -p sequencer_core -p bridge_core --all-features
```

New coverage:
- **storage**: checkpoint lands with an orphan-only update; a checkpoint-only event does not rewrite head state; several deposits in one event are all recorded; re-delivered deposit is not re-accepted; repeated withdrawal key folds to just once per occurrence; a produced block persists its publish checkpoint.
- **sequencer_core**: a follow update persists its checkpoint alongside its effects; observed deposits are recorded for the production drain; unfulfilled deposits are drained from the store (not the mempool) on production; a drained deposit is not minted twice across turns.

## 🔗 Dependencies

None.

## 🔜 Future Work

- **#634**: durable resubmit queue. The store-as-queue drain established here is where it plugs in; the `try_push` orphan-resubmit still drops on a full mempool (marked with a TODO).
- **Shared `unchanged_post_states` helper**: it's duplicated across bridge, vault, faucet and the inbox. Extracting it into lee_core::program would touch three unrelated guests and rebuild their images, so it's left for its own change.
- **Receipt state growth**: one permanent account per deposit, deliberately unbounded rather than epoch-sharded like the inbox — for a component that mints native funds, a forgetful design's worst case is a double-mint of real value.
- **Bounded finalization scan**: `finalized_up_to` still scans the whole block CF per finalizing event. A persisted watermark (`LastFinalizedBlockIdCell` already exists, unused) would bound it, but changes backfill semantics — deferred as its own decision. Not a regression: the prior code scanned per finalized block rather than per event.
- **Pre-existing off-by-one** in unseen-withdraw reconciliation (N intents permit N+1 consumes) is replicated as-is here, not fixed; worth its own issue.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


lez/storage/src/sequencer/mod.rs
```diff
@@ -396,88 +460,145 @@ impl RocksDBIO {
...
         }
+        let accepted = records.len().saturating_sub(before);
+
+        let removed = if remove_op_ids.is_empty() {
+            0
+        } else {
+            let after_append = records.len();
+            records.retain(|record| !remove_op_ids.contains(&record.deposit_op_id));
+            after_append.saturating_sub(records.len())
+        };
```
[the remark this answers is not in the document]
erhant: `retain` membership switched to a `HashSet` in https://github.com/logos-blockchain/logos-execution-zone/pull/642/commits/ede6ca5468d170f966b1d44e37a8faf4410cdfb6

lez/storage/src/sequencer/mod.rs
```diff
@@ -396,88 +460,145 @@ impl RocksDBIO {
...
         self.db.write(batch).map_err(|rerr| {
             DbError::rocksdb_cast_message(
                 rerr,
-                Some("Failed to mark deposit events submitted".to_owned()),
+                Some("Failed to add pending deposit event".to_owned()),
             )
-        })
+        })?;
+        Ok(accepted > 0)
     }
```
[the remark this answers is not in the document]
erhant: Empty-write guard added to `add_pending_deposit_event` in https://github.com/logos-blockchain/logos-execution-zone/pull/642/commits/ede6ca5468d170f966b1d44e37a8faf4410cdfb6

This mirrors the `store_update`'s existing `batch.is_empty()` short-circuit; matters because re-delivery is the steady state.

lez/sequencer/core/src/lib.rs
```diff
@@ -742,17 +609,9 @@ impl<BP: BlockPublisherTrait> SequencerCore<BP> {
...
-                    }
-                    deposit_event_ids.push(deposit_op_id);
-                }
-
+                // Bridge deposits are deduped by their receipt PDA in chain
+                // state (drained only when unminted, no-op on replay), so no
+                // node-local guard is needed here.
                 state
                     .transition_from_public_transaction(public_tx, block_height, timestamp)
                     .context("Failed to execute sequencer-generated transaction")?;
```
moudyellaz: the Sequencer branch of apply_mempool_transaction returns Err on execution failure and it propagates all the way to `produce_new_block().await?` in the service loop, which ends the production task. with the drain now reading from the store every turn, and a record only clearing when its own deposit finalizes, a mint that can never execute fails identically on every turn instead of once per restart like it did from the mempool. bridge liquidity is the case i'd worry about, the bridge account is funded once at genesis and every deposit transfers out of it, and since all sequencers see the same L1 deposit they'd all halt together. could the Sequencer-origin failure be a skip-and-log like the User branch above? the record stays queued for retry either way.

lez/sequencer/core/src/lib.rs
```diff
@@ -780,9 +639,36 @@ impl<BP: BlockPublisherTrait> SequencerCore<BP> {
...
+        // Bridge deposit mints are drained from the store, not the mempool: the
+        // follow path records the event durably but cannot enqueue the mint
+        // itself (it runs on the publisher's drive task, where an await stalls
+        // the very task production needs). Draining here also subsumes the old
+        // startup replay.
+        //
+        // Skip any deposit whose receipt PDA already exists in the state we
+        // build on — it was minted by us or by a peer whose block we adopted.
+        // An orphan reverts the receipt with the block, so the next turn
+        // re-mints without any bookkeeping of our own.
```
moudyellaz, line 666: nice coverage on the other three acceptance tests from #639. the orphan one (manifestation 2) looks untested though, and it's the case with the least backstop since deposits aren't requeued and the whole recovery rests on the receipt reverting with the block. could you add one that orphans a deposit-carrying block and checks the mint lands exactly once in the replacement?

lez/storage/src/sequencer/mod.rs
```diff
@@ -396,135 +460,210 @@ impl RocksDBIO {
...
+        for event in new_events {
+            if records
+                .iter()
+                .any(|record| record.deposit_op_id == event.deposit_op_id)
+            {
+                continue;
+            }
+            records.push(event.clone());
         }
+        let accepted = records.len().saturating_sub(before);
```
moudyellaz: `accepted` is counted before the retain runs, so if one event both sees a deposit and finalizes it (backfill delivers both at once), it still counts as accepted. the record is correctly gone, but the log then says a mint is coming when it already happened. counting after the retain fixes it.

lez/sequencer/core/src/lib.rs
erhant: Added `an_orphaned_deposit_is_reminted_exactly_once_in_the_replacement` in https://github.com/logos-blockchain/logos-execution-zone/pull/642/commits/dca87670a0398dd8fd6d795c1e12512c45ca57e7: mints, orphans the block via the follow path, asserts the receipt reverts (`deposit_already_minted` is false), then re-produces and asserts exactly one mint + vault credited once.

lez/storage/src/sequencer/mod.rs
erhant: Reworked `stage_pending_deposit_events` in https://github.com/logos-blockchain/logos-execution-zone/pull/642/commits/dca87670a0398dd8fd6d795c1e12512c45ca57e7
: an op-id present in `remove_op_ids` is neither appended nor counted, so accepted only reflects records that will actually be drained.

lez/sequencer/core/src/lib.rs
erhant: Changed the Sequencer branch to skip-and-log in https://github.com/logos-blockchain/logos-execution-zone/pull/642/commits/dca87670a0398dd8fd6d795c1e12512c45ca57e7 (now mirroring the User branch as you also mentioned), record stays queued for retry.

Also with this we no longer need `Result` return type!

lez/programs/bridge/src/main.rs
```diff
@@ -54,20 +54,47 @@ fn main() {
...
 
-            let mut bridge_for_vault = bridge;
-            bridge_for_vault.is_authorized = true;
+            assert_eq!(
+                receipt.account_id,
+                bridge_core::deposit_receipt_account_id(self_program_id, l1_deposit_op_id),
+                "Third account must be the deposit-receipt PDA"
+            );
+
+            // Replay protection: the receipt PDA exists iff this op id was
```
moudyellaz, line 63: agree with no-op over fail here, for the reason you give in the description. one consequence worth noting: a replay and a real mint are both successful txs with identical shape (post states = pre states), so an indexer can't tell "credited" from "already credited by a peer" without deriving the receipt id and checking whether it existed before the block. probably fine for now, just worth knowing before the explorer starts showing deposits.

moudyellaz (approved):

lez/programs/bridge/src/main.rs
erhant: Good point! Added a comment on https://github.com/logos-blockchain/logos-execution-zone/pull/642/commits/40e319af7fbf274741e3f745ca88a4f39741f9c0

lee/state_machine/src/state/mod.rs
```diff
@@ -270,6 +270,13 @@ impl V03State {
             .unwrap_or_else(Account::default)
     }
 
+    /// Borrowing counterpart of [`Self::get_account_by_id`], for callers that
+    /// only inspect the account and would otherwise clone it to drop it.
+    #[must_use]
+    pub fn get_public_account(&self, account_id: AccountId) -> Option<&Account> {
```
schouhy: can we rename this to `get_account_by_id_ref`? or `find_account_by_id`?

lez/programs/bridge/src/main.rs
schouhy: I didn't understand the need of a noop. Each sequencer adds the bridge tx when they see a deposit event landing on finalized bedrock. If each sequencer adds the bridge tx to their mempool, only one of those will land if the tx fails instead of noop, no? because failed txs are dropped on block production.

schouhy (approved):
lgtm thanks!

lee/state_machine/src/state/mod.rs
erhant: `get_account_by_id_ref` makes much sense, changed in https://github.com/logos-blockchain/logos-execution-zone/pull/642/commits/5c16419515476835869b640e5b579c36be74ef52

lez/programs/bridge/src/main.rs
moudyellaz: deposit mints don't go through the mempool, they're drained from the store at build time and that drain already skips op ids whose receipt exists. so you're right that a failing tx is dropped on the producer's own build path. but blocks are all-or-nothing on apply: apply_block_to_state propagates any tx error, so a duplicate that fails inside an adopted or replayed block parks it on every follower instead of being dropped. no-op keeps the chain moving; fail turns a benign race into a stall (the second NOTE in the description covers this). @schouhy @erhant

erhant: Drop-on-failure only covers the **producer** assembling its own block (`build_block_from_mempool` skips a failed tx).

But a mint also reaches every other node the second way: **applying a block already published** to the channel, which can come from adopting a peer's block, replaying finalized history, or a backfill/reconstruction. That path (`apply_block_to_state`) executes every tx and rejects the **whole block** on the first failure; no per-tx drop.

Consider two sequencers each draining the deposit from their own store before adopting the other's block (the window where neither `head` has the receipt yet), so both mint `D`. Once one is finalized first, the receipt exists — and the other block, which L1 has already finalized and every node must apply, now carries a mint whose receipt exists.

- **If this is a fail:** that block becomes invalid and every follower the chain halts because this is a finalized block with a duplicate mint, and we can't skip it.
- **If this is a no-op:** the block stays valid with the duplicate tx skipped, all nodes reach the same state.

erhant: oh I just saw @moudyellaz 's reply, my internet died while I was writing this so I guess I missed it hahah

schouhy: > Consider two sequencers each draining the deposit from their own store before adopting the other's block (the window where neither head has the receipt yet), so both mint D

right, that's what I think can't happen. because sequencers are not independently building blocks. they only build blocks on their turn.

schouhy: while they are waiting for their turn, sequencers follow blocks posted by the sequencer in turn. all those should be valid as they come. No sequencer should apply a peer's block to a lez state different to the one  used by the peer to construct the block

erhant: you are correct; there is only a small edge-case to this is right at the turn-change:

- A construct a block with mint, and publishes it (almost at the same time as there is a turn change)
- B receives turn notification, and published with block with mint before it sees that of A's

I can't recall now if L1 actually handles this edge case, but no-op is a cheap solution to this edge case

It is also safe against a malicious sequencer double-minting on it's own turn.

I think we can revisit no-op vs fail & park in a new issue, along with the general sequencer overhauling.

PR #643, opened 2026-07-24
feat(integration_tests): adding multi-sequencer support for integration tests

Pravdyvy:
## 🎯 Purpose

Now, we have ability to startup multiple sequencers and zones, therefore it is possible to represent it in integration tests.

## ⚙️ Approach

- [x] Modification of `TestContext` to fit multiple sequencers.
- [x] Making wallet optional for cases if there is no wallet support available.
- [x] Cross-zone integration tests migration. 
- [x] Multi-sequencer integration tests migration.

## 🧪 How to Test

Refactor of an old tests. CI must pass.

## 🔗 Dependencies

None

## 🔜 Future Work

- When possible, write a test, which will start both multiple zones and multiple sequencers in zone.
- Add wallet support for multi-zones.
- Write multi sequencer client tests. Would requite mocked clients for completion.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


.github/workflows/ci.yml
```diff
@@ -306,12 +320,19 @@ jobs:
...
 
-      - name: Test valid proof
+      - name: Test valid proof (RETRY TEST)
         uses: ./.github/actions/run-in-ci-image
         with:
           image: ${{ needs.ci-image.outputs.image }}
           env: RUST_LOG=info
-          run: cargo test -p integration_tests -- --exact private::private_transfer_to_owned_account
+          run: |
+            for i in 1 2 3; do
```
erhant: this shape is pretty common, you think we can use a common .sh (or a workflow helper) for it? e.g. it would look like `with_retry $some_command [$max_attempts (default: 3)]`

test_fixtures/src/config.rs
```diff
@@ -77,6 +79,28 @@ impl std::fmt::Display for UrlProtocol {
     }
 }
 
+#[derive(Debug, Clone, Copy)]
+/// Config for test context in multi-node case.
+///
+/// For now have only one field: `num_nodes`.
```
erhant: comment no longer correct with the addition of `bedrock_channel` field

test_fixtures/src/config.rs
```diff
@@ -268,6 +299,32 @@ pub fn bedrock_channel_id_b() -> ChannelId {
...
+    root.to_le_bytes()
+        .repeat(8)
+        .try_into()
+        .unwrap_or_else(|_| unreachable!())
+}
+
+#[must_use]
+#[expect(
+    clippy::big_endian_bytes,
+    reason = "Make channels and bedrock keys different bytewise"
```
erhant: cant we have a simpler way of doing this maybe? without thinking much on it, I wonder if XOR works for instance

- signing key: `(root) * 8`
- bedrock channel id: `(0xFF..FF ^ root) * 8`

also I think I have seen some testing utility like this in other code where it was called `seed: u32` instead, up for your preference but I think that naming fits better for the context

test_fixtures/src/lib.rs
```diff
@@ -152,82 +290,174 @@ impl TestContext {
...
+            .wallet
             .storage()
             .key_chain()
             .public_account_ids()
             .map(|(account_id, _idx)| account_id)
             .collect()
     }
 
-    /// Get existing private account IDs in the wallet.
+    /// Get default(1 zone) existing private account IDs in the wallet.
```
erhant:
```diff
+ default (1 zone)
```

test_fixtures/src/lib.rs
```diff
@@ -70,56 +81,181 @@ pub struct DiskSizes {
...
+        Self::builder(vec![MultiNodeTestContextConfig::default()])
+            .build()
+            .await
     }
 
     /// Get a builder for the test context to customize its configuration.
     #[must_use]
-    pub fn builder() -> TestContextBuilder {
-        TestContextBuilder::new()
+    pub fn builder(configs: Vec<MultiNodeTestContextConfig>) -> MultiZoneTestContextBuilder {
```
erhant: isn't this technically anti-pattern for builder? we could init with empty vec and then use `with_configs(...)` maybe

lez/sequencer/service/src/lib.rs
```diff
@@ -296,6 +296,7 @@ async fn main_loop(seq_core: Arc<Mutex<SequencerCore>>, block_timeout: Duration)
 
         info!("Our turn: collecting transactions from mempool, creating block");
         let id = state.produce_new_block().await?;
-        info!("Block with id {id} created");
+        let author_identity = hex::encode(state.sequencer_config().signing_key);
+        info!("Block with id {id} created by {author_identity:?}");
```
erhant: one nit (for any code now or later :D) I think we can use `log::info` instead, we had done it elsewhere (and the code has both usages)

erhant (approved):
LGTM, few comments

.github/workflows/ci.yml
Pravdyvy: Added a script

test_fixtures/src/config.rs
Pravdyvy: Fixed

test_fixtures/src/config.rs
Pravdyvy: Resolved it through substraction
now bedrock channel id is `u32::MAX.saturating_sub(seed)`.
Looks fine for you?

test_fixtures/src/lib.rs
Pravdyvy: Fixed

test_fixtures/src/config.rs
erhant: yesss

test_fixtures/src/lib.rs
Pravdyvy: Maybe, but it is very useful to think about topology of a system you want to setup beforehand.

lez/sequencer/service/src/lib.rs
Pravdyvy: Why not, updated it everywhere.

test_fixtures/src/lib.rs
erhant: Im fine as is, leaving to you 🙏🏻

lez/explorer_service/src/api.rs
```diff
@@ -151,9 +151,8 @@ pub async fn get_transactions_by_account(
 #[cfg(feature = "ssr")]
 pub fn create_indexer_rpc_client(url: &url::Url) -> Result<IndexerRpcClient, String> {
     use jsonrpsee::http_client::HttpClientBuilder;
-    use log::info;
 
-    info!("Connecting to Indexer RPC on URL: {url}");
+    log::info!("Connecting to Indexer RPC on URL: {url}");
```
Arjentix, line 155: I don't get why we have these changes in this PR

test_fixtures/src/lib.rs
Arjentix:
```suggestion
    pub fn builder(configs: impl IntoIterator<Item = MultiNodeTestContextConfig>) -> MultiZoneTestContextBuilder {
```

Since you'll iterate over it anyway this will let user pass `[]` as well

test_fixtures/src/lib.rs
```diff
@@ -70,56 +81,181 @@ pub struct DiskSizes {
...
+        channel_id: ChannelId,
+    ) -> Option<impl Iterator<Item = &SequencerComponents>> {
+        self.zones
+            .get(&channel_id)
+            .map(|zone| zone.sequencers.iter())
+    }
+
+    #[must_use]
+    /// Reference for the default sequencer component for a zone (in case, if only one sequencer
+    /// exists).
```
Arjentix:
```suggestion
    /// Reference for the default sequencer component for a zone (in case, if only one sequencer
    /// exists).
    #[must_use]
```

test_fixtures/src/lib.rs
```diff
@@ -70,56 +81,181 @@ pub struct DiskSizes {
...
+            .iter_mut()
+            .next()
+            .expect("Must be at least one integration component")
+    }
+
+    /// Get reference to the default wallet.
+    #[must_use]
+    pub fn wallet(&self) -> &WalletCore {
+        &self.default_zone().wallet.as_ref().unwrap().wallet
+    }
```
Arjentix, line 191: Maybe let's better have just `default_zone()` and `default_zone_mut()` and create getters for `TestContextZone` instead of putting every getter here.

test_fixtures/src/lib.rs
```diff
@@ -357,67 +599,252 @@ impl TestContextBuilder {
...
+
+        let mut sequencer_keys = vec![];
+
+        sequencer_keys.push(config::SEQUENCER_SIGNING_KEY);
+
+        for i in 1..mn_config.num_nodes {
+            sequencer_keys.push(config::sequencer_signing_key_from_seed(
+                u32::try_from(i).expect("Not being able to fit is realistically impossible"),
+            ));
         }
```
Arjentix: Let's do it with collect instead

test_fixtures/tests/prebuilt_fixture.rs
```diff
@@ -7,14 +7,20 @@ use lee::{AccountId, PublicKey};
...
 };
 
 /// Builds from genesis (no prebuilt database) and checks the on-chain state follows the config.
 #[tokio::test]
 async fn genesis_from_scratch_follows_config() -> Result<()> {
-    let ctx = TestContext::builder().from_scratch().build().await?;
+    let ctx = TestContext::builder(vec![MultiNodeTestContextConfig::default()]);
+    let default_channel_id = ctx.default_channel_id();
+
+    let ctx = ctx.from_scratch(default_channel_id).build().await?;
```
Arjentix: This pattern looks strange to me. We use default everywhere, but still no to first use getter from builder to continue building...

test_fixtures/src/config.rs
```diff
@@ -268,6 +297,31 @@ pub fn bedrock_channel_id_b() -> ChannelId {
     ChannelId::from(channel_id)
 }
 
+#[must_use]
+/// Generate sequencer signing key from `u32` number via repeating le bytes 8 times.
```
Arjentix:
```suggestion
/// Generate sequencer signing key from `u32` number via repeating le bytes 8 times.
#[must_use]
```

test_fixtures/src/config.rs
```diff
@@ -268,6 +297,31 @@ pub fn bedrock_channel_id_b() -> ChannelId {
...
+        .repeat(8)
+        .try_into()
+        .unwrap_or_else(|_| unreachable!())
+}
+
+#[must_use]
+/// Generate bedrock channel id from `u32` number via repeating le bytes 8 times.
+///
+/// Counting from the end of `u32` to guarantee, that it is different from
+/// `sequencer_signing_key_from_seed`.
```
Arjentix:
```suggestion
/// Generate bedrock channel id from `u32` number via repeating le bytes 8 times.
///
/// Counting from the end of `u32` to guarantee, that it is different from
/// `sequencer_signing_key_from_seed`.
#[must_use]
```

integration_tests/tests/cross_zone_bridge.rs
```diff
@@ -72,37 +67,45 @@ async fn lock_on_zone_a_mints_wrapped_token_on_zone_b() -> Result<()> {
...
+    .disable_wallet(channel_a)
+    .disable_wallet(channel_b)
+    .disable_indexer(channel_a)
+    .with_sequencer_partial_config(channel_a, partial)
+    .with_sequencer_partial_config(channel_b, partial)
+    .with_genesis(channel_a, genesis_a)
+    .with_genesis(channel_b, vec![])
+    .with_cross_zone(channel_b, cross_zone)
+    .build()
+    .await?;
```
Arjentix: I'm trying to think about overall approach... I think something like this fits better:

```rust
Topology::builder()
    .with_zone(
        Zone::builder(channel_a)
            .with_sequeencer(
                Sequencer::builder()
                    .with_partial_config(partial)
            )
            .disable_wallet()
    )
```

So that we don't expand currently existing `TestContext` with new functionality. Instead we make it a builder for a zone and create topology builder on top moving bedrock there.

Arjentix (changes requested):
I suggest to revisit the look of the builder. I think it's possible to set it up in more "one-shot style" instead of reconfiguring every field with something like `.disable_wallet(channel_a)`

lez/explorer_service/src/api.rs
Arjentix: Maybe this is something from Erhan's comments. But I'd consider removing this commit at all, here is my reasoning: [link](https://github.com/logos-blockchain/logos-execution-zone/pull/691#discussion_r3736304256)

test_fixtures/src/lib.rs
Pravdyvy: Done with collect.

lez/explorer_service/src/api.rs
Pravdyvy: First of all, For me there is really no difference, if we use fully qualified syntax for extern crates calls, or not.

Secondly, apparently, `info!` is not the only one `log` method that we use in this two ways, we use all methods from there differently. 

Your reasoning, I think, does not connect with the problem, transition from `info!` -> `log::info!` is just a simple find-replace operation in every text redactor(inverse also true).
In transition into tracing we will need to rewrite log messages, not calls. Fully qualified suntax does not affect it anyhow, again, because we will just find-replace `log::info!` with `tracing::info!` everywhere.

test_fixtures/src/lib.rs
Pravdyvy: With this default functions we could continue use default `TestContext` setup like before. With those, most of our functions does not change.

lez/explorer_service/src/api.rs
erhant: It really boils down to a matter of preference, but in the wild I had mostly seen the fully qualified usage; and the text-replacement is as straightforward as @Pravdyvy says.

I wondered and checked 3 random repos from our Cargo.toml:

- libp2p uses fully-qualified `tracing::` logs ([ref](https://github.com/search?q=repo%3Alibp2p%2Frust-libp2p+debug%21&type=code))
- kameo uses both fully-qualified `tracing::` logs ([ref](https://github.com/search?q=repo%3Atqwewe%2Fkameo+debug%21&type=code)) and imported logs ([ref](https://github.com/search?q=repo%3Atqwewe%2Fkameo+info%21&type=code))
- reqwest uses fully-qualified logs ([ref](https://github.com/search?q=repo%3Aseanmonstar%2Freqwest+debug%21&type=code))

integration_tests/tests/cross_zone_bridge.rs
Pravdyvy: Updated in this fashion, naming is different, used existing infrastructure. Take a look.

test_fixtures/tests/prebuilt_fixture.rs
Pravdyvy: Approach changed to look more like in the last thread, we can continue here.

lez/sequencer/service/src/lib.rs
```diff
@@ -307,8 +307,9 @@ async fn main_loop(seq_core: Arc<Mutex<SequencerCore>>, block_timeout: Duration)
             continue;
         }
 
-        info!("Our turn: collecting transactions from mempool, creating block");
+        log::info!("Our turn: collecting transactions from mempool, creating block");
         let id = state.produce_new_block().await?;
-        info!("Block with id {id} created");
+        let author_identity = hex::encode(state.sequencer_config().signing_key);
```
moudyellaz: Doesn't this log the sequencer's private block-signing key at info on every block?

lez/explorer_service/src/api.rs
Arjentix: At least all of them use tracing, hah. Okay, agreed

integration_tests/tests/multi_sequencer.rs
```diff
@@ -34,117 +30,114 @@ const ROTATION_BLOCKS: u64 = 8;
...
-        .await
-        .context("Failed to set up sequencer B")?;
-    let b = sequencer_client(seq_b.addr())?;
+    log::info!(
+        "Passed wait for height A to be at least {}",
+        height_at_config + 1
+    );
+
+    let join_height = seq_client_a.get_last_block_id().await?;
+    wait_for_height(seq_client_b, join_height, "B to sync to A's height at join").await?;
```
moudyellaz, line 77: Does B still join late here, or is it up from the start now?

test_fixtures/src/lib.rs
```diff
@@ -70,56 +81,164 @@ pub struct DiskSizes {
...
+    #[must_use]
+    pub fn sequencer_client(&self) -> &SequencerClient {
+        &self.default_sequencer_component().sequencer_client
     }
 
     /// Get reference to the sequencer client.
     #[must_use]
-    pub const fn sequencer_client(&self) -> &SequencerClient {
-        &self.sequencer_client
+    pub fn sequencer_client_getter(
```
Arjentix:
```suggestion
    pub fn default_sequencer_client(&self) -> &SequencerClient {
        &self.default_sequencer_component().sequencer_client
}

/// Get reference to the sequencer client.
   #[must_use]
    pub fn sequencer_client(
```

Maybe like that so that we don't have to use word "getter"?

Arjentix (approved):
LGTM, thanks

test_fixtures/src/lib.rs
```diff
@@ -357,67 +607,176 @@ impl TestContextBuilder {
...
 
-            setup_private_accounts_with_initial_supply(&mut wallet, &initial_private_accounts)
-                .await
-                .context("Failed to initialize private accounts in wallet")?;
+        log::info!("Passed wait untill genesis");
+
+        sequencer_addrs.push(leader_addr);
+        sequencer_components.push(leader_components);
+
+        post_chain_config_with_default_parameters(
```
moudyellaz: Should the channel-config post be skipped when num_nodes == 1?

test_fixtures/src/lib.rs
```diff
@@ -152,82 +273,174 @@ impl TestContext {
...
+                }))
+            }),
+            indexer_bytes: self.zones.values().fold(0, |acc, zone| {
+                acc.saturating_add(
+                    zone.indexer
+                        .as_ref()
+                        .map_or(0, |val| dir_size_bytes(val.temp_dir.path())),
+                )
+            }),
+            wallet_bytes: self.zones.values().fold(0, |acc, zone| {
```
moudyellaz, line 371: Should this be map_or(0, ...) like indexer_bytes above, now that the wallet is optional?

Pravdyvy: True, fixed.

test_fixtures/src/lib.rs
Pravdyvy: Yes, for one node should work without it, skipped if just one node present.

test_fixtures/src/lib.rs
Pravdyvy: We use symbol `sequencer_client` in a lot of places, I do not want to break convention. In this case replaced `getter` with `by_node_ids`. Replaced other uses of getter by `zone`.

lez/sequencer/service/src/lib.rs
Pravdyvy: True, replaced with public key.

integration_tests/tests/multi_sequencer.rs
Pravdyvy: Still late, during setup we just for at least A genesis to be posted and only after that we post chain config, so there can be delay.

test_fixtures/src/lib.rs
```diff
@@ -70,56 +81,164 @@ pub struct DiskSizes {
...
+            .with_zone(ZoneTestContextBuilder::new(
+                MultiNodeTestContextConfig::default(),
+            ))
+            .build()
+            .await
+    }
+
+    /// Reference for the default zone(in case if only one present).
+    #[must_use]
+    pub fn default_zone(&self) -> &TestContextZone {
```
moudyellaz, line 128: Maybe assert zones.len() == 1 here so a multi-zone context can't silently get an arbitrary zone?

test_fixtures/src/lib.rs
```diff
@@ -357,67 +609,179 @@ impl TestContextBuilder {
...
             bedrock_addr,
-            sequencer_handle: Some(sequencer_handle),
-            indexer_components,
-            temp_sequencer_dir,
-            temp_wallet_dir,
         })
     }
 
+    #[must_use]
+    pub fn with_zone(mut self, zone_builder: ZoneTestContextBuilder) -> Self {
```
moudyellaz, line 799: Should a duplicate channel id be rejected here instead of replacing the existing zone?

test_fixtures/src/config.rs
```diff
@@ -268,6 +297,31 @@ pub fn bedrock_channel_id_b() -> ChannelId {
...
+        .try_into()
+        .unwrap_or_else(|_| unreachable!())
+}
+
+/// Generate bedrock channel id from `u32` number via repeating le bytes 8 times.
+///
+/// Counting from the end of `u32` to guarantee, that it is different from
+/// `sequencer_signing_key_from_seed`.
+#[must_use]
+pub fn bedrock_channel_id_from_seed(seed: u32) -> ChannelId {
```
moudyellaz, line 312: Maybe drop this since nothing calls it?

test_fixtures/src/lib.rs
```diff
@@ -547,3 +976,121 @@ fn dir_size_bytes(path: &Path) -> u64 {
...
+        POSTING_TIMEFRAME_SLOTS,
+        POSTING_TIMEOUT_SLOTS,
+        1,
+        1,
+    )
+    .await
+    .context("Failed to configure the channel committee")
+}
+
+async fn wait_untill_genesis(client: &SequencerClient) -> Result<()> {
```
moudyellaz: Typos worth a pass: wait_untill_genesis, "singulat" (114), "sequecner" (861), "comitee" (986), "thanreplace" (1056).

test_fixtures/src/config.rs
```diff
@@ -12,6 +12,8 @@ use sequencer_core::config::{BedrockConfig, CrossZoneConfig, GenesisAction, Sequ
 use url::Url;
 use wallet::config::{MultiSequencerClientConfig, SequencerConnectionData, WalletConfig};
 
+use crate::config;
```
moudyellaz: Maybe drop this self-import since it's inside config.rs already?

Pravdyvy: Dropped

test_fixtures/src/lib.rs
Pravdyvy: Fixed

test_fixtures/src/config.rs
Pravdyvy: Let's leave, will be useful in the future. For cases with 3 and more zones.

test_fixtures/src/lib.rs
Pravdyvy: Added assert.

test_fixtures/src/lib.rs
Pravdyvy: Added assert and corrsponding comments in all depending docs.

moudyellaz (approved):

Issue #648, opened 2026-07-28
Cross-zone peer block cache accepts unvalidated peer-asserted blocks, allowing a remote indexer halt


moudyellaz:
`read_peer` / `consume_peer_stream` in `lez/indexer/core/src/cross_zone_verifier.rs` cache a peer's finalized blocks with no validation of what the peer asserts about its own chain. Three consequences, the first of which is a live remote brick:

Last-write-wins overwrite. `PeerBlocks::insert` keys on `block.header.block_id` taken from the decoded block and overwrites any existing entry. An equivocating peer that inscribes a second block claiming an id it already used replaces the real block in the cache. The next legitimate dispatch naming that id re-derives against the wrong block, mismatches, and is rejected as forged, so the ingest loop halts permanently. One inscription, remote halt.

No pinned-key check on insert. The sequencer's watcher drops peer blocks not signed by the configured key before use (`lez/sequencer/core/src/cross_zone_watcher.rs`), and the verifier applies the same check in `rederive`, but only to the block a dispatch actually names. The reader applies no check at all, so anything inscribed on the peer channel enters the cache.

Hash links are self-asserted. `header.hash` is not recomputed on decode, so the `prev_block_hash == prev.header.hash` test that advances `PeerChain::verified_prefix` compares two fields the peer wrote. The verified prefix therefore bounds how far this reader has read; it does not establish that the chain is authentic. This is currently inert, since the prefix only gates a branch that is unreachable while the cache is unbounded, but it will not stay inert.

Also in scope: `verified_prefix` never retreats. A peer chain reset (supported via `allow_chain_reset`) restarts ids at 1 and overwrites cached entries last-write-wins while the prefix still certifies the old chain, so a dispatch can re-derive against a not-yet-overwritten old-chain block and halt ingestion the same way.

Suggested direction: apply the pinned-key check in the reader before inserting, reject an insert that would overwrite a differing block at an existing id rather than replacing it, and decide explicitly what a chain reset should do to the cache and the prefix.

All of this predates the verified-prefix change; that change neither introduced nor fixed it.

Issue #647, opened 2026-07-28
Cross-zone verifier cannot distinguish a permanently unavailable peer block from a lagging peer


moudyellaz:
The indexer's cross-zone verifier resolves a dispatch's source block in `wait_for_peer_block` (`lez/indexer/core/src/cross_zone_verifier.rs`). Anything it cannot place inside the run verified from the peer's genesis waits and then returns `PeerUnavailable`, which the ingest loop treats as retriable: it holds its cursor and retries indefinitely.

That is correct and fail-closed, the block is never applied, but one genuinely permanent condition is still reported as ordinary lag: a dispatch referencing a block the peer will never produce, a far-future or fabricated id. Deciding this needs the peer's actual tip, which neither the previous `max(cached block ids)` high-water mark nor the current verified prefix provides. Read the peer channel's tip at LIB and reject a reference above it once the reader is demonstrably caught up. The indexer tip freezes on the first local block carrying such a dispatch, with a message suggesting the peer is merely behind.

Context: the previous `max(cached ids)` mark did reject this, but unsoundly. A peer picks its own block ids, so one inscribed block claiming `u64::MAX` drove the mark past every real id and made every later dispatch look forged, halting ingestion permanently. That is why the mark was replaced by a hash-linked verified prefix.

Two corrections after #670 and #686.

The skipped-slot half of this issue no longer exists. Neither side reads past a slot it cannot decode: `consume_peer_stream` holds the cursor at the last fully consumed slot, and the watcher returns `PassOutcome::Undecodable` and holds its floor. `DECODE_RETRY_LIMIT` is gone from both. So `wait_for_peer_block` reporting "the reader has not reached it" is now accurate rather than false, and there is nothing left to record or surface. What replaced it is a different and larger problem: one undecodable slot stalls that peer's whole channel indefinitely.

The forged signal is reachable, so the claim that it "cannot occur while the peer block cache is unbounded" is wrong. `PeerBlocks::insert` refuses any id below `GENESIS_BLOCK_ID`, which is 1, so a dispatch naming `src_block_id` 0 resolves as inside the verified run and absent, which is the forgery signal, and that terminates the ingest task and survives restart. Every id inside the run is still cached, so this is the only route to it.

[...]

PR #649, opened 2026-07-28
fix(cross-zone): gate the forgery test on a verified peer-chain prefix

moudyellaz:
## 🎯 Purpose

Fixes a live bug that permanently bricks indexer ingestion. The verifier decided whether a cross-zone dispatch referenced a real peer block by comparing against the highest block id it had cached. That mark was not a sound signal for "the reader has read this far".

A peer block that failed to deserialize was skipped while the cursor advanced, so later blocks were cached and the mark climbed over an id that was never read. A legitimate dispatch referencing that id was then judged forged and the indexer halted for good. The same mark could also be driven past every real id by a single peer block claiming a huge `block_id`, since a peer picks its own ids.

## ⚙️ Approach

- [x] Track a verified prefix per peer: the highest id such that every block from genesis up to it has been read and hash-links to its predecessor. The forgery test gates on that instead of the highest cached id, so an id that does not continue the run cannot affect it.
- [x] On a decode failure, hold the cursor and re-read the slot so a transient failure heals, then read past it after a few attempts so a permanently undecodable inscription cannot wedge the reader. Skipping only leaves a hole, and a hole cannot advance the verified prefix past itself.
- [x] Resolve a referenced peer block under a single read lock. Answering "cached?" and "inside the verified run?" separately races with the peer reader: an insert landing between the two reads looks like absent-and-inside-the-run, which is the forgery signal, for a block that is in fact cached.
- [x] Split verification failures: `verify_block` returns `Forged` (terminal, halt as before) or `PeerUnavailable` (the reader is behind, hold the cursor and retry). `wait_for_peer_block` gained a bounded wait to produce the latter instead of blocking ingestion forever.
- [x] Advance the read cursor only on a slot boundary. One L1 slot can carry several L2 blocks and the channel stream resumes after the stored slot, so advancing as each block was handled stranded every later block in the same slot whenever a pass ended early: the `PeerUnavailable` retry above, and also `park_undeserializable` and both `Parked` arms. Extracted as `SlotProgress` with unit tests.

## 🧪 How to Test

```
cargo test -p indexer_core
cargo test --release --test cross_zone_verified   # also cross_zone_bridge, cross_zone_ping
```

## 🔗 Dependencies

None.

## 🔜 Future Work

A reference to a block the peer will never produce now stalls rather than being rejected, since a destination cannot distinguish that from a peer that has not produced it yet. Rejecting it needs the peer's tip read at LIB. Tracked in #647.

The peer block cache accepts what a peer asserts about its own chain, which predates this change. Tracked in #648.

`cross_zone_watcher.rs` has the same skip-on-decode-failure shape. It self-heals today because its cursor is in memory and resets on restart, so it is fixed alongside the change that makes that cursor durable.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests

### DIFF
--- a/lez/indexer/core/src/lib.rs
+++ b/lez/indexer/core/src/lib.rs
@@ -29,6 +29,16 @@ pub mod status;
 /// Consecutive failed apply attempts of the same block before parking.
 const APPLY_RETRY_LIMIT: u32 = 3;
 
+/// Which slot the ingest loop is currently inside, so the read cursor only ever
+/// moves on a slot boundary.
+///
+/// One L1 slot can carry several L2 blocks, and the channel stream resumes
+/// *after* the stored slot. Advancing the cursor as each block is handled would
+/// therefore put a later block in the same slot beyond the cursor whenever a
+/// pass ends early, and nothing would ever read it again.
+#[derive(Default)]
+struct SlotProgress(Option<Slot>);
+
 #[derive(Clone)]
 pub struct IndexerCore {
     pub zone_indexer: Arc<ZoneIndexer<NodeHttpClient>>,
@@ -40,6 +50,23 @@ pub struct IndexerCore {
     pub status: Arc<ArcSwap<IndexerSyncStatus>>,
     /// Option B cross-zone verifier; `None` when cross-zone messaging is disabled.
     pub verifier: Option<CrossZoneVerifier>,
+}
+
+impl SlotProgress {
+    /// Records that a message from `slot` is being handled, returning the slot
+    /// that just completed, if this message begins a new one.
+    fn enter(&mut self, slot: Slot) -> Option<Slot> {
+        if self.0 == Some(slot) {
+            return None;
+        }
+        self.0.replace(slot)
+    }
+
+    /// The slot in progress when the stream drained cleanly, which is therefore
+    /// complete. Not called when a pass ends early: that slot must be re-read.
+    const fn drained(self) -> Option<Slot> {
+        self.0
+    }
 }
 
 impl IndexerCore {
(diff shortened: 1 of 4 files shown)


agureev (approved):
looks good from the overall semantic perspective!

lez/indexer/core/src/lib.rs
erhant: is this Option B outdated comment maybe? if not, are there comments on what were the other options?

erhant: I noticed a bug that is outside this PR, but affects it; lets say we have two blocks `A` and `B` inscribed in the same L1 block; we process `A` and it is applied, advancing the cursor;

now `B` arrives and lets say we got `CrossZoneVerifyError::PeerUnavailable` so it will retry

however, we had advanced the cursor while processing `A`, so `B` is lost beyond our advanced cursor now!

potential fix is to maybe reset the cursor on `had_cycle_error` to the current position for retrial?

erhant (approved):
LGTM! found an issue that is older than this PR but affects it 🤔

lez/indexer/core/src/lib.rs
erhant: one question to ask is how likely is it to have >1 L2 blocks in a single L1 block?

moudyellaz: Fixed by never advancing inside a slot: the cursor moves when a message from a later slot arrives, or when the stream drains. That is the same rule the sequencer's watcher uses, for exactly this reason.

moudyellaz: Not exotic. It is guaranteed whenever an L1 slot spans more than one L2 block, and the sequencer side already treats it as the normal case: the watcher has a test called `watcher_does_not_persist_inside_a_partially_failed_slot` covering the identical interleaving. Resubmission can also put several inscriptions in one slot.

moudyellaz: "Option B" names the design locked when we were discussing the initial approach: the destination's indexer independently re-derives each dispatch from the peer's finalized blocks and checks it, so delivery does not require trusting the source sequencer. I reworded it to state what the code does and why.

erhant (approved):
LGTM, ty for addressing comments!

PR #650, opened 2026-07-28
refactor!: unify the public and private execution traversals behind a Backend trait

agureev:
## 🎯 Purpose

The public state machine and the privacy preserving circuit walked the same call tree with two hand-maintained copies of the same traversal. A dozen checks were duplicated and the divergence was enforced only by comments and tests. This collapses them into one traversal in `lee_core`, behind a `Backend` trait.

## ⚙️ Approach

`lee_core::validation::validate_state_diff` owns the walk. Three things genuinely differ between environments, and they are the trait:

- `output_for_call`, execute the program versus verify a proof of it
- `expected_first_sight`, an account's authoritative value, or `None` where there is no chain state
- `judge_authorization`, the authorization rule and the journal view it exports

plus `observe_windows`, `observe_events` and `finish`. `PublicBackend` executes under the cycle budget against `V03State`; `PrivateBackend` verifies the prover's output and owns the private-PDA binding machinery. `execution_state.rs` is deleted.

The circuit's error type is an uninhabited `enum Fatal {}`, so no `Err` can be constructed in-guest and a rejection can only be the panic in `From<ValidationError> for Fatal`.

One behaviour change: the `instruction_data` echo was checked only by the circuit and now applies to both. Without it a chained call to `transfer(1)` could be answered by a genuine receipt for `transfer(1_000_000)`.

Two orderings are load-bearing. The subtree's authorized set is extended from the journalled `is_authorized`, captured before `judge_authorization` may export a different value, so masking a journal cannot widen what a program was allowed to do. And first sight is keyed on the position map rather than the touched map, so an output naming one account twice cannot claim two witness positions.

## 🧪 How to Test

`RISC0_DEV_MODE=1 cargo nextest run --workspace --exclude integration_tests`. Behaviour is pinned by the existing suites, not by new tests. `lee_core::validation::tests` adds a hook-order test, since the check ordering is what the public error-variant assertions depend on.

## 🔜 Future Work

The prover-side mirror in `privacy_preserving_transaction/circuit/mod.rs` still has its own partial traversal and fits the trait as a third backend.

Holding the top-level call to the same `pre_states_match_accounts` as every other call was implemented and measured, then left out because it changes what a program may journal. It costs one rule for program authors: a top-level program must echo the accounts it was handed, including ones it only forwards. No shipped program violates that today. It would also make `root_output_is_confined` and the per-account containment check redundant, and make `account_identities` positions the declared order rather than execution-dependent first-sight order.

## 📋 Notes

Touching `lee_core` rebuilds every guest artifact, so image ids move, which changes the builtin genesis account ids, `genesis_fingerprint` and the committed fixture. Requires a testnet reset. Circuit ELF grows about 15 KB against the 700 KiB ceiling.


lee/state_machine/core/src/error.rs
```diff
@@ -10,3 +17,83 @@ pub enum LeeCoreError {
     #[error("IO error: {0}")]
     Io(#[from] io::Error),
 }
+
+/// Ways a program can violate the execution rules the environment enforces on it.
+///
+/// Lives here rather than in `lee` so the shared traversal in [`crate::validation`] can reject
+/// on it from either environment: the public state machine surfaces it as
+/// `LeeError::InvalidProgramBehavior`, the privacy preserving circuit panics on it in-guest.
```
jonesmarvin8: Can we remove the second paragraph of this docstring? I don't think it's informative to know what the code state used to be when reading the current version (I'm sure there's exceptions).

lee/privacy_preserving_circuit/src/main.rs
jonesmarvin8, line 17: Both `dummy_inputs` and `ciphertext_padding` are used to obfuscate the circuit's private account outputs. It appears these are usually passed together, so it might make sense to introduce a struct that bundles these together.

lee/privacy_preserving_circuit/src/main.rs
```diff
@@ -15,16 +19,43 @@ fn main() {
         program_image_claims,
     } = borsh::from_slice(&read_input_frame()).expect("circuit input must be valid borsh");
 
-    let execution_state = execution_state::ExecutionState::derive_from_outputs(
-        &account_identities,
+    let Some(first_output) = program_outputs.first() else {
+        panic!("No program outputs provided");
```
jonesmarvin8: Should this provide context to the file that panicked?

jonesmarvin8 (approved):
Lgtm. A couple minor comments.

lee/state_machine/core/src/error.rs
moudyellaz: Done.

lee/privacy_preserving_circuit/src/main.rs
moudyellaz: Message now names the offending input: `PrivacyPreservingCircuitInput::program_outputs is empty: nothing to validate`.

lee/privacy_preserving_circuit/src/main.rs
moudyellaz: I would hold off here, because `ciphertext_padding` came in with #630, and bundling them changes `PrivacyPreservingCircuitInput`, which is the prover-to-guest wire format, so it would widen a behaviour-preserving refactor into a format change on someone else's field.

schouhy (approved):
lgtm thanks! great refactor

jonesmarvin8 (approved):
lgtm

PR #651, opened 2026-07-28
feat(metrics): add sequencer metrics & grafana dashboard generation tool

Arjentix:
## 🎯 Purpose

This PR introduces metrics to the Sequencer as well as a tool to generate Grafana dashboards with code-first approach.

## ⚙️ Approach

- Add two `metric` crates: one for `sequencer_core` and one for `sequencer_service`
- Add `tool/dashboard_gen` which generates Grafana dashboard from rust specification
- Add `docker compose` and `just` setup to run Prometheus + Grafana setup locally
- Add documentation as `metrics.md`
- Add CI job to ensure grafana dashboard is up to date with rust code

Screenshot (not full):

image

## 🧪 How to Test

- Run the whole setup with `just`
- Open `http://localhost:3000` and go to the `Sequencer` dashboard
- Play around, submit some transactions, see metrics changing

## 🔗 Dependencies

None

## 🔜 Future Work

- Indexer metrics
- `dashboard_gen` tool only supports things we need right now, so on every new non-supported Grafana json syntax we need to expand it

## 📋 PR Completion Checklist

- [x] Complete PR description
~- [ ] Add/update tests~


lez/sequencer/core/metrics/src/names.rs
erhant, line 1: is `names.rs` a standard? I usually see `constants.rs` for this kind of thing, up to you of course 🙏🏻

lez/sequencer/service/src/main.rs
```diff
@@ -65,6 +68,36 @@ async fn main() -> Result<()> {
...
+    /// Fallback ladder for histograms that count things rather than measure time.
+    const COUNT_BUCKETS: &[f64] = &[1.0, 2.0, 5.0, 10.0, 25.0, 50.0, 100.0, 250.0, 500.0, 1000.0];
+
+    PrometheusBuilder::new()
+        .with_recommended_naming(true)
+        .set_buckets(COUNT_BUCKETS)
+        .context("Failed to set default histogram buckets")?
+        .set_buckets_for_metric(Matcher::Suffix("_seconds".to_owned()), LATENCY_BUCKETS)
+        .context("Failed to set latency histogram buckets")?
+        .install()
```
erhant, line 111: where does this bind to? would it fail if we run multiple sequencers, due to port re-use?

lez/common/src/transaction.rs
```diff
@@ -200,7 +209,17 @@ impl From<lee::ProgramDeploymentTransaction> for LeeTransaction {
...
+    Debug,
+    Clone,
+    Copy,
+    PartialEq,
+    Eq,
+    Serialize,
+    Deserialize,
+    BorshSerialize,
+    BorshDeserialize,
+    strum::IntoStaticStr,
```
erhant: nit: prometheus uses `snake_case` [convention](https://prometheus.io/docs/practices/naming/) for metric names usually, maybe we could add `#[strum(serialize_all = "snake_case")]` to our enums in favor of that?

lez/sequencer/service/metrics/src/lib.rs
```diff
@@ -0,0 +1,9 @@
+//! This crate provides all metrics exposed by the sequencer core crate.
```
erhant:
```diff
- core
+ service
```

docker-compose.yml
erhant, line 1: btw we should add `9000:9000` to our other docker-compose's too, so that the in-docker prometheus can be reached from outside (i.e. `lez/sequencer/service/docker-compose.yml`)

lez/common/src/transaction.rs
erhant: oh I see you wrote about this exact convention in metrics.md too, so it could be ok as-is too 🙏🏻

erhant (approved):
LGTM nice work! few small comments

lez/sequencer/service/src/main.rs
Arjentix: Good catch! Indeed it would fail because it binds to `9000` by default. Fixed by adding new optional config field `metrics_address` and a cli option to override it so that we can run different sequencers from the same config but with different metrics ports

lez/common/src/transaction.rs
Arjentix: Ah, indeed, didn't know the same rules applies for labels. Agreed, let's keep it uniformed, applied your suggestion, but created a separate `TxKind` in `sequencer_core_metrics` so that we don't create multiple string representations for `common::transaction::TxKind`.

Same applied to `sequencer_core_metrics::TxOrigin`.

docker-compose.yml
Arjentix: Indeed, fixed

lez/sequencer/core/metrics/src/names.rs
Arjentix: Ah, not a standard but I just wanted to highlight that this module contains only metric names. So not every constant would fit here. So I'd prefer keep `names.rs`

lez/sequencer/core/src/lib.rs
```diff
@@ -861,6 +887,9 @@ impl<BP: BlockPublisherTrait> SequencerCore<BP> {
             now.elapsed().as_secs()
         );
 
+        sequencer_core_metrics::record_block_creation_time(now.elapsed());
+        sequencer_core_metrics::increment_blocks_total();
```
moudyellaz: `increment_blocks_total()` fires in `build_block_from_mempool` before publish, so it counts build attempts, parked blocks included, and misses peer-adopted ones. Since startup seeds it from `chain_height()`, "Chain height" drifts once a second sequencer produces. Could we move it to `record_produced_block`'s `Applied` arm plus the follow path, or rename it `blocks_produced_total`?

tools/dashboard_gen/src/lib.rs
```diff
@@ -0,0 +1,591 @@
...
+        "rate({metric}_sum[{RATE_WINDOW}]) / rate({metric}_count[{RATE_WINDOW}])"
+    ))
+    .legend("avg")
+}
+
+/// A per-minute rate target for a counter metric.
+#[must_use]
+pub fn rate_per_min(metric: &str, legend: &str) -> Target {
+    Target::new(format!("rate({metric}[1m]) * 60")).legend(legend)
+}
```
moudyellaz, line 594: Maybe `rate({metric}[{RATE_WINDOW}]) * 60` instead?

lez/common/src/transaction.rs
moudyellaz: Maybe drop `IntoStaticStr` and `EnumIter` since they are not used?

lez/sequencer/core/src/lib.rs
```diff
@@ -830,8 +846,17 @@ impl<BP: BlockPublisherTrait> SequencerCore<BP> {
                 new_block_timestamp,
                 &mut deposit_event_ids,
                 &mut withdrawals,
-            )? {
+            )?;
+            sequencer_core_metrics::record_mempool_transaction_application_time(
+                origin.into(),
+                tx.kind().into(),
```
moudyellaz: Maybe record this only when `applied`, or add a `status` label, so failed applications do not skew the histogram?

tools/dashboard_gen/src/lib.rs
```diff
@@ -0,0 +1,591 @@
...
+    Target::new(format!(
+        "histogram_quantile(${{{PERCENTILE_VAR}}}, sum by ({grouping}) (rate({metric}_bucket[{RATE_WINDOW}])))"
+    ))
+    .legend(legend)
+}
+
+/// A percentile as its `histogram_quantile` argument, derived without float
+/// math: zero-pad to two digits then drop trailing zeros (50 → `0.5`).
+fn quantile(percentile: u32) -> String {
+    let quantile = format!("0.{percentile:02}");
```
moudyellaz, line 575: `quantile(100)` gives `0.1` and `quantile(0)` gives `0.`, maybe assert `1..=99` like the `default` check above?

moudyellaz (approved):
lgtm, thanks! Left some minor comments/questions.

lez/sequencer/core/src/lib.rs
```diff
@@ -168,6 +177,8 @@ impl<BP: BlockPublisherTrait> SequencerCore<BP> {
     pub async fn start_from_config(
         config: SequencerConfig,
     ) -> (Self, MemPoolHandle<(TransactionOrigin, LeeTransaction)>) {
+        sequencer_core_metrics::init();
```
Pravdyvy, line 210: It looks like we make metrics collection mandatory for every sequencer. For this PR it is fine, but it would be good to make it configurable. Not providing `metrics_address` in config should be a sufficient trigger.

Pravdyvy (approved):
lgtm. One comment, probably a future work if it is not trivial.

lez/sequencer/core/src/lib.rs
Arjentix: Indeed, fixed. Now chain height and produced blocks are different metrics

tools/dashboard_gen/src/lib.rs
Arjentix: This should be more correct, thanks. Fixed

lez/sequencer/core/src/lib.rs
Arjentix: Nice idea, fixed

lez/sequencer/core/src/lib.rs
Arjentix: After some consideration I agree. I now made this field optional, but it's `Some()` by default. So user should explicitly disable metrics by providing `null` in config.

And I implemented this conditional logic in `sequencer/service/main.rs` instead of this file. Because there's no problems in metric recording in libraries until there are no global recorder which is set in service main.

PR #652, opened 2026-07-29
feat(cross-zone): persist each watcher's peer read cursor

moudyellaz:
## 🎯 Purpose

A cross-zone delivery could be lost. The watcher's read position lived only in memory, so a restart re-read the peer channel from genesis, and the dispatch it built lived only in the mempool, so a restart between injecting it and a block carrying it dropped the message with nothing left to recover it from.

## ⚙️ Approach

- [x] Persist a delivery floor per peer channel, so a restart resumes instead of replaying
- [x] Record each delivery in the store rather than injecting it into the mempool
- [x] Drain those records during block production, skipping any the inbox seen shard already holds in the state being built on
- [x] Drop a record once its delivery lands in an irreversible block, keyed by message key
- [x] Give up on a delivery that repeatedly fails to execute, and on one too large for any block
- [x] Wait for background tasks on shutdown, so a restart can reopen the store

## 🧪 How to Test

`cargo test -p integration_tests --test cross_zone_watcher_restart` covers the wiring end to end: zone B delivers a ping, restarts on the same home, and must produce no further inbox transactions, which is what a watcher that lost its cursor would emit.

`cargo test -p sequencer_core --features mock` and `cargo test -p storage` cover the watcher state machine, the drain and its two-state skip, the give-up policy, and settlement on both the follow and reconstruction paths.

## 🔗 Dependencies

None.

## 🔜 Future Work

None.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


lez/sequencer/core/src/block_store.rs
```diff
@@ -239,6 +255,32 @@ pub(crate) fn block_to_transactions_map(block: &Block) -> HashMap<HashType, u64>
...
+pub fn set_cross_zone_peer_floor(
+    dbio: &RocksDBIO,
+    peer_zone: PeerZoneKey,
+    cursor: Slot,
+) -> Result<()> {
+    let bytes =
+        serde_json::to_vec(&cursor).context("Failed to serialize cross-zone peer cursor")?;
+    dbio.put_cross_zone_peer_floor_bytes(peer_zone, &bytes)?;
+    Ok(())
+}
```
Pravdyvy, line 266: This operations look a little redundant, isn't it? `Slot` is basically u64, there is no need for runtime serialization via serde, `Slot::to_le_bytes()` should be enough.

lez/sequencer/core/src/cross_zone_watcher.rs
```diff
@@ -1,34 +1,84 @@
...
+    self_zone: [u8; 32],
+    allowed_targets: Vec<ProgramId>,
+    expected_pubkey: Option<PublicKey>,
+}
+
+/// The outcome of one pass over a peer's message stream.
+struct WatcherPass {
+    /// Set when the pass ended early on a message that would not decode.
+    stalled_at: Option<Slot>,
+}
```
Pravdyvy: This struct carries very little additional information. Will it be extended in future? Right now just `Option<Slot>` probably enough?

lez/sequencer/core/src/cross_zone_watcher.rs
```diff
@@ -179,7 +428,403 @@ async fn deliver_block(
...
+        }
+    }
+
+    /// A pass reading past `slot`, which freezes the floor.
+    fn skipping(slot: u64) -> SkipPolicy {
+        SkipPolicy {
+            skip_slot: Some(Slot::from(slot)),
+            persist_floor: false,
+        }
+    }
```
Pravdyvy: This functions live in tests, but it is representable to how `SkipPolicy` is used in the `watch_peer` function. It looks like can be replaced with enum, something like
```
pub enum SkipPolicy {
    DeliverAll,
    SkipSlot(u64),
}
```
or just `Option<Slot>` but it is not really informative.

lez/storage/src/sequencer/mod.rs
```diff
@@ -97,8 +108,22 @@ impl DbDump {
...
+    /// Serializes the read-modify-write cycles over the pending-record lists.
+    ///
+    /// Each list is a single value holding the whole `Vec`, so every add, mark,
+    /// reset and prune reads it, edits it and writes it back. Those callers run
+    /// on different tasks (a watcher, the production loop, the publisher's drive
+    /// task) against clones of one `Arc<RocksDBIO>`, and rocksdb only makes the
+    /// write atomic, not the cycle. Without this lock the writer that read first
+    /// silently drops the other's record, which loses a delivery outright now
+    /// that pruning is one of the writers.
+    pending_records: Mutex<()>,
```
Pravdyvy: I really do not like this approach, however I understand that it can be necessary.
In this exact case we can live without it, but it would require some refactor if you wish to do it cleaner. Both `PendingCrossZoneDispatchesCellOwned` and `PendingDepositEventsCellOwned` use constant key, therefore we can use `TransactionDB` instead of `DBCommon`, this will allow similar functionality to a `Mutex` but just for one key-value pair.
I understand if it is too much for this PR, then maybe in the future. For now we can probably just replace Mutex with Rwlock.

Pravdyvy (commented):
Thank you for the PR. Left couple of comments, mostly structural questions.

lez/sequencer/core/src/lib.rs
```diff
@@ -1036,14 +1148,254 @@ fn apply_follow_update(
...
+/// after the persist above, the block sitting at that height is the canonical
+/// one, so a delivery it still carries is on chain and must be left marked. For
+/// a deposit, clearing it wrongly would mint twice, since the bridge has no
+/// on-chain replay guard.
+///
+/// Re-queued transactions come from our own pending records, never from the
+/// orphaned block: those bytes were authored by whichever sequencer produced
+/// that block, and re-queueing them as [`TransactionOrigin::Sequencer`] would
+/// walk them straight past the ingress guard into a block we sign.
+fn refork_pending_records(
```
erhant: there may be a double-mint bug in this function; consider the scenario:

1. A bridge deposit `d` is finalized on the L1. Every sequencer of the zone watches L1 deposit events independently, so both we (`A`) and co-sequencer `B` record `d` and queue a mint tx, which is a normal operation, not a fault.

2. It's our turn; we produce block 5 carrying `d`, persist it optimistically, and publish the inscription. Our record: `d` is submitted in 5.

3. Our inscription loses (can be a turn boundary, lost race, or an L1 reorg of the inscription). That's the whole reason `record_produced_block_skips_persistence_on_lost_race` and the orphaned field
of `FollowUpdate` exist: our locally persisted block can turn out **not** to be the channel's block.

4. `B`'s chain wins. `B` saw `d`'s event slightly _later_ than we (`A`) did (independent event streams, independent block cadence), so `B`'s block 5′ doesn't carry `d` but block 6′ does!

5. The channel delivers one update: `adopted = [5′, 6′]`, `orphaned = [our 5]`. Now if we walk `refork_pending_records`:
    - **Canonical loop**: marks `d` as submitted in 6′. Correct because `d` is indeed on chain.
    - **Orphan loop**: _our_ block 5 carries `d`, so it reads the settled block at height 5 (that's 5′), doesn't find `d` there, concludes that `d` fell off the chain, and unmarks the record (i.e. `submitted_in_block_id = Some(6)`) it marked two statements earlier.

6. `d` is re-queued, the production guard passes (record is unsubmitted), our next block mints `d` again while 6′ already minted it.

in short, our own deposit that gets orphaned but then adopted at a later block by another sequencer causes us to requeue it even though its already on chain

lez/sequencer/core/src/lib.rs
```diff
@@ -753,9 +770,59 @@ impl<BP: BlockPublisherTrait> SequencerCore<BP> {
...
+                    match store
+                        .dbio()
+                        .record_dispatch_failure(message_key, RETIRE_DISPATCH_AFTER_FAILURES)
+                    {
+                        Ok(true) => error!(
+                            "Retiring cross-zone dispatch {} after {RETIRE_DISPATCH_AFTER_FAILURES} failed attempts: {err:#}",
+                            hex::encode(message_key)
+                        ),
+                        Ok(false) => warn!(
+                            "Cross-zone dispatch {} failed to execute, will retry: {err:#}",
```
erhant: does this actually retry the deposit? the `while let Some((origin, tx)) = self.mempool.pop() { ... }` loop pops from mempool but here (where we are within `apply_mempool_transaction`) this will return `Ok(false)`, there should be something to push the dispatch back into the mempool so that we can retry it again.

I could be missing the part where the push happens though

erhant (commented):
2 comments! All good other than that, but will re-review after these are resolved 🙏🏻

lez/sequencer/core/src/block_store.rs
moudyellaz: Agreed, changed to `Slot::to_le_bytes` on write and a length-checked `u64::from_le_bytes` on read.

lez/sequencer/core/src/cross_zone_watcher.rs
moudyellaz: Agreed, gone. The pass now reports why it ended rather than wrapping a single field.

lez/sequencer/core/src/cross_zone_watcher.rs
moudyellaz: Taken, with a third variant. Two cannot express "nothing is being skipped now, but the floor stays frozen", which is the state a run lands in after it reads past a slot it gave up on. Losing it would let the floor advance over the gap, which is exactly what makes a skipped slot unrecoverable across restarts. So: `DeliverAll`, `Skipping(Slot)`, `FloorFrozen`, and the illegal combination the two loose fields allowed is now unrepresentable.

lez/storage/src/sequencer/mod.rs
moudyellaz: `RwLock` would not help here: every holder of this lock is doing a read-modify-write, so they all need exclusive access and it would behave as a `Mutex` with more ceremony. The readers take no lock and get their consistency from rocksdb's single-key atomicity.

`TransactionDB` is the right shape and I agree it is out of scope here. It is subsumed by moving these lists to per-key entries, which removes the whole-blob rewrite that makes the lock necessary at all. Filing that separately.

Pravdyvy (approved):
lgtm

lez/sequencer/core/src/lib.rs
erhant: fixed in https://github.com/logos-blockchain/logos-execution-zone/commit/7abf4fe17f849dd1c5926ad48edd04fa0df03d7f

lez/sequencer/core/src/lib.rs
erhant: fixed in https://github.com/logos-blockchain/logos-execution-zone/commit/7abf4fe17f849dd1c5926ad48edd04fa0df03d7f

erhant (approved):
LGTM, thanks for addressing the comments 🙏🏻

lez/sequencer/core/src/lib.rs
moudyellaz: Correct. The mark is gone: deposits read the receipt PDA from #642, deliveries the inbox seen shard, both out of the state being built on. An orphan reverts that entry with the block, so there is nothing to unmark.

lez/sequencer/core/src/lib.rs
moudyellaz: Correct, and worse: only the block-size branch called push_front, so the counter advanced once per process start, not per attempt. Deliveries are now drained from the store every turn, so a failure retries on the next block, and the record is dropped at the limit.

PR #653, opened 2026-07-30
feat!: lez sequencer self join

schouhy:
## 🎯 Purpose

Sequencers can now join or leave the Bedrock committee on their own, just by staking or unstaking. No need to manually edit the committee config. This also fixes a startup bug: a sequencer would break if it pointed at a Bedrock channel that already existed, because it always assumed it was the first (bootstrap) sequencer.

## ⚙️ Approach

- [x] Add the self-join/leave mechanism: new `Stake`, `ConfirmStake`, `UnstakeRequest`, `FinalizeUnstake` instructions. The sequencer watches the stake config on-chain and submits a committee update whenever it disagrees with what Bedrock currently has.
- [x] Startup behavior change: the sequencer now checks if the Bedrock channel already exists before building its genesis state. If it doesn't exist, it creates it (bootstrap). If it does, it just joins normally instead of assuming it's the first one.
- [x] Bootstrap sequencer's own stake it's a real, replayable transaction at genesis.
- [x] Add a `submit_stake` CLI tool to stake or unstake a sequencer by hand.

## 🧪 How to Test

**Option A — integration test (easiest)**

```
RISC0_DEV_MODE=1 RUST_LOG=info cargo test -p integration_tests --test sequencer_stake_demo -- --nocapture
```

This spins up a local Bedrock node and one sequencer, stakes a demo key, waits for it to join the committee, then fully unstakes it and waits for it to be removed and the funds released.

**Option B — manual, with two real sequencers**
1. Clean: `just clean`
1. Start Bedrock: `just run-bedrock`
1. Start sequencer A: `just run-sequencer`. It creates the channel since it doesn't exist yet.
3. Start sequencer B in a new terminal, pointing at the same channel but its own home dir:
   ```
   just run-sequencer --home /tmp/seq-b --port 3041 --metrics-address 0.0.0.0:9091
   ```
   B should log `"Channel already exists; joining as a non channel creator"` and just sit there.
1. Get B's Bedrock key: 
   ```
   cargo run --release --bin bedrock_pubkey -- lez/sequencer/service/configs/debug/sequencer_config.json --home /tmp/seq-b
   ```
1. Fund a wallet account with the wallet CLI (e.g. by `wallet auth-transfer init` + `wallet pinata claim`), make a fresh ownership account (`wallet account new public`), then stake B:
   ```
   cargo run --release -p sequencer_service --bin submit_stake --features submit_stake -- --wallet <wallet-home> stake \
     --funding-account <funded-account> \
     --ownership-account <fresh-account> \
     --sequencer-key <B's key> \
     --amount <amount>
   ```
1. Wait ~20-30s. B should start alternating blocks with A (`"Our turn"` logs on both).
1. To leave:
   ```
   cargo run --release -p sequencer_service --bin submit_stake --features submit_stake -- --wallet <wallet-home> unstake-request \
     --ownership-account <account> --amount <amount> --destination <account>
   ```
   B gets removed from the committee once fully drained.

## 🔗 Dependencies

None.

## 🔜 Future Work

- Move `sequencer_stake_signing_key` out of the config file into its own file, like `bedrock_signing_key` already works.
- Replace the genesis faucet pass-through key (`GENESIS_STAKE_FUNDING_KEY`) with a real "deposit from Bedrock" flow instead of a fixed key.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


lez/sequencer/core/src/lib.rs
```diff
@@ -595,6 +640,10 @@ impl<BP: BlockPublisherTrait> SequencerCore<BP> {
             .raise_published_high_water(block.header.block_id)
             .context("Failed to persist published high water mark")?;
 
+        // Independent Mantle tx, not bundled with the block above — join/exit
+        // config updates don't need to be.
+        self.submit_committee_update(committee_update).await;
+
```
Pravdyvy, line 665: If it is independent, maybe move it one level above, into `main_loop`, otherwise its usage is hidden in block production.

lez/sequencer/core/src/lib.rs
```diff
@@ -980,7 +1102,10 @@ impl<BP: BlockPublisherTrait> SequencerCore<BP> {
 
         sequencer_core_metrics::record_block_creation_time(now.elapsed());
 
-        Ok(BlockWithMeta { block, withdrawals })
+        let committee_update = live_accredited_keys
+            .and_then(|keys| committee_discovery::committee_update(&working_state, keys));
+
+        Ok((BlockWithMeta { block, withdrawals }, committee_update))
```
Pravdyvy: If withdrawals is a metainformation, then maybe comitee updates are as well? Maybe move them into `BlockWithMeta`?

lez/sequencer/service/src/bin/submit_stake.rs
Pravdyvy, line 1: It goes through a `WalletCore`, is there any reason why it can not live in wallet?

Pravdyvy (approved):
lgtm, a couple of questions.

lez/programs/sequencer_stake/core/src/lib.rs
```diff
@@ -0,0 +1,290 @@
...
+use lee_core::{
+    account::AccountId,
+    program::{InstructionData, ProgramId},
+};
+use serde::{Deserialize, Serialize};
+
+const SEQUENCER_STAKE_CONFIG_SEED_DOMAIN: [u8; 32] = *b"/LEZ/v0.3/MinSequencerStake/0000";
+
+/// The Bedrock sequencer identity a stake backs.
+pub type SequencerKey = [u8; 32];
```
erhant, line 15: this may be problematic, as here in `guest` program we accept plain 32 bytes, but on `host` it goes through `Ed25519PublicKey::from_bytes(&key)` which will fail for non curve-point byte arrays

e.g. what would happen if we inscribe an execution of this program with a non-curve point of 32 bytes?

lez/sequencer/core/src/committee_discovery.rs
```diff
@@ -0,0 +1,340 @@
...
+/// those are fixed constants supplied separately when building the
+/// `ChannelConfigOp`.
+#[must_use]
+pub fn committee_update(
+    state: &lee::V03State,
+    live_accredited_keys: &[SequencerKey],
+) -> Option<Vec<SequencerKey>> {
+    let config = read_config(state)?;
+
+    // Sorted by key bytes so the list is deterministic across calls: a
```
erhant, line 23: question: would BTreeMap be a more ergonomic option here? allowing for efficient inclusion checks while preserving order as well

lez/programs/sequencer_stake/Cargo.toml
erhant, line 1: this also needs `[lints] workspace = true` i think

lez/system_accounts/src/lib.rs
```diff
@@ -1,10 +1,24 @@
...
 
 use clock_core::ClockAccountData;
 use lee_core::account::{Account, AccountId, Nonce};
 
+/// Minimum summed stake for a Bedrock sequencer key to be a committee candidate.
+pub const DEFAULT_MINIMUM_SEQUENCER_STAKE: u128 = 149;
+
+/// Channel administration defaults.
+///
+/// Slots, not seconds (1 slot = 1s on the current devnet): 20-slot turns,
```
erhant, line 14: can maybe use `Slot` type here?
if its not imported already, maybe a tiny type alias would be readable too instead of the comment

lez/sequencer/core/src/config.rs
```diff
@@ -56,6 +56,11 @@ pub struct SequencerConfig {
     pub retry_pending_blocks_timeout: Duration,
     /// Sequencer own signing key.
     pub signing_key: [u8; 32],
+    /// Signing key of the LEZ account backing this sequencer's genesis stake.
+    /// Needed to sign the genesis `Stake` transaction, so top-up/unstake later
+    /// use the same account.
+    // TODO: move out of the config file and into its own file, like `signing_key`.
+    pub sequencer_stake_signing_key: [u8; 32],
```
erhant, line 63: this is a braking change due to having no default, we should indicate the PR title for that

lez/system_accounts/src/lib.rs
```diff
@@ -1,10 +1,24 @@
 //! This crate provides system accounts used by LEZ.
 
-use std::str::FromStr as _;
+use std::{collections::BTreeMap, str::FromStr as _};
 
 use clock_core::ClockAccountData;
 use lee_core::account::{Account, AccountId, Nonce};
 
+/// Minimum summed stake for a Bedrock sequencer key to be a committee candidate.
+pub const DEFAULT_MINIMUM_SEQUENCER_STAKE: u128 = 149;
```
erhant, line 10: can we have a comment on how this number was derived?

Cargo.toml
```diff
@@ -375,3 +377,4 @@ clippy.cargo-common-metadata = "allow"
 # Reason: hard to address right now and mostly comes from dependencies
 # so the fix would be just a long list of exceptions.
 clippy.multiple-crate-versions = "allow"
+
```
erhant: nit: trailing line

lez/sequencer/core/src/lib.rs
```diff
@@ -217,7 +244,22 @@ impl<BP: BlockPublisherTrait> SequencerCore<BP> {
             hex::encode(bedrock_signing_key.public_key().to_bytes())
         );
 
-        let (store, state) = Self::open_or_create_store(&config);
+        let own_sequencer_key = bedrock_signing_key.public_key().to_bytes();
+
+        // Only seed our own key into genesis as the bootstrap sequencer if the
+        // channel doesn't exist yet. Otherwise it's someone else's channel and
+        // we join later, the normal self-join way.
+        let channel_already_exists = BP::channel_exists(&config.bedrock_config)
```
erhant: will this play nicely with `docker compose`? i.e. if we launch the services together, would this call fail and exit the sequencer immedaitely? or shall we add a retry here

lez/sequencer/core/src/committee_discovery.rs
```diff
@@ -0,0 +1,340 @@
...
+        .iter()
+        .filter(|(_, entry)| entry.net_stake() >= config.minimum_sequencer_stake)
+        .map(|(key, _)| *key)
+        .collect();
+    desired.sort_unstable();
+
+    let mut live = live_accredited_keys.to_vec();
+    live.sort_unstable();
+
+    (desired != live).then_some(desired)
```
erhant, line 46: `Keys` used for the config is a `NonEmptyBoundedVec`, which may error on empty list, is it possible for `desired` to return `[]` here? that would fail the code elsewhere within `Keys::try_from`

erhant (changes requested):
few comments, one important one is about `sequencer_key` validity within the program (bytes may be an invalid curve point)

lez/sequencer/core/src/lib.rs
schouhy: It's already at the same level as the block submission. Isn't that ok? Maybe the name of the method `produce_new_block` is not great. But it builds the block, submits it and in case of success and in the presence of config updates needed, it also submits the config update.

lez/sequencer/core/src/lib.rs
schouhy: done

lez/sequencer/service/src/bin/submit_stake.rs
schouhy: to keep the PR scoped to just the self-join/leave mechanism. In the future we can add the wallet commands. I thought for the time being it was ok to have it as a thin binary just for staking.

lez/programs/sequencer_stake/core/src/lib.rs
schouhy: thanks. fixed it

lez/sequencer/core/src/committee_discovery.rs
schouhy: I think committees will be in the order of the tens or something like that. Not sure how much we can gain with btreemap? I can change it if you insist though, but I'd personally wait for this to be flagged as a bottleneck to start optimizing it.

lez/programs/sequencer_stake/Cargo.toml
schouhy: done

lez/system_accounts/src/lib.rs
schouhy: done

lez/system_accounts/src/lib.rs
schouhy: added a TODO to properly set it. It's now set to 149 because pinata transfers 150 and so a single call to it is enough for manual testing

Cargo.toml
schouhy: done

lez/sequencer/core/src/lib.rs
schouhy: true. Added retries

lez/sequencer/core/src/committee_discovery.rs
schouhy: haven't thought about this. Added a few guards to it.

lez/sequencer/core/src/lib.rs
Pravdyvy: In this case, let's change name and doc of the function and will be fine.

lez/sequencer/service/src/bin/submit_stake.rs
Pravdyvy: Ok

lez/sequencer/core/src/committee_discovery.rs
erhant: ah oki, I was thinking it may go much beyond that hahah, let's keep it as is

erhant (approved):
LGTM, thanks for addressing the comments 🙏🏻

lez/sequencer/core/src/committee_discovery.rs
```diff
@@ -0,0 +1,377 @@
...
+    // positional.
+    let mut desired: Vec<SequencerKey> = config
+        .entries
+        .iter()
+        .filter(|(_, entry)| entry.net_stake() >= config.minimum_sequencer_stake)
+        .map(|(key, _)| *key)
+        .collect();
+    desired.sort_unstable();
+
+    if desired.is_empty() {
```
moudyellaz, line 35: With no cancel instruction, doesn't the last sequencer's full UnstakeRequest wedge here permanently (can't finalize, can't top up)?

lez/sequencer/core/src/committee_discovery.rs
```diff
@@ -0,0 +1,377 @@
...
+    let config = read_config(state)?;
+
+    // Sorted by key bytes so the list is deterministic across calls: a
+    // `ChannelConfigOp`'s `keys` field must reproduce the same order every
+    // time given the same state, since Bedrock's accredited-key index is
+    // positional.
+    let mut desired: Vec<SequencerKey> = config
+        .entries
+        .iter()
+        .filter(|(_, entry)| entry.net_stake() >= config.minimum_sequencer_stake)
```
moudyellaz, line 30: Anything stopping someone staking keys they never run, so each dead key's turn stalls until the posting timeout?

lez/sequencer/core/src/lib.rs
```diff
@@ -919,6 +1046,17 @@ impl<BP: BlockPublisherTrait> SequencerCore<BP> {
...
             }
 
+            // Block-validity rule: a not-yet-valid FinalizeUnstake is dropped
+            // outright, not applied — whether it arrived via the mempool
+            // (anyone may submit one, per spec) or from this sequencer's own
+            // discovery above. It re-appears on its own once conditions are
+            // met (mempool: whoever wants it finalized resubmits;
+            // discovery-sourced: reconstructed fresh next block), so it
+            // doesn't need requeuing here.
+            if !finalize_unstake_is_includable(&working_state, &tx, live_accredited_keys) {
```
moudyellaz, line 1135: Since this only runs while building, what stops a peer's block from carrying a full drain that was never valid?

lez/sequencer/core/src/lib.rs
```diff
@@ -1488,10 +1638,14 @@ fn build_genesis_state(config: &SequencerConfig) -> (lee::V03State, Vec<LeeTrans
         // genesis tx.
         GenesisAction::SupplyBridgeLockHolding { .. } => None,
     });
+    let bootstrap_stake_txs = bootstrap_sequencer_key.into_iter().flat_map(|key| {
```
moudyellaz: Is a joiner guaranteed to reorg onto the creator's genesis here, given validate_against_tip rejects a mismatched prev link?

lez/sequencer/service/configs/debug/sequencer_config.json
```diff
@@ -78,5 +78,39 @@
         37,
         37,
         37
+    ],
+    "sequencer_stake_signing_key": [
```
moudyellaz: Since this key alone can unstake the bootstrap sequencer, move it out of the config now rather than as future work?

lez/programs/sequencer_stake/src/main.rs
```diff
@@ -0,0 +1,405 @@
...
+    let mut ownership_account_data = ownership_account.account.clone();
+    ownership_account_data.data = StakeRecord {
+        sequencer_key,
+        pending_unstake: None,
+    }
+    .to_bytes()
+    .try_into()
+    .expect("StakeRecord should fit in account data");
+    let ownership_account_post =
+        AccountPostState::new_claimed_if_default(ownership_account_data.clone(), Claim::Authorized);
```
moudyellaz, line 198: Should a claim require a zero balance, so an account funded beforehand doesn't lock the excess forever?

lez/sequencer/core/src/committee_discovery.rs
schouhy: the last sequencer can't leave from the bedrock channel either. Bedrock refuses config updates with empty accredited keys I think

lez/sequencer/core/src/committee_discovery.rs
schouhy: The bad behavior of joining and not producing blocks will be slashed. The idea is to go further and have sequencers gossip the mempool so that even sequencers that join and produce empty blocks are slashed

lez/sequencer/core/src/lib.rs
schouhy: That would be visible to the other sequencers to ignore the block and slash the bad behavior no?

test_fixtures/src/config.rs
```diff
@@ -18,6 +18,13 @@ pub const INITIAL_PRIVATE_BALANCES_FOR_WALLET: [u128; 2] = [10_000, 20_000];
 /// Fixed sequencer signing key; exposed so the fixture generator can reopen the produced store.
 pub const SEQUENCER_SIGNING_KEY: [u8; 32] = [37; 32];
 
+/// Key of the account holding the sequencer's genesis stake. Separate from
+/// [`SEQUENCER_SIGNING_KEY`]: block signing and stake control are distinct roles.
+pub const SEQUENCER_STAKE_KEY: [u8; 32] = [55; 32];
+
+/// Bedrock signing key used by the prebuilt dump as first accredited key.
+pub const SEQUENCER_BEDROCK_SIGNING_KEY: [u8; 32] = [77; 32];
```
Arjentix, line 30: Hm, shouldn't we unify `SEQUENCER_BEDROCK_SIGNING_KEY` and `SEQUENCER_SIGNING_KEY`?

lez/testnet_initial_state/src/lib.rs
```diff
@@ -235,6 +239,7 @@ fn initial_programs() -> Vec<Program> {
         programs::ping_receiver(),
         programs::bridge_lock(),
         programs::wrapped_token(),
+        programs::sequencer_stake(),
```
Arjentix: I can see there is a cross-zone related comment above. Let's either clear this comment or move `sequencer_stake()` above it

lez/programs/sequencer_stake/core/src/lib.rs
Arjentix: Sorry, can't see how it's fixed, this is still plain `[u8; 32]`, maybe introduce strong type here?

lez/programs/sequencer_stake/core/src/lib.rs
```diff
@@ -0,0 +1,295 @@
...
+    /// Records a request to release `amount` to `destination`; no balance
+    /// moves yet. Must leave the account at zero or at/above the minimum.
+    UnstakeRequest {
+        amount: u128,
+        destination: AccountId,
+    },
+
+    /// Unsigned, permissionless: releases a pending `UnstakeRequest`.
+    /// Block-inclusion validity is enforced outside this program.
+    FinalizeUnstake,
```
Arjentix, line 91: Can you add a comment please why these are two different instructions and not just `Unstake`?

lez/programs/sequencer_stake/src/main.rs
```diff
@@ -0,0 +1,405 @@
...
+        let record = StakeRecord::from_bytes(ownership_account.account.data.as_ref())
+            .expect("claimed ownership account should decode as StakeRecord");
+        assert_eq!(
+            record.sequencer_key, sequencer_key,
+            "ownership account backs a different sequencer key"
+        );
+        assert!(
+            record.pending_unstake.is_none(),
+            "cannot top up while an unstake request is pending"
+        );
```
Arjentix, line 151: Interesting. Then there should be a way to drop it?

lez/programs/sequencer_stake/src/main.rs
```diff
@@ -0,0 +1,405 @@
...
+    }
+
+    if let Some(entry) = config.entries.get_mut(&sequencer_key) {
```
Arjentix: I think you can use [`entry()`](https://doc.rust-lang.org/std/collections/struct.BTreeMap.html#method.entry) to make it more efficient

lez/programs/sequencer_stake/src/main.rs
```diff
@@ -0,0 +1,405 @@
...
+            "an initial stake must already meet the minimum"
+        );
+        config.entries.insert(
+            sequencer_key,
+            SequencerEntry {
+                account_id: ownership_account.account_id,
+                total_staked: amount,
+                total_pending_unstake: 0,
+            },
+        );
```
Arjentix: You make additions and insertions into the config before stake was confirmed. Just looks a bit strange to me, is this safe?

lez/sequencer/core/src/lib.rs
```diff
@@ -217,7 +247,34 @@ impl<BP: BlockPublisherTrait> SequencerCore<BP> {
...
+        // we join later, the normal self-join way.
+        let channel_probe_retry_strategy =
+            FixedInterval::new(Self::CHANNEL_PROBE_RETRY_DELAY).take(Self::CHANNEL_PROBE_RETRIES);
+        let channel_already_exists = Retry::start(channel_probe_retry_strategy, || async {
+            BP::channel_exists(&config.bedrock_config)
+                .await
+                .inspect_err(|err| warn!("Failed to probe Bedrock channel: {err:#}"))
+        })
+        .await
+        .expect("Failed to probe Bedrock channel");
```
Arjentix, line 283: Why we do retries here? Isn't one check enough?

lez/sequencer/core/src/lib.rs
```diff
@@ -595,6 +659,10 @@ impl<BP: BlockPublisherTrait> SequencerCore<BP> {
             .raise_published_high_water(block.header.block_id)
             .context("Failed to persist published high water mark")?;
 
+        // Independent Mantle tx, not bundled with the block above — join/exit
+        // config updates don't need to be.
+        self.submit_committee_update(committee_update).await;
```
Arjentix, line 697: I wonder if this is okay... Can't there happen any races? For example our sequencer can loose it's time window? IMO better to keep everything derived from block in the same tx like withdrawals. Maybe ask this feature from Bedrock team?

lez/sequencer/core/src/lib.rs
```diff
@@ -610,6 +678,49 @@ impl<BP: BlockPublisherTrait> SequencerCore<BP> {
...
+    ) {
+        let Some(new_keys) = committee_update else {
+            return;
+        };
+        if self
+            .last_committee_submission_at
+            .is_some_and(|at| at.elapsed() < Self::COMMITTEE_SUBMISSION_COOLDOWN)
+        {
+            return;
+        }
```
Arjentix: Just runtime state? Shouldn't it be in db?

Arjentix: Btw why we update by timeout?

lez/sequencer/core/src/lib.rs
```diff
@@ -724,7 +835,10 @@ impl<BP: BlockPublisherTrait> SequencerCore<BP> {
         true
     }
 
-    fn build_block_from_mempool(&mut self) -> Result<BlockWithMeta> {
+    fn build_block_from_mempool(
```
Arjentix, line 913: Omg, this function became so long I have very little idea what's going on inside... Not a remark for this PR only, just my thoughts

Arjentix (commented):
Haven't fully reviewed yet, will revisit, here are some comments

lez/sequencer/core/src/lib.rs
schouhy: it should be enough with one check, yes. But to avoid getting into flaky tests where both sequencer and bedrock nodes are spin up at the same time, I added the retries.

lez/programs/sequencer_stake/src/main.rs
schouhy: lez programs dont write live to the lez state. The config updates will take effect only if the entire execution succeeds and the changes are recorded in the post state. I should be safe yes.

lez/sequencer/core/src/lib.rs
schouhy: when we move to K of N threshold, the config will need to be distributed and signed by at least K sequencers. The signatures for a config are over the entire mantle tx that contains them. If we put the config in same mantle tx as the inscription with the stake/unstake tx, that will effectively stop the chain until all the signatures are collected. So it needs to be a separate tx. A next version of this will include a rule that the config is only valid if the inscription with the stake/unstake tx is already finalized (meaning honest sequencers will refuse to sign it if too early). That altogether should solve the race problems I think

lez/programs/sequencer_stake/core/src/lib.rs
schouhy: sure. for completeness I include it here too. There will always be a delay between the intent of unstaking and the removal of the exiting sequencer key from the Bedrock accredited_keys. That's because the channel config update is not instant and goes through signature collection and bedrock inclusion times. A single instruction that unstakes and unfreezes the funds instantly means there will be a sequencer with no funds at stake during that time window where the channel config didnt kick in yet. For the unstaking process to be sound, it needs to be split into two functions. The one that records the intent of unstaking, and the one that finishes it and releases the funds only once the sequencer is no longer part of the committee

lez/programs/sequencer_stake/src/main.rs
schouhy: finialize unstake is the intended way

lez/sequencer/core/src/lib.rs
schouhy: > Just runtime state? Shouldn't it be in db?

I think it's ok to be runtime state. The channel config update is triggered by the sequencer seeing a difference between the lez desired committee and the actual bedrock committee. Without this cooldown it would resubmit the same channel config at every block until it takes effect

lez/programs/sequencer_stake/core/src/lib.rs
schouhy: > Sorry, can't see how it's fixed, this is still plain [u8; 32], maybe introduce strong type here?

The fix is on the `is_valid_sequencer_key` call.

test_fixtures/src/config.rs
schouhy: I think we should remove the SEQUENCER_SIGNING_KEY actually. That one is signing the blocks and predates our integration to Bedrock. Posting to bedrock already involves signing the mantle tx and that should be enough, no need to sign the blocks also.

lez/testnet_initial_state/src/lib.rs
schouhy: thanks. done

lez/programs/sequencer_stake/core/src/lib.rs
schouhy: The fix is on the is_valid_sequencer_key call.

erhant: perhaps we can add a referencing comment here for that

test_fixtures/src/config.rs
erhant: on gossip im using the bedrock signing key as well btw, not the `SEQUENCER_SIGNING_KEY`

lez/programs/sequencer_stake/core/src/lib.rs
Arjentix: > The fix is on the is_valid_sequencer_key call.

But that requires to always remember calling this function, while this type is also used in `sequencer_core`. I think it should be possible to introduce a type which will check validity in deserialization

lez/sequencer/core/src/lib.rs
Arjentix: Then timeout should be implemented in blocks I think. Not in time.

lez/sequencer/core/src/config.rs
```diff
@@ -104,7 +115,14 @@ const fn default_metrics_address() -> Option<SocketAddr> {
...
+/// Covers a possible rise in gas prices before it gets mined — a stuck
+/// transaction keeps retrying with the same fee, so if prices rise past this
+/// it never gets mined. 10k still covers a normal price jump on a block
+/// transaction; the sdk's default of 200 is too low for large transactions
+/// and can leave them stuck.
 #[must_use]
 pub const fn default_priority_fee() -> u64 {
-    logos_blockchain_zone_sdk::sequencer::FundingConfig::DEFAULT_PRIORITY_FEE
+    10_000
 }
```
Arjentix: This Claude's comment can stale pretty quickly, let's just leave some short description

lez/sequencer/core/src/lib.rs
moudyellaz: ok makes sense, i keep not considering slashing since it is not implemented!

Arjentix (approved):
LGTM, thanks. Seems quite non-trivial, glad we have a lot of new tests for this.

One thing that again scares me is how complex sequencer has become and how complex it will be to debug something on stage. Hopefully with actors we will be able to at least trace messages between different parts

lez/sequencer/core/src/config.rs
schouhy: thanks. done

lez/sequencer/core/src/lib.rs
schouhy: I implemented it in bedrock slots. let me know if that's ok.

lez/programs/sequencer_stake/src/main.rs
schouhy: thanks. done

lez/programs/sequencer_stake/core/src/lib.rs
schouhy: done!

lez/programs/sequencer_stake/core/src/lib.rs
schouhy: implemented the type safe approach suggested by @Arjentix  [here](https://github.com/logos-blockchain/logos-execution-zone/pull/653#discussion_r3765810154)

lez/sequencer/core/src/lib.rs
schouhy: I'm not sure what you mean sorry. Did the multi sequencer at genesis refactor solved this by chance?

lez/sequencer/service/configs/debug/sequencer_config.json
schouhy: done! Moved it to a file

lez/programs/sequencer_stake/src/main.rs
schouhy: excess can happen even after if someone sends native tokens to one of those accounts. Since sending native tokens doesnt require recipient authorization and can be done by any program. The design treats every extra native token in those accounts as burned. Does it make sense? The stake program only reasons about the funds introduced by calls to stake. The rest is ignored

lez/sequencer/core/src/lib.rs
schouhy: done, renamed it to `run_production_turn`

moudyellaz (approved):
lgtm, thanks for addressing comments!

Arjentix (approved):

PR #660, opened 2026-07-31
refactor(lee): refactor private kinds

agureev:
## 🎯 Purpose

This refactor is the first step in the environment unification task with the "north star" being #650.

Its goal is twofold:

1) reduce amount of production code by getting rid of combinatorial enum structure in favor of a united `PrivateWitness` struct carrying information about init/updates and common info such as nullifier and key information all account carry regardless of kind.

2) ensure an interface which is easier to update if we introduce more keys (either shared or specific to the account kind) such as the authorization key for the regular private accounts. this way there is a centralized struct to update instead of disjoint enums.

Hence this will make authorization key addition more seemless.

As authorization information is currently determined protocol-wide (e.g. enforcing authorization for all regular private accounts) removing separate authorization-varying enum information does not impact the logic.

## ⚙️ Approach

The approach is to replace the combinatorial private `InputAccountIdentity` enum by a `Private(WitnessKind)` enum variant with `WitnessKind` struct specifying whether an account is init or updated.

- [x] Define `NullifierWitness` struct carrying information about init/update of the account plaintext
- [x] Define `PrivateWitness` struct carrying common private account information such as identifier and npk/vpk
- [x] Replace `InputAccountIdentity` private enums with one `Private(PrivateWitness)` enum.

## 🧪 How to Test

## 🔗 Dependencies

#604

## 🔜 Future Work

Further environment unification tasks guiding towards #650.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


lee/state_machine/core/src/circuit_io.rs
```diff
@@ -24,74 +24,63 @@ pub struct PrivacyPreservingCircuitInput {
...
-    /// as the 4th input.
-    PrivatePdaInit {
-        vpk: ViewingPublicKey,
-        random_seed: [u8; 32],
-        npk: NullifierPublicKey,
-        identifier: Identifier,
-        commitment_root: CommitmentSetDigest,
-        /// When `Some((seed, authority_program_id))`, the circuit binds this position via the
+    Private(PrivateWitness),
+}
```
Pravdyvy, line 36: Would there be another kinds in this enum in the future which will use `PrivateWitness`? If no, then I do not see a purpose in encapsulation `Private(PrivateWitness)`, we can just use `PrivateWitness` instead.

Pravdyvy (approved):
lgtm, small comment.

lee/state_machine/core/src/circuit_io.rs
agureev: That's a good point! I wanted to have a private-public enum and then a struct with `PrivateWitness` name as that is descriptive enough, but I may just use `Private`, let me try it out

erhant (approved):
LGTM!

lee/privacy_preserving_circuit/src/output.rs
```diff
@@ -40,220 +40,114 @@ pub fn compute_circuit_output(
...
+                    // The npk-to-account_id binding is established upstream in
+                    // `validate_and_sync_states` via `Claim::Pda(seed)` or a caller `pda_seeds`
+                    // match. Here we only enforce the lifecycle pre-conditions. The supplied npk
+                    // on the witness has been recorded into `private_pda_by_position` and used
+                    // for the binding check; we use `pre_state.account_id` directly for nullifier
+                    // and commitment derivation.
+                    WitnessKind::Pda { .. } => pre_state.account_id,
+                };
+
+                match (kind, nullifier) {
```
moudyellaz, line 69: Since `kind` and `nullifier` are now independent axes, can we keep this match wildcard-free so a future variant can't silently fall into an existing arm?

moudyellaz (approved):
lgtm, thanks! Left some minor comments.

lee/state_machine/core/src/circuit_io.rs
moudyellaz: #669 keeps `PrivateWitness` as a named struct and grows `Regular` into `Regular { ask }`, worth keeping the wrapper rather than inlining and re-churning?

agureev: Well, the churn isn't a problem imo, I am fine with that, but if we are indeed looking at future changes instead of just locally, it is worth noting that #650 removes the enum completely.

I was thinking that every PR should be self-sufficient but if we can be motivated by further progresss, maybe indeed let's keep it this way for the sake of ease as the final product will just delete the enum either way. So if we inline, we indeed will have to re-extract the struct.

Wdyt @Pravdyvy @moudyellaz? Is it worth to minimize the Enum here given that it will be possibly removed given the current plan in the future?

Pravdyvy: @agureev Your call.

agureev: Then I propose to indeed keep the enum wrapper due to further removal plans.

lee/privacy_preserving_circuit/src/output.rs
agureev: @moudyellaz did you mean something like this: [ca23f12](https://github.com/logos-blockchain/logos-execution-zone/pull/660/commits/ca23f12f7adb81d2f6aa631a271a20234e084d53)?

agureev: Note that this matching pattern will be completely gone after #685

moudyellaz (approved):
lgtm, thanks!

PR #662, opened 2026-07-31
feat(sequencer): gossip module

erhant:
## 🎯 Purpose

Adds **phase 1 of direct sequencer-to-sequencer p2p: open mempool gossip.** Every node floods transactions to its peers (standard blockchain-style), so the current **leader** — the sequencer whose L1 turn it is to write — already holds the pending transactions it needs, no matter which node received them.

This solves a real gap: today only the leader produces blocks, and a transaction submitted to a non-leader's RPC sits in that node's local mempool until its own turn comes around; the current leader never sees it. Mempool gossip gets it to whoever is leading now.

The mesh is **open and permissionless** by design — transactions are self-authenticating (signed by the user, validated against state), so relayer identity is irrelevant and there is nothing to gate on. It is a latency optimization layered over L1: **L1 remains the source of truth**, gossip being down degrades cleanly to today's L1-only behavior, and a gossip failure after startup never halts the node.

## ⚙️ Approach

- [x] Discover peers via libp2p **Kademlia + Identify + bootstrap** (mDNS behind a `mdns` cargo feature for local dev) — replaces the custom signed-address-announcement discovery
- [x] Gossip individual `LeeTransaction`s on a per-channel GossipSub topic `/lez/<channel_id_hex>/v1/txs`, Signed + Strict validation, message-id = tx hash
- [x] Validate a received transaction with the **same stateless admission the RPC uses** (size cap → `transaction_stateless_check` → sequencer-only-program guard); Accept (forward + admit) / Reject (do not forward) / Ignore
- [x] Add a bounded **seen-cache** (FIFO over tx hashes, metrics-ready counters) since the mempool has no dedup — prevents re-admitting duplicates from multiple peers or local echoes
- [x] Bridge local → mesh: RPC `send_transaction` publishes admitted txs to the topic (fire-and-forget, non-blocking); gossip-received txs enter the same `MemPool` as `TransactionOrigin::Gossip` and are forwarded automatically by GossipSub (no re-publish loop)
- [x] Remove the accredited-address machinery (announcements, peer directory, targeted dialing, keys-refresher); relocate the accredited-keys provider to `gossip/accreditation/`, retained off-mesh for future ChannelConfig work
- [x] Keep block production leader-only and unchanged

## 🧪 How to Test

Unit + multi-node QUIC integration tests (real nodes on `127.0.0.1`):

```bash
RISC0_DEV_MODE=1 cargo test -p sequencer_core --all-features
```

Key integration cases in `gossip/tests.rs`:
- `nodes_discover_each_other_via_bootstrap` — B and C bootstrap only to A, and A ends up connected to both via Kademlia/Identify (no hand-fed addresses).
- `transaction_submitted_to_one_node_reaches_others` — a valid tx published on node A appears in B's and C's mempools (hash-matched) via gossip.
- `invalid_transaction_is_not_propagated` — an invalidly-signed tx driven through real QUIC is `Reject`ed and never reaches any mempool.

Compile the local-dev discovery feature and both service configs:
```bash
cargo clippy -p sequencer_core --all-features --features mdns --tests
cargo check -p sequencer_service --features standalone
```

Manual: run multiple sequencers with `gossip.bootstrap_peers` pointed at one bootstrap node, submit a tx to a non-leader's RPC, and confirm it is included by the current leader.

## 🔗 Dependencies

None.

## 🔜 Future Work

- **ChannelConfig signature gathering:** the retained `AccreditedKeysProvider` (`gossip/accreditation/`) is the accessor for a later gossip subsystem that gossips `ChannelConfig` ops and collects `k` threshold signatures from accredited keys. Not built here. Can be revisited after #653 
- **One deferred minor (reviewed safe):** `publish_transaction` marks a tx "seen" just before the publish call (benign — only gates the echo path).
- **Block gossip**, if a later phase wants to beat L1 latency for block propagation too.
- Optional: richer mempool-full backpressure than drop-on-full; expose the seen-cache counters to metrics. Mempool stuff can be revisited after #691 as well.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


lez/sequencer/core/src/gossip/network.rs
```diff
@@ -0,0 +1,692 @@
...
+
+    fn publish_announcement(&mut self) {
+        // Announce concrete listener addresses; unspecified IPs (0.0.0.0)
+        // are useless to remote peers.
+        let addrs: Vec<String> = self
+            .swarm
+            .listeners()
+            .filter(|addr| !is_unspecified(addr))
+            .chain(self.swarm.external_addresses())
+            .map(ToString::to_string)
```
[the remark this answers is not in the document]
erhant: Good catch, done in https://github.com/logos-blockchain/logos-execution-zone/pull/662/commits/49a70127ae98cc164ef978c93a6e3bc638fd72ee

lez/sequencer/core/src/gossip/validation.rs
```diff
@@ -0,0 +1,231 @@
...
+    // The key was verified as a valid curve point by `decode_and_verify`.
+    let Some(peer_id) = peer_id_from_ed25519(&announcement.public_key) else {
+        return Evaluation::Reject(RejectReason::BadSignature);
+    };
+
+    if directory
+        .iter()
+        .any(|(key, entry)| key == &announcement.public_key && entry.seq >= announcement.seq)
+    {
+        return Evaluation::IgnoreStale;
```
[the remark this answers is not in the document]
erhant: Done in https://github.com/logos-blockchain/logos-execution-zone/pull/662/commits/49a70127ae98cc164ef978c93a6e3bc638fd72ee, added a `PeerDirectory::seq_of` helper for it

lez/sequencer/service/src/lib.rs
```diff
@@ -216,11 +224,41 @@ pub async fn run(config: SequencerConfig, listen_addr: SocketAddr) -> Result<Seq
...
+    // sockets and no tasks. Startup failure here is a hard error
+    // (misconfiguration); after startup, gossip never halts the node.
+    let gossip_network = match gossip_config {
+        None => None,
+        Some(gossip_config) => {
+            let key_path = sequencer_home.join("bedrock_signing_key");
+            let secret: [u8; 32] = std::fs::read(&key_path)
+                .with_context(|| format!("Failed to read {}", key_path.display()))?
+                .try_into()
+                .map_err(|_bytes| anyhow!("Bedrock signing key has incorrect length"))?;
```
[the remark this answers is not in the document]
erhant: Done in https://github.com/logos-blockchain/logos-execution-zone/commit/49a70127ae98cc164ef978c93a6e3bc638fd72ee

lez/sequencer/core/src/config.rs
```diff
@@ -67,6 +79,9 @@ pub struct SequencerConfig {
     /// Address the Prometheus metrics exporter binds to.
     #[serde(default = "default_metrics_address")]
     pub metrics_address: Option<SocketAddr>,
+    /// Sequencer p2p gossip configuration. `None` disables gossip.
```
erhant, line 82: im allowing `None` here so that it's easier to migrate while we are working on other stuff

lez/sequencer/core/src/lib.rs
```diff
@@ -78,13 +79,16 @@ pub enum TransactionOrigin {
     User,
     /// Transactions generated by the sequencer itself.
     Sequencer,
+    /// Transactions received via p2p gossip from a peer sequencer.
+    Gossip,
```
schouhy, line 86: do we really need this? these txs will always be user txs no?

lez/sequencer/core/src/gossip/network.rs
```diff
@@ -0,0 +1,582 @@
...
+                    .add_address(&peer_id, addr.clone());
+                continue;
+            }
+            if let Err(err) = swarm.dial(addr.clone()) {
+                log::warn!("Failed to dial gossip bootstrap peer {addr}: {err}");
+            }
+        }
+        if let Err(err) = swarm.behaviour_mut().kademlia.bootstrap() {
+            log::debug!("Kademlia bootstrap skipped (no known peers yet): {err}");
+        }
```
moudyellaz, line 186: bootstrap runs once and its failure is only logged at debug, so a node that starts before its bootstrap peer never joins; should this retry on an interval while there are no connected peers?

lez/sequencer/core/src/gossip/network.rs
```diff
@@ -0,0 +1,582 @@
...
+            .swarm
+            .behaviour_mut()
+            .gossipsub
+            .report_message_validation_result(message_id, &source, acceptance);
+    }
+
+    /// Publishes a locally-submitted transaction to the mesh.
+    fn publish_transaction(&mut self, tx: &LeeTransaction) {
+        let hash = tx.hash();
+        self.seen.insert(hash);
```
moudyellaz, line 399: a publish that fails with `InsufficientPeers` is dropped and already marked seen, so a tx submitted while the mesh is still forming never propagates; can `seen.insert` move after a successful publish and failures be requeued?

lez/sequencer/core/src/gossip/validation.rs
```diff
@@ -0,0 +1,90 @@
...
+        Ok(tx) => tx,
+        Err(err) => return TxEvaluation::Reject(format!("undecodable transaction: {err}")),
+    };
+
+    let authenticated = match tx.transaction_stateless_check() {
+        Ok(tx) => tx,
+        Err(err) => return TxEvaluation::Reject(format!("stateless check failed: {err:?}")),
+    };
+
+    if let LeeTransaction::Public(public_tx) = &authenticated
```
moudyellaz, line 41: the sequencer-only guard matches `Public` only and `transaction_stateless_check` returns `Ok` unconditionally for `ProgramDeployment`, so an unsigned ~1MB deployment is Accepted and relayed into every node's mempool; is that intended before the mesh is public?

lez/sequencer/core/src/gossip/tests.rs
```diff
@@ -0,0 +1,169 @@
...
+        tokio::time::sleep(Duration::from_millis(200)).await;
+    }
+    false
+}
+
+#[tokio::test]
+async fn nodes_discover_each_other_via_bootstrap() {
+    let secrets = [[10; 32], [11; 32], [12; 32]];
+    let (node_a, _mempool_a) = start_node(secrets[0], vec![]).await;
+    let a_addr = node_a.listen_addrs()[0].clone();
```
moudyellaz, line 80: `listen_addrs()[0]` carries no `/p2p/` suffix so all three nodes take the plain-dial branch at `network.rs:180`; can one case bootstrap with a peer-id-suffixed addr to cover the Kademlia path operators actually configure?

test_fixtures/src/config.rs
```diff
@@ -112,6 +112,7 @@ pub fn sequencer_config(
         },
         cross_zone,
         metrics_address: Some(SequencerConfig::DEFAULT_METRICS_ADDRESS),
+        gossip: None, // no gossip needed for standalone sequencers
```
moudyellaz, line 115: gossip is `None` across the integration matrix, so nothing asserts the PR's premise end to end; can `multi_sequencer` submit to a non-leader and assert the leader includes it?

lez/sequencer/core/src/gossip/tests.rs
```diff
@@ -0,0 +1,169 @@
...
+    // publish a structurally well-formed transaction with an invalid
+    // signature, which still exercises the real gossip pipeline's rejection
+    // path (`evaluate_transaction`'s stateless check, then
+    // `MessageAcceptance::Reject`) end-to-end.
+    let secrets = [[30; 32], [31; 32]];
+    let (node_a, _mempool_a) = start_node(secrets[0], vec![]).await;
+    let a_addr = node_a.listen_addrs()[0].clone();
+    let (node_b, mut mempool_b) = start_node(secrets[1], vec![a_addr]).await;
+
+    assert!(
```
moudyellaz, line 151: the precondition is swarm connectivity, not gossipsub mesh membership, so this absence assertion can pass before the mesh grafts; can it publish a valid tx first and observe it arrive to prove the link is live?

lez/sequencer/core/src/gossip/network.rs
```diff
@@ -0,0 +1,582 @@
...
+        config: GossipConfig,
+        channel_id: [u8; 32],
+        signing_key: Ed25519Key,
+        mempool_handle: MemPoolHandle<(TransactionOrigin, LeeTransaction)>,
+        max_block_size: u64,
+    ) -> Result<Self> {
+        // Reuse the node's L1 bedrock signing key as the libp2p identity. The
+        // secret stays in a `Zeroizing` buffer that both `ed25519_from_bytes`
+        // and drop wipe.
+        let mut secret = signing_key.into_unsecured().to_bytes();
```
schouhy, line 92: The `into_unsecured` doesnt look good. Maybe we could instead of reusing the Bedrock key (that requires unwrapping it with the unsecured method) use a gossip specific key and sign a message with the bedrock key as part of the accreditation so that other nodes can build the map { node_key -> Bedrock key }?

schouhy (approved):
Looks good! Thanks. Left a couple of comments

lez/sequencer/core/src/gossip/tests.rs
erhant: indeed, added it in https://github.com/logos-blockchain/logos-execution-zone/pull/662/commits/bc2302b19be1c7038a536064277006d6927964b1

lez/sequencer/core/src/gossip/tests.rs
erhant: good idea, added https://github.com/logos-blockchain/logos-execution-zone/pull/662/commits/bc2302b19be1c7038a536064277006d6927964b1

lez/sequencer/core/src/gossip/validation.rs
erhant: im keeping this as is right now, as we may have `ProgramDeployment` become an actual `Public` tx as well (as per today's discussions w.r.t fees)

we can visit this later when that PR lands

lez/sequencer/core/src/gossip/network.rs
erhant: good idea, added a queue-based retrial here in https://github.com/logos-blockchain/logos-execution-zone/pull/662/commits/bc2302b19be1c7038a536064277006d6927964b1

lez/sequencer/core/src/gossip/network.rs
erhant: good point! added retrial in https://github.com/logos-blockchain/logos-execution-zone/pull/662/commits/bc2302b19be1c7038a536064277006d6927964b1

lez/sequencer/core/src/gossip/network.rs
erhant: indeed `into_unsecured` isn't pretty, but I'd defer the gossip-specific key to the accreditation phase rather than add it now.

Nothing on the mesh currently relies on the peer identity: txs are self-authenticating and the mesh is deliberately open, so the libp2p key is only a transport identity right now

a separate gossip key *without* the bedrock-signed binding buys us nothing yet in particular for this PR, but it could be useful when we bring in the accredited keys checks (with sequencer Join/Leave and `k of n` sigs), which will bring that kind of `node_key -> Bedrock key` mapping.

I think we can keep it as is (also to save time for fees rn), if ok?

lez/sequencer/core/src/lib.rs
erhant: that is true, but it will allow us to expose some p2p related metrics this way

also, in `apply_mempool_transaction` the gossip-origin validation failures log at `debug!` instead of `error!`.

a gossiped tx that the current leader already included is _expected_ to fail the nonce check on every other node when its own turn comes, without the variant that would always `error!` here

test_fixtures/src/config.rs
erhant: added one test for this at https://github.com/logos-blockchain/logos-execution-zone/pull/662/commits/0203b8c55fb98bac418b01fbe2d4908fd4016546

moudyellaz (approved):

lez/sequencer/core/src/gossip/network.rs
schouhy: sure sounds good. Can we add a `FIXME` to remove the `into_unsecured` in the future?

erhant: ofc, added in https://github.com/logos-blockchain/logos-execution-zone/pull/662/commits/5171a305522e2e6ec1ba6483d23bae36c7ca38ab

PR #667, opened 2026-08-03
feat(cross-zone)!: authorize deliveries by source program and target

moudyellaz:
## 🎯 Purpose

A peer could mint wrapped tokens with nothing locked behind them.

The inbox authorized a delivery by its target program alone and never read `src_program_id`. A zone that bridges has to allow `wrapped_token` as a target, and `ping_sender` lets its caller choose the target program and the payload freely, so any user on a peer zone could send a `wrapped_token::Mint` payload through the ping emitter and have the inbox deliver it. `wrapped_token::mint` only checks that its caller is the inbox, so it minted.

Authorization now names the pair, so the emitting program is part of it.

## ⚙️ Approach

- [ ] Add `CrossZoneRoute { src_program_id, target_program_id }` as the unit of authorization
- [ ] Replace the per-peer target list with per-peer routes, in config and in the inbox config PDA
- [ ] Assert the route in the inbox guest
- [ ] Share one `routes_permit` between the guest and the sequencer's watcher, so the two cannot drift
- [ ] Drop `InboxConfig::allowed_peers`, which was never read

## 🧪 How to Test

`cargo test -p integration_tests --test cross_zone_state_machine` drives the guest against the config a bridging zone writes. `a_mint_from_an_unrouted_emitter_is_rejected` sends the mint through `ping_sender` and must fail; `a_mint_from_the_routed_emitter_is_accepted` sends the same mint through `bridge_lock` and must still credit the recipient.

`cargo test -p cross_zone_inbox_core` covers the predicate, including that a lock-to-wrapped-token route and a ping-to-receiver route do not compose into ping-to-wrapped-token.

`cargo test -p sequencer_core --features mock` covers the watcher dropping a message with no route.

## 🔗 Dependencies

None.

## 🔜 Future Work

None.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


agureev (approved):
looks good! this seems to me the correct way forward.

to elaborate on this, as an aside, what cross-zone messaging now reminds me of (being a sender-receiver-instruction protocol) is an asynchronous `ChainedCall`

especially given our discussions regading making the `pre_states` in chained calls be a vector of account IDs instead, it seems that we maybe could make an interface where we have something like

```
ChainedCall {
  zone: Option<ZoneID>,
  program_id: ProgramId,
  account_ids: Vec<AccountId>,
  instruction_data: Vec<u32>,
  pda_seeds: Vec<PdaSeed>
}
```

where `zone` is `None` is interpreted as it being a local (synchronous) call and otherwise is a remote call. this was just a general thought i wanted to record.

Pravdyvy (approved):
lgtm

moudyellaz:
> looks good! this seems to me the correct way forward.
> 
> to elaborate on this, as an aside, what cross-zone messaging now reminds me of (being a sender-receiver-instruction protocol) is an asynchronous `ChainedCall`
> 
> especially given our discussions regading making the `pre_states` in chained calls be a vector of account IDs instead, it seems that we maybe could make an interface where we have something like
> 
> ```
> ChainedCall {
>   zone: Option<ZoneID>,
>   program_id: ProgramId,
>   account_ids: Vec<AccountId>,
>   instruction_data: Vec<u32>,
>   pda_seeds: Vec<PdaSeed>
> }
> ```
> 
> where `zone` is `None` is interpreted as it being a local (synchronous) call and otherwise is a remote call. this was just a general thought i wanted to record.

@agureev  Agreed, and that shape is close to where Phase 2 is heading: "call any program on the other zone" wants one call surface rather than two.

Worth noting the ordering it implies. A remote ChainedCall is only the same primitive as a local one if the caller can rely on the callee having actually run, and today B cannot confirm the source call even succeeded, it infers it from the shape of A's transaction. Phase 1, replaying peer blocks, is what makes zone: Some(..) mean the same thing as zone: None.

PR #669, opened 2026-08-03
feat!: private authorization key

agureev:
## 🎯 Purpose

Second PR in the stack towards #650 environment unification.

Previously, private accounts could not be updated while being unauthorized. This was a difference between the private and public semantics, locking one of the update paths.

OUT OF SCOPE: Claiming semantics unification and authorization longevity unification is out of scope of this PR and will be added on top later. 

## ⚙️ Approach

- [x] Update shared account keychain derivation to `ssk -> ask -> nsk, ssk -> vsk`
- [x] Update the circuit i/o to accept Option<AuthorizationSecretKey>
- [x] Update the ID check to derive the `npk` from the `ask` if provided
- [x] Enforce proper authorization semantics given correct `ask`

## 🔗 Dependencies

#660 

## 🔜 Future Work

With this PR, shared accounts can be shared restrictively. E.g. a user may share `nsk/vsk` knowledge without sharing the `ask`. There should be an interface to do that.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


lee/key_protocol/src/key_management/secret_holders.rs
```diff
@@ -87,41 +89,35 @@ impl SeedHolder {
...
+        <NullifierSecretKey>::from(&self.generate_authorization_secret_key(index))
     }
 
     #[must_use]
     #[expect(clippy::big_endian_bytes, reason = "BIP-032 uses big endian")]
     pub fn generate_viewing_secret_seed_key(&self, index: Option<u32>) -> ViewingSecretKey {
-        const PREFIX: &[u8; 8] = b"LEE/keys";
-        const SUFFIX_1: &[u8; 1] = &[2];
-        const SUFFIX_2: &[u8; 19] = &[0; 19];
+        const DOMAIN: &[u8; 29] = b"/LEE/v0.3/Keys/Viewing/Secret";
```
moudyellaz, line 113: Body says `ask -> vsk`, but `generate_viewing_secret_seed_key` derives vsk from ssk, is the body stale or is vsk meant to move under ask?

lee/key_protocol/src/key_management/secret_holders.rs
```diff
@@ -36,7 +38,7 @@ impl ViewingSecretKey {
 /// for recepient.
 #[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
 pub struct PrivateKeyHolder {
-    pub nullifier_secret_key: NullifierSecretKey,
+    pub authorization_secret_key: AuthorizationSecretKey,
```
moudyellaz, line 41: `PrivateKeyHolder` now only stores `ask`, so the Future Work case of sharing nsk/vsk without ask isn't representable, is a second holder shape planned for that?

lee/key_protocol/src/key_management/key_tree/keys_private.rs
```diff
@@ -36,7 +38,7 @@ impl ChildKeysPrivate {
         // to generate the `ssk` and `ccc` values.
         let mut parent_hash = sha2::Sha256::new();
         parent_hash.update(b"LEE/keys");
-        parent_hash.update(self.value.0.private_key_holder.nullifier_secret_key);
+        parent_hash.update(self.value.0.private_key_holder.nullifier_secret_key());
```
moudyellaz, line 43: `nth_child` still uses the legacy `LEE/keys` label while the leaf KDFs moved to `/LEE/v0.3/...`, fold it into this reset rather than paying for a second one?

lez/wallet/src/account_manager.rs
```diff
@@ -589,15 +552,40 @@ fn private_key_tree_acc_preparation(
...
+    AccountPreparedData {
+        // The wallet holds no key for a recipient, so it can neither spend the account nor
+        // consent on its behalf. The program still claims it: a private claim never requires
+        // authorization.
+        ask: None,
+        nsk: None,
+        npk,
+        identifier,
+        vpk,
+        pre_state: AccountWithMetadata::new(Account::default(), false, account_id),
```
moudyellaz, line 572: Foreign init now yields `is_authorized: false` where it was `true`, has lez-programs been checked for programs that assert authorization on an account they initialize?

lez/wallet-ffi/src/types.rs
```diff
@@ -460,8 +464,7 @@ impl From<AccountIdentity> for FfiAccountIdentity {
 
                 Self {
                     kind: FfiAccountIdentityKind::PrivateShared,
-                    nullifier_secret_key: nsk.into(),
-                    nullifier_public_key: npk.0.into(),
+                    authorization_secret_key: ask.0.into(),
```
moudyellaz, line 469: `PrivateShared` now leaves `nullifier_secret_key` and `nullifier_public_key` at `Default::default()`, so C callers read 32 zero bytes, intentional or should they still be populated?

moudyellaz (approved):
lgtm, thanks! Left some questions/comments.

lee/key_protocol/src/key_management/secret_holders.rs
agureev: Ah, sorry, that was just one of my ideas. In the end, I opted to keep it separated like the nullifier, viewing pair in the original design.

Will fix

lez/wallet/src/account_manager.rs
agureev: I have not explicitly checked, yet the authorization for specifically initialized accounts was - I thought - assumed to be protocol level, e.g. when we assert authorization for Claimed regular accounts in the public case.

But I will double-check.

lee/key_protocol/src/key_management/secret_holders.rs
agureev: I do not think that a separate KeyHolder struct is necessary, but we may `Option` the key fields. Regardkess, I think this should be a separate PR where we provide an interface for solely distributing selected keypairs for "partially shared" accounts instead of a shared secret and then it would make sense to share the partial key info as well in the struct.

agureev: Appropriate issue #695 filed

lee/key_protocol/src/key_management/secret_holders.rs
agureev: Fixed in PR description

lez/wallet/src/account_manager.rs
agureev: Went through all programs with Claude, nothing seems to currently reach that branch in a practical way

lez/wallet-ffi/src/types.rs
agureev: The idea was that if the ask is supplied, nsk can be re-derived user-side without possible misalignment risk.

However, I'm now thinking of maybe looking at breaking the interface here for betterment

agureev: Decided against breaking changes since I am unsure what exact infra relies on this. Probably worth pairing with #695, recorded a plausible solution there

lee/key_protocol/src/key_management/key_tree/keys_private.rs
agureev: fixed in [c8812a6](https://github.com/logos-blockchain/logos-execution-zone/pull/669/commits/c8812a6da8087e0165b93513bd37ee93d0cebc1f)

lez/wallet-ffi/src/types.rs
agureev: reverted here: [64ddd74](https://github.com/logos-blockchain/logos-execution-zone/pull/669/commits/64ddd7440fc5b53bc86b3fb341ffc2d01c73e236)

with an extra semantic check making sure that the derived account keys are correct

moudyellaz (approved):
lgtm, thanks!

lee/state_machine/src/privacy_preserving_transaction/circuit/tests.rs
```diff
@@ -718,6 +738,215 @@ fn private_authorized_update_encrypts_regular_kind_with_identifier() {
...
+                membership_proof,
+            },
+        })],
+        &program.into(),
+    );
+
+    assert!(matches!(result, Err(LeeError::CircuitProvingError(_))));
+}
+
+/// An `ask` tthat does not derive this account's `npk` is not a credential for it.
```
jonesmarvin8: Typo `tthat` in docstring

jonesmarvin8 (approved):
Lgtm; left a minor comment.

lee/state_machine/src/privacy_preserving_transaction/circuit/tests.rs
agureev: Fixed in [1f7a4c4](https://github.com/logos-blockchain/logos-execution-zone/pull/669/commits/1f7a4c4162405c90bcfa830c787e1bb8c64172d3)

PR #670, opened 2026-08-04
fix(indexer): validate peer blocks before caching them

### LINKED ISSUE #648: Cross-zone peer block cache accepts unvalidated peer-asserted blocks, allowing a remote indexer halt

moudyellaz:
## 🎯 Purpose

Closes #648.

The indexer cached a peer's blocks with no validation of what the peer asserted. `PeerBlocks::insert` keyed on the `block_id` in the decoded block and overwrote last-write-wins, so a peer inscribing a second block at an id it had already used replaced the real one. The next dispatch naming that id re-derived against the wrong block, mismatched, and was rejected as forged, halting ingestion permanently. One inscription, remote halt.

Two things let that block reach the cache at all: `header.hash` was never recomputed, so the hash links the verified prefix walks were fields the peer wrote, and the pinned block-signing key was checked only when a dispatch named a block, never on the way in.

## ⚙️ Approach

- [ ] Recompute `header.hash` from the block's contents before caching, unconditionally
- [ ] Apply the pinned block-signing key in the reader, as the sequencer's watcher already does
- [ ] Make `PeerBlocks::insert` first write wins: an identical re-read is a no-op, a differing block at a held id is refused and logged as equivocation

Behaviour change worth knowing: a peer that legitimately resets its chain is indistinguishable from equivocation here, so it is now refused and logged rather than silently followed. The cache is process-local, so adopting a reset peer means restarting the indexer.

## 🧪 How to Test

`cargo test -p indexer_core cross_zone` covers the three rules and the halt itself. `an_equivocating_peer_cannot_replace_a_cached_block` inserts a second differing block at a used id and asserts the first is kept and an honest dispatch naming that id still verifies.

`cargo test -p integration_tests --test cross_zone_verified` exercises the verifier end to end, so the new gate does not reject legitimate peer blocks.

## 🔗 Dependencies

None.

## 🔜 Future Work

`PeerChain::blocks` is an unbounded `HashMap` holding every peer block ever read, never evicted, which is an OOM on a long-lived peer. Folded into the Phase 1 plan, since the peer-state replica builds on this cache and makes it grow faster. Worth noting the trap: `resolve` returns `InsideRun` for a block that is absent but at or below `verified_prefix`, and the verifier reads `InsideRun` as forged, so naive eviction would reintroduce the halt this PR fixes. Eviction has to distinguish evicted-from-cache from absent-from-chain.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests

### DIFF
--- a/lez/indexer/core/src/cross_zone_verifier.rs
+++ b/lez/indexer/core/src/cross_zone_verifier.rs
@@ -12,7 +12,7 @@ use cross_zone_inbox_core::{
 };
 use futures::{Stream, StreamExt as _};
 use lee::{GENESIS_BLOCK_ID, PublicKey};
-use log::{debug, error, info};
+use log::{debug, error, info, warn};
 use logos_blockchain_core::mantle::ops::channel::ChannelId;
 use logos_blockchain_zone_sdk::{
     CommonHttpClient, Slot, ZoneMessage, adapter::NodeHttpClient, indexer::ZoneIndexer,
@@ -73,11 +73,11 @@ struct PeerChain {
     ///
     /// This, not `max(blocks.keys())`, is what the forgery test gates on: a peer
     /// picks its own `block_id`s, and an id that does not continue the run
-    /// cannot advance the run. It bounds how far this reader has read, not that
-    /// the chain is authentic: the link is self-asserted, since `header.hash` is
-    /// not recomputed on decode and the reader does not apply the pinned-key
-    /// check that [`crate::cross_zone_verifier::CrossZoneVerifier::rederive`]
-    /// applies to the block a dispatch actually names.
+    /// cannot advance the run.
+    ///
+    /// The link it walks means something only because [`accept_peer_block`]
+    /// recomputes `header.hash` and checks the pinned key before anything is
+    /// cached. Without that it compared two fields the peer wrote.
     verified_prefix: Option<u64>,
 }
 
@@ -126,11 +126,38 @@ struct PeerBlocks {
 }
 
 impl PeerBlocks {
-    async fn insert(&self, zone: ZoneId, block: Block) {
+    /// Caches `block` unless a different block is already held at its id, and
+    /// says whether it was newly cached.
+    ///
+    /// First write wins. An identical re-read is a no-op, which the reader does
+    /// on every slot retry. A differing block at a held id is equivocation, and
+    /// replacing the entry is what makes it a remote halt: the prefix already
+    /// certified the old value, so the next dispatch naming that id re-derives
+    /// against the new one and reads as forged.
+    ///
+    /// A peer resetting its chain is indistinguishable from this and is refused
+    /// the same way. The cache is process-local, so a restart adopts it.
+    async fn insert(&self, zone: ZoneId, block: Block) -> bool {
         let mut chains = self.chains.write().await;
         let chain = chains.entry(zone).or_default();
+
+        if let Some(held) = chain.blocks.get(&block.header.block_id) {
+            if held.header.hash == block.header.hash {
+                return false;
+            }
+            error!(
+                "Peer zone {} equivocated at block {}: holding {}, refusing {}. Restart the indexer if this peer legitimately reset its chain.",
+                hex::encode(zone),
+                block.header.block_id,
+                held.header.hash,
+                block.header.hash
+            );
+            return false;
+        }
+
         chain.blocks.insert(block.header.block_id, block);
         chain.extend_prefix();
+        true
     }
 
     /// Resolves `block_id` under a single read lock.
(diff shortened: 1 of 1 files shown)


lez/indexer/core/src/cross_zone_verifier.rs
```diff
@@ -126,11 +126,38 @@ struct PeerBlocks {
...
+    async fn insert(&self, zone: ZoneId, block: Block) -> bool {
         let mut chains = self.chains.write().await;
         let chain = chains.entry(zone).or_default();
+
+        if let Some(held) = chain.blocks.get(&block.header.block_id) {
+            if held.header.hash == block.header.hash {
+                return false;
+            }
+            error!(
+                "Peer zone {} equivocated at block {}: holding {}, refusing {}. Restart the indexer if this peer legitimately reset its chain.",
```
erhant, line 149: does restart alone help here? perhaps indexer should wipe its own store (or restart with `allow_chain_reset` true?)

erhant (approved):
LGTM!

Arjentix (approved):
lgtm

lez/indexer/core/src/cross_zone_verifier.rs
```diff
@@ -444,6 +472,43 @@ struct PeerPass {
...
+    if let Some(expected) = expected_pubkey
+        && !block.is_signed_by(expected)
+    {
+        warn!(
+            "Peer reader dropping block {} from {}: not signed by the pinned block-signing key",
+            block.header.block_id,
+            hex::encode(peer_zone)
+        );
+        return false;
+    }
```
Pravdyvy, line 507: So, I suppose, it is assumes absence of decentralization? If other zone have 2 sequencers it could fire regularly. Probably a Future Work item?

Pravdyvy (approved):
lgtm, one question.

lez/indexer/core/src/cross_zone_verifier.rs
moudyellaz: Restart is enough: `PeerBlocks` holds no store handle and `read_peer` starts from `cursor = None`, so a restart drops the cache and re-reads the peer channel from the start. Nothing is persisted to wipe.

A store wipe would be worse. `allow_chain_reset` keys on the indexer's own chain, not a peer's, and destroys the store; re-indexing from scratch re-verifies historical blocks, so any dispatch derived against the peer's old chain re-derives against the new one, returns `Forged`, and halts ingestion unrecoverably.

lez/indexer/core/src/cross_zone_verifier.rs
moudyellaz: Yes. The pin is a single key and any block not signed by it is dropped, so a peer zone running more than one sequencer would have every block from the other signers rejected: a stall on the indexer, a silently undelivered message on the watcher.

It is contained today only because the pin is optional (a peer with no key still gets the unconditional `header.hash` recomputation) and every zone runs one sequencer. Filed as #684: read the peer channel's accredited key set from L1 at LIB and accept any currently-accredited signer, with a static multi-key `Vec` as the interim.

PR #683, opened 2026-08-04
fix!(lee): scope private PDA authorization to the callee subtree

agureev:
## 🎯 Purpose

Continue implementing LEE environment unification on the road to #650.

This PR implements unification of authorization longevity semantics for private accounts. Before private PDAs were authorized for the rest of the calls after the initial one. Now, the semantics matches the public path: PDAs are authorized for the rest of the call subtree.

OUT OF SCOPE: Claiming unification for regular accounts, API unification, PDA authorization derivation.

## ⚙️ Approach

- [x] Use `CallerData` type for private chained calls
- [x] Differentiate globally authorized regular accounts versus PDA authorization.
- [x] Unify semantics.

## 🧪 How to Test

Check `tests/circuit.rs`

## 🔗 Dependencies

#669 

## 🔜 Future Work

Rest of LEE unification.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


lee/privacy_preserving_circuit/src/execution_state.rs
```diff
@@ -444,6 +491,12 @@ impl ExecutionState {
 
             post_states_entry.insert_entry(post.into_account());
         }
+
+        caller
```
moudyellaz: Maybe skip building this set when the call has no chained calls, so leaf calls do not pay the hashing?

lee/privacy_preserving_circuit/src/execution_state.rs
```diff
@@ -357,10 +364,50 @@ impl ExecutionState {
...
+                        )
+                    {
+                        // authorize_public_first_sight is only true for PDAs
+                        // which will be recorded in output journal.
+                        //
+                        // Since we are in a privacy circuit, the verifier cannot
+                        // replay the transaction to see which public PDAs were
+                        // actually authorized. We mark them false as the
+                        // verifier checks regular account signatures as well.
+                        pre.is_authorized = false;
```
moudyellaz, line 400: Maybe gate this on `account_identities[pre_state_position].is_public()` so the mask can never reach a private position?

lee/privacy_preserving_circuit/src/execution_state.rs
```diff
@@ -357,10 +364,50 @@ impl ExecutionState {
...
+                        && authorize_first_sight_without_pda_witness(
+                            &mut self.pda_family_binding,
+                            &mut self.globally_authorized,
+                            &caller,
+                            caller_pda_seeds,
+                            pre_account_id,
+                            pre_is_authorized,
+                        )
+                    {
+                        // authorize_public_first_sight is only true for PDAs
```
moudyellaz: Maybe update the name here, `authorize_public_first_sight` no longer exists?

lee/privacy_preserving_circuit/src/execution_state.rs
```diff
@@ -51,7 +51,9 @@ pub struct ExecutionState {
     /// `AccountId::for_private_pda(program_id, seed, npk, vpk, identifier) ==
     /// pre_state.account_id`.
     private_pda_by_position: HashMap<usize, (NullifierPublicKey, ViewingPublicKey, Identifier)>,
-    authorized_accounts: HashSet<AccountId>,
+    /// The set containing regular accounts authorized at the root of the call-tree, remaining
```
moudyellaz: Maybe say first sight anywhere in the tree, since insertion is not root-only?

lez/wallet/src/account_manager.rs
```diff
@@ -527,20 +527,20 @@ fn private_key_tree_acc_preparation(
...
+    // A PDA is program-authorized and carries no credential of its own.
+    let ask = (!is_pda).then_some(from_keys.private_key_holder.authorization_secret_key);
     let nsk = from_keys.private_key_holder.nullifier_secret_key();
     let from_npk = from_keys.nullifier_public_key;
     let from_vpk = from_keys.viewing_public_key.clone();
 
     // TODO: Technically we could allow unauthorized owned accounts, but currently we don't have
     // support from that in the wallet.
-    let sender_pre = AccountWithMetadata::new(from_acc.account.clone(), true, account_id);
+    let sender_pre = AccountWithMetadata::new(from_acc.account.clone(), ask.is_some(), account_id);
```
moudyellaz, line 538: Does anything in lez-programs assert `is_authorized` on a PDA it receives at top level?

moudyellaz (approved):

lee/privacy_preserving_circuit/src/execution_state.rs
agureev: I am unsure. Here, the logic of recognizing public vs private PDAs rests on `has_private_pda_witness` being a correct flag, which rests upon `private_pda_by_position` being correct. So a private account positioning as public would make the proof fail on the global level, where the verifier would not find it in its public on-chain state.

Seems like a strong enough assumption given the rest of the protocol. 

And it does not stop e.g. a private account that mascarades as public downstream, since the data is prover-supplied.

lez/wallet/src/account_manager.rs
agureev: Hm.. No valid program can, right? Since it should be caller-authorized which can only happen in a chained-call down the line. So none of our legal programs do.

lee/privacy_preserving_circuit/src/execution_state.rs
agureev: done: ab1053f

lee/privacy_preserving_circuit/src/execution_state.rs
agureev: done: [fbdd794](https://github.com/logos-blockchain/logos-execution-zone/pull/683/commits/fbdd7947a0e8c41235bbd605c8bb43ce9e55ece2)

lee/privacy_preserving_circuit/src/execution_state.rs
agureev: so, sadly there is no minimal way of understanding whether a done call is a leaf call. the evident thing would be to check the vector of chained calls to be inserted into the queue. however, to form these, we have to indeed form the authorized set already to pass that alongside the call. so we either would have to somehow encode this information about the tree beforehand in a flag e.g. or have an unpleasant mutable call vector which we then update with the authorized accounts by hand.

also worth saying that given the ML KEM cost, the optimized hashing calls are minimal and here readability would be preferred in my opinion.

however, there is an evident minimization to be done via extensions which I did here :[d4e95ce](https://github.com/logos-blockchain/logos-execution-zone/pull/683/commits/d4e95ce009aab28e5e19b2a7a110adfc2b693e4d)

Extending the set, we only sha whatever new we add to the authorized set, which may optimize the construction for deep trees without sacrificing readability.

agureev: new commit: f83cf9b

lee/privacy_preserving_circuit/src/execution_state.rs
agureev: [49a8d00](https://github.com/logos-blockchain/logos-execution-zone/pull/683/commits/49a8d002602431e6c49b7b26eeaf6c69f82e97ec)

lee/privacy_preserving_circuit/src/execution_state.rs
agureev: [05c72e2](https://github.com/logos-blockchain/logos-execution-zone/pull/683/commits/05c72e286336a016a26fc6bbaf91225924b207b4)

lee/privacy_preserving_circuit/src/execution_state.rs
```diff
@@ -357,10 +364,50 @@ impl ExecutionState {
...
+                    }
+                    if !has_private_pda_witness
+                        && authorize_first_sight_without_pda_witness(
+                            &mut self.pda_family_binding,
+                            &mut self.globally_authorized,
+                            &caller,
+                            caller_pda_seeds,
+                            pre_account_id,
+                            pre_is_authorized,
+                        )
```
schouhy, line 391: cant we use else if here?
```suggestion
                    else if authorize_first_sight_without_pda_witness(
                            &mut self.pda_family_binding,
                            &mut self.globally_authorized,
                            &caller,
                            caller_pda_seeds,
                            pre_account_id,
                            pre_is_authorized,
                        )
```

schouhy (approved):
LGTM thanks. Left one minor comment.

Also, I feel like `globally_authorized` is reimplementing the same logic as the caller data: you grant authorization to the children. Just that all the calls are children of the root, so it ends up being global. Did you think about a more unifying approach?

agureev:
> LGTM thanks. Left one minor comment.
> 
> Also, I feel like `globally_authorized` is reimplementing the same logic as the caller data: you grant authorization to the children. Just that all the calls are children of the root, so it ends up being global. Did you think about a more unifying approach?

So `globally_authorized` semantics rely on the fact that an authorized regular (non-pda) account is authorized across the entire tree. On the public path, this is ensured by the fact that before running the state-diff validation, the sequencer feeds an initial `authorized_accounts` set filled by all regular accounts that can ever be authorized in the TX. This happens because the authorization validation happens outside of `validate_state_diff` function. For the private case, this is different. The sequencer should indeed when generating the journal to check the proof against, check the signatures for public accounts used for PPTX out-of-circuit and generate an initial set of authorized public accounts. However private regular accounts may only be checked for authorization in the transaction mid-flight (by checking the `ask`), so that the `globally_authorized` set can grow even down a branch of the call tree. If we only extended the authorized set we pass in the chained calls, these regular accounts would dissapear and not be authorized across branches.

Does that make sense?

lee/privacy_preserving_circuit/src/execution_state.rs
agureev: good catch!

agureev: ah, sadly out lint does not allow this:

```
error: `if` expression with an `else if`, but without a final `else`
   --> lee/privacy_preserving_circuit/src/execution_state.rs:383:28
    |
383 |                       } else if authorize_first_sight_without_pda_witness(
    |  ____________________________^
384 | |                         &mut self.pda_family_binding,
385 | |                         &mut self.globally_authorized,
386 | |                         &caller,
...   |
398 | |                         pre.is_authorized = false;
399 | |                     }
    | |_____________________^
 ```
 
 so keeping as-is

Issue #684, opened 2026-08-05
Peer block-signing key is a single pinned key, so a decentralized peer zone's blocks are rejected


moudyellaz:
The cross-zone peer reader accepts a peer block only if it is signed by one pinned key, `expected_pubkey: Option<&PublicKey>` in `accept_peer_block` (`lez/indexer/core/src/cross_zone_verifier.rs`), and the sequencer watcher applies the same single-key check (`lez/sequencer/core/src/cross_zone_watcher.rs`). A peer zone running more than one sequencer signs its blocks with more than one key, so every block signed by a key other than the pinned one is dropped.

On the indexer that is a stall. On the watcher it is not one undelivered message: `link_against` returns `Link::OffChain` for a block failing the key check, the block is read past and the stored chain tip does not advance, so the peer's next block trips the "skips past" branch and so does every block after it. Delivery from that peer ends permanently, and the tip is durable, so a restart does not recover it.

Survivable today because each zone runs a single sequencer, and because the pin is optional: a peer configured with no key still gets the unconditional `header.hash` recomputation. It stops being survivable as soon as peer zones decentralize, which #603 is building toward. Nothing pins a key today either: `expected_block_signing_pubkey` is `None` in every config file, every fixture, every integration test and the `cross_zone_chat` tool. So the permanent-halt consequence above is latent, and goes live the moment one is set.

The fix direction previously recorded here does not work, and both halves fail. `channel_state(peer).accredited_keys` are Ed25519 channel-write keys, while the block signature `Block::is_signed_by` verifies is k256 BIP-340 x-only from a separate keystore, `SequencerConfig.signing_key` against `<home>/bedrock_signing_key`. No mapping between the two key spaces exists anywhere, on chain or in config. And there is no way to read channel state at a finalized point: `/channel/:id` takes no tip parameter and the handler queries at `cryptarchia_info.tip`, while the only finality-anchored read in use is `next_messages`' own `lib_slot` bound. So this needs a design decision about what identifies a peer sequencer before any code is worth writing.

Two further facts. The two sides configure the key separately, `CrossZonePeer` deserialized independently from the sequencer's config file and the indexer's, with no cross-check at startup, on chain or over RPC, so they can diverge (#674). And `test_fixtures` clones one `CrossZoneConfig` into both services, so that divergence is structurally impossible to reproduce in the harness; a test for it has to build two configs.

[...]

Issue #676, opened 2026-08-04
Inbox seen-shard saturates at ~3200 messages per peer per epoch and then panics the guest


moudyellaz:
The inbox records delivered message keys in a `SeenShard`, one per `(src_zone, src_block_id / EPOCH_BLOCKS)` with `EPOCH_BLOCKS = 10_000` (`lez/programs/cross_zone_inbox/core/src/lib.rs`). The whole `BTreeSet<[u8; 32]>` lives in one account, written back through `.try_into().expect("seen shard fits in account data")` (`lez/programs/cross_zone_inbox/src/main.rs`) against `DATA_MAX_LENGTH = 100 KiB` (`lee/state_machine/core/src/account/data.rs`).

That saturates near 3,200 keys. Once a shard is full, every later message from that peer in that epoch panics the guest, so the dispatch fails, is retried once per block, and is given up on after `RETIRE_DISPATCH_AFTER_FAILURES`. Those messages are dropped permanently with only a log line.

Reachable by any ordinary user on a configured peer zone, with no privileges and no fees beyond their own zone: roughly 3,200 messages inside one epoch, which is about 28 hours at a 10s peer block time. It is the cheapest denial of service in the cross-zone path.

It needs a design decision rather than a patch, because each obvious option trades something: rejecting past the cap reintroduces a halt, dropping loses deliveries silently, and rolling to a fresh shard changes the PDA derivation and so the addresses both sides compute.

moudyellaz:
Checked what the panic costs. It does not halt the zone: `apply_mempool_transaction`'s sequencer arm catches the error and returns false rather than propagating (`lez/sequencer/core/src/lib.rs:686-692`), deliberately, so one unexecutable transaction cannot abort production. `count_dispatch_failure` then retires the dispatch after `RETIRE_DISPATCH_AFTER_FAILURES = 3` attempts and it is never retried (`:1053-1070`).

So block production continues and every delivery arriving while the shard is saturated is permanently discarded rather than queued. A fresh shard at the epoch boundary does not recover anything dropped in the window. For a bridge transfer that is value locked on the source and never minted on the destination, with no refund and no on-chain record of the failure. Message loss, not denial of service.

Issue #674, opened 2026-08-04
Sequencer watcher accepts peer blocks the indexer would reject, and the two use different decode-skip budgets


moudyellaz:
The sequencer watcher and the indexer verifier both read the same peer channels and must derive byte-identical dispatches, or the verifier treats the sequencer delivery as forged and halts ingestion. They validate peer blocks differently.

Indexer (`lez/indexer/core/src/cross_zone_verifier.rs`, after #648): `accept_peer_block` recomputes `header.hash` from the block contents unconditionally, applies the pinned block-signing key when configured, and `PeerBlocks::insert` refuses a differing block at an id it already holds. It also keeps a genesis-rooted hash-linked `verified_prefix`.

Sequencer (`lez/sequencer/core/src/cross_zone_watcher.rs`): checks the pinned key only, and only when one is configured. It does not recompute `header.hash`, keeps no peer block cache, and has no equivocation refusal.

So a peer block whose `header.hash` does not match its own contents is dropped by the indexer and accepted by the sequencer.

Second asymmetry: `DECODE_RETRY_LIMIT` is 3 in the verifier and 20 in the watcher. On a peer block neither can decode, the indexer reads past it seventeen passes before the sequencer does, so for that window the two are working from different block sets.

Both are survivable today, because derivation is a pure function of a single peer block and a skipped block costs only the messages in it. They stop being survivable under any design where the two sides accumulate state from the peer stream (the Phase 1 replica), where a difference in which blocks were applied forks the derivation permanently.

Suggested direction: lift the acceptance rule into one shared place used by both, share the retry limit, and consider whether the watcher should keep a `PeerChain` too.

moudyellaz:
#686 removes both asymmetries named here: the watcher recomputes `header.hash`, refuses a second block at an id it has delivered from, and keeps a genesis-rooted linked chain, and neither side reads past a slot it cannot decode.

Leaving this open for the suggested direction, which matters more after that PR than before it. Each side now carries accumulated per-peer state, so a difference in which blocks were applied forks the derivation permanently rather than costing one block's messages, and the two rules are still implemented separately.

PR #686, opened 2026-08-05
fix(lez): deliver only from peer blocks on the peer's verified chain

### LINKED ISSUE #677: Message keys bind nothing about block content, so a peer sequencer can pre-burn a key and suppress a message
The replay key is `message_key(src_zone, src_block_id, src_tx_index)` (`lez/programs/cross_zone_inbox/core/src/lib.rs`). It commits to coordinates only: nothing about the peer block hash, the emitting transaction, or the payload. `src_block_id` is read straight out of the peer block header by both the watcher and the verifier.

So a peer sequencer can publish a block claiming a future `block_id` carrying a trivial emission at index 0, let it deliver normally, and burn that key. When the real block with that id later carries a victim emission at index 0, the inbox finds the key already in its seen shard, treats the delivery as a replay, and no-ops it. The indexer waves it through too, because its seen-set short circuit skips re-derivation for a key already recorded.

The result is total and undetectable suppression of a chosen message, including a bridge lock whose funds are already escrowed on the source side. It defeats the guarantee the design exists to provide, that delivery does not require trusting the source zone sequencer.

Direction to consider: include the peer block hash, or the source transaction hash, in the key preimage, and reject a peer block whose `block_id` has already been delivered from. Note this changes an on-chain value, so existing seen shards, stored pending dispatch records and the indexer seen set all have to migrate together, and in-flight messages would re-deliver once under the new key unless that is handled.

moudyellaz:
## 🎯 Purpose

Closes #677.

Cross-zone deliveries deduplicate on `message_key(src_zone, src_block_id, src_tx_index)`, and neither side required a peer block to be on the peer's verified chain before delivering from it. A peer sequencer could therefore inscribe a block claiming a future `block_id`, have it delivered, and burn the key an honest block would later use, leaving the real message to no-op on chain as a replay.

The issue offers two directions: bind the block hash into the key preimage, and refuse a block whose `block_id` was already delivered from. This implements the second and reaches the first's goal without changing an on-chain value, which would have relocated every inbox PDA and split the key that currently collapses two blocks claiming one id into a single delivery.

## ⚙️ Approach

Both sides now deliver only from blocks on the peer's verified chain, and run the same rule: anchor at the peer's genesis, advance by one block whose `prev_block_hash` matches the head, refuse everything else. A block claiming an id the chain has not reached links to nothing, and an id already consumed cannot recur, so the suppression and the double-delivery close together.

- [x] Add a durable per-peer chain tip, so a restart cannot re-anchor on whatever block it reads first
- [x] Deliver from a peer block only when it is the next link off that tip
- [x] Re-derive against a peer block only when it is inside the run verified from the peer's genesis
- [x] Stop reading past a slot that will not decode, on both sides, since nothing after a hole can be placed
- [x] Let a block that continues the run displace one that never could, at the single id that would extend it

Two asymmetries are deliberate. A block that does not decode holds the delivery floor and is retried indefinitely, because it may be the peer's next link. A block that decodes and does not link is read past, because nothing is delivered from it either way and stopping would let a peer end its own deliveries permanently with one inscription. The indexer refuses to cache ahead of the run for the same reason: caching ahead let its prefix later walk into a block the watcher had already passed over, and the two would then hold different blocks at one id.

A store holding a delivery floor but no chain tip re-reads that peer's channel from its genesis once, to rebuild the tip. Deliveries deduplicate by message key, so nothing is delivered twice.

## 🧪 How to Test

```
RISC0_DEV_MODE=1 cargo test -p sequencer_core -p indexer_core -p storage --lib
just test-cross-zone-verified
just test-cross-zone-bridge
```

## 🔜 Future Work

- #676: the inbox seen shard saturates and every later delivery from that peer in the epoch is then lost. It is the next change here, and it is where binding the block hash into the key preimage becomes cheap, since sharding per `(src_zone, src_block_id)` makes the shard account itself the anchor.
- #674: both asymmetries it names are fixed here, but it should stay open. Each side now carries accumulated per-peer chain state, a durable tip in the watcher and the verified run in the verifier, so the forked-derivation risk it anticipates is live rather than hypothetical, and the two acceptance rules are still implemented separately.
- #684: the peer block-signing key is still a single pinned key, and it is configured separately for the sequencer and the indexer with no cross-check between them.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests

### DIFF
--- a/lez/indexer/core/src/cross_zone_verifier.rs
+++ b/lez/indexer/core/src/cross_zone_verifier.rs
@@ -36,9 +36,9 @@ const PEER_BLOCK_WAIT_TIMEOUT: Duration = Duration::from_secs(300);
 /// sleep can overshoot, understates it.
 const PEER_BLOCK_POLL_INTERVAL: Duration = Duration::from_secs(1);
 
-/// Consecutive passes a peer reader re-reads the same undecodable slot before
-/// giving up and reading past it.
-const DECODE_RETRY_LIMIT: u32 = 3;
+/// Consecutive passes a peer reader spends stuck on one slot before it says so
+/// as something more than the per-pass failure. It never reads past the slot.
+const STUCK_SLOT_ALERT_PASSES: u32 = 3;
 
 /// Why a cross-zone dispatch could not be verified.
 ///
@@ -111,6 +111,7 @@ impl PeerChain {
 
 /// What one consistent look at the peer cache says about a referenced block.
 enum PeerLookup {
+    /// Held, and inside the run verified from the peer's genesis.
     Cached(Box<Block>),
     /// Inside the verified run but not held, so it is not on the peer chain.
     InsideRun,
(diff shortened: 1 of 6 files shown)


Pravdyvy (approved):
lgtm

agureev (approved):

PR #691, opened 2026-08-06
refactor(sequencer): actor architecture, phase 1: RpcServer and Executor actors

Arjentix:
## 🎯 Purpose

First phase of transitioning to the actor-based architecture.

## ⚙️ Approach

- Added `ExecutorActor` which wraps `SequencerCore` handling all sequencer logic
- Added `RpcServerActor` which listens for jsonRPC queries and transmits them to `ExecutorActor`
- Utilized `kameo_actors::Scheduler` to run drive periodical block building instead of a separate tokio task
- `SequencerHandle` rewritten and now contains all three actors and their setup

## 🧪 How to Test

See CI. Or you can try with `just` commands

## 🔗 Dependencies

None

## 🔜 Future Work

There are 4 more phases planed.

But in particular currently `ExecutorActor` still have some weird stuff like `cancellation_token` from BlockPublisher and `background_tasks`. These needs to be removed in future.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


lez/sequencer/actors/executor/src/lib.rs
```diff
@@ -0,0 +1,302 @@
...
+    type Reply = Result<()>;
+
+    async fn handle(
+        &mut self,
+        ProduceBlock: ProduceBlock,
+        _ctx: &mut Context<Self, Self::Reply>,
+    ) -> Self::Reply {
+        // Only produce on our turn.
+        if !self.sequencer.is_our_turn() {
+            info!("Not our turn to produce a block, skipping");
```
erhant, line 280: for logging, should we prefer `log::info` etc.? the code currently has both usages, but I think `log::*` is more readable

Arjentix: I don't quite get what you mean. Do you mean explicit crate name `log` in every macro invocation?

erhant: yeah exactly

erhant: I think it provides a slight readability / pattern advantage when it comes to differentiating the log macro from every other, but nevertheless we have both uses in our repo overall

Arjentix: Actually I think the opposite is good because I want one day to get rid of `log` and use `tracing` instead. `tracing` have `log`-compatible macros, so just changing imports will work.

Although there will be some other changes to improve logs, like instead of:

```rust
info!("Built new block {block_number} with {tx_count} transactions");
```

It's advised to use:

```rust
info!(block_number = block_number, tx_count = tx_count, "New block built");
```

lez/sequencer/actors/executor/src/lib.rs
```diff
@@ -0,0 +1,302 @@
...
+impl Message<Transaction> for ExecutorActor {
+    type Reply = ();
+
+    async fn handle(
+        &mut self,
+        Transaction { transaction }: Transaction,
+        _ctx: &mut Context<Self, Self::Reply>,
+    ) -> Self::Reply {
+        self.mempool_handle
+            .push((TransactionOrigin::User, transaction))
```
erhant, line 122: This may be causing a dead-lock (similar to how we run into the same issue before, where I had introduced a `try_push` with an acknowledgement log), because `push` will block if the mempool is full, and the only thing that can deplete the mempool is a `ProduceBlock` to this same actor (where `self.sequencer.produce_new_block().await?;` is called), so it will be stuck.

Maybe we can return an error instead to indicate that mempool is full, and should be retried later?

lez/sequencer/actors/executor/src/lib.rs
```diff
@@ -0,0 +1,302 @@
...
+    message::{Context, Message},
+};
+use lee_core::{
+    BlockId,
+    account::{Balance, Nonce},
+};
+use log::{info, warn};
+use mempool::MemPoolHandle;
+use sequencer_core::{
+    SequencerCore, TransactionOrigin,
```
erhant, line 19: doesn't this miss the standalone sequencer core? e.g. we had:

```rs
#[cfg(not(feature = "standalone"))]
use sequencer_core::SequencerCore;
#[cfg(feature = "standalone")]
use sequencer_core::SequencerCoreWithMockClients as SequencerCore;
```

lez/sequencer/actors/executor/src/lib.rs
```diff
@@ -0,0 +1,302 @@
...
+impl Message<GetBlockRange> for ExecutorActor {
+    type Reply = Result<Vec<Block>>;
+
+    async fn handle(
+        &mut self,
+        GetBlockRange { range }: GetBlockRange,
+        _ctx: &mut Context<Self, Self::Reply>,
+    ) -> Self::Reply {
+        range
+            .map_while(|block_id| {
```
erhant, line 152: this now returns a shorter range of blocks when a block id is missing, instead of returning an error with `NOT_FOUND_ERROR_CODE` as before, is this handled correctly?

lez/sequencer/service/src/lib.rs
```diff
@@ -107,28 +77,21 @@ impl SequencerHandle {
...
-            store: _,
         } = self;
 
-        // Cloned rather than taken: `stopped()` consumes a handle, and taking
-        // this one would leave `shutdown` with no way to stop the server.
-        let server_handle = server_handle.clone();
-        tokio::select! {
-            () = server_handle.stopped() => {
-                Err(anyhow!("RPC Server stopped"))
+        select! {
```
erhant, line 86: for the `wait_*`'s here, wouldn't `wait_for_shutdown_with_result` be better? we could carry the error object for exactly why the actor had shutdown, instead of the generic `* actor has been stopped`

lez/sequencer/actors/executor/src/lib.rs
```diff
@@ -0,0 +1,302 @@
...
+                Err(anyhow!("The sequencer's block publisher has stopped unexpectedly"))
+            }
+        }
+    }
+
+    async fn on_stop(
+        &mut self,
+        _actor_ref: WeakActorRef<Self>,
+        _reason: ActorStopReason,
+    ) -> std::prelude::v1::Result<(), Self::Error> {
```
erhant, line 104: i guess this is because of `use anyhow::{Result}` above?

lez/sequencer/actors/executor/src/lib.rs
```diff
@@ -0,0 +1,302 @@
...
+            mempool_handle,
+            driver_cancellation,
+            background_tasks,
+        }
+    }
+}
+
+impl Actor for ExecutorActor {
+    type Args = Self;
+    type Error = anyhow::Error;
```
erhant, line 68: I wonder if we should instead define an error enum for each actor, and impl anyhow to that, instead of bailing everything to anyhow

erhant: basically an `errors.rs` for each actor

lez/sequencer/actors/rpc_server/src/lib.rs
```diff
@@ -0,0 +1,105 @@
+//! RPC Server Actor server RPC queries and retranslates them to Executor.
```
erhant, line 1: typo here

`serves..` and maybe `forwards them to`?

erhant (changes requested):
excited for this! several comments 🙏🏻

lez/sequencer/service/src/lib.rs
```diff
@@ -1,103 +1,73 @@
...
-use jsonrpsee::server::ServerHandle;
-use log::{error, info, warn};
-use mempool::MemPoolHandle;
-#[cfg(not(feature = "standalone"))]
-use sequencer_core::SequencerCore;
-#[cfg(feature = "standalone")]
-use sequencer_core::SequencerCoreWithMockClients as SequencerCore;
+use kameo::actor::{ActorRef, Spawn as _};
+use kameo_actors::scheduler::{Scheduler, SetInterval};
+use log::{error, info};
```
schouhy, line 7: Could we keep kameo isolated away from this crate? Using a wrapper of `ActorRef<ExecutorActor>`. So that our dependency on kameo does not lurk everywhere over time and is kept under the actor crate only.

lez/sequencer/actors/executor/src/lib.rs
schouhy: is second this comment

schouhy (approved):
lgtm! agree with @erhant comments.

lez/sequencer/actors/executor/src/lib.rs
Arjentix: Indeed, good idea, fixed that and added a test

lez/sequencer/actors/executor/src/lib.rs
Arjentix: Good idea. Actually while writing a test for your another comment I found out that `anyhow::Error` surprisingly doesn't implement `std::error::Error`, which makes wrapped `kameo::SendError` also not implement `std::error::Error` which is hard to work with in tests. So indeed having strongly typed errors for every actor would also fix that

lez/sequencer/service/src/lib.rs
```diff
@@ -1,103 +1,73 @@
...
 
-        // Nothing this handle owns holds the store, so waiting here rather than
-        // after the drop is the same thing, and it keeps the guarantee inside
-        // the call the caller awaits.
-        wait_for_store_release(&self.store).await;
+        info!("Stopping Executor Actor...");
+        if let Err(err) = executor_ref.stop_gracefully().await {
+            error!("Failed to stop Executor Actor gracefully: {err}");
+        }
+        executor_ref.wait_for_shutdown().await;
```
moudyellaz, line 69: One on ordering: `wait_for_shutdown()` fires when the mailbox closes, which kameo does before `on_stop` runs, so we return before the background tasks and the store are actually gone. Could we use `wait_for_shutdown_result()` (it waits past `on_stop`) and keep a store-release wait? `cross_zone_watcher_restart` reopens the same home right after, so it'd race the RocksDB lock.

lez/sequencer/service/src/lib.rs
```diff
@@ -163,152 +118,59 @@ impl SequencerHandle {
...
+    info!("Executor Actor spawned");
+
+    let rpc_server = RpcServerActor::new(executor_ref.clone(), listen_addr, max_block_size).await?;
+    let addr = rpc_server.addr();
+    let rpc_server_ref = RpcServerActor::spawn(rpc_server);
+    info!("RPC Server Actor spawned");
+
+    let scheduler_ref = Scheduler::spawn(Scheduler::new());
+    scheduler_ref
+        .tell(SetInterval::new(
```
moudyellaz, line 162: `tokio::time::interval` fires its first tick immediately and defaults to `MissedTickBehavior::Burst`, so we produce at startup and catch up back to back after an overrun. `SetInterval::start_delay()` + `set_missed_tick_behaviour()` to match the old cadence?

moudyellaz (approved):
lgtm, thanks! Left some comments/questions.

lez/sequencer/actors/executor/src/lib.rs
moudyellaz: Confirming this one: there's precedent in the repo at `lez/sequencer/core/src/lib.rs:1412`, where the orphan resubmit path uses `try_push` with a comment spelling out that a blocking push stalls the only thing that drains the mempool.

lez/sequencer/actors/executor/src/lib.rs
Arjentix: Ah, missed that weird qualified name. Gone when I implemented strongly-typed errors

lez/sequencer/actors/executor/src/lib.rs
Arjentix: Yes, I removed it but now while writing test I brought it back. Initially I thought it will be easier to start without standalone feature and add it later. But then I thought about removing it completely as I thought we don't use it at all.

So restored currently the same way it was before. But one day I plan to move this bedrock communication part to another actor and then the `BP` generic should be gone and we will be able to control it by just enabling/disabling actor

lez/sequencer/actors/executor/src/lib.rs
Arjentix: Yes, I changed that logic here just because it looks simpler and might still provide existing data to the user if they really need it. Not sure which approach is better, the code just looks cleaner to me

lez/sequencer/service/src/lib.rs
Arjentix: Didn't know that exists, good idea, implemented. Surprisingly this required to make some shenanigans to convert error to something meaningful

lez/sequencer/service/src/lib.rs
Arjentix: Maybe you confuse this crate with `sequencer_core`, no? I don't think it's possible to remove `kameo` from there as we need to define actor topology and initialize them somewhere. And I think this service is exactly the place for it. It basically just sets up actors and that's it.

lez/sequencer/service/src/lib.rs
Arjentix: Wow, indeed looks like so, nice catch. No way to tell this from docs, weird. Didn't know about `wait_for_shutdown_result()` but already did this change as Erhan suggested. But that's yet another reason to use `wait_for_shutdown_result()`

lez/sequencer/service/src/lib.rs
Arjentix: Interesting catch, thanks, fixed

Arjentix: Choosing between `Delay` or `Skip` is hard as they're not so different in our case IMO, as `Scheduler` should not block on something and I expect it to call `interval.tick()` on time. But just in case I set it to `Delay`, so that we won't produce too much blocks too close to each other in some weird situation

lez/sequencer/service/src/lib.rs
schouhy: ah no I meant this crate actually. I was proposing to wrap all the ActorRef< __ > usages here in types defined in the actors crate to keep kameo as an implementation detail of the actor design. That's not feasible?

erhant (approved):
LGTM!

lez/sequencer/service/src/lib.rs
Arjentix: We had a discussion in DM. As I understood the main point it that kameo appears as a very high dependency which make create difficulties in future if we ever decide to move to non-kameo-compatible framework (not even our fork of it).

I agree that having such a heavy dependency on 3rd party crate is not ideal. But this was the lowest effort to achieve actor architecture at current stage. IMO if we ever need our own stuff we can fork it or write our own framework (which will be a very big pain).

Regarding just code encapsulation... I don't think we can fully encapsulate `kameo` in `*_actor` crates and let `sequencer_service` forget about it. There are two reasons:
1. We rely not only on our actors, but also on kameo's `Scheduler`. And there will be others in future. Creating another crate like `sequencer_scheduler_actor` just to wrap `kameo_actors::Scheduler` seems a waste of code. If one day we find them not meeting our needs we can write our own replacement. But for now this looks good IMO.
2. We need to define actor topology somewhere. And it must be the top-level thing. Currently it's encapsulated in `sequencer_service::SequencerHandle` which let's everyone else (like integration tests) use it like before without any changes. But `sequencer_service` must "understand" actor topology.

I tried to refactor it a bit, but what I ended up doing is just new `ActorHandle<T: kameo::Actor>` struct which encapsulates common logic like shutting down. This makes the code a bit easier, but still that's all inside `sequencer_service`.

If there will be any ideas regarding this topic, I'm happy to discuss them and implement in future. For now I propose to close this as you also wrote in DM that you're okay with that.

schouhy: thanks for considering this!

Issue #692, opened 2026-08-06
Any account id can be made permanently unclaimable, so a PDA can be bricked before it exists


moudyellaz:
`pinata` credits an arbitrary account with no claim and no authorization check (`lez/programs/pinata/src/main.rs`): `winner` is `pre_states[1]`, taken straight from `message.account_ids`, and its post state is `AccountPostState::new(winner_post)` with no `Claim` and no `assert!(winner.is_authorized)`. Any 32 bytes can be given a balance of `PRIZE` = 150.

That is enough to make the account permanently unclaimable. `validate_execution` runs before the claim loop assigns ownership (`lee/state_machine/src/validated_state_diff/mod.rs`), so a claiming post state still carries `DEFAULT_PROGRAM_ID` when rule 7 in `lee/state_machine/core/src/program/mod.rs` checks it:

```rust
if post.account.program_owner == DEFAULT_PROGRAM_ID && pre.account != Account::default() {
```

Default owner and non-default value, so every later `new_claimed_if_default` against it fails, for ever. No recovery either: rule 5 forbids decreasing the balance of an account the executing program does not own, and no program owns it.

The target need not exist yet. PDA addresses are public functions of a program id and a seed, so accounts nobody has created can be bricked in advance: every `cross_zone_inbox` seen shard, every `wrapped_token` holding, any PDA whose seed is derivable. Cost is one pinata win each, about 2^24 hashes, and the seeded balance covers about 10,000.

Shipped: `pinata` is registered by `initial_state_testnet()` behind `#[cfg(feature = "testnet")]` and `lez/sequencer/service/Cargo.toml` enables that feature. Nothing restricts which account ids a transaction may name.

Direction to consider: make `pinata` claim the winner the way `authenticated_transfer` does, `Claim::Authorized` on a default-owner recipient, which the claim loop gates on `pre.is_authorized`, so crediting a non-signing account fails the transaction instead of wedging it. Whether rules 6 and 7 should also admit a PDA claim against a pre-state differing from default only in balance is a separate question.

moudyellaz:
[...]

Issue #678, opened 2026-08-04
wrapped_token mint panics on overflow, so one crafted cross-zone message permanently bricks a recipient holding


moudyellaz:
A mint adds to the recipient holding with `.checked_add(amount).expect("wrapped-token balance overflow")` (`lez/programs/wrapped_token/src/main.rs:74`). A panic in the guest is an execution failure, not a recoverable error.

`cross_zone_outbox` asserts only that the caller is a program and writes the payload verbatim, so a program on a peer zone can emit `wrapped_token::Mint { recipient, amount: u128::MAX }` at a victim. Given a route that permits the emitter to reach `wrapped_token`, the first such mint sets the holding near the maximum, and every later legitimate mint to that recipient overflows and fails.

Effect: the victim holding is permanently unusable for inbound bridge transfers. It is a griefing vector rather than theft, and it costs the attacker one message.

Direction to consider: saturate, or fail the delivery cleanly rather than panicking, so a single message cannot poison a holding for good. Note the route pin added in #667 narrows who can reach `wrapped_token`, but does not help where the emitter is legitimately routed.

PR #693, opened 2026-08-07
fix(cross-zone)!: shard the inbox seen-set per peer block and bind deliveries to their source block

### LINKED ISSUE #676: Inbox seen-shard saturates at ~3200 messages per peer per epoch and then panics the guest

moudyellaz:
## 🎯 Purpose

Closes #676 and #678, and takes the half of #677 that #686 left behind.

The inbox records delivered messages in a seen shard per `(src_zone, src_block_id / 10_000)`, holding 32-byte message keys in one 100 KiB account. It saturates near 3,199 keys, after which the guest panics on every further delivery from that peer in that epoch, the sequencer retries three times and drops the message. Permanent message loss, and on a bridge route permanent value loss, reachable by any ordinary user.

Separately, `wrapped_token::Mint` adds to a holding with `checked_add(..).expect(..)`, so one crafted message pins a holding near `u128::MAX` and every later legitimate mint to that recipient fails for good.

This is a breaking change and needs a fresh genesis: both the inbox and wrapped-token image ids move, relocating the inbox config, every seen shard, the wrapped-token config and every holding, and `CrossZoneMessage`'s encoding changes so a pending dispatch written by earlier code no longer decodes.

## ⚙️ Approach

Sharding per exact `src_block_id` alone does not fix #676. A minimal emitting transaction is about 257 bytes, so one peer block can carry thousands of emissions, and the peer's own transaction cap is invisible to us. At 32 bytes per key that overflows a single block's shard.

The shard address already binds `(src_zone, src_block_id)`, so a key stored inside it records nothing the address does not. The shard becomes one account per exact peer block holding a 32-byte source block hash and a set of `u32` transaction indices, which raises the ceiling to 25,591 against an L1 inscription cap that admits roughly 7,100 per block.

The block hash header is #677's residual, done fail-closed. Two peer blocks claiming one block id resolve to the same shard account, so the first binds it and the second aborts, rather than splitting one coordinate into two keys and two deliveries.

- [x] Dead-letter the deliveries this node gives up on, so a lost message is identifiable rather than one log line
- [x] Count and expose them: a metric, a `getCrossZoneDeadLetters` RPC and a dashboard row
- [x] Bind a delivery to the recomputed peer block hash it came from
- [x] Shard the inbox seen-set per peer block, keyed by transaction index
- [x] Cap a single `wrapped_token` mint, and refuse an over-cap lock at the source so it fails before escrowing
- [x] Rebuild artifacts and regenerate the prebuilt fixture

## 🧪 How to Test

```
RISC0_SKIP_BUILD=1 cargo clippy --workspace --all-targets --all-features -- -D warnings
RISC0_DEV_MODE=1 cargo nextest run -p storage -p indexer_core -p cross_zone_inbox_core
RISC0_DEV_MODE=1 cargo nextest run -p integration_tests -E 'binary(cross_zone_state_machine) + binary(cross_zone_ping) + binary(cross_zone_bridge) + binary(cross_zone_verified) + binary(cross_zone_watcher_restart)'
```

## 🔗 Dependencies

None.

## 🔜 Future Work

The indexer's seen-key set is still in memory, so a resumed sync re-derives what a fresh one short-circuits.

Filed #692: any account id can be made permanently unclaimable, which can pre-brick a seen shard or a `wrapped_token` holding before it exists.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


erhant (approved):
lgtm!

one minor comment, since this is already marked breaking; we should maybe mark `rocksdb` folder with a channel id suffix like we do on indexer (e.g. `rocksdb-01010..01`), I think with future tests where we run multiple sequencers and may selectively wipe state etc. it would come in handy

lez/indexer/core/src/cross_zone_verifier.rs
```diff
@@ -63,6 +63,11 @@ pub enum CrossZoneVerifyError {
     },
 }
 
+/// The replay key plus the source block, which is what the inbox treats as one
+/// delivery. Skipping re-derivation on the key alone would wave through a
+/// dispatch the guest refuses, parking the block and holding ingestion.
+type SeenKey = (MessageKey, [u8; 32]);
```
agureev, line 69: Maybe would be good to have a `BlockHash` wrapper instead of opaque 32 bytes?

agureev (approved):
Looks correct, but I am a bit confused regarding the threat model.

I thought at this point, we assumed the sequencer is honest. Can an honest sequencer claim two different shards at the same block ID?

Pravdyvy (approved):
lgtm

lez/indexer/core/src/cross_zone_verifier.rs
moudyellaz: Agreed the opaque bytes deserve a type, but they originate in `CrossZoneMessage` in the guest core, where a newtype moves every image id; a verifier-only wrapper just adds conversions at the boundary. Deferring to the next genesis-moving change to the core types.

moudyellaz:
> Looks correct, but I am a bit confused regarding the threat model.
> 
> I thought at this point, we assumed the sequencer is honest. Can an honest sequencer claim two different shards at the same block ID?

@agureev The actor is the peer's sequencer, not ours. A hostile peer can inscribe two blocks at one height; the shard pins the first-seen block hash (`SeenShard::binds`), so exactly one of them can ever deliver, and since #686 the watcher only delivers from the peer's verified hash-linked run anyway. Our own sequencer never claims two shards at one id: there is one shard per peer and block id, bound to a single hash on first use. The shard is there so peer equivocation cannot reach state even through a buggy watcher, and so the indexer re-derivation stays deterministic.

moudyellaz:
> lgtm!
> 
> one minor comment, since this is already marked breaking; we should maybe mark `rocksdb` folder with a channel id suffix like we do on indexer (e.g. `rocksdb-01010..01`), I think with future tests where we run multiple sequencers and may selectively wipe state etc. it would come in handy

@erhant done in f846e7ac: the sequencer db is now `rocksdb-<channel id>` matching the indexer's naming, with a startup warning naming a pre-suffix directory so an in-place upgrade does not silently fresh-start. The bootstrap test that relied on a foreign channel id finding the old store was reworked to wipe the channel instead, which is the invariant it was actually pinning.

PR #696, opened 2026-08-07
feat(wallet_ffi): Tx status polling

Pravdyvy:
## 🎯 Purpose

Users may want to know inclusion status of an already send transaction.

## ⚙️ Approach

- [x] Adding `wallet_ffi_poll_transaction_status`. It return `bool` value which is `true` if transaction found and false otherwise.

## 🧪 How to Test

Updated `test_wallet_ffi_transfer_public` with poll call. CI must pass.

## 🔗 Dependencies

None

## 🔜 Future Work

When sequecner would be able to sho transaction status, replace `bool` with enum.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


erhant (approved):
lgtm,

I think the correct place for this is the explorer though, and we will soon have that ability (indexer being able to follow non-finalized blocks as well, with sequencer - indexer unification)

Pravdyvy:
> lgtm,
> 
> I think the correct place for this is the explorer though, and we will soon have that ability (indexer being able to follow non-finalized blocks as well, with sequencer - indexer unification)

Well, first of all, when is soon?)

But if not joking, it is kind of a complex topic. Because optimally program developers would like to know causes of failing state transition, which means that failing transactions probably should be recorded somewhere, and it somewhere is unfortunately, probably, sequencers db as we would not like to store it on bedrock.

It is unfortunate, because we will probably have to stall program developers to a point, where we have p2p.

moudyellaz (approved):
lgtm, thanks!

erhant:
> > lgtm,
> > I think the correct place for this is the explorer though, and we will soon have that ability (indexer being able to follow non-finalized blocks as well, with sequencer - indexer unification)
> 
> Well, first of all, when is soon?)
> 
> But if not joking, it is kind of a complex topic. Because optimally program developers would like to know causes of failing state transition, which means that failing transactions probably should be recorded somewhere, and it somewhere is unfortunately, probably, sequencers db as we would not like to store it on bedrock.
> 
> It is unfortunate, because we will probably have to stall program developers to a point, where we have p2p.

I think with the recent developments and the requests, soon = ASAP :D 

In that rework, a very simple map of txhash to failure reasons would also be nice as well.

schouhy (approved):
lgtm

PR #697, opened 2026-08-08
fix(cross-zone)!: make an outbox slot a write-once log entry keyed by its emitter

### LINKED ISSUE #675: Outbox PDA is not namespaced by emitter and is overwritten in place, so it cannot serve as a message log
The outbox PDA seed is domain, `target_zone`, `ordinal` (`lez/programs/cross_zone_outbox/core/src/lib.rs`, `outbox_pda_seed`). It carries no component identifying the emitting program, and the outbox guest asserts only `caller_program_id.is_some()` (`lez/programs/cross_zone_outbox/src/main.rs`), so any program reachable by a chained call may write any `(target_zone, ordinal)` PDA. `OutboxRecord` stores `target_zone`, `target_program_id`, `target_accounts`, `payload`, with no emitter, counter or nonce.

Latent today: delivery is derived by decoding the emitter transaction (`extract_emission`), so nothing reads the PDA and its contents cannot influence a delivery. It is effectively write-only.

It stops being latent under any design that reads the PDA to discover emissions, which is what Phase 1 proposes. Then:

Overwrite. `bridge_lock` writes `(zone_B, 0)` in one transaction; any other program writes `(zone_B, 0)` later in the same block. Reading the post-state yields the second record. The lock is escrowed and never minted, and any peer user can censor a bridge transfer with one transaction.

Splice. Reading identity per transaction while resolving the PDA later can combine `bridge_lock` identity, which the allowlist permits to reach `wrapped_token`, with a payload written by someone else.

Repeat sends are invisible. `ordinal` is caller-supplied and the outbox validates it only against the account the caller passed in (`ping_sender` and `bridge_lock` both forward it unchecked), so a caller may target any slot, including one another emitter is using. Sending the same payload to the same zone and ordinal twice leaves the post-state byte-identical to the pre-state, so change detection sees nothing and the second message is dropped. Today both are delivered under distinct `src_tx_index`.

Two emissions in one transaction collapse. The per-transaction diff is a `HashMap<AccountId, Account>`, so a second write to the same PDA in one transaction overwrites the first before anything can read it.

Directions to consider: put the emitter in the seed and/or `caller_program_id` in `OutboxRecord`, and give the outbox an ordinal discipline (monotonic per emitter and target, or append-only per message) so the post-state is a faithful log rather than a mutable cell. Both are guest changes, so they move program ids and need to land before anything depends on reading the outbox.

moudyellaz:
## 🎯 Purpose

Closes #675.

The outbox PDA seed is `(target_zone, ordinal)` with no emitter, and the guest checks only that a caller exists. So `ping_sender` and `bridge_lock` collide on one slot, `OutboxRecord` does not say who wrote it, and a second `Emit` at the same slot silently overwrites the first, because by then the account is outbox-owned and passes the state machine's rule on data writes.

Latent today: delivery is discovered by decoding the emitter transaction, so nothing reads the PDA and its contents cannot influence a delivery. It stops being latent under any design that reads records to discover emissions, which is why the outbox has to be a faithful log before that work starts.

This is a breaking change and needs a fresh genesis: the outbox image id moves, so every outbox record address changes, and `OutboxRecord`'s borsh layout changes so records written by earlier code no longer decode.

## ⚙️ Approach

- [x] Key a slot by the emitting program as well as the destination zone and ordinal, so two programs cannot share one
- [x] Record the emitter and ordinal, so an entry identifies its writer and its slot without inverting an address that is a hash
- [x] Write a slot once, so a record cannot be overwritten, spliced, or repeated invisibly
- [x] Start the chat demo's ordinals from a free slot rather than zero
- [x] Rebuild artifacts and regenerate the prebuilt fixture

The emitter is `caller_program_id`, the only emitter identity the state machine verifies, since it checks a guest's claimed caller against the real one. It is the immediate chained caller, while cross-zone discovery names the top-level program; the two coincide only while every emitter refuses to be called by another program, which both do.

Write-once is one assert that the pre-state is `Account::default()`, the same predicate the state machine already requires of a first write. It also covers two emissions in one transaction, since a chained callee's pre-states are checked against the running state diff before the underlying state.

`ping_sender` and `bridge_lock` are unchanged. Their compiled bytes move only because both link `cross_zone_outbox_core`.

## 🧪 How to Test

```
RISC0_SKIP_BUILD=1 cargo clippy --workspace --all-targets --all-features -- -D warnings
RISC0_DEV_MODE=1 cargo nextest run -p cross_zone_outbox_core
RISC0_DEV_MODE=1 cargo nextest run -p integration_tests -E 'binary(cross_zone_state_machine) + binary(cross_zone_ping) + binary(cross_zone_bridge) + binary(cross_zone_verified) + binary(cross_zone_watcher_restart)'
```

## 🔗 Dependencies

None.

## 🔜 Future Work

An ordinal is caller-chosen in a namespace every user of an emitter shares, and an emission needs no signature, so a slot can be occupied by anyone and write-once then fails the honest sender's transaction. A client picks a slot the chain does not already hold, which is what the chat demo now does. Removing the contention rather than working around it would mean assigning the ordinal in the guest.

`outbox_program_id` is caller-supplied in both emitters and forwarded to the chained call unchecked, so an emission can occur with no outbox record at all. This guarantees at most one record per slot, not that every emission has one. That gap has to close before anything depends on reading records, since `bridge_lock` debits escrow in its own execution regardless.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


Pravdyvy (approved):
lgtm

agureev (approved):

Issue #673, opened 2026-08-04
Cross-zone inbox config is immutable after genesis, so no peer or route can be added to a running zone


moudyellaz:
The inbox allowlist is written by an `InitConfig` genesis transaction (`lez/cross_zone/src/lib.rs`, `build_inbox_init_config_tx`, called from `build_genesis_state`). The guest applies it init-once: a config PDA that is not `Account::default()` must already hold exactly the same allowlists, otherwise it asserts (`lez/programs/cross_zone_inbox/src/main.rs`, `init_config`).

So `InboxConfig.allowed_routes` is fixed at a zone genesis. Adding a peer, or adding a `(src_program_id, target_program_id)` route to an existing peer, is impossible on a running zone: there is no instruction that can rewrite the config, and re-running `InitConfig` with different allowlists is refused by design (the refusal is deliberate, it stops a post-genesis allowlist hijack).

Two consequences.

Operationally, onboarding a new peer zone or a new cross-zone program pair requires a new genesis, i.e. restarting the zone from scratch.

For the roadmap, it undercuts Phase 1 (emitter-agnostic delivery). Phase 1 aims for "a new cross-zone program needs zero changes to the sequencer or indexer", but delivery stays gated by this genesis-fixed allowlist, so the emitters an emitter-agnostic host would newly recognise are exactly the ones the inbox refuses. Worth resolving before committing to that phase.

Direction to consider: an authorized config-update instruction, in the same shape as the `adminConfigureChannel` roster mechanism, so allowlists can change under an explicit authority rather than not at all.

moudyellaz:
Scope is wider than this issue states.

`ProgramId` is the risc0 image id, computed at build time from the compiled guest (`risc0_binfmt::compute_image_id`, `build_utils/src/lib.rs:51`). So a program's identity is a hash of its own code, and changing one line changes its id. PDAs derive from the program id, so a changed program's accounts all relocate.

`allowed_routes` is a list of `CrossZoneRoute { src_program_id, target_program_id }` fixed by the genesis `InitConfig`. So a changed emitter's id no longer matches its genesis-written route and deliveries from it stop, and a changed inbox relocates the config PDA itself, making the genesis config unreachable.

So it is not only that a new peer or route needs a new genesis. No cross-zone program can be patched on a running zone at all. #675 and #676 are each guest changes, so on a live zone each would need a reset until this lands. #677 turned out not to be one: it shipped host-only in #686.

moudyellaz:
[...]

PR #699, opened 2026-08-09
feat(cross-zone)!: pin the outbox in both emitters and bridge_lock's mint destination

### LINKED ISSUE #698: Emitters forward a caller-supplied outbox program id, so an emission can leave no record
`ping_sender` and `bridge_lock` take `outbox_program_id` as an instruction field and pass it straight into `ChainedCall::new(outbox_program_id, ...)` (`lez/programs/ping_sender/src/main.rs`, `lez/programs/bridge_lock/src/main.rs`). Neither guest checks it, so a caller can chain the emission into any program it likes. The watcher and the verifier both read the emission off the top-level instruction (`extract_emission` in `lez/cross_zone/src/lib.rs`) and never look at the outbox record, so the message is delivered either way: the record can disagree with what was delivered, or not exist at all, while `bridge_lock` debits escrow regardless.

Harmless while nothing depends on records. It stops being harmless the moment anything does, which is why it blocks #673 and record-driven discovery: neither can rely on an emission having been recorded until block inclusion implies it.

A second defect in the same guest: `bridge_lock::Lock` takes `target_program_id` and `target_accounts` from the caller and validates only that the payload is a `wrapped_token::Mint` whose amount equals the locked amount. A wrong target program or a wrong account list debits escrow on the source zone, then fails on the destination, is retried, and is dead-lettered. Nothing releases an escrow, so that balance is burned.

A guest cannot import another guest's image id, so the fix is the genesis-seeded config PDA the repo already uses for the wrapped-token minter (`lez/programs/wrapped_token/core/src/lib.rs`): one config account per emitter pinning the outbox, with `bridge_lock`'s also pinning its mint target and deriving the account list from the payload recipient. `target_zone` stays caller-chosen. A lock to a zone that will not route it burns escrow the same way, but a source-side egress allowlist is a separate decision.

moudyellaz:
## 🎯 Purpose

Closes #698.

`ping_sender` and `bridge_lock` take `outbox_program_id` as an instruction field and pass it unchecked into `ChainedCall::new`, so a caller can route an emission into a program of their choosing. The transaction still succeeds, `bridge_lock` still debits escrow, and the delivery still happens, because discovery reads the top-level instruction and never the outbox record. An emission can therefore leave no record at all, or a record that disagrees with what was delivered.

`bridge_lock::Lock` separately takes `target_program_id` and `target_accounts` from the caller and validates only that the payload mints the locked amount. A wrong value in either debits escrow on the source zone, fails on the destination, and is dead-lettered. Nothing releases an escrow, so that balance is burned.

## ⚙️ Approach

A guest cannot import another guest's image id, so each emitter reads its pins from a genesis-seeded config PDA, the idiom `wrapped_token` already uses for its minter.

- [x] Pin the outbox program in `ping_sender` and drop `outbox_program_id` from `SenderInstruction::Send`
- [x] Pin the outbox program and the mint target in `bridge_lock` and drop `outbox_program_id` from `Instruction::Lock`
- [x] Assert in `bridge_lock` that `target_program_id` and `target_accounts` match the pin and the payload's recipient, before the escrow debit
- [x] Seed both configs by genesis transaction on every zone, since a config PDA left default is claimable by the first initializer
- [x] Regenerate artifacts and the prebuilt sequencer fixture

`target_program_id` and `target_accounts` stay in `Lock` rather than being removed, because `cross_zone::extract_emission` reads them off the transaction and decodes every emitter through one shape.

Both emitters' image ids move, which relocates their PDAs and requires a fresh genesis. Sequencer, indexer and every peer zone have to move together.

## 🧪 How to Test

`RISC0_DEV_MODE=1 cargo test --release -p integration_tests --test cross_zone_state_machine` covers the pins: a send into a foreign outbox slot, a substituted config account, a send and a lock with nothing pinned, a lock naming another target program, a lock naming other mint accounts, and the init-once behaviour of both configs.

## 🔗 Dependencies

#697

## 🔜 Future Work

`target_zone` is still the caller's, so a lock aimed at a zone that will not route it escrows and never mints. It burns only the signer's own balance. Bounding it needs source-side egress configuration and is an input to #673.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


Pravdyvy (approved):
lgtm

lez/programs/ping_sender/src/main.rs
```diff
@@ -21,19 +26,63 @@ fn main() {
...
+        .expect("Send requires the config and outbox accounts");
+
+    // Pinned rather than caller-named: chaining elsewhere would let an emission
+    // skip the real outbox and leave no record of itself.
+    assert_eq!(
+        config.account_id,
+        sender_config_account_id(self_program_id),
+        "first account must be the ping-sender config PDA"
+    );
+    let outbox_program_id = read_outbox(&config.account.data.clone().into_inner())
```
erhant, line 84: `read_outbox` uses a `&[u8]`, do we really need to `clone().into_inner()` here? It costs R0 cycles too

erhant: NOTE: there may be other similar cases in other programs

lez/programs/ping_core/src/lib.rs
```diff
@@ -36,3 +49,65 @@ pub fn ping_record_pda(receiver_id: ProgramId) -> AccountId {
...
+    for (word, chunk) in outbox_program_id.iter().zip(bytes.chunks_exact_mut(4)) {
+        chunk.copy_from_slice(&word.to_le_bytes());
+    }
+    bytes
+}
+
+/// Decodes the pinned outbox program id from the config account's data.
+#[must_use]
+pub fn read_outbox(data: &[u8]) -> Option<ProgramId> {
+    if data.len() < 32 {
```
erhant, line 132: is there a reason to allow `> 32` as well?

since below we do `data[..32]` we could very well check `data.len() == 32` maybe

integration_tests/tests/cross_zone_state_machine.rs
```diff
@@ -94,6 +96,63 @@ fn seed_wrapped_config(state: &mut V03State) {
     )]);
 }
 
+/// Seeds the ping-sender config account pinning the real outbox, matching what
+/// genesis seeds for a real zone.
+fn seed_ping_sender_config(state: &mut V03State) {
+    let sender_id = programs::ping_sender().id();
+    *state = std::mem::replace(state, V03State::new()).with_public_accounts([(
```
erhant, line 116: we have a `test-utils` feature elsewhere:

```rs
#[cfg(any(test, feature = "test-utils"))]
impl V03State {
    pub fn force_insert_account(&mut self, account_id: AccountId, account: Account) {
        self.public_state.insert(account_id, account);
    }
}
```

Maybe we can introduce such things to avoid `std::mem::replace` here?

erhant (approved):
LGTM, two suggestions only

lez/programs/ping_core/src/lib.rs
erhant: ah nvm, my suggestion was wrong

lez/programs/ping_sender/src/main.rs
moudyellaz: Real, and there are more of these (the config reads in wrapped_token and ping_receiver clone the same way). Any fix moves every image id, so rather than rebuild the three stacked branches for it, I will batch this whole class into one cycle-hygiene pass right after the stack merges, same genesis window.

integration_tests/tests/cross_zone_state_machine.rs
moudyellaz: Agreed, a test-utils helper reads better than the mem::replace idiom. Deferring to right after the stack merges, since #700 and #712 extend these helpers and would conflict over the change.

PR #700, opened 2026-08-10
fix(cross-zone)!: authorize deliveries at the target instead of the inbox

### LINKED ISSUE #673: Cross-zone inbox config is immutable after genesis, so no peer or route can be added to a running zone

moudyellaz:
## 🎯 Purpose

Closes #673.

The cross-zone inbox decides who may deliver what. `InboxConfig.allowed_routes` maps each peer zone to the `(src_program_id, target_program_id)` pairs it may use, and the inbox refuses anything else. That makes the inbox the authorization point for value arriving from peer zones, and it is a table only genesis can write, so adding a cross-zone program means a fresh genesis rather than a deployment.

It exists because the inbox destroys the one value a target would need to decide for itself. The chained call carries the target and the payload, so a target sees `caller_program_id = Some(inbox)` and nothing about who on the peer sent it. The central table reconstructs that identity on the target's behalf.

## ⚙️ Approach

The inbox becomes authenticated transport. It passes an unclaimed marker account, whose address derives from `(src_zone, src_program_id)`, at position 0 of the chained call, and each target authenticates its own sources the way in-zone programs already authenticate their callers.

- [x] Derive a source marker PDA in the inbox core and have the inbox require and forward it
- [x] Declare the marker in `build_inbox_dispatch_tx`, so the watcher and the verifier cannot disagree about it
- [x] Give `wrapped_token` a config holding its minter plus the peer sources it will mint for, and check the marker against it
- [x] Give `ping_receiver` the same, which it had no config to do before
- [x] Remove `allowed_routes`, `InboxConfig::permits` and `routes_permit`
- [x] Keep a watcher-only hygiene filter refusing a peer that names a sequencer-only program

Operator configuration is unchanged: the same per-peer routes, fanned into each target's config at genesis rather than into one table.

The marker is never written or claimed and its address is public, so it is not a capability. It means something only because a target checks it after pinning its caller to the inbox.

Every guest image id moves, not only the changed ones, so this needs a fresh genesis with all peers moving together.

## 🧪 How to Test

`RISC0_DEV_MODE=1 cargo test --release -p integration_tests --test cross_zone_state_machine` covers the target checks: a delivery from a source the token did not authorize, a token authorizing nothing, and a directly invoked `Mint`, which must refuse before it looks at the marker.

## 🔗 Dependencies

#699

## 🔜 Future Work

The peer set is still genesis-fixed, so adding a peer needs a fresh genesis. Making it operator-updatable needs the sequencer-only gate split per instruction.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


lez/cross_zone/src/lib.rs
```diff
@@ -252,6 +267,38 @@ pub fn build_bridge_lock_init_config_tx() -> lee::PublicTransaction {
     )
 }
 
+/// The genesis transaction naming the peer sources `ping_receiver` accepts a
+/// delivery from, fanned out of the operator's routes exactly as the wrapped
+/// token's is.
+#[must_use]
+pub fn build_ping_receiver_init_config_tx(
```
erhant, line 274: if i see correctly, this is very similar to `build_wrapped_token_init_config_tx` as well, just a target program id is different

nitpick but, maybe worth grouping the logic here to a helper?

lez/sequencer/core/src/lib.rs
erhant, line 1526:
```diff
- three
```

there are 4 config txes here, could be stale doc

lez/programs/cross_zone_inbox/core/src/lib.rs
```diff
@@ -84,32 +85,20 @@ pub struct CrossZoneMessage {
     pub l1_inclusion_witness: Option<Vec<u8>>,
 }
 
-/// Per-peer delivery routes, plus this inbox's own zone id.
+/// This inbox's own zone id.
+///
+/// It no longer decides who may deliver what. Each target program authorizes its
```
erhant, line 90: we can maybe get rid of past-related doc here

lez/programs/cross_zone_inbox/core/src/lib.rs
```diff
@@ -286,6 +256,48 @@ pub fn inbox_seen_shard_seed(src_zone: &ZoneId, src_block_id: u64) -> PdaSeed {
...
+    bytes[..32].copy_from_slice(&SOURCE_MARKER_SEED_DOMAIN);
+    bytes[32..64].copy_from_slice(src_zone);
+    for (word, chunk) in src_program_id.iter().zip(bytes[64..].chunks_exact_mut(4)) {
+        chunk.copy_from_slice(&word.to_le_bytes());
+    }
+
+    let seed: [u8; 32] = Impl::hash_bytes(&bytes)
+        .as_bytes()
+        .try_into()
+        .unwrap_or_else(|_| unreachable!());
```
erhant, line 298: how does this work in risc0 w.r.t just `unwrap()`?

erhant (approved):
LGTM!

PR #705, opened 2026-08-11
feat!: events

agureev:
## 🎯 Purpose

The purpose of the PR is to introduce the start of the event system based on the specified notion document.

Specifically, every program output now carries an `events` field that encodes a vector of program-specific bytes that can be used by indexing services to understand program behavior easier similarly to other chains. Afterwards, they get inserted as transaction-wide `events` vector where said data gets stamped with the program id that emitted it.

As discussed, private transactions do not commit the events to journal and hence events are not visible to the chain for PPTXs. For a public transaction, events get added via the order of their emission as chained calls.

Excluded in this PR: selector semantics, indexer, RPC work, event placement in existing programs.

## ⚙️ Approach

- [x] Introduce `events` field into `ProgramOutput`
- [x] Introduce `Event` struct
- [x] Make `validate_state_diff` output a vector of `Event`s
- [x] Update current programs to emit empty event vector.

## 🧪 How to Test

See the `tests/events.rs` file

## 🔜 Future Work

Further work, as specified in the [issue](https://app.notion.com/p/LEZ-events-exposure-3a58f96fb65c804cbdbbe4d5bf88984d?v=d358f96fb65c8343ad11880171a1e245&source=copy_link). Next up: selector semantics.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


lee/state_machine/core/src/program/mod.rs
```diff
@@ -445,6 +445,8 @@ pub struct ProgramOutput {
     pub block_validity_window: BlockValidityWindow,
     /// The timestamp window where the program output is valid.
     pub timestamp_validity_window: TimestampValidityWindow,
+    /// A vector of event data.
+    pub events: Vec<Vec<u8>>,
```
moudyellaz: Should there be a protocol-level cap on event count or total size here, given nothing charges for them?

lee/state_machine/core/src/program/mod.rs
```diff
@@ -445,6 +445,8 @@ pub struct ProgramOutput {
     pub block_validity_window: BlockValidityWindow,
     /// The timestamp window where the program output is valid.
     pub timestamp_validity_window: TimestampValidityWindow,
+    /// A vector of event data.
```
moudyellaz: Worth saying here that private execution proves these but discards them, so a program used in both contexts silently emits nothing?

moudyellaz (approved):

lee/state_machine/core/src/program/mod.rs
agureev: Hm... Is there a reason they introduce any sort of problem? The sequencer does not store them and does not push onto the L1, so there is no blockchain storage cost. The computation for events will be covered via fees and does not seem to introduce additional vectors of attack.

The indexer can indeed store these but realistically you should configure the indexer to only notice the events that are interesting to you and drop everything else most probably.

We can, of course, introduce an arbitrary limit but I am having a hard time figuring out what we would prevent there.

lee/state_machine/core/src/program/mod.rs
agureev: done: 7f47398

agureev: after rebase: [e99e254](https://github.com/logos-blockchain/logos-execution-zone/pull/705/commits/e99e25470881edf67448c9e52ae59a6f1380ac3f)

lee/state_machine/core/src/program/mod.rs
```diff
@@ -526,6 +535,16 @@ impl ProgramOutput {
...
 }
 
+/// A struct holding an event-output of a program.
+#[cfg(feature = "host")]
+#[derive(Debug, Clone, PartialEq, Eq, BorshSerialize, BorshDeserialize)]
+pub struct Event {
+    // Which program emitted the event.
+    pub program_id: ProgramId,
+    // The arbitrary event-data emitted in the program output.
+    pub data: Vec<u8>,
```
```suggestion
    /// Which program emitted the event.
    pub program_id: ProgramId,
    /// The arbitrary event-data emitted in the program output.
    pub data: Vec<u8>,
```

schouhy (approved):
LGTM thanks!

lee/state_machine/core/src/program/mod.rs
moudyellaz: The indexer-side filter is the part I can't find: in #709 `apply_block_to_state` collects every event and `write_block_atomic` persists all of them, with no filter and no pruning, which is the TODO in your own #709 description. Fees aren't in yet either. So one of the three has to land before the other two hold.

lee/state_machine/core/src/program/mod.rs
```diff
@@ -445,6 +445,9 @@ pub struct ProgramOutput {
     pub block_validity_window: BlockValidityWindow,
     /// The timestamp window where the program output is valid.
     pub timestamp_validity_window: TimestampValidityWindow,
+    /// A vector of event data. Dropped for private transaction for function
+    /// privacy.
+    pub events: Vec<Vec<u8>>,
```
jonesmarvin8: Is it possible to introduce a datatype for events? E.g., `Event: vec<u8>`? (perhaps `Event` is similar to `Data`)? I'm just thinking in terms of readability of code and to make sure datatypes aren't mixed up.

lee/state_machine/core/src/program/mod.rs
```diff
@@ -526,6 +535,16 @@ impl ProgramOutput {
...
 }
 
+/// A struct holding an event-output of a program.
+#[cfg(feature = "host")]
+#[derive(Debug, Clone, PartialEq, Eq, BorshSerialize, BorshDeserialize)]
+pub struct Event {
+    /// Which program emitted the event.
+    pub program_id: ProgramId,
+    /// The arbitrary event-data emitted in the program output.
+    pub data: Vec<u8>,
```
jonesmarvin8: Slightly confusing since we have a datatype for `Data`.

lee/state_machine/src/state/tests/events.rs
```diff
@@ -0,0 +1,172 @@
+use super::*;
+
+fn program_transaction<T: serde::Serialize>(
+    program_id: ProgramId,
+    account_id: AccountId,
+    instruction: T,
+) -> PublicTransaction {
+    let message =
+        public_transaction::Message::try_new(program_id, vec![account_id], vec![], instruction)
+            .unwrap();
```
jonesmarvin8: Preferred to avoid `unwrap()` in non-test code; at least from my understanding. `expect` returns a message if something strange happens.

lee/state_machine/src/state/tests/events.rs
```diff
@@ -0,0 +1,172 @@
...
+    let witness_set = public_transaction::WitnessSet::for_message(&message, &[]);
+    PublicTransaction::new(message, witness_set)
+}
+
+fn payloads(events: &[lee_core::program::Event]) -> Vec<Vec<u8>> {
+    events.iter().map(|event| event.data.clone()).collect()
```
jonesmarvin8: Is clone necessary here?

lee/state_machine/src/state/tests/mod.rs
```diff
@@ -179,6 +181,12 @@ enum FlashSwapInstruction {
     },
 }
 
+#[derive(serde::Serialize, serde::Deserialize)]
+struct EmitterInstruction {
+    events: Vec<Vec<u8>>,
+    chain: Vec<(ProgramId, Vec<u32>)>,
```
jonesmarvin8: What's Vec<u32> used to convey?

lee/state_machine/core/src/program/mod.rs
jonesmarvin8: Correction: misread the comment, this is a list of (data from events)...not events themselves. Point still stands, but instead of `Event: vec<u8>`, it should be `EventData: vec<u8>` (or `Data`).

jonesmarvin8 (approved):
Lgtm. I left some comments for readability.

lee/state_machine/core/src/program/mod.rs
agureev: Hm, but this does not have the same semantics as `Data`, no? For example, this does not need to be at all constrained by `MAX_DATA_LENGTH`

`Data` already comes with baggage of the impls on top

jonesmarvin8: True, just thinking for readability, it might make sense to use a label.

lee/state_machine/core/src/program/mod.rs
agureev: right, i agree, i still have to fix 709 with this, i had this thought afterwards and will address this now during rebasing

basically, there is no storage problem introduced in *this* specific PR. when rebasing 709, i will introduce a filter-based system similar to application-level indexing for Ethereum, where an indexer can subscribe to a particular program and selector to store them.

archival indexers can then set a flag to record all events

then it is fully dependent on the indexer maintainer to choose what to store, which seems like correct semantics

wdyt?

lee/state_machine/core/src/program/mod.rs
agureev: Fair point, but in #707 this is deprecated in favor of `Vec<ProgramEvent>` which seems to address this concern

Given that, I propose to skip this concern as it will be immediately overturned in the next PR

lee/state_machine/core/src/program/mod.rs
agureev: so I do not think it is a good idea to use `Data`, again, because the semantics differ. a struct's semantics come from its impls. `data: Data` or `data: EventData`, i would argue, would either way not be more readable and imo invite just more questions from the reader. `data: Vec<u8>` is honest and shows the implementation straightforwardly

since the access to the field is either way governed by `.data` access, i think this is the correct level of abstraction

lee/state_machine/src/state/tests/events.rs
agureev: haha I would not say "necessary" but a lot of tests using the `payloads` helper e.g. to compare literal values, than use the un-reffed events, so we would need to compare refs if this were to return a refs, which is ok but kinda weird

if we were to pass a vec to this then we would either way need to clone to create a vec to pass outside

generally, as this is a testing function, i am unsure whether it is worth much more thought

lee/state_machine/src/state/tests/mod.rs
agureev: it's supposed to be the intruction set. you're totally right that this is very opaque, let me change this up

lee/state_machine/src/state/tests/events.rs
agureev: hm but this is indeed a test function

i will nevertheless do an `expect` because it may be theoretically used by somebody in production by mistake

lee/state_machine/src/state/tests/mod.rs
agureev: done: [ed0415d](https://github.com/logos-blockchain/logos-execution-zone/pull/705/commits/ed0415db879f6c6c90372fd3762bfa1629c77139)

lee/state_machine/src/state/tests/events.rs
agureev: done: [4ed6fcc](https://github.com/logos-blockchain/logos-execution-zone/pull/705/commits/4ed6fcc1cda75a79c1ff6ddc50f87cef949e9667)

lee/state_machine/core/src/program/mod.rs
jonesmarvin8: Reasonable.

lee/state_machine/core/src/program/mod.rs
jonesmarvin8: Fair. This can always be a refactor that is brought back up later if need be. But I do agree, it's readable enough for the moment.

lee/state_machine/src/state/tests/events.rs
jonesmarvin8: Sorry about that, for some reason when I left this comment I thought it was outside of a test function. Thanks for the overkill expect inclusion. :-)

PR #707, opened 2026-08-11
feat!(lee): event selectors

agureev:
## 🎯 Purpose

Continuing the event implementation, introduce protocol-level selector bytes into the program-emitted events.

This is closer to EVM and Cosmos based design rather than Solana-based one. This has its merits, e.g. encouraging better design patters and higher servillability.

The convention to follow may be something like the selector being the first 8 bytes of a `sha` of a string `"<program_name>::<event_name>"`.

## ⚙️ Approach
- [x] Define `ProgramEvent`, containing 8 selector bytes
- [x] Define `TransactionEvent` containing the program events and its emitting program
- [x] Add a convention example

## 🧪 How to Test

See `tests/events.rs`

## 🔗 Dependencies

#705 

## 🔜 Future Work

Further event work. RPC/Indexer

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Implement the core functionality
- [x] Add/update tests
- [x] Add/update documentation and inline comments

PR #709, opened 2026-08-11
feat(indexer): event rederivation

agureev:
## 🎯 Purpose

Continue the events integration with indexer and event-storage changes.

In particular, expose events from transactions in blocks when applying them, storing them alongside indices pointing to the specific transaction they were emitted from.

## ⚙️ Approach

- [x] Add `TxEvent` struct to carry info on events with some origin metadata.
- [x] Add event returns during block processing.
- [x] Add functionality to persist gotten events.

## 🧪 How to Test

See `block_store` tests.

## 🔗 Dependencies

#707 

## 🔜 Future Work

Continue event integration.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


lez/storage/src/indexer/read_multiple.rs
```diff
@@ -76,6 +76,51 @@ impl RocksDBIO {
         Ok(block_batch)
     }
 
+    // A block whose transactions emitted nothing has no row at all, so unlike
+    // `get_block_batch_seq` a missing key is expected here and must be skipped rather
+    // than terminate the scan.
+    //
+    // Callers bound the span: this materializes one key per block in `from..=to`, so an
+    // unbounded range allocates proportionally. The policy bound is
+    // `MAX_EVENT_QUERY_BLOCK_SPAN`, enforced by every query surface before calling here.
```
moudyellaz: Should the clamp live inside this function? `MAX_EVENT_QUERY_BLOCK_SPAN` does not exist yet at this commit, and there is no non-test caller to enforce it.

agureev: Yeah, this was born out of splitting a finalized branch into smaller parts. 

Generally, I don't think there should be any restrictions that truncate data in automatic services like the indexer. Silent truncations (or even loud ones for that matter) leave room for a lot of confusion imo

agureev: The const mention is now cleared after the rebase

lez/indexer/core/src/block_store.rs
```diff
@@ -82,6 +82,27 @@ impl IndexerStore {
             .find(|enc_tx| enc_tx.hash().0 == tx_hash))
     }
 
+    pub fn get_events_for_block(&self, block_id: u64) -> Result<Option<Vec<TxEvents>>> {
+        Ok(self.dbio.get_block_events(block_id)?)
+    }
+
+    pub fn get_events_range(&self, from: u64, to: u64) -> Result<Vec<(u64, Vec<TxEvents>)>> {
```
jonesmarvin8: For the u64's referring to block_id could we use/introduce `BlockId` alias?

lez/storage/src/indexer/read_multiple.rs
```diff
@@ -76,6 +76,49 @@ impl RocksDBIO {
...
+    // Callers bound the span: this materializes one key per block in `from..=to`, so an
+    // unbounded range allocates proportionally.
+    pub fn get_block_events_range(
+        &self,
+        from: u64,
+        to: u64,
+    ) -> DbResult<Vec<(u64, Vec<TxEvents>)>> {
+        let cf_events = self.events_column();
+
+        let mut keys = Vec::new();
```
jonesmarvin8, line 93: Would this have issues with blocks that are skipped?

jonesmarvin8 (approved):
Lgtm. minor comments.

schouhy (approved):
lgtm

moudyellaz (approved):
lgtm, thanks!

lez/indexer/core/src/block_store.rs
agureev: fixed on integration branch [here](https://github.com/logos-blockchain/logos-execution-zone/commit/a0f2efbb5fdbaa4dc5c9693f9977acd8fb3a9d4e)

lez/storage/src/indexer/read_multiple.rs
agureev: iirc `validate_against_tip` demands that the incoming block in exactly the next one in the indexing process, so the indexer simply does not have a notion of skipped blocks, it just parks when a block is skipped until the appropriate one arrives.

PR #712, opened 2026-08-12
feat(cross-zone)!: let a target's peer sources be updated by a named authority

moudyellaz:
## 🎯 Purpose

Target source lists are fixed at genesis, so authorizing a new peer source means a fresh genesis. Make them updatable as new zones join: an operator key now, a governance program later, folded onto #700's config format so it changes once.

## ⚙️ Approach

- [x] Give each target config an optional `authority` allowed to change its source list, seeded from the new `CrossZoneConfig.source_authority`, unset by default, which behaves exactly as #700 does
- [x] Add `UpdateSources` and one-way `RenounceAuthority` to both targets, top-level and signed by the authority
- [x] Claim the authority account on first use, so it can act more than once
- [x] Let a configured governance program carry the authority's action on a chained call by delegating its PDA, for the later phase where a PDA holds the authority
- [x] Add an `authority_proxy` test guest that chain-calls an arbitrary target with delegation, which is what makes the governance path testable

Two operator constraints, enforced by nothing but this paragraph: set `source_governance` only with a PDA authority owned by that program, because with a keypair authority any transaction the key signs that reaches the governance program would let it act, authorization being ambient down a call chain. And the authority account must be fresh and never used: any prior history makes it permanently unusable, renounce included.

Every guest image id moves, so this rides the same genesis window as #700.

## 🧪 How to Test

`RISC0_DEV_MODE=1 cargo test --release -p integration_tests --test cross_zone_state_machine` covers the keypair path (repeat updates, renounce), the governance path (update, renounce, per-handler guard negatives), one authority shared by both targets, the inbox's structural inability to reach these instructions, and refusal of an authority account with history.

## 🔗 Dependencies

#700

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


Pravdyvy (approved):
lgtm

schouhy (approved):
lgtm

PR #713, opened 2026-08-12
feat(indexer): RPC for events

agureev:
## 🎯 Purpose

Continue event implementation for LEZ with updates to Indexer's RPC.

I am not at all proficient with anything RPC-ish, so feel free to critique to your hearts' content :)

## ⚙️ Approach

- [x] Update the `Block` schema to `Protocol` schema 
- [x] Separate event and block subscriptions through a `NewSubscription` enum
- [x] Add `get_events` rpc and appropriate filter for `EventRecord`

## 🧪 How to Test

See added module tests.

## 🔗 Dependencies

#709 

## 🔜 Future Work

Finalize event integration via FFI work.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


lez/indexer/service/src/service.rs
```diff
@@ -156,6 +177,50 @@ impl indexer_service_rpc::RpcServer for IndexerService {
...
+                        records.extend(EventRecord::from_tx_events(block_id, group));
+                    }
+                }
+                records
+            }
+        };
+
+        Ok(records
+            .into_iter()
+            .filter(|record| matches_filter(record, &filter))
```
moudyellaz: Worth a cap on returned records too? The span bound is in blocks, so a `program_id`/`selector` query still materializes every event in the range before filtering.

lez/indexer/service/src/service.rs
```diff
@@ -348,6 +478,75 @@ pub fn not_yet_implemented_error() -> ErrorObjectOwned {
...
+fn resolve_block_range(
+    filter: &GetEventsFilter,
+    tip: BlockId,
+) -> Result<(BlockId, BlockId), ErrorObjectOwned> {
+    let Some(from_block) = filter.from_block else {
+        return Err(invalid_params_error(
+            "getEvents requires either `tx_hash` or `from_block`",
+        ));
+    };
+    let to_block = filter.to_block.unwrap_or(tip);
```
moudyellaz: Should an omitted `to_block` clamp to `from_block + MAX_EVENT_QUERY_BLOCK_SPAN` rather than defaulting to tip and then erroring?

lez/indexer/service/src/service.rs
agureev: Yes, introduced a cap in 10 MB, seemingly a standard number. Let me know if this is fine.

lez/indexer/service/src/service.rs
agureev: hm, I would probably be against that. usual behavior of event-bearing RPC will not allow you to recover relevant information. it is rare that application-info is relevant if not at the tip, at least most application-state is malleable.

i think introducing some other semantics will only complicate things. if things cannot be queried up to the tip, then just fail

lez/indexer/service/src/mock_service.rs
```diff
@@ -329,6 +362,39 @@ impl indexer_service_rpc::RpcServer for MockIndexerService {
...
+        // contract as the real service.
+        let state = self.state.read().await;
+        let tip = state.blocks.last().map_or(0, |block| block.header.block_id);
+        let records = state.blocks.iter().filter_map(mock_event_record);
+
+        let records: Vec<EventRecord> = match plan_query(&filter, tip)? {
+            EventQuery::ByTxHash(tx_hash) => {
+                let matched: Vec<EventRecord> =
+                    records.filter(|record| tx_hash == record.tx_hash).collect();
+                if matched.is_empty() {
```
erhant: the real service does:

```rs
let block_id = self
                    .indexer
                    .store
                    .block_id_by_tx_hash(tx_hash.0)
                    .map_err(db_error)?
                    .ok_or_else(unknown_transaction_error)?;
```

which means it returns `unknown_transaction_error` when there is no block for the given tx; and `Ok([])` when the tx exists but has no events

but here, instead of `Ok([])` we return `unknown_transaction_error` when there is no record; it should instead cover the `Ok([])` case too

erhant (approved):
LGTM, one comment only

lez/indexer/service/src/mock_service.rs
agureev: good catch! fix on [64f5981](https://github.com/logos-blockchain/logos-execution-zone/pull/713/commits/64f598107dda05c39e24bcf71256065195b95b56)

erhant: looks good ty!

moudyellaz (approved):
lgtm, thanks!

PR #714, opened 2026-08-12
feat(indexer): expose events over FFI

agureev:
## 🎯 Purpose

Finalize core event integration by FFI updates.

## ⚙️ Approach

- [x] Update FFI

## 🔗 Dependencies

#713 

## 🔜 Future Work

Program event upgrades.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Implement the core functionality
- ~~[ ] Add/update tests!~~
- [x] Add/update documentation and inline comments

PR #716, opened 2026-08-12
feat(bridge): emit deposit/withdraw events

agureev:
## 🎯 Purpose

Add deposit and withdraw emits.

## ⚙️ Approach

- [x] Add event for deposits
- [x] Add event for withdraws in a commented-out state

## 🧪 How to Test

See new `lib.rs` tests

## 🔗 Dependencies

#714

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


lez/programs/bridge/core/src/lib.rs
```diff
@@ -39,6 +40,49 @@ pub enum Instruction {
...
+        borsh::to_vec(self).expect("Deposit serializes")
+    }
+
+    pub fn from_bytes(bytes: &[u8]) -> borsh::io::Result<Self> {
+        borsh::from_slice(bytes)
+    }
+}
+
+#[derive(Debug, PartialEq, Eq, BorshSerialize, BorshDeserialize)]
+pub struct Withdraw {
```
moudyellaz: Keep `Withdraw` here while its only emitter is commented out above a `panic!`, or defer it until withdrawals are re-enabled?

agureev: I think we can keep it here so that whoever will re-enable the withdrawals does not have to do any extra work and understand the event format.

Wdyt?

lez/programs/bridge/core/src/lib.rs
```diff
@@ -39,6 +40,49 @@ pub enum Instruction {
     },
 }
 
+#[derive(Debug, PartialEq, Eq, BorshSerialize, BorshDeserialize)]
+pub struct Deposit {
```
schouhy: can we add this under a `event.rs` submodule?

schouhy (approved):

lez/programs/bridge/core/src/lib.rs
moudyellaz: Makes sense, keep it. Worth a line on `Withdraw` saying the emitter is disabled, so nobody greps the selector and assumes withdraw events already flow?

moudyellaz (approved):
lgtm, thanks!

lez/programs/bridge/core/src/lib.rs
agureev: done on the integration branch [here](https://github.com/logos-blockchain/logos-execution-zone/commit/165d8078a434c972a79e23223323133f2872edca)

lez/programs/bridge/core/src/lib.rs
agureev: comment added [here](https://github.com/logos-blockchain/logos-execution-zone/commit/cbd0595b23bf2e8cc878eeb18655e488ee3271cc)

PR #717, opened 2026-08-12
refactor(key-protocol)!: introduce constitent domain separators for key protocol

jonesmarvin8:
## 🎯 Purpose

Prior to this PR, key protocol have inconsistent domain separators. Moreover, some of the domain separators is LEE version specific. This PR introduces a consistent format of domain separators for (regular) LEE accounts that is independent of LEE version: `LEE-Keys/v1/...`

This guarantees that future updates to LEE protocol (state-machine) does not require changes to the key protocol (in terms of domain separator). The inclusion of a version number (`v1`) does not prevent future changes to the protocol.

Why `v1`? Convenient pick since this is the first time of using "consistent" versioning within the key protocol.

## ⚙️ Approach

Only changes to the domain separators used for regular keys generated using LEE key protocol; group shared accounts not included.

- [X] Changed domain separators.
- [X] Updated tests

## 🧪 How to Test

No new tests introduced; simple updates to key protocol related to reflect the changes. E.g., updates to hardcoded test cases.

## 🔗 Dependencies

N/A

## 🔜 Future Work

N/A

## 📋 PR Completion Checklist

- [X] Complete PR description
- [X] Add/update tests


lee/key_protocol/src/key_management/key_tree/keys_public.rs
```diff
@@ -20,7 +20,7 @@ pub struct ChildKeysPublic {
 impl ChildKeysPublic {
     #[must_use]
     pub fn root(seed: [u8; 64]) -> Self {
-        let hash_value = hmac_sha512::HMAC::mac(seed, "LEE_master_pub");
+        let hash_value = hmac_sha512::HMAC::mac(seed, "/LEE-Keys/v1/Master/Public");
```
agureev, line 23: Several questions:

1) Is there a reason to make it `LEE-Keys` instead of `LEE/Keys`?

2) Do you maybe think it would be better to put the version as the last post-fix as in `...Keys/Master/Public/v1`?

3) Why the `Public` post-fix? I don't think there are any other e.. private master keys. Are any planned? Do these make sense?

lee/key_protocol/src/key_management/secret_holders.rs
```diff
@@ -90,7 +90,7 @@ impl SecretSpendingKey {
     #[must_use]
     #[expect(clippy::big_endian_bytes, reason = "BIP-032 uses big endian")]
     pub fn generate_authorization_secret_key(&self, index: Option<u32>) -> AuthorizationSecretKey {
-        const DOMAIN: &[u8; 35] = b"/LEE/v0.3/Keys/Authorization/Secret";
+        const DOMAIN: &[u8; 33] = b"/LEE-Keys/v1/Authorization/Secret";
```
agureev, line 93: Similar questions here given the prefix revamp

agureev (commented):
Left some comments

lee/key_protocol/src/key_management/key_tree/keys_public.rs
jonesmarvin8: 1. `LEE/Keys` indicates to me that we would have something like `LEE/State`. I'm not opposed to this style of domain separators if we want to consistently label individual components with their own subdoman.

2. I was thinking in terms key protocol version as a whole. I'm not sure how to interpret this suggested reordering.

3. Do you mean have `Master-Public` and `Master-Private`? These keys are meant to be root of the wallet hd keys, so no others are planned.

agureev (approved):

lee/key_protocol/src/key_management/key_tree/keys_private.rs
```diff
@@ -48,7 +48,7 @@ impl ChildKeysPrivate {
         // Each child (of the same parent node) share the same `parent_pt`.
         // To ensure that each child generates unique keys, we include the child index.
         let mut input = vec![];
-        input.extend_from_slice(b"LEE_seed_priv");
+        input.extend_from_slice(b"/LEE-Keys/v1/Seed/Private");
```
schouhy, line 51: why not LEE/Keys?

schouhy (approved):
lgtm thanks. left a comment

lee/key_protocol/src/key_management/key_tree/keys_private.rs
jonesmarvin8: Artem and I discussed this. LEE/Keys to me indicates that LEE/State...etc would follow the same format. If that's the case, then happy to make the change.

schouhy: I'm fine with this version too. Just curious

PR #718, opened 2026-08-13
fix(storage): store pending cross-zone dispatches as per-message entries

### LINKED ISSUE #679: Pending cross-zone dispatch records are one blob rewritten on every change
The pending delivery records live in a single store value holding the whole `Vec<PendingCrossZoneDispatchRecord>` (`lez/storage/src/sequencer/mod.rs`). Every read decodes the whole list and every change rewrites it: recording a peer block delivery, counting a failed production attempt, dropping settled records. Block production decodes the list once per turn.

Cost is O(n) per operation with n the number of pending records, under a mutex shared with the block-write path, so it is also lock-held IO on the path production needs.

Bounded rather than fixed by #652: `MAX_PENDING_CROSS_ZONE_DISPATCHES` is 4096, and recording past it fails so the watcher holds its delivery floor and re-reads the slot later. That caps the blob but does not make the access cheap, and it converts pressure into a stalled peer.

After #686 that stalled-peer consequence is the whole peer channel rather than one slot. `record_block_deliveries` returning false holds the floor, and the durable chain tip is only written after the deliveries are recorded, so a full pending list stops the tip advancing alongside the floor. The watcher then re-reads the same slot indefinitely and nothing from that peer is ever delivered again, rather than one slot waiting for room.

Direction to consider: per-key entries rather than one value, so add, update and remove touch one record. That also subsumes the earlier `TransactionDB` suggestion, since the read-modify-write cycle that the mutex exists to serialise largely disappears.

moudyellaz:
## 🎯 Purpose

Closes #679.

The pending deliveries were one borsh blob decoded and rewritten on every change, under the mutex the block-write path shares; since #686 a full list also stalls the whole peer channel while every operation stays O(n).

## ⚙️ Approach

- [x] Store each record as its own cell keyed by message key, with a count cell maintained in the same write, so add, retry, retire and settle touch only the records involved and the cap check reads one value
- [x] Keep the mutex serializing mutations, now over short read-modify-write cycles instead of whole-list IO
- [x] Keep the cap, all-or-nothing per peer block, dedupe, retirement and settlement contracts unchanged
- [x] Migrate a legacy blob into per-message entries on open, folding additively so a restore into a used store cannot break the cap

Drain order becomes message-key order; the cap's backpressure bounds any delivery's wait. A corrupt legacy blob refuses to open with an error naming the key.

## 🧪 How to Test

`cargo test -p storage`, `RISC0_DEV_MODE=1 cargo test --release -p sequencer_core --features mock`, and the `cross_zone_watcher_restart` e2e for records surviving a real reopen.

## 🔗 Dependencies

#697

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


lez/sequencer/core/src/cross_zone_watcher.rs
```diff
@@ -792,27 +790,36 @@ mod tests {
...
+                .map(|record| record.message_key)
+                .collect(),
+        )
+    }
+
+    /// `keys` as [`recorded_keys`] reports them.
+    fn sorted(mut keys: Vec<[u8; 32]>) -> Vec<[u8; 32]> {
+        keys.sort_unstable();
+        keys
     }
```
Pravdyvy: Most of this functions calls are in tests, so I think better move there and sort explicitly in `recorded_keys` for clarity.

lez/storage/src/sequencer/mod.rs
```diff
@@ -384,9 +389,77 @@ impl RocksDBIO {
...
+    /// dump lands after the open-time pass. Without the legacy key it is a
+    /// no-op read.
+    fn migrate_legacy_pending_dispatches(&self) -> DbResult<()> {
+        let legacy = self
+            .get_opt::<LegacyPendingCrossZoneDispatchesCellOwned>(())
+            .map_err(|err| {
+                DbError::db_interaction_error(format!(
+                    "Legacy pending-dispatch blob at key {DB_META_PENDING_CROSS_ZONE_DISPATCHES_KEY:?} does not decode; delete that key to start without it: {err}"
+                ))
+            })?;
```
Pravdyvy, line 409: I see here that we read `LegacyPendingCrossZoneDispatchesCellOwned` but it looks like we never write into this cell. How we can get some legacy dispatches? Or I missed the moment of a write?

Pravdyvy (commented):
Thank you for the PR. Got confused a little, I do not see writes of a legacy dispatches. Also small comment.

lez/sequencer/core/src/cross_zone_watcher.rs
moudyellaz: Both were already inside `mod tests`; folded the sort into `recorded_keys` and dropped the helper in 54c834a1.

lez/storage/src/sequencer/mod.rs
moudyellaz: Correct, nothing writes it anymore, which is the point: every pre-this-PR build wrote the blob under that key (the old `put_pending_cross_zone_dispatches`), so a store created before the upgrade still carries it on disk; the migration reads it once, converts, and deletes it. A fresh store takes the no-op read. The migration test writes the blob raw to stand in for such a store.

Pravdyvy (approved):
lgtm

agureev (approved):

PR #721, opened 2026-08-13
fix(cross-zone): one peer-block acceptance policy for watcher and verifier

### LINKED ISSUE #674: Sequencer watcher accepts peer blocks the indexer would reject, and the two use different decode-skip budgets

moudyellaz:
## 🎯 Purpose

Closes #674.

The sequencer's watcher and the indexer's verifier must derive identical dispatch sets from the same peer stream, or the indexer calls the sequencer's delivery forged and halts. They judged peer blocks with duplicated code: the same checks in opposite order, different alert cadences, and drift room wherever one side changed and the other did not. Both sides now accumulate per-peer chain state, where any divergence forks derivation permanently.

## ⚙️ Approach

- [x] One `cross_zone::acceptance` module: block screening (hash recomputation, then pinned key), chain linking (genesis anchor, next id, predecessor hash), and the stall counter with one alert cadence, replacing 20 versus 3 with 5
- [x] The watcher adopts it, and gains the verifier's explicit equivocation report for a differing hash at its tip id
- [x] The verifier adopts it: cache admission, run extension and screening route through the same predicates; its run tip now carries its own hash, so a future cache bound cannot freeze the run
- [x] `PeerChainTip` moves to `common` (storage re-exports it, schema unchanged), keeping the policy crate free of rocksdb
- [x] The pure predicate tests and the duplicated chain-builder test helpers are shared

Stated behavior changes, all log or robustness side: unified check order and refusal texts; both alert cadences at 5, including the watcher's stranded alert; the watcher reports equivocation at its tip id where it previously stayed silent (below the tip it has no held hashes, so the verifier remains the louder of the two); the verifier's stall counter no longer resets when a stream truncates before the stalled slot.

Out of scope, deliberately: the two stream loops stay separate (the watcher persists floors and records deliveries mid-loop, the verifier is a pure cache), gap handling keeps its per-side semantics (the watcher skips to keep delivering, the verifier waits for canonicality; both act only on hash-linked runs), and the sequencer-only target filter stays watcher-side only, as its comment already requires.

## 🧪 How to Test

`cargo test -p cross_zone -p indexer_core`, `RISC0_DEV_MODE=1 cargo test --release -p sequencer_core --features mock`, and the `cross_zone_verified`, `cross_zone_watcher_restart` and `cross_zone_ping` e2e.

## 🔗 Dependencies

#697

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests

### DIFF
--- a/lez/cross_zone/src/acceptance.rs
+++ b/lez/cross_zone/src/acceptance.rs
@@ -1,0 +1,472 @@
+//! The one peer-block acceptance policy.
+//!
+//! The sequencer's watcher and the indexer's verifier both admit peer blocks
+//! through here, so they cannot disagree about which block holds an id: a
+//! disagreement makes the verifier re-derive a delivery against a block the
+//! watcher never delivered from, and ingestion halts.
+
+use std::fmt::{self, Display, Formatter};
+
+use common::{
+    HashType,
+    block::{Block, PeerChainTip},
+};
+use cross_zone_inbox_core::ZoneId;
+use lee::{GENESIS_BLOCK_ID, PublicKey};
+
+/// Consecutive passes a reader spends stuck on one slot before it says so as
+/// something more than the per-pass failure. It never reads past the slot.
+///
+/// The cadence bounds log volume only: at one pass per poll interval, 5 passes
+/// is roughly seconds to tens of seconds depending on each side's interval.
+pub const STUCK_SLOT_ALERT_PASSES: u32 = 5;
+
+/// Where a screened peer block sits relative to the chain pinned by `tip`.
+#[derive(Debug, PartialEq, Eq)]
+pub enum Link {
+    /// The next block on the peer's chain, carrying its recomputed hash.
+    Next(HashType),
+    /// At or below the tip, so already accepted from. The ordinary shape of a
+    /// re-read slot. `equivocates` is set when the block claims the tip's own
+    /// id under a different hash, so callers can say so; below the tip there is
+    /// no held hash to compare against and it is never set.
+    AlreadySeen { equivocates: bool },
+    /// Not on the chain the tip pins, so not acceptable. Callers read on: the
+    /// peer's own next block still links to the tip, and treating this as
+    /// terminal would hand the peer a way to stop its deliveries permanently.
+    OffChain(OffChain),
+}
+
+/// Why a block is not on the chain the tip pins.
+#[derive(Debug, PartialEq, Eq)]
+pub enum OffChain {
+    /// The next id off the tip, but linking to some other predecessor.
+    DoesNotLink {
+        block_id: u64,
+        tip_id: u64,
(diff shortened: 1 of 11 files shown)


Pravdyvy (approved):
lgtm

lez/cross_zone/src/acceptance.rs
```diff
@@ -0,0 +1,472 @@
...
+            ),
+        }
+    }
+}
+
+/// The pass-to-pass stall of one peer reader: the slot it is stuck on and how
+/// many consecutive passes it has spent there. Keyed by slot so a failure at a
+/// new slot does not inherit an older slot's count.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct StallState<S> {
```
agureev, line 125: Do I understand correctly that the reason to make this type-dependent is simply not to putt `Slot`'s entire sdk crate with it into the dependencies?

agureev (approved):
looks good, left minor question

lez/cross_zone/src/acceptance.rs
moudyellaz: Yes, exactly. The counter needs only `Copy + PartialEq + PartialOrd` from the slot. Same reasoning as `PeerChainTip` coming from `common` rather than `storage`: the crate stays pure policy, and each adopter instantiates `S` with its own `Slot`.

PR #725, opened 2026-08-14
feat(sequencer): Implement sequencer join/exit queues

schouhy:
## 🎯 Purpose

Computes the committee from finalized state, so a key is added or removed only once the change is on a chain that cannot be orphaned. This covers both joining and leaving.

Joins and exits queue naturally, with no queue to maintain: the wait is just the finality delay.

## ⚙️ Approach

- [x] Build committee updates from finalized state instead of the head
- [x] Add `CommitteeAbsence`: records on which LEZ block each leaving key was first seen gone from the committee
- [x] Read the committee on every tick, on turn or not, so a key seen back in it resets its clock
- [x] Hold back a `FinalizeUnstake` that removes a key until a block built after the removal has finalized
- [x] Add `sequencer_rejoin` integration test: B exits in full, gets its stake back, stakes again, and produces again
- [x] Add unit tests for both gates

### Entry path

1. `Stake` lands in a block. Head state already shows the key above the minimum, but `committee_update` reads finalized state, so nothing happens yet.
2. That block finalizes → the key enters `desired` → `desired != live` → `ChannelConfigOp` submitted, as its own Mantle tx, not bundled with the block.
3. Bedrock includes it; the key appears in the live committee.
4. The round-robin reaches the key → it produces.

One finality delay. A joining sequencer cannot submit the config op itself: only an accredited sequencer on its turn does.

### Exit path

1. `UnstakeRequest` lands in a block. Head state already shows `net_stake() == 0`, but `committee_update` reads finalized state, so nothing happens yet.
2. That block finalizes → the key drops out of `desired` → `desired != live` → `ChannelConfigOp` submitted, as its own Mantle tx, not bundled with the block.
3. Bedrock **includes** it; the key disappears from the live committee — on inclusion, not finalization.
4. `update_committee_absence` sees it gone and stamps `absent_since`. If a later read shows the key back, the stamp is dropped and this returns to step 2: `committee_update` re-fires and the op is submitted again.
5. A block built after that height finalizes →  the removal is considered final.
6. The `FinalizeUnstake` becomes includable and the stake is released.
7. The config entry is gone, so the next observation drops the note. A future exit on this key starts from scratch.

Two finality delays: step 2 to leave the committee, step 5 to get the stake back, with the Bedrock round trip in between.

Partial releases are unaffected. A release that leaves the key at or above the minimum never removes it from the committee, so there is nothing to wait for.

The check only runs on blocks we build. It is not re-checked when adopting someone else's block.

## 🧪 How to Test

```
cargo test -p sequencer_core --lib
```

```
RISC0_DEV_MODE=1 cargo test -p integration_tests --test sequencer_rejoin
```

## 🔗 Dependencies

None.

## 🔜 Future Work

None.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


Pravdyvy (approved):
lgtm

lez/sequencer/core/src/lib.rs
```diff
@@ -662,11 +665,28 @@ impl<BP: BlockPublisherTrait> SequencerCore<BP> {
...
+    ///
+    /// Worth calling off-turn: a key seen back in the committee resets its clock.
+    pub async fn update_committee_absence(
+        &mut self,
+    ) -> Option<Vec<sequencer_stake_core::SequencerKey>> {
+        let live_accredited_keys = self.live_accredited_sequencer_keys().await;
+        if let Some(live) = live_accredited_keys.as_deref() {
+            let leaving = self.with_state(committee_discovery::keys_leaving_the_committee);
+            self.committee_absence
+                .observe(live, &leaving, self.chain_height());
```
moudyellaz: Stamping the LEZ head height rather than the channel tip slot: can a peer's block at H+1, inscribed before the config op landed, open this gate early?

lez/sequencer/core/src/lib.rs
```diff
@@ -662,11 +665,28 @@ impl<BP: BlockPublisherTrait> SequencerCore<BP> {
...
 
+    /// Updates the absence record from a fresh read of the live committee, and
+    /// returns that committee. `None` if the read failed, which is not the same
+    /// as an empty committee.
+    ///
+    /// Worth calling off-turn: a key seen back in the committee resets its clock.
+    pub async fn update_committee_absence(
+        &mut self,
+    ) -> Option<Vec<sequencer_stake_core::SequencerKey>> {
+        let live_accredited_keys = self.live_accredited_sequencer_keys().await;
```
moudyellaz: `accredited_keys()` turns a missing channel state into `Ok(vec![])`, so should an unreadable channel really stamp every leaving key as absent?

lez/sequencer/core/src/committee_discovery.rs
```diff
@@ -70,40 +105,45 @@ pub fn finalize_unstake_candidates(state: &lee::V03State) -> Vec<(lee::AccountId
...
-/// which anyone can inflate — is valid only once the key is no longer
-/// accredited. An unknown account, no pending request, or no config entry
-/// also counts as valid; the program itself rejects those cases anyway.
-///
-/// TODO: checks live Bedrock membership, not an ordered history walk like the
-/// spec calls for. Fine for the sequencer building the next block, but a
-/// follower re-checking an already-adopted block has no independent way to
-/// verify it this way. Switch once zone-sdk exposes ordered `ChannelConfigOp`
-/// data to LEZ.
+// TODO: Only checked on blocks we build, never re-checked on adoption.
```
moudyellaz, line 112: With no re-check on adoption, is this a rule or builder policy a malicious producer can simply ignore?

lez/sequencer/core/src/committee_discovery.rs
```diff
@@ -1,8 +1,43 @@
 //! Discovery process for the `sequencer_stake` committee.
 
+use std::collections::{BTreeMap, BTreeSet};
+
 use log::warn;
 use sequencer_stake_core::{PendingUnstake, SequencerKey, SequencerStakeConfig, StakeRecord};
 
+/// When each leaving key was first seen gone from the live committee.
+#[derive(Debug, Default)]
+pub struct CommitteeAbsence {
```
moudyellaz, line 13: Worth a line saying a restart drops these stamps and costs a pending exit one more finality delay?

moudyellaz (approved):
lgtm, thanks! Left some minor comments

lez/sequencer/core/src/lib.rs
erhant: i agree that this is a valid concern

lez/sequencer/core/src/lib.rs
erhant: this is also a valid problem I think, it stems from https://github.com/logos-blockchain/logos-execution-zone/blob/schouhy/lez-seqeuncer-self-join-finality-queue/lez/sequencer/core/src/block_publisher.rs#L572 where we return default `[]` on error

erhant (changes requested):
two comments to be addressed

lez/sequencer/core/src/committee_discovery.rs
schouhy: yes, but this will be fixed when we move to k of N thresholds and introduce slashing. Such behavior I think can be surfaced with the gossip system and slashed

lez/sequencer/core/src/lib.rs
schouhy: agreed. Fixed in [053bc0c](https://github.com/logos-blockchain/logos-execution-zone/pull/725/commits/053bc0c374cc017e20dde5694d5a1f28283c751c). Also, when [#3272](https://github.com/logos-blockchain/logos-blockchain/pull/3272) lands we'll be able to remove the slot arithmetic altogether in favor tracking the channel config separate linage

lez/sequencer/core/src/lib.rs
schouhy: fixed in [053bc0c](https://github.com/logos-blockchain/logos-execution-zone/pull/725/commits/053bc0c374cc017e20dde5694d5a1f28283c751c). Although I think the branch triggered when reading accredited keys from a non existing channel is unreachable, since sequencers will create the channel on startup if they don't find it on bedrock

lez/sequencer/core/src/committee_discovery.rs
schouhy: added in [053bc0c](https://github.com/logos-blockchain/logos-execution-zone/pull/725/commits/053bc0c374cc017e20dde5694d5a1f28283c751c)

erhant (approved):
LGTM, ty for the fixes!

moudyellaz (approved):

PR #726, opened 2026-08-14
feat!: require authorization for balance decrease

agureev:
## 🎯 Purpose

Continuing LEE environment unification. One thing that we agreed upon is to allow claiming to be permissionless in order to align public and private semantics (permissionless claiming is what happens currently on the private level).

However, permissionless claiming introduces a possible theft vector, where a new regular account gets credited, a malicious actor can then claim the account for a malicious program that allows permissionless balance transfer to them. This theft would be allowed using the current (Solana-like) permission system where decrease of balance is allowed by the owning program.

To close this hole beforehand, this PR changes rule 5 of the protocol to require the account whose balance becomes decreased to be authorized. Now malicious programs cannot spend a regular account's balance without the owner's signature.

NOTE!

- Breaking for program development e.g. for pinata where prior the regular account's balance was spent without signature, now shifted
- Now a regular account's balance can be decreased by a program that does not own it, but only with the owner's signature, which should minimize the attack surface.

## ⚙️ Approach

- [x] Change the pinata program to not store the balance on a regular account. Now, pinata chains to the authorized transfer to disseminate the prize through a PDA.
- [x] Change rule 5 to require authorization for balance spends.

## 🧪 How to Test

See `public_program_rules.rs`

## 🔗 Dependencies

#683 

## 🔜 Future Work

System program and protocol-level authorized program resets.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


lee/state_machine/core/src/program/mod.rs
```diff
@@ -710,10 +710,8 @@ pub fn validate_execution(
 
         let account_program_owner = pre.account.program_owner;
 
-        // 5. Decreasing balance only allowed if owned by executing program
-        if post.account.balance < pre.account.balance
-            && account_program_owner != executing_program_id
-        {
+        // 5. Decreasing balance requires the account to be authorized
```
moudyellaz, line 741: Should this keep the ownership term too, so a debit needs both authorization and `account_program_owner == executing_program_id`?

lee/state_machine/core/src/program/mod.rs
```diff
@@ -710,10 +710,8 @@ pub fn validate_execution(
...
+        if post.account.balance < pre.account.balance && !pre.is_authorized {
             return Err(ExecutionValidationError::UnauthorizedBalanceDecrease {
                 account_id: pre.account_id,
                 owner_program_id: account_program_owner,
```
moudyellaz: Maybe drop `owner_program_id` and `executing_program_id` now that neither takes part in the check?

lee/state_machine/src/state/tests/public_program_rules.rs
```diff
@@ -290,6 +290,41 @@ fn program_should_fail_if_transfers_balance_from_non_owned_account() {
     ));
```
moudyellaz, line 290: Worth a companion test where the sender is signed, so the new permissive direction is pinned rather than just the unsigned one?

lee/state_machine/core/src/program/mod.rs
agureev: Opened a thread to discuss this: https://discord.com/channels/973324189794697286/1537746133886050315

PR #731, opened 2026-08-14
fix(indexer)!: make cross-zone verification halts durable, diagnosable and recoverable

### LINKED ISSUE #647: Cross-zone verifier cannot distinguish a permanently unavailable peer block from a lagging peer

moudyellaz:
## 🎯 Purpose

Closes #647.

A cross-zone verification halt is fail-closed and correct, but blind and irreversible: nothing durable records it, `getStatus` freezes, a restart silently re-derives and re-halts, a dispatch naming a block the peer will never produce hides as ordinary lag forever, one undecodable peer slot bricks that channel while reporting as a retry loop, and the only lever is disabling all verification.

## ⚙️ Approach

- [x] Add a `Halted` indexer state with a persisted halt record naming the local block and the dispatch coordinates, written before the stream ends and reported at boot, cleared only when the recorded block applies
- [x] Add per-peer diagnosis to `getStatus`: verified tip, read cursor, stuck-slot attempts, and a health classification (Live, Lagging, Holed, Halted)
- [x] Escalate a dispatch naming a block above the peer's chain to a permanent verdict only on evidence: the reader drained at the channel tip for the full wait window, with an observed LIB at or past the halting block's inscription slot, so a stale replica can never refute an honest dispatch
- [x] Add `cross_zone_accept_unverified`, an operator accept-list of local block hashes taken before verification, so it also clears dead-peer stalls; acceptance marks the block's dispatch keys seen unverified and can permanently consume a real message's delivery slot, which its doc states

An undecodable peer slot stays unskippable (nothing past a hole can link); it is now diagnosable as Holed rather than log noise.

## 🧪 How to Test

`cargo test -p indexer_core -p indexer_service_protocol -p storage` and the `indexer_stall`, `cross_zone_verified`, `cross_zone_ping` e2e.

## 🔗 Dependencies

#721

## 🔜 Future Work

None

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


Pravdyvy (approved):
lgtm

schouhy (approved):
lgtm

PR #736, opened 2026-08-14
refactor(sequencer): actor architecture, phase 2: Storage Actor

Arjentix:
## 🎯 Purpose

Phase 2 of Actor Architecture. Encapsulate sequencer database into the Storage Actor.

## ⚙️ Approach

A lot of things were done in this PR, here is a quick recap.

- Created `StorageActor` encapsulating `storage` crate usage
- Created `StorageActorTrait` and `MockStorageActor` (with the help of [`mockall`](https://crates.io/crates/mockall) crate). This let's us write powerful tests for client-crates such as `Executor` (will be more in future). See `executor/src/tests.rs` for example.
- SequencerCore now utilizes only `StorageActor` as storage. `storage` crate becomes `StorageActor` implementation detail and once should disappear as a separate crate.
 
## 🧪 How to Test

See CI.

## 🔗 Dependencies

None

## 🔜 Future Work

- Address several TODOs I left on the way. This is not really related to this PR.
- Remove `storage` as it is one day. Make it inner-module of `sequencer_storage_actor`
- Probably remove `sequencer_cells` and just keep protocol types with encoding/decoding as actor inner implementation detail

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


lez/sequencer/actors/storage/src/actor.rs
```diff
@@ -0,0 +1,578 @@
...
+        GetCrossZonePeerFloorBytes, GetCrossZonePeerTip, GetDeadLetterDispatchCount,
+        GetDeadLetterDispatches, GetFinalSnapshot, GetFirstBlockId, GetLastBlockId,
+        GetLatestBlockMeta, GetLeeState, GetPendingCrossZoneDispatches, GetPendingDepositEvents,
+        GetPublishedHighWater, GetZoneAnchor, GetZoneCheckpointBytes, MarkBlockAsFinalized,
+        RaisePublishedHighWater, RecordDispatchFailure, RecordNewBlock, ResetAllBlocksToPending,
+        SetCrossZonePeerFloorBytes, SetCrossZonePeerTip, SetZoneAnchor, SetZoneCheckpointBytes,
+    },
+};
+
+pub trait StorageActorTrait:
```
erhant: could be a nit, but maybe this can go under `trait.rs` and all the other stuff that it requires along with it are exported via `prelude.rs`, so we only need to do `use crate::storage::prelude::*` here?

Arjentix: At first I thought that Rust won't let me use `pub mod trait` because it's a reserved keyword, but then I remembered about raw identifiers and `pub mod r#trait` works. Although forcing all users typing this would be very bad, so I exposed it in `lib.rs` as `pub use r#trait::StorageActorTrait`. So looks good, thanks.

Regarding `prelude::*`... I'm against this pattern in production code and especially for private internal crates. This make sense for public libraries probably, but I see no reason to use it here. Better to explicitly import everything. We even have [`wildcard_imports`](https://rust-lang.github.io/rust-clippy/rust-1.94.0/index.html#wildcard_imports) lint enabled.

lez/sequencer/core/src/lib.rs
```diff
@@ -252,8 +195,41 @@ impl<BP: BlockPublisherTrait> SequencerCore<BP> {
...
+                state,
+                checkpoint_bytes: None,
+            })
+            .await
+            .expect("Failed to seed the database with the genesis block");
+    }
+
+    #[expect(
+        clippy::cognitive_complexity,
+        reason = "Slop has won the battle, but our war is not over"
```
erhant, line 230: ahahhaha

lez/sequencer/actors/executor/src/actor.rs
```diff
@@ -199,30 +216,37 @@ impl<BP: BlockPublisherTrait + Send + 'static> Message<GetBlockRange> for Execut
         _ctx: &mut Context<Self, Self::Reply>,
     ) -> Self::Reply {
         range
-            .map_while(|block_id| {
-                self.sequencer
-                    .block_store()
-                    .get_block_at_id(block_id)
+            .map(|block_id| {
```
erhant: i think using `map` instead of `map_while` here opens up an exploit for the RPC caller to provide a HUGE range, and that entire range will be mapped to `Future`; it will `take_while` from there below but still the entire Future array was created.

we can try a more efficient approach here perhaps 🙏🏻

lez/sequencer/actors/storage/src/protocol.rs
```diff
@@ -0,0 +1,505 @@
...
+    /// Highest block id this event made irreversible: stored blocks at or below
+    /// it become finalized.
+    pub finalized_up_to: Option<BlockId>,
+
+    /// Deposit events observed on L1, recorded unless already pending.
+    pub new_deposit_events: Vec<PendingDepositEventRecord>,
+
+    /// Deposit op ids whose mint finalized: their pending records are dropped.
+    pub remove_deposit_records: Vec<HashType>,
+    /// Deposit op ids whose mint finalized: their pending records are dropped.
```
erhant: typo here about `Deposit` -> `Dispatch`

lez/sequencer/actors/executor/src/error.rs
```diff
@@ -9,9 +9,22 @@ pub enum Error {
     #[error("The mempool is full")]
     MempoolIsFull,
 
-    #[error("Storage error")]
-    StorageError(#[from] storage::error::DbError),
+    #[error("Storage request failed")]
+    StorageRequestFailed(kameo::error::SendError<NoMatter, sequencer_storage_actor::error::Error>),
```
erhant: could we preserve the `#[source]` here as well about the underlying storage error?

lez/sequencer/actors/executor/src/error.rs
```diff
@@ -9,9 +9,22 @@ pub enum Error {
...
+    #[error("Block production failed")]
+    BlockProductionFailed(#[source] anyhow::Error),
+
+    #[error("Failed to read the cross-zone dead letter")]
+    CrossZoneDeadLettersUnavailable(#[source] anyhow::Error),
+}
+
+/// A dummy struct replacing message type in [`kameo::error::SendError`]
+/// as we don't want to expose the message type in the  public API.
+pub struct NoMatter;
```
erhant, line 26:
```diff
- ...in the  public...
+ ...in the public...
```

lez/sequencer/actors/storage/src/protocol.rs
Arjentix: Thanks, copy-pasted wrong line, fixed

lez/sequencer/actors/executor/src/error.rs
Arjentix: Missed that, thanks

lez/sequencer/actors/executor/src/actor.rs
Arjentix: Nice observation. It turned out to be a deeper rabbit hole than I expected. So in the end I made it:

1. Non-executor-blocking by utilizing `DelegatedReply` and `ctx.spawn()`. Added a test for it which was really easy thanks to mocking.
2. Allocating only fixed amount of futures (16) per time by utilizing [`buffered()`](https://docs.rs/futures/latest/futures/stream/trait.StreamExt.html#method.buffered) which I didn't know about before
3. Clamped to 1024 blocks max

lez/sequencer/core/src/block_publisher.rs
```diff
@@ -417,18 +417,15 @@ impl BlockPublisherTrait for ZoneSdkPublisher {
...
                                     on_follow(FollowUpdate {
                                         checkpoint,
                                         adopted,
                                         orphaned,
                                         finalized: finalized_blocks,
                                         deposits,
                                         withdrawals,
-                                    });
+                                    })
+                                    .await;
```
moudyellaz: This drops the `Nothing is awaited here` comment and the property it guarded. What makes awaiting `on_follow` here safe now that the drive loop can't service `Command::Publish` while it runs?

lez/sequencer/core/src/lib.rs
```diff
@@ -792,29 +803,30 @@ impl<BP: BlockPublisherTrait> SequencerCore<BP> {
...
-                    withdrawal_reconciliation_keys,
-                    chain.head_state(),
-                    Some(&checkpoint_bytes),
-                )?;
+                let block_id = block.header.block_id;
+                self.store
+                    .record_new_block(
+                        block,
+                        withdrawal_reconciliation_keys,
+                        chain.head_state().clone(),
```
moudyellaz: `head_state().clone()` plus `final_state().clone()` deep-copies the whole `V03State` on every persist (also 644 and 1667). Can the message carry `Arc<V03State>` instead?

lez/sequencer/core/src/block_store.rs
```diff
@@ -1,127 +1,139 @@
...
         let mut tx_hash_to_block_map = HashMap::new();
 
-        if let Some(last_id) = last_id {
-            log::info!("Preparing block cache");
-            for i in genesis_id..=last_id {
-                let block = dbio
-                    .get_block(i)?
-                    .expect("Block should be present in the database");
+        log::info!("Preparing block cache");
+        for i in genesis_id..=last_id {
```
moudyellaz: One `GetBlock` round trip per block to build the cache, and `restore_chain_state` walks the chain again right after. Could both share a single `GetAllBlocks`?

lez/sequencer/core/src/lib.rs
```diff
@@ -1700,6 +1716,26 @@ fn apply_follow_update(
...
+    let (genesis_state, genesis_txs) = build_genesis_state(config, bootstrap_sequencer_key);
+    let genesis_block = HashableBlockData {
+        block_id: GENESIS_BLOCK_ID,
+        transactions: genesis_txs,
+        prev_block_hash: HashType([0; 32]),
+        timestamp: 0,
+    }
+    .into_pending_block(signing_key);
+
+    sequencer_core_metrics::increment_blocks_produced_total();
```
moudyellaz: Should this increment live in `seed_genesis_if_absent` next to the guard, rather than in a function that builds a value?

lez/storage/src/sequencer/mod.rs
```diff
@@ -448,38 +439,6 @@ impl RocksDBIO {
...
-            DbError::rocksdb_cast_message(
-                rerr,
-                Some("Failed to write first block in db".to_owned()),
-            )
-        })?;
-
-        Ok(())
-    }
-
     pub fn put_meta_last_block_in_db(&self, block_id: u64) -> DbResult<()> {
```
moudyellaz: `put_block`'s `first` arm, `put_meta_last_block_in_db`, `put_lee_state_in_db`, `put_meta_last_finalized_block_id` and `open` have no callers left after dropping `create`. Drop them here?

lez/sequencer/actors/executor/src/tests.rs
```diff
@@ -1,25 +1,33 @@
...
 use crate::{ExecutorActor, protocol};
 
-fn sequencer_config() -> (SequencerConfig, tempfile::TempDir) {
-    let home = tempfile::tempdir().expect("Failed to create tmp home dir");
-
+fn sequencer_config() -> (SequencerConfig, TempDir) {
+    let home = TempDir::new().expect("Failed to create temporary home directory");
     let config = SequencerConfig {
-        home: home.path().to_path_buf(),
+        home: "not-used".into(),
```
moudyellaz: `home` is `not-used` now, so the returned `TempDir` is never written to. Drop it, or point `home` at it?

lez/sequencer/actors/storage/src/actor.rs
erhant, line 1: one note about this actor w.r.t how things were previously;

we had a `Arc`-shared handle where those who needed `dbio` cloned it, and were able to make requests. RocksDB is thread-safe, so there was no issue in calling it from multiple places at once anyways.

now with the storage actor, if I understand correctly we have a single `tokio` worker for that actor, and each handler runs in the same worker; consequentially this will put all RocksDB operations in a sequential order in the order of messages that come to the actor

RocksDB ops are not that expensive so this is ok, but there are several ops like `GetAllBlocks` and `DumpDb` that will keep the thread blocking a bit and their "cost" will grow linearly with the chain length itself

`spawn_blocking` for each call could help maybe, but I'm completely fine with leaving this out of scope of this PR 🙏🏻

erhant (approved):
LGTM

one particular note about the `storage` actor [here](https://github.com/logos-blockchain/logos-execution-zone/pull/736/changes#r3810915154)

lez/sequencer/core/src/block_publisher.rs
Arjentix: Interesting observation, indeed a poorly implemented `on_follow()` could block this task. However IMO this was possible before as well with synchronous code with something like `std::thread::sleep()`.

Not sure we're really protecting from something here, but just in case I now fixed it with `tokio::spawn()`.

Btw rewriting block publisher as an actor is the next step (aka phase 3).

lez/sequencer/core/src/lib.rs
Arjentix: Thanks, after some thinking on this comment I've even closed https://github.com/tqwewe/kameo/pull/396. Agreed, state is heavy, and looks like `Arc` is the only way to go. Fixed

lez/sequencer/core/src/block_store.rs
Arjentix: Indeed. I dug deeper and moved this transaction index into storage actor it-self. I think it makes more sense there as it has full control over the data.

lez/sequencer/core/src/lib.rs
Arjentix: Agreed, moved to `seed_genesis_if_absent`

lez/storage/src/sequencer/mod.rs
Arjentix: Good catch, removed

lez/sequencer/actors/executor/src/tests.rs
Arjentix: Thanks, missed that during another round of refactoring, now home tempdir is used

lez/sequencer/actors/storage/src/actor.rs
Arjentix: Hm, yes, but actually I think this is more a benefit rather than a degradation.

Remember we discussed an issue where multiple writers could potentially break consistency if writing at the same time because we haven't used batches enough? I think now this is solved by strict sequential order of message processing.

Dispatching things like `GetAllBlocks`  into a separate task would mean that result is non-deterministic if we have a mailbox full of messages requiring to write data.

Although I agree that this may lead to writer-starvation... I'd revisit this once we encounter a real issue like that

erhant (approved):

lez/sequencer/core/src/block_publisher.rs
moudyellaz: Ah, sorry, my comment was a question rather than a request, so I think I sent you down a longer path than needed here. One thing to watch with the spawn: the panic at lib.rs:1680 is meant to take the node down on a failed persist, and it does that by killing the drive task so _driver_guard drops. From my understanding, in a spawned task the panic stays local, so that halt no longer happens. There's also a smaller ordering point, two BlocksProcessed deltas can now apply in either order and the older checkpoint can win, though that looks recoverable given the re-delivery idempotency you note at 1519. IMO going back to plain .await would give you both properties back and is the smallest change, and phase 3 turns the publisher into an actor anyway. Happy either way, just didn't want the halt to go quiet.

Arjentix: Ah, I missed that panic, yes, we don't want it to go quiet. Removed spawn and bring back await. Yes, will be fixed in a better way in phase 3

moudyellaz (approved):

PR #737, opened 2026-08-15
feat: slashing v1

schouhy:
## 🎯 Purpose

This PR adds a first slashing mechanism: the committee can burn a sequencer's stake and remove it from the channel.

It covers one offence: inscribing a payload that is not a block.

Threshold is 1 of N for now.

### EDIT
Surfaced by @erhant's [comment](https://github.com/logos-blockchain/logos-execution-zone/pull/737#discussion_r3822003443). This PR also includes a fix (commit [f0e98f1](https://github.com/logos-blockchain/logos-execution-zone/pull/737/commits/f0e98f1fe7f8fb0252881f4c769b45c890dcc5a7)) to how mantle txs are built. 

**The issue**: the block id and the inscription parent were chosen at two different times. The sequencer picked the block id when it built the block, but the parent was filled in later by zone sdk, using whatever the channel tip was when the transaction went out. If the tip changed in between, for example a reorg dropped some blocks, the inscription was still valid for Bedrock and it landed anyway. The result was a block on the channel whose id does not match its position, and it stays there forever. Under slashing this is dangerous, because such a block counts as a fault, and the race could make an honest sequencer produce one.

**The solution**: pick the block id and the parent together, at build time, and freeze both into the transaction. If the tip is still the same when the transaction arrives, the block lands and its id is correct by construction. If the tip changed, Bedrock rejects the transaction and the sequencer just loses that turn. Nothing wrong can land anymore.

For this to work the sequencer tracks the tip itself, in a cursor updated as it processes channel events. The cursor follows every inscription, including garbage and invalid blocks, so junk on the channel does not block publishing. The next valid block just uses it as parent. The cursor is persisted, so restarts keep working.

## ⚙️ Approach

A `Slash` is authorized only by the signatures it carries. The program never learns why the inscription was an offence. It checks that enough accredited keys signed an approval naming the offender and the inscription, then burns.

This keeps a `Slash` valid or invalid based only on the transaction and on chain state, so every node agrees on whether it can be included.

- [x] Add `Instruction::Slash` to `sequencer_stake`, with `SlashApproval` and a signed approval message
- [x] Verify approvals before burning: each signer accredited, no duplicates, count over the threshold
- [x] Burn the whole tracked stake to a sink account and drop the config entry
- [x] Add `slashing.rs` in the sequencer: record offences, ask L1 who wrote them, build the `Slash`
- [x] Report non block inscriptions from the follow path, so no extra scanner is needed
- [x] Persist the offence record so it survives a restart
- [x] Add an integration test where one sequencer is slashed by its peer

## 🧪 How to Test

```
cargo test -p sequencer_core --lib --features mock
RISC0_DEV_MODE=1 cargo test -p integration_tests --test slashing
```

The integration test runs two sequencers. The follower writes garbage and never produces blocks, so only the leader can slash it. It checks the stake moves to the sink, the entry leaves the config, the key leaves the committee, and no block id is skipped.

## 🔗 Dependencies

Depends on #725 

## 🔜 Future Work

- More slashing reasons. 
- Thresholds above 1, which need approvals from peers over gossip.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


Pravdyvy (approved):
lgtm

lez/programs/sequencer_stake/src/main.rs
```diff
@@ -325,6 +346,119 @@ fn unstake_request(
...
+    sequencer_key: SequencerKey,
+    inscription: [u8; 32],
+    approvals: &[SlashApproval],
+) {
+    let message = slash_approval_message(sequencer_key, inscription);
+
+    let mut approvers: Vec<SequencerKey> = Vec::with_capacity(approvals.len());
+    for approval in approvals {
+        assert!(
+            config.entries.contains_key(&approval.signer),
```
moudyellaz, line 365: Authority is any config entry rather than the live committee: at threshold 1, can't anyone stake the 149 minimum with a fresh key and burn every other sequencer?

lez/sequencer/core/src/block_publisher.rs
```diff
@@ -401,13 +429,22 @@ impl BlockPublisherTrait for ZoneSdkPublisher {
...
-                                                    finalized_blocks.push(entry);
+                                                match block_from_inscription(&inscription) {
+                                                    Some(entry) => finalized_blocks.push(entry),
+                                                    // A `ChannelConfig` arrives
+                                                    // empty and offends nobody.
+                                                    None if <Inscription as AsRef<[u8]>>::as_ref(
+                                                        &inscription.payload,
+                                                    )
+                                                    .is_empty() => {}
+                                                    None => undecodable
```
moudyellaz: This makes the offence "this build failed to decode it": what stops a rolling `Block` layout change from slashing every upgraded sequencer?

lez/programs/sequencer_stake/core/src/lib.rs
```diff
@@ -172,6 +196,27 @@ impl SequencerEntry {
     }
 }
 
+/// Bytes an approver signs. Naming the inscription keeps the approval single use.
+#[must_use]
+pub fn slash_approval_message(sequencer_key: SequencerKey, inscription: [u8; 32]) -> Vec<u8> {
```
moudyellaz, line 201: No channel id, no expiry, and nothing on chain records a consumed approval: what stops this signature burning the same key again after a re-stake, or on another zone?

lez/sequencer/core/src/slashing.rs
```diff
@@ -0,0 +1,357 @@
...
+    /// Absent in tests.
+    store: Option<Arc<RocksDBIO>>,
+}
+
+#[derive(Debug, Default, borsh::BorshSerialize, borsh::BorshDeserialize)]
+struct Offences {
+    /// Reported by the follow path, still to be attributed to a writer.
+    unattributed: BTreeSet<([u8; 32], u64)>,
+    /// Never removed: an offending key stays liable for good.
+    found: BTreeSet<(SequencerKey, [u8; 32])>,
```
moudyellaz, line 32: Never pruned, so does a slashed key that re-stakes get burned again automatically by every node that still remembers?

lez/sequencer/core/Cargo.toml
```diff
@@ -58,6 +58,8 @@ testnet = []
 mock = []
 # Enable mDNS-based local peer discovery for gossip.
 mdns = []
+# Expose hooks integration tests need to provoke misbehaviour. Never in a release build.
+test-utils = []
```
moudyellaz: Feature unification means a workspace build turns this on for the node binary too: gate `publish_raw_inscription` on `cfg(test)` instead?

moudyellaz (changes requested):
Keeping a `Slash` valid or invalid from the transaction and chain state alone is right: adoption needs no re-derivation, and the program-side checks (caller is top level, config account identified by id not owner, entry pointed at the right ownership account, duplicate and unaccredited approvers rejected before any signature verification) all hold up.

Approval authority is `config.entries`, not the live accredited committee, and the threshold is 1. An entry appears the moment a `Stake` executes, and the initial stake only has to meet `DEFAULT_MINIMUM_SEQUENCER_STAKE`, which is 149. So anyone with 149 units and a freshly generated Ed25519 key can obtain an approval-capable identity without running a sequencer or ever being accredited on Bedrock, then sign approvals that burn every other sequencer's stake and drop their entries. Their own 149 is not burned. That makes taking over the channel cheaper than any offence slashing is meant to punish.

The offence predicate is `borsh::from_slice::<Block>` failing on the observing build, which is build-dependent rather than objective. Any change to `Block`'s borsh layout makes every block written by an upgraded sequencer an offence in the eyes of a node that has not upgraded yet, and at threshold 1 that node burns the stake on its own. With the number of wire-format changes currently in flight this fires on a rolling upgrade, not in some adversarial corner.

The replay note inline is third but it compounds both: an approval is good forever, on any zone, and `found` is never pruned.

schouhy:
> Approval authority is `config.entries`, not the live accredited committee, and the threshold is 1. An entry appears the moment a `Stake` executes, and the initial stake only has to meet `DEFAULT_MINIMUM_SEQUENCER_STAKE`, which is 149. So anyone with 149 units and a freshly generated Ed25519 key can obtain an approval-capable identity without running a sequencer or ever being accredited on Bedrock, then sign approvals that burn every other sequencer's stake and drop their entries. Their own 149 is not burned. That makes taking over the channel cheaper than any offence slashing is meant to punish.

This PR still lives in the 1 of N threshold world and assumes no sequencer will abuse its L1 key. Going to k of N is not in the scope of this PR. This PR aims to introduce the foundations of slashing to later move in two directions: add more slashing reasons and move to k of N thresholds

lez/programs/sequencer_stake/src/main.rs
schouhy: yes, but it's not in the scope of this PR to cover cases outside of the 1 of N threshold. So this is still insecure in a scenario where at least one sequencer is malicious. The only case covered by slashing is the case of a faulty sequencer that inscribes garbage to the L1 channel.

lez/sequencer/core/src/block_publisher.rs
schouhy: I think this is to be handled when we tackle lez version migrations. But it's out of the scope of this PR. I'd assume for now that all sequencers are running the same lez version

lez/programs/sequencer_stake/core/src/lib.rs
schouhy: we can add the channel id to the message to cover the cross zone case

lez/sequencer/core/src/slashing.rs
schouhy: yes. For simplicity I'm assuming a slashed key is unsafe to be reused for staking again

lez/programs/sequencer_stake/src/main.rs
schouhy: this method is general for any config, but the slash function calls it on a config which passed through `decode_config` method. This would panic if anything but the unique stake config is passed to it

schouhy: the unique stake config will have the correct committee in it. The only difference with bedrock view of the committee are those sequencers that staked but their stake tx is not yet finalized (and so not yet in the bedrock committee). But those sequencers won't be able to gossip txs probably during the entry finality delay, since the gossip system will only allow messages from sequencers in bedrock committe. But it could also be a slashing reason to gossip a tx while not in the bedrock committee (if even possible). So I think this design scales well to k of N thresholds

moudyellaz (approved):
lgtm, thanks for answering the question. I did not realize it is v1 and i remember that we discussed about 1-N and k-N and i thought this would be implemented. But it is out of the scope in this version.

lez/storage/src/sequencer/mod.rs
```diff
@@ -1333,6 +1349,11 @@ impl RocksDBIO {
         if let Some(anchor) = zone_anchor {
             self.put_batch(&ZoneAnchorCell(*anchor), (), &mut batch)?;
         }
+        if let Some(cap) = lower_published_high_water
```
erhant, line 1442: when `run_production_turn` is active in `lib.rs` here, we build a block and then immediately put the new high water

here, we are in an `apply_follow_update` and batch-write things, including the new high water if `mark > cap`

it is possible that there is a race condition here (because RocksDB does "last-write-wins"); here is an example:

1. Suppose the **cell is 10**.
1. A **follow update** lands: head rewound to 7, predicate `10 > 7` passes, reads 10, stages `put_batch(7)` **Cell is still 10**.
1. **Production task**: `publish_block(11)` completes (inscription is on L1) and immediately does `put(11)`. **Cell is now 11**.
1. `store_update` from step 1 finishes, batch commits, applying the staged 7. **Cell is now 7**.

The mark now says 7 while our block-11 inscription is in flight, so nothing holds the turn: we produce 8', 9', 10', 11' ; ending up with two different blocks at the same height

erhant (approved):
LGTM, one important comment!

lez/storage/src/sequencer/mod.rs
schouhy: hey good catch thanks, you are right. And this is specially relevant on slashing, since that kind of race condition would result in a block with invalid id being published by a honest sequencer (hence slashed in the future). I added a fix in commit [f0e98f1](https://github.com/logos-blockchain/logos-execution-zone/pull/737/commits/f0e98f1fe7f8fb0252881f4c769b45c890dcc5a7) and explained it in the PR description.

lez/sequencer/core/src/mock.rs
```diff
@@ -32,6 +38,11 @@ pub struct MockBlockPublisher {
     messages: Vec<(ZoneMessage, Slot)>,
     /// Canned signer; `None` means nothing can be attributed.
     inscription_signer: Option<Ed25519PublicKey>,
+    /// Last entry a publish left the channel at, for the pinned-parent check.
+    channel_tip: Arc<Mutex<Option<MsgId>>>,
+    /// Fails every publish, as zone-sdk does for an atomic withdraw at this
```
erhant: maybe better worded as `Fails every publish if `true`, ...`?

lez/sequencer/core/src/lib.rs
```diff
@@ -1634,12 +1706,14 @@ async fn apply_follow_update<S: StorageActorTrait>(
             );
         }
 
-        // Outcomes align with `adopted`.
-        let outcomes = chain.apply_channel_update(&orphaned, &adopted);
+        // Orphans first, then adopted entries in channel order. Outcomes align
+        // with `adopted_blocks`.
+        let _no_adoptions = chain.apply_channel_update(&orphaned, &[]);
```
erhant: shouldn't we also rewind the tip on orphans?

erhant (approved):
LGTM, one important comment about rewinding the tip on orphans

lez/programs/sequencer_stake/core/src/lib.rs
moudyellaz: Still open on the current head, or landing separately?

lez/chain_state/src/chain.rs
```diff
@@ -97,6 +103,30 @@ impl ChainState {
...
+    /// blocks.
+    #[must_use]
+    pub fn head_tip_msg(&self) -> Option<MsgId> {
+        self.head_blocks.last().map(|entry| entry.this_msg)
+    }
+
+    /// Parent the next inscription must be pinned on: the channel cursor, or
+    /// the head's inscription when no cursor is set.
+    #[must_use]
+    pub fn pin_parent(&self) -> Option<MsgId> {
```
moudyellaz: `restore_head_block` gives a restored block a placeholder `MsgId` derived from its hash, so this fallback can pin on a parent Bedrock never saw. On a fresh node the bootstrap loop at `lib.rs:397` discards the real `MsgId` from the publish outcome, so nothing sets the cursor and block 2 pins on genesis's placeholder.

lez/sequencer/core/src/lib.rs
```diff
@@ -717,11 +765,23 @@ impl<BP: BlockPublisherTrait, S: StorageActorTrait> SequencerCore<BP, S> {
...
-            .block_publisher
-            .publish_block(&block, withdrawals)
-            .await
-            .context("Failed to publish block to Bedrock")?;
+        } = match parent {
+            // Chained on the channel tip the cursor sat on when the block was
+            // built, so a tip that moved since costs this turn instead of
+            // inscribing a second block at a height the channel already
+            // carries.
+            Some(parent) if withdrawals.is_empty() => {
```
moudyellaz, line 753: Blocks carrying withdrawals fall through to the unpinned path, so the wrong-id race stays open for them. Does that also undercut the lowering in 2bb38c43, whose comment assumes an in-flight publish cannot land because it is pinned on a dropped entry?

lez/sequencer/core/src/lib.rs
```diff
@@ -614,8 +640,12 @@ impl<BP: BlockPublisherTrait, S: StorageActorTrait> SequencerCore<BP, S> {
         // Above the final tier the head is reorg-able, so finalized history
         // wins: `apply_finalized` finalizes the matching prefix and rebases the
         // head onto what the channel settled. Validation happens inside it.
-        match chain.apply_finalized(MsgId::from(block.header.hash.0), block, slot) {
-            AcceptOutcome::Applied | AcceptOutcome::AlreadyApplied => {}
+        match chain.apply_finalized(this_msg, block, slot) {
+            AcceptOutcome::Applied | AcceptOutcome::AlreadyApplied => {
+                // The channel replays in order, so this inscription is the tip
+                // so far.
+                chain.set_channel_cursor(this_msg);
```
moudyellaz: Reconstruction runs after the stored cursor is loaded, so re-stamping here rewinds it to the last finalized block on every restart, and the replay skips non-block messages so a config-op tip is never reached. Should the cursor seed from the channel's own `tip_message` instead?

lez/sequencer/core/src/lib.rs
```diff
@@ -186,6 +191,13 @@ impl<BP: BlockPublisherTrait, S: StorageActorTrait> SequencerCore<BP, S> {
                 panic!("Stored block {block_id} does not replay while restoring chain state: {err}")
             });
         }
+        if let Some(cursor) = store
```
moudyellaz, line 193: This unconditional ask has no mock expectation, so `get_block_range_keeps_executor_responsive` panics on `handle_get_channel_cursor`; needs one in `prepare_mock_storage_with_empty_genesis`.

moudyellaz (approved):
lgtm, thanks! I left some questions

lez/sequencer/core/src/lib.rs
schouhy: thanks. using last_msg_id from the checkpoint now

lez/chain_state/src/chain.rs
schouhy: I refactored things and removed the smelly MsgId placeholders. The pin now comes only from the cursor. The bootstrap loop also sets the cursor from its publish outcomes now, so the fresh node case is covered too.

lez/sequencer/core/src/lib.rs
schouhy: withdrawals are still disabled and their tests commented out. So i left them out from this work

lez/sequencer/core/src/lib.rs
schouhy: thanks. the issues should be solved now

lez/sequencer/core/src/lib.rs
schouhy: thanks. fixed

lez/sequencer/core/src/mock.rs
schouhy: fixed

PR #733, opened 2026-08-14
feat(lee)!:  program pubtx - deploy program

jonesmarvin8:
## 🎯 Purpose

This PR introduces `Deploy` program for public transactions. Programs can now be deployed as `Account`s through a native Rust program; this program is executed through native Rust to due to massive computational overhead of forming `Account` representation of programs in Risc0 (without proof generation). This is due to `instruction_data` size including `elf`.

`Deploy` designed with future PRs (program upgradeability) in mind; `update_auth` included in `ProgramData` and PDA id generation.

## ⚙️ Approach

- [X] `ProgramData` struct for `Deploy`
- [X] `Deploy` logic; runs in native Rust instead of guest for practical purposes.
- [X] PDA seed/address derivation

## 🧪 How to Test

New tests:
`loader_deploys_program`
`loader_rejects_redeploying_an_already_deployed_program`
`loader_rejects_invalid_bytecode`
`loader_rejects_wrong_target_account`
`loader_rejects_wrong_number_of_accounts`.
`loader_deploys_program_via_chained_call` (added, but ignored; currently not possible. Part of future works)

## 🔗 Dependencies

Builds off of [PR 720](https://github.com/logos-blockchain/logos-execution-zone/pull/720), [PR 722](https://github.com/logos-blockchain/logos-execution-zone/pull/722), [PR 723](https://github.com/logos-blockchain/logos-execution-zone/pull/723), and [PR 735](https://github.com/logos-blockchain/logos-execution-zone/pull/735).

## 🔜 Future Work

Future PRs will expand on deploying programs:
- Split ProgramData's identity fields from its bytecode into separate accounts, so authenticating a program's identity never requires touching its (potentially large) bytecode.
- Segment bytecode across multiple accounts as deployed programs grow.
- ProgramDeploymentTransaction removal.
- Address issue with calling `Deploy` via chain call.
- Program updateability

## 📋 PR Completion Checklist

- [X] Complete PR description
- [X] Implement the core functionality
- [X] Add/update tests
- [X] Add/update documentation and inline comments

PR #738, opened 2026-08-15
feat(lee)!: program pubtx - anchor privacy-circuit image ids and migrate all programs …

jonesmarvin8:
## 🎯 Purpose

(very out of date description due to a few merges)

This PR builds off PR 733. Programs in Genesis are now deployed using public transactions instead of program transactions. Additionally, this PR anchors privacy transactions to programs stored on chain (genesis and legacy `ProgramDeploymentTransaction`) onto the `Deploy` two-account shape. Unifies every program's dispatch address onto the real Deploy-PDA formula.

Breaking change: Inclusion of `raw_payload` entry for public transaction messages. This is required for a later PR to efficiently handle deploying programs through public transactions. Unused in this branch, but introduced here to avoid breaking changes after RC.

## ⚙️ Approach

- [X] Privacy-preserving circuit: add ProgramImageClaim, anchoring a Deploy-created program's real image_id to its account for env::verify; the sequencer independently re-derives the real image_id via get_program to authenticate the claim.
- [X] Fix compute_public_authorized_pdas and the private circuit's resolve_authorization_and_record_bindings to use the caller's real recovered image_id instead of bijection-guessing it from the caller's account_id — wrong for Deploy-created callers.
- [X] V03State::insert_program (genesis and live ProgramDeploymentTransaction) now always writes the Deploy two-account shape (header + segment).
- [X] Unify every program's dispatch address onto the real Deploy-PDA formula (loader_core::immutable_deploy_account_id), not the bijection AccountId::from(program_id) — including the header address itself, superseding this PR's own initial approach of leaving the header at the bijection address.
- [X] Sweep and fix 7 dispatch-address bugs the bijection assumption was masking: lez/wallet's native_token_transfer facade, integration_tests' auth_transfer/private and private_pda suites, wallet-ffi's generic_transaction FFI boundary, a stale assertion in program_deployment.rs, and a stale expected-error string in cross_zone_state_machine.rs.
- [X] Drop the now-redundant ProgramAlreadyExists pre-check in ProgramDeploymentTransaction validation — PDA claiming already prevents redeploying an account.
- [X] Migrate remaining tests off ProgramDeploymentTransaction onto native Deploy.
- [X] Fix a further PDA-collision vector: `AccountId::for_public_pda`/`for_private_pda` and the privacy circuit's `CallerData` keyed PDA authorization off the caller's image_id (ProgramId) rather than its real dispatch AccountId, so two deployments of identical bytecode with different `update_auth` collided on the same PDA. Retype PDA derivation, wire fields, and comparisons onto AccountId throughout host, guest, and test-harness code.
- [X] Flip `ProgramWithDependencies`'s (privacy-circuit test harness) default dispatch address from the bijection to the real PDA-based address, matching genesis dispatch, and fix every test relying on the old default.
- [X] Remove now-unnecessary `self_program_id`/`outbox_program_id`/etc. wire fields across ping, wrapped_token, bridge, bridge_lock, cross_zone_inbox/outbox, sequencer_stake, amm, and associated_token_account — every guest already receives `self_account_id` for free.
- [X] Fix wrapped_token/ping_receiver's governance-authorization check, which compared the caller against a bijection-derived address instead of the governance program's real dispatch address, silently rejecting any real governance program.
- [X] Fix remaining stray ProgramId→AccountId bijection call sites the earlier sweep missed (cross_zone_state_machine.rs, cross_zone_chat, the example tail-call guest binaries) and a stale test-assertion string.
- [X] Add `raw_payload` to public transaction's messages; this is used in a later PR. Moved up to avoid breaking changes after RC.

## 🧪 How to Test

All previous tests should work.

New tests:
- `loader_deployed_program_is_invocable_via_dispatch`
- `manually_segmented_program_reconstructs_and_executes_identically` — ad hoc proof that a
  program's bytecode split across multiple PDA accounts (using the segment index
  `segment_pda_seed` already accepts) reconstructs and executes identically to the original.
  No production path exercises multi-segment storage yet; see Future Work.

## 🔗 Dependencies
Builds off of [PR 733](https://github.com/logos-blockchain/logos-execution-zone/pull/733).

## 🔜 Future Work

- Move Deploy's bytecode transfer off instruction_data's word-serde encoding, and split out the shared RISC0 kernel so only per-program bytecode needs to be transmitted/stored.
- Segment a deployed program's bytecode across multiple PDA accounts, so the temporarily-raised
DATA_MAX_LENGTH can come back down to 100 KiB.
- ProgramDeploymentTransaction removal, once wallet's deploy-program is repointed at native Deploy.
- Finish removing `ProgramId` and `Program` structures

## 📋 PR Completion Checklist

- [X] Complete PR description
- [X] Add/update tests


schouhy (commented):
The method `transition_from_program_deployment_transaction` is dead code. Not specific to this branch, already dead on `dev`. But can we use this PR to remove it?

lez/programs/program_loader/core/src/lib.rs
```diff
@@ -1,62 +1,79 @@
...
+/// Derives the PDA seed for a deployed program's `ProgramData` header account.
+///
+/// Combines the program's content-derived identity (`image_id`), its position in a (currently
+/// always single-segment) split (`segment_number`), and who may redeploy it (`update_auth`).
+///
+/// Domain-separated from other PDA-seed derivations in the codebase, including
+/// [`deploy_segment_pda_seed`], so a header seed can never collide with a segment seed (or
+/// anything else) even when the input triple coincides.
+#[must_use]
+pub fn deploy_header_pda_seed(
```
schouhy: can we remove the `deploy_*` prefix of all these new functions? I tend to read them as a verb and it confused me: looks like a functions that deploys as opposed to a function that computes a value without mutating anything

lez/programs/program_loader/core/src/lib.rs
schouhy, line 1: Doesnt this program belong to `lee` actually? As the state_machine has special treatment for it

lee/state_machine/core/src/program/mod.rs
```diff
@@ -31,12 +43,51 @@ pub const MAX_NUMBER_CHAINED_CALLS: usize = 10;
...
+///
+/// Lives here rather than in `program_loader_core` so that `V03State::get_program` — a generic,
+/// program-agnostic lookup — can decode it without depending on a specific program's crate.
+/// `program_loader_core` re-exports this type.
+#[derive(Debug, PartialEq, Eq, BorshSerialize, BorshDeserialize)]
+pub struct ProgramData {
+    pub image_id: ProgramId,
+    pub segment_number: u32,
+    pub update_auth: AccountId,
+}
```
schouhy, line 63: I think this should live on the program_loader_core crate. Probably moving that program to `lee` too.

lee/state_machine/core/src/program/mod.rs
```diff
@@ -31,12 +43,51 @@ pub const MAX_NUMBER_CHAINED_CALLS: usize = 10;
...
+/// The account-data layout of a program's header account, deployed via the `Deploy` native
+/// dispatch shortcut (see [`DEPLOYMENT_PROGRAM_ACCOUNT_ID`]).
+///
+/// Deliberately holds only small, fixed-size fields — never the program's bytecode, which lives
+/// in a separate account (see `program_loader_core::deploy_segment_account_id`). Keeping the two
+/// apart means anything that needs to authenticate a program's *identity* (e.g. the
+/// privacy-preserving circuit confirming which `image_id` an `AccountId` currently maps to) only
+/// ever has to read this handful of bytes, not the full program — the only account-authentication
+/// primitive available today is whole-account equality, so what's bundled into one account sets
+/// the floor for how cheap that authentication can be.
```
schouhy: I'm assuming this is a code comment generated by AI. I find such comments very hard to chew. can we use a more direct and simple style?

lee/state_machine/core/src/program/mod.rs
```diff
@@ -31,12 +43,51 @@ pub const MAX_NUMBER_CHAINED_CALLS: usize = 10;
...
+/// new `image_id` into the same, stable `AccountId`) is the entire upgrade mechanism — no
+/// separate "which version" bookkeeping needed.
+///
+/// Lives here rather than in `program_loader_core` so that `V03State::get_program` — a generic,
+/// program-agnostic lookup — can decode it without depending on a specific program's crate.
+/// `program_loader_core` re-exports this type.
+#[derive(Debug, PartialEq, Eq, BorshSerialize, BorshDeserialize)]
+pub struct ProgramData {
+    pub image_id: ProgramId,
+    pub segment_number: u32,
```
schouhy: At first I thought `segment_number` was an index. But here it means the total number of segments for this program, no? If so can we rename the variable?

```rust
    pub number_of_segments: u32, // or `segment_count`?
```

lez/programs/program_loader/core/src/lib.rs
```diff
@@ -1,62 +1,79 @@
...
+    image_id: ProgramId,
```
schouhy: isn't the image_id something that can change with program updates? The header should be a global identifier independent of any version of the program, no?

lez/programs/program_loader/core/src/lib.rs
```diff
@@ -1,62 +1,79 @@
...
+/// Same inputs as [`deploy_header_pda_seed`], domain-separated so the two never collide. Kept as
+/// a distinct account from the header specifically so that authenticating a program's identity
+/// (e.g. for privacy-preserving proof verification) never has to touch its bytecode: the only
+/// account-authentication primitive available is whole-account equality, so what's bundled into
+/// one account sets the floor for how cheap that authentication can be.
 #[must_use]
-pub fn deploy_pda_seed(
+pub fn deploy_segment_pda_seed(
+    image_id: ProgramId,
+    segment_number: u32,
```
schouhy: this segment_number should be the index of this segment no? If a program is split into 10 segments, there will be a pda for each and the seed for those pdas depend on the index of each part. It's confusing that the ProgramData also has a field `segment_number` that I think should be semantically different: not an index, the actual count.

lee/state_machine/src/state/mod.rs
```diff
@@ -281,16 +322,44 @@ impl V03State {
...
     /// this is the single place that distinction is enforced, so callers never have to remember
     /// to re-check it themselves.
     #[must_use]
-    pub fn get_program(&self, program_account_id: AccountId) -> Option<&Account> {
+    pub fn get_program(&self, program_account_id: AccountId) -> Option<(ProgramId, Vec<u8>)> {
         let account = self.get_account_by_id_ref(program_account_id)?;
-        (account.program_owner == PROGRAM_STORAGE_OWNER).then_some(account)
+        if account.program_owner == PROGRAM_STORAGE_OWNER {
+            return Some((ProgramId::from(program_account_id), account.data.to_vec()));
+        }
```
schouhy: there are no paths to create an account with `PROGRAM_STORAGE_OWNER` as programowner. Let's remove it in favor of the new program loader. With a header that has only one segment

lee/state_machine/src/state/mod.rs
```diff
@@ -281,16 +322,44 @@ impl V03State {
...
+        if account.program_owner == DEPLOYMENT_PROGRAM_ACCOUNT_ID {
+            let header = ProgramData::try_from(&account.data).ok()?;
+            let segment_account_id = program_loader_core::deploy_segment_account_id(
+                DEPLOYMENT_PROGRAM_ACCOUNT_ID,
+                header.image_id,
+                header.segment_number,
+                header.update_auth,
+            );
```
schouhy: Shouldn't this read the header segment number and iterate through all segments and concatenate them?
If we want to postpone supporting multi segment programs let's just add a comment and a warning in case this sees header of a program that claims to be multi segmented

lee/state_machine/core/src/program/mod.rs
schouhy, line 1: not specifically from this branch, but the program is called `program_loader` which makes sense, but the constant with its reserved id is called `DEPLOYMENT_PROGRAM_ACCOUNT_ID`. Let's be more consistent with it. Let's rename the const to `PROGRAM_LOADER_ACCOUNT_ID`.

lez/programs/program_loader/core/src/lib.rs
```diff
@@ -72,23 +89,54 @@ pub fn deploy_pda_seed(
...
+/// upgrade authority.
+///
+/// `segment_number` 0, `update_auth` `AccountId::default()`. What every genesis-seeded builtin,
+/// and any `Deploy` submitted with a default `update_auth`, dispatches at.
+#[must_use]
+pub fn immutable_deploy_account_id(image_id: ProgramId) -> AccountId {
+    deploy_header_account_id(
+        lee_core::program::DEPLOYMENT_PROGRAM_ACCOUNT_ID,
+        image_id,
+        0,
```
schouhy, line 121: shouldn't this be 1? it has one segment.

schouhy (changes requested):
Hey thanks for the PR. I'm leaving some comments. Overall I have the impression that the implementation attempts to set the ground for multi-segmented, but it's not clear to me that what's left for later is just adding the paths to exercise these. There are a few details in the design that I don't understand.

IMO If this PR is what will land in RC I think it should have:
- Only one owner of program accounts: the program loader program (delete `STORAGE_PROGRAM_OWNER`)
- A design that supports multisegmented programs.

Last one tested in unit tests with multi segment test programs. Leaving the machinery to actually exercise it on the live system for later. Otherwise it's very hard to understand if the approach is sound and it feels incomplete

lez/programs/program_loader/core/src/lib.rs
jonesmarvin8: Taken care of.

lee/state_machine/src/state/mod.rs
jonesmarvin8: Fixed.

lee/state_machine/core/src/program/mod.rs
jonesmarvin8: Renamed.

lez/programs/program_loader/core/src/lib.rs
jonesmarvin8: This is actually updated later in the stack. I don't think this update is necessary now though. But if you want, I can include the update version of `ProgramData` in this PR: it includes `genesis_image_id`, and `genesis_update_auth` which are the values for `image_id` and `update_auth` at the time of deployment.

Let me know what you think.

lez/programs/program_loader/core/src/lib.rs
jonesmarvin8: Same as previous [comment](https://github.com/logos-blockchain/logos-execution-zone/pull/738#discussion_r3876712034).

lez/programs/program_loader/core/src/lib.rs
jonesmarvin8: For context of how I was thinking...
Program updates would have `ProgramData` header account and then accounts that contain the various program segments.
ProgramData account would be marked as 0...then each segment account itself numbered.

Looks a little weird here since we're not using the segments. Let me know what you think

lee/state_machine/core/src/program/mod.rs
jonesmarvin8: Renamed.

lez/programs/cross_zone_inbox/core/src/lib.rs
```diff
@@ -106,8 +109,13 @@ pub struct CrossZoneMessage {
     /// without either trusting what the peer wrote.
     pub src_block_hash: [u8; 32],
     pub src_tx_index: u32,
-    pub src_program_id: ProgramId,
+    /// The emitting program's dispatch address on the peer zone (its `OutboxRecord.emitter`).
+    pub src_account_id: AccountId,
     pub target_program_id: ProgramId,
```
agureev: Hm, shouldn't the target accountID completely replace the program ID?

Isn't the entire point that the protocol now does not worry about ProgramIDs and that it is an implementation detail?

lez/programs/cross_zone_inbox/core/src/lib.rs
```diff
@@ -30,8 +31,10 @@ pub type MessageKey = [u8; 32];
...
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize, BorshSerialize, BorshDeserialize)]
 pub struct CrossZoneRoute {
-    /// The program on the peer zone that emitted the message.
-    pub src_program_id: ProgramId,
+    /// The dispatch address of the program on the peer zone that emitted the message (matched
+    /// against its `OutboxRecord.emitter`, itself the peer's state-machine-verified caller
+    /// address — not derivable from any `ProgramId`).
+    pub src_account_id: AccountId,
     /// The program on this zone it may be delivered to.
     pub target_program_id: ProgramId,
```
agureev: Similar concern here

lee/state_machine/test_methods/guest/src/bin/flash_swap_callback.rs
```diff
@@ -24,16 +24,20 @@
...
 };
 
 #[derive(borsh::BorshSerialize, borsh::BorshDeserialize)]
 pub struct CallbackInstruction {
     /// If true, return the borrowed funds to the vault (happy path).
     /// If false, keep the funds (simulates a malicious callback, triggers rollback).
     pub return_funds: bool,
-    pub token_program_id: ProgramId,
+    /// The dispatch address of the token program.
+    pub token_program_id: AccountId,
```
agureev: should this be `token_account_id`?

lee/state_machine/src/state/mod.rs
```diff
@@ -112,6 +112,12 @@ impl BorshDeserialize for NullifierSet {
 #[derive(Clone, PartialEq, Eq, BorshSerialize, BorshDeserialize)]
 #[cfg_attr(test, derive(Debug))]
 pub struct V03State {
+    /// Deployed programs live here too, as `Account`s findable via [`Self::get_program`]:
```
agureev: I feel like this comments is PR-specific. Wdyt? Maybe better to remove it?

lee/state_machine/src/public_transaction/message.rs
```diff
@@ -12,6 +12,20 @@ pub struct Message {
     pub account_ids: Vec<AccountId>,
     pub nonces: Vec<Nonce>,
     pub instruction_data: InstructionData,
+    /// An optional large raw byte payload, carried alongside `instruction_data` rather than
+    /// packed into it.
+    ///
+    /// `instruction_data` is word-serialized (`risc0_zkvm::serde`) so it can be read inside a
+    /// RISC0 guest; that format encodes a `Vec<u8>` at 4 bytes per word, since it doesn't route
```
agureev: Wait, aren't we skipping the RISC0 guest? Why do we need to encode it this way then?

lee/state_machine/src/public_transaction/message.rs
```diff
@@ -12,6 +12,20 @@ pub struct Message {
...
+    /// through the serializer's `serialize_bytes`. `Message` itself, in contrast, is
+    /// borsh-encoded on the wire, which packs `Vec<u8>` byte-for-byte. So a large payload that
+    /// only needs to reach *native* dispatch logic (never a real guest — e.g. `Deploy`'s program
+    /// bytecode, which is handled natively, see `PROGRAM_LOADER_ACCOUNT_ID`'s doc
+    /// comment) belongs here instead of in `instruction_data`, avoiding that ~4x bloat entirely
+    /// rather than just packing it more efficiently.
+    ///
+    /// Unused for now — no constructor sets it and no dispatch logic reads it yet.
+    pub raw_payload: Option<Vec<u8>>,
```
agureev: This seems like a bad change to make for the RC. Breaking changes for wire format are very normal. We should not just plug unused fields just because the wire format may change on the frontend.

integration_tests/tests/auth_transfer/private.rs
```diff
@@ -672,6 +704,7 @@ async fn prove_init_with_commitment_root(
     let recipient_account_id = AccountId::for_regular_private_account(&npk, &vpk, 0);
     let recipient = AccountWithMetadata::new(Account::default(), true, recipient_account_id);
 
+    let program_id = program.id();
```
agureev: a nit and/or stupid question: but why abstract into a varable if used only once?

integration_tests/tests/auth_transfer/public.rs
```diff
@@ -387,12 +396,27 @@ async fn cannot_execute_faucet_program() -> Result<()> {
...
+        faucet_chain_caller_segment,
+        bytecode,
     ));
+
+    let tx_size = encoded_tx_size(&deploy_tx);
+    let ctx = MultiZoneTestContextBuilder::default()
+        .with_zone(
+            ZoneTestContextBuilder::new(MultiNodeTestContextConfig::default())
+                .with_sequencer_partial_config(SequencerPartialConfig {
+                    max_block_size: ByteSize::b(tx_size + 10 * 1024),
```
agureev: Why specifically this? Why not set some max size? Is that the tx size + clock tx? Maybe I am being stupid here...

integration_tests/tests/cross_zone_ingress_guard.rs
```diff
@@ -45,22 +45,25 @@ async fn user_origin_inbox_call_rejected() -> Result<()> {
...
+    let ping_receiver_id = programs::ping_receiver().id();
     let msg = CrossZoneMessage {
         src_zone: [2; 32],
         src_block_id: 1,
         src_block_hash: [7; 32],
         src_tx_index: 0,
-        src_program_id: [9; 8],
-        target_program_id: programs::ping_receiver().id(),
+        src_account_id: lee::AccountId::new([9; 32]),
+        target_program_id: ping_receiver_id,
```
agureev: Returning to the point, seems like this should just use the rager account id

lee/state_machine/core/src/program/mod.rs
jonesmarvin8: Made an initial pass (with AI) to make comments simpler (without losing meaning). I'll go back through and manually work on the comments as well.

lee/state_machine/core/src/program/mod.rs
jonesmarvin8: Agreed. Can I save this for the moment? (before PR merged, it'll be done)

lez/programs/program_loader/core/src/lib.rs
jonesmarvin8: Agreed. Can I save this for the moment? (before the PR is merged, make the change then)

lee/state_machine/src/validated_state_diff/mod.rs
```diff
@@ -543,10 +538,30 @@ fn authenticate_public_transaction_signers(
...
+    // Anchor each claimed image_id to real chain state: reconstruct the claims using the
+    // program's *actual* current image_id (via `get_program`), not the message's own claim. If
+    // the claim was wrong, the reconstructed journal won't match what the receipt actually
+    // committed to, and `proof.is_valid_for` below fails — the same mechanism `public_actions`
+    // already relies on for authenticating account content against real state.
+    let program_image_claims = message
+        .program_image_claims
+        .iter()
+        .map(|claim| {
+            let (image_id, _elf) = state.get_program(claim.account_id).ok_or_else(|| {
```
erhant: maybe we can add a helper to read the image id only but not the huge `elf` here?

lee/state_machine/core/src/program/mod.rs
```diff
@@ -16,27 +16,71 @@ pub const DEFAULT_PROGRAM_ID: ProgramId = [0; 8];
...
+        BorshDeserialize::try_from_slice(data.as_ref())
+    }
+}
+
+impl From<&ProgramData> for crate::account::Data {
+    fn from(program_data: &ProgramData) -> Self {
+        let mut data = Vec::with_capacity(std::mem::size_of_val(program_data));
+        BorshSerialize::serialize(program_data, &mut data)
+            .expect("borsh serialization should not fail");
+        Self::try_from(data).expect("elf must fit under DATA_MAX_LENGTH")
```
erhant: i think this is not `elf must fit` but more like `program header must fit`

lee/state_machine/src/validated_state_diff/mod.rs
jonesmarvin8: Added helper function `get_program_image_id`.

lee/state_machine/core/src/program/mod.rs

jonesmarvin8:
Stale PR; this approach has been superseded by the linked-list approach of PR 812.

PR #741, opened 2026-08-16
test: add cucumber environment with initial ported integration tests

hansieodendaal:
## 🎯 Purpose

*What problem does this PR solve or what feature does it add? Mention issues related to it*

This PR introduces the initial Cucumber-based integration test framework for the LEZ, building on the testing-framework integration work started by Andrus, and adds the first set of Cucumber integration scenarios.

The goal is to provide readable, scenario-driven integration coverage while using the Logos testing framework for deployment and lifecycle management rather than embedding the existing monolithic `TestContext` into Cucumber state.

The initial coverage includes:

- LEZ environment smoke tests
- authenticated public and private transfers
- indexer state consistency
- multi-sequencer committee convergence

The implementation also establishes reusable infrastructure for future Cucumber scenarios, including scenario-owned deployments, named transaction artifacts, deterministic teardown, per-scenario diagnostics, structured tracing, and concurrent scenario execution.

This PR builds on the initial testing-framework integration work started by @andrussal and extends that foundation into the Cucumber integration test environment and scenario coverage.

Co-authored-by: @andrussal

## ⚙️ Approach

*Describe the core changes introduced by this PR.*

- [X] Add a Cucumber test target behind the `cucumber` feature.
- [X] Add reusable Cucumber world, context, error, step, and scenario-state infrastructure.
- [X] Add testing-framework applications and handles for Bedrock, sequencer, indexer, wallet, multi-sequencer registry, and the complete local LEZ stack.
- [X] Keep service ownership and lifecycle management in the testing-framework deployment context while exposing cloneable handles to Cucumber.
- [X] Preserve the testing framework's existing `Send` deployment contract, rather than changing the block-publisher API throughout.
- [X] Add a wallet actor interface so a single `WalletCore` owner can safely service async Cucumber operations.
- [X] Track successful transfers as named artifacts so subsequent steps can assert against the exact transaction, participants, amount, kind, and inclusion block.
- [X] Add transaction-aware indexer convergence checks before comparing finalized state.
- [X] Add parameterized multi-sequencer support where sequencer registration and committee authorization are separate concepts.
- [X] Derive shared multi-sequencer genesis stakes from the committee members explicitly authorized in Gherkin.
- [X] Verify Bedrock committee membership and wait for the requested sequencer's posting turn before exercising transaction submission through that sequencer.
- [X] Add explicit service shutdown and after-hook fallback cleanup so scenarios can be repeatedly deployed and torn down without arbitrary sleeps.
- [X] Add per-scenario artifact directories and unique scenario contexts suitable for concurrent execution.
- [X] Add structured Cucumber lifecycle and step tracing controlled through `RUST_LOG`.
- [X] Add initial smoke, authenticated-transfer, indexer-consistency, and multi-sequencer scenarios.

The remaining explicit futures are not Cucumber-specific production changes; they are compile-time assertions required because rustc cannot currently propagate the higher-ranked Send proof through this generic async chain (rust-lang/rust#100013). trait_variant is used for the LEZ-owned `BlockPublisherTrait` so its implementations retain native async fn methods while exposing the required `Send` future contract. The explicit workaround is therefore limited to `ExecutorActor::new`, `sequencer_service::run`, and `SequencerApp::deploy_inner`.

The new testing-framework applications reuse existing low-level `test_fixtures` primitives such as sequencer setup, indexer/wallet setup, account initialization, and genesis helpers.

Some composition and lifecycle orchestration remains alongside the existing `TestContext` implementation. Consolidating those two paths is intentionally left for a follow-up PR rather than broadening this initial Cucumber migration.

## 🧪 How to Test

Copy and paste these all-together or one at a time in the repo root on a local machine:

```bash
RUSTFLAGS="-A single-use-lifetimes" RISC0_DEV_MODE=1 RUST_LOG='info,logos_blockchain_chain_leader_service=trace,risc0_zkvm::host::server::exec::executor=off' MAX_CUCUMBER_CONCURRENT_SCENARIOS=1 cargo test -p integration_tests --features cucumber --test cucumber -- --tags "@smoke_ci"

RUSTFLAGS="-A single-use-lifetimes" RISC0_DEV_MODE=1 RUST_LOG='info,logos_blockchain_chain_leader_service=trace,risc0_zkvm::host::server::exec::executor=off' MAX_CUCUMBER_CONCURRENT_SCENARIOS=1 cargo test -p integration_tests --features cucumber --test cucumber -- --tags "@auth_transfer_ci"

RUSTFLAGS="-A single-use-lifetimes" RISC0_DEV_MODE=1 RUST_LOG='info,logos_blockchain_chain_leader_service=trace,risc0_zkvm::host::server::exec::executor=off' MAX_CUCUMBER_CONCURRENT_SCENARIOS=1 cargo test -p integration_tests --features cucumber --test cucumber -- --tags "@indexer_state_ci"

RUSTFLAGS="-A single-use-lifetimes" RISC0_DEV_MODE=1 RUST_LOG='info,logos_blockchain_chain_leader_service=trace,risc0_zkvm::host::server::exec::executor=off' MAX_CUCUMBER_CONCURRENT_SCENARIOS=1 cargo test -p integration_tests --features cucumber --test cucumber -- --tags "@multi_sequencer_ci"
```

The Cucumber environment supports concurrent runs, so something like `MAX_CUCUMBER_CONCURRENT_SCENARIOS=3` is also supported and was validated.

Validated scenario results:

- Smoke: **3 scenarios / 15 steps passed**
- Authenticated transfers: **5 scenarios / 45 steps passed**
- Indexer state consistency: **2 scenarios / 17 steps passed**
- Multi-sequencer committee convergence: **1 scenario / 19 steps passed**

Total: **11 scenarios / 96 steps passed**.

Concurrent execution with `MAX_CUCUMBER_CONCURRENT_SCENARIOS=3` was also validated, see consol logs: [cucumber_console.log](https://github.com/user-attachments/files/31171546/cucumber_console.log)

Additional development validation included targeted compilation/Clippy, formatting, and `git diff --check`.

## 🔗 Dependencies

- None

## 🔜 Future Work

- Integration into GitHub CI was intentionally left out as work for a consecutive PR.
- Consolidate the remaining composition/lifecycle overlap between the existing `test_fixtures::TestContext` orchestration and the new testing-framework applications. The TF applications already reuse the existing low-level fixture primitives; the follow-up can focus on sharing the remaining owned-component builders and lifecycle abstractions without coupling that refactor to this initial Cucumber migration.

## 📋 PR Completion Checklist

- [X] Complete PR description
- [X] Add/update tests


integration_tests/src/cucumber/error.rs
```diff
@@ -0,0 +1,62 @@
+use thiserror::Error;
+
+/// Errors reported by Cucumber step implementations.
+#[derive(Debug, Error)]
+pub enum StepError {
```
andrussal: I think we can add this to StepError:

```rust
impl From<DynError> for StepError {
    fn from(source: DynError) -> Self {
        Self::QueryFailed { source }
    }
}
```

and change QueryFailed to carry the source instead of a pre-rendered string:

```rust
#[error("failed to query a LEZ service")]
QueryFailed {
    #[source]
    source: DynError,
},
```

then most of these ~78 map_errs collapse from

```rust
.map_err(|error| StepError::QueryFailed { message: error.to_string() })?;
```

to just `?`(sites with String/concrete error types need .map_err(Into::into)? or their own From). 

#[source] also displays the full cause chain shows up in CI output.

integration_tests/src/cucumber/world.rs
```diff
@@ -0,0 +1,420 @@
...
+    /// Balance returned by the sequencer.
+    pub observed_balance: Option<u128>,
+    /// Balance configured for the selected account.
+    pub expected_balance: Option<u128>,
+    /// Last indexer block observed by the convergence step.
+    pub observed_indexer_height: Option<u64>,
+    /// Fresh recipient account created for the new-account transfer scenario.
+    pub new_public_account: Option<lee::AccountId>,
+    /// Sender balance recorded before the public transfer.
+    pub sender_initial_balance: Option<u128>,
```
andrussal: Could we store the sender/receiver balances (captured at submission time) inside the TransferArtifact instead of the shared sender_initial_balance/receiver_initial_balance fields on the world?

integration_tests/src/cucumber/steps/mod.rs
```diff
@@ -0,0 +1,13 @@
+use cucumber::gherkin::Step;
+
```
andrussal, line 2: A couple of notes on the step defs:

- "included in a block" has two implementations with different semantics — one-shot in `transfers/assertions.rs`, polled in `committee/assertions.rs`. Could we keep just the polled one (sequencer optional)? Same for the three indexer-convergence variants.
- the balance-delta steps (6 near-identical bodies in `transfers/assertions.rs`) could share one helper — keeping the current Gherkin phrasings, just deduping the implementation
- "a LEZ smoke stack" / "a LEZ stack with configured public accounts" have identical bodies (`environment/setup.rs`, private pair too) — could be `#[given]` aliases on one fn so they can't drift
- "only the sender signs transfers {string} and {string}" (`transfers/assertions.rs` ~591) isn't used by any feature — can be removed

lez/sequencer/actors/executor/src/actor.rs
```diff
@@ -47,17 +49,23 @@ pub struct ExecutorActor<BP: BlockPublisherTrait> {
...
-        Self {
-            sequencer,
-            mempool_handle,
-            driver_cancellation,
-            background_tasks,
+    #[expect(
+        clippy::manual_async_fn,
+        reason = "An explicit Send + 'static future is required by the TF deployment boundary"
+    )]
+    pub fn new(config: SequencerConfig) -> impl Future<Output = Self> + Send + 'static {
```
andrussal: Could the required `Send` future be handled inside the TF app adapter? This otherwise changes the production sequencer and block-publisher APIs specifically for the testing-framework deployment boundary.

lez/wallet-ffi/wallet_ffi.h
```diff
@@ -249,7 +249,6 @@ typedef struct FfiAccountIdentity {
    * C-compatible string.
    */
   char *key_path;
-  struct FfiBytes32 authorization_secret_key;
```
andrussal, line 245: This header looks regenerated against a stale base — the Rust side still has `authorization_secret_key` in `FfiAccountIdentity` (types.rs)

integration_tests/src/tf/stack.rs
```diff
@@ -0,0 +1,198 @@
...
+        } else {
+            wallet.without_private_account_initialization()
+        };
+        let wallet = scenario_base_dir.as_ref().map_or_else(
+            || wallet.clone(),
+            |dir| wallet.clone().with_state_dir(dir.join("lez/wallet")),
+        );
+        ctx.deploy_and_expose(wallet).await?;
+
+        Ok(LezStackHandle)
```
andrussal: Could we return the deployed components here?

```rust
Ok(LezStackHandle {
    bedrock,
    indexer,
    sequencer,
    wallet,
})
```

Cucumber could then use the returned handle directly instead of looking up the same components again in `LezScenarioContext::from_deployment`. If `LezScenarioContext` is still useful for step-specific helpers, it could wrap `LezStackHandle` instead.

lez/wallet/src/lib.rs
```diff
@@ -53,6 +53,8 @@ pub mod poller;
 pub mod program_facades;
 pub mod storage;
 
+const LEZ_CUCUMBER_RUN: &str = "LEZ_CUCUMBER_RUN";
```
andrussal: Would it make sense to use `tracing` for these messages instead? That might let Cucumber control the output through the normal log level and avoid introducing a test-specific environment variable into the wallet code.

integration_tests/src/cucumber/world.rs
```diff
@@ -0,0 +1,420 @@
...
+pub struct CommitteeConfiguration {
+    /// Alias of the sequencer that declares the configuration.
+    pub leader_alias: String,
+    /// Registered aliases authorized in the founding committee.
+    pub authorized_sequencers: Vec<String>,
+}
+
+#[derive(Clone, Debug, Default)]
+/// Observable account and indexer state recorded during a scenario.
+pub struct EnvironmentState {
```
andrussal: Would it make sense to split `EnvironmentState` into smaller areas such as accounts, transfers, indexer, and committee state? It is already holding quite a few unrelated optional values, and I think this may become difficult to follow as more scenarios are added. This could also be done as a follow-up if it is too much for the initial integration.

integration_tests/src/testing_framework/stack.rs
```diff
@@ -0,0 +1,198 @@
...
+        let bedrock_app = scenario_base_dir.as_ref().map_or_else(
+            || self.bedrock.clone(),
+            |dir| {
+                self.bedrock
+                    .clone()
+                    .with_scenario_base_dir(dir.join("node"))
+            },
+        );
+        let bedrock = ctx.deploy_and_expose(bedrock_app).await?;
+        wait_for_bedrock_ready(&bedrock).await?;
```
andrussal, line 149: Should this also `wait_for_first_block()` before deploying the sequencer, like the registry path does?

lez/sequencer/actors/executor/src/actor.rs
hansieodendaal: **Note:** This requested change is deferred until the TF has been updated, see https://github.com/logos-blockchain/logos-blockchain-testing/pull/64.

integration_tests/src/cucumber/world.rs
hansieodendaal: See https://github.com/logos-blockchain/logos-execution-zone/pull/741/commits/4502a9e37bfc5f21f56bf2cce148d02a1cf6d333.

integration_tests/src/cucumber/steps/mod.rs
hansieodendaal: See https://github.com/logos-blockchain/logos-execution-zone/pull/741/commits/4502a9e37bfc5f21f56bf2cce148d02a1cf6d333.

lez/wallet-ffi/wallet_ffi.h
hansieodendaal: See https://github.com/logos-blockchain/logos-execution-zone/pull/741/commits/4502a9e37bfc5f21f56bf2cce148d02a1cf6d333.

integration_tests/src/tf/stack.rs
hansieodendaal: See https://github.com/logos-blockchain/logos-execution-zone/pull/741/commits/4502a9e37bfc5f21f56bf2cce148d02a1cf6d333.

lez/wallet/src/lib.rs
hansieodendaal: I changed the env var to be test agnostic - adding the print to trace logs will just add the spam to the logs.

integration_tests/src/cucumber/world.rs
hansieodendaal: See https://github.com/logos-blockchain/logos-execution-zone/pull/741/commits/4502a9e37bfc5f21f56bf2cce148d02a1cf6d333.

integration_tests/src/testing_framework/stack.rs
hansieodendaal: I tried it out, but it introduces a race condition. The way it's coded is robust and support concurrent scenarios.

integration_tests/src/utils.rs
erhant, line 1: this file seems to provide many of the utils existing in `lib.rs`

we should remove those in `lib.rs` and re-export via this one

integration_tests/src/cucumber/world.rs
```diff
@@ -0,0 +1,469 @@
...
+    /// Multi-sequencer committee observations.
+    pub committee: CommitteeState,
+}
+
+/// Per-scenario state for Cucumber tests that deploy LEZ applications.
+///
+/// Cucumber creates a fresh world for each scenario. Its deployment context
+/// starts empty; `Given` steps can deploy and expose only the applications the
+/// scenario needs. Dropping the world releases all TF-managed resources.
+#[derive(World, Derivative)]
```
erhant: what is the usage of this here?
`derivative` is also [unmaintained](https://rustsec.org/advisories/RUSTSEC-2024-0388.html)

integration_tests/Cargo.toml
```diff
@@ -38,11 +42,29 @@ test_programs.workspace = true
...
+tokio = { workspace = true, features = ["rt-multi-thread", "macros", "time"] }
 hex.workspace = true
 tempfile.workspace = true
 bytesize.workspace = true
+num-bigint.workspace = true
+
+testing-framework-app.workspace = true
+testing-framework-core.workspace = true
+derivative = "2.2.0"
+thiserror = "2.0.18"
```
erhant: we prefer to have these at root level, and use `*.workspace = true` instead

besides few of the crates here, we have `thiserror` at root already too

lez/sequencer/actors/executor/src/actor.rs
hansieodendaal: I investigated this further, including the TF-side alternative in logos-blockchain-testing#64. As discussed, that would introduce a second deployment API which we don’t want to maintain.

As per your suggestion:

The remaining explicit futures are not Cucumber-specific production changes; they are compile-time assertions required because of `rust-lang/rust#100013`: `rustc` cannot currently propagate the required Send proof through this generic async chain.

Using `#[async_trait]` on `BlockPublisherTrait` and its implementations lets the publisher API keep its normal `async fn` shape. Explicit `impl Future + Send + 'static` is now limited to the three boundaries where `rustc` still needs the assertion:

- `ExecutorActor::new`
- `sequencer_service::run`
- `SequencerApp::deploy_inner`

The Cucumber target and all 11 scenarios / 97 steps pass with this minimized form. TF PR #64 is no longer required.

.gitignore
```diff
@@ -20,3 +20,6 @@ bedrock_signing_key
 integration_tests/configs/debug/
 
 .DS_Store
+
+/integration_tests/cucumber_tests/temp/
+MyLogFile.log
```
Arjentix, line 25: Where does this come from? If this is not something our project generates I'd suggest using local `.git/info/exclude ` instead of committing this to `.gitignore`

lez/sequencer/core/src/block_publisher.rs
```diff
@@ -120,10 +121,7 @@ enum Command {
 
 type CommandSender = mpsc::Sender<Command>;
 
-#[expect(
-    async_fn_in_trait,
-    reason = "Only the methods reached from the executor actor need an explicitly Send future"
-)]
+#[async_trait]
```
Arjentix: `async_trait` is outdated since async traits supported by Rust natively, so let's not use this and better write bounds explicitly or with [`trait_variant`](https://docs.rs/trait-variant/latest/trait_variant/attr.make.html) (owned by rust lang org)

lez/wallet/src/lib.rs
```diff
@@ -53,6 +53,8 @@ pub mod poller;
 pub mod program_facades;
 pub mod storage;
 
+pub const SUPPRESS_VERBOSE_PRINTS: &str = "SUPPRESS_VERBOSE_PRINTS";
```
Arjentix, line 56: Looks useful, but let's better create a follow-up issue to use `-v` cli arg instead of env var

lez/sequencer/core/src/block_publisher.rs
Arjentix: `async_trait` exists in our codebase just because `jsonrpsee` crate still forces us to use it. But I see no reason to use it for our own traits

lez/indexer/service/src/service.rs
```diff
@@ -35,6 +39,11 @@ impl IndexerService {
             indexer,
         })
     }
+
+    #[cfg(not(feature = "mock-responses"))]
+    pub(crate) fn subscription_shutdown(&self) -> SubscriptionShutdown {
+        self.subscription_service.shutdown_handle()
+    }
```
Arjentix, line 46: Can you elaborate why we need this? I don't get it...

test_fixtures/src/setup.rs
```diff
@@ -108,15 +108,32 @@ impl SequencerSetup {
...
+    }
+
     /// Set up the sequencer in an explicit `home` directory owned by the caller.
     ///
     /// Useful for tests that restart the sequencer against the same on-disk store.
     pub async fn setup_at(self, home: &Path) -> Result<SequencerHandle> {
+        self.setup_owned(home.to_owned()).await
+    }
+
+    async fn setup_owned(self, home: PathBuf) -> Result<SequencerHandle> {
```
Arjentix: Too many similar functions IMO.

Can we simplify that? Maybe let's assume `home` just always must exist otherwise it's an error? So that it's a caller responsibility like it was before

integration_tests/Cargo.toml
```diff
@@ -8,6 +8,8 @@ license = { workspace = true }
 workspace = true
 
 [dependencies]
+async-trait = "0.1"
+cucumber = { features = ["macros", "output-junit"], workspace = true }
```
Arjentix: We try to keep our own crates on top of dependencies and keep externals at the bottom

integration_tests/Cargo.toml
```diff
@@ -38,11 +42,29 @@ test_programs.workspace = true
...
 tempfile.workspace = true
 bytesize.workspace = true
+num-bigint.workspace = true
+
+testing-framework-app.workspace = true
+testing-framework-core.workspace = true
+derivative = "2.2.0"
+thiserror = "2.0.18"
+tracing = "0.1.44"
+tracing-subscriber = { version = "0.3.23", features = ["env-filter", "fmt"] }
```
Arjentix: One day we should also transfer to `tracing` ecosystem instead of `log`. Hope this things right now won't contradict

integration_tests/src/lib.rs
```diff
@@ -23,6 +23,13 @@ use wallet::{
     storage::key_chain::FoundPrivateAccount,
 };
 
+/// Cucumber world, configuration, and step support for integration tests.
+pub mod cucumber;
+/// Testing-framework application deployments and lifecycle helpers.
+pub mod tf;
```
Arjentix: Honestly without reading this description I had to idea what `tf` means. Can we rename it? I think full `testing_framework` would work. Or just `framework`. Client-sites can reexport as `tf` if needed

.gitignore
erhant, line 25: this was a local artifact I think

integration_tests/src/testing_framework/mod.rs
```diff
@@ -0,0 +1,21 @@
+//! Testing-framework deployments and Cucumber support for LEZ tests.
```
Arjentix, line 1: Maybe it should be moved into `cucumber` module then? Or am I missing something?

.gitignore
hansieodendaal: `MyLogFile.log` comes from the underlying circuits

erhant: hmm i didnt know that, is it easy to rename that to something like `{circuitname}.log` or such?

hansieodendaal: We would have to change the underlying library for that

integration_tests/tests/tf_app_deployments.rs
```diff
@@ -0,0 +1,165 @@
...
+async fn complete_lez_stack_can_be_deployed_as_one_app() -> Result<(), DynError> {
+    let mut deployment = new_deployment();
+
+    deployment
+        .deploy(LezLocalApp::new().with_bedrock_nodes(2))
+        .await?;
+
+    assert_lez_stack_works(&deployment).await?;
+    shutdown_lez_deployment(&deployment).await?;
+    shutdown_lez_deployment(&deployment).await?;
```
Arjentix: Why shutdown two times?

integration_tests/tests/tf_app_deployments.rs
```diff
@@ -0,0 +1,165 @@
...
+        wait_for_indexer_to_catch_up,
+    },
+};
+use sequencer_service_rpc::RpcClient as _;
+use tempfile::tempdir;
+use testing_framework_app::{AppHostEnv, AppHostTopology, DeployContext};
+use testing_framework_core::scenario::{DynError, NodeClients};
+
+#[tokio::test]
+async fn complete_lez_stack_can_be_deployed_as_one_app() -> Result<(), DynError> {
```
Arjentix: I'm not sure about this 3rd party `DynError` in our tests. I mean, we have a well-known `anyhow::Result` we use everywhere and I'd prefer stick to it

integration_tests/tests/tf_app_deployments.rs
```diff
@@ -0,0 +1,165 @@
...
+    let bedrock = deployment
+        .deploy_and_expose(BedrockApp::nodes(
+            2,
+            "lez_apps_can_be_deployed_individually".to_owned(),
+        ))
+        .await?;
+
+    deployment
+        .deploy_and_expose(IndexerApp::new(bedrock.primary_api_addr()))
+        .await?;
```
Arjentix: Don't you need `indexer` var to shutdown it as well?

integration_tests/tests/tf_app_deployments.rs
```diff
@@ -0,0 +1,165 @@
...
+        ))
+        .await?;
+
+    deployment
+        .deploy_and_expose(WalletApp::from_sequencer(&sequencer))
+        .await?;
+
+    assert_lez_stack_works(&deployment).await?;
+    shutdown_lez_deployment(&deployment).await?;
+    shutdown_lez_deployment(&deployment).await?;
```
Arjentix: Again shutdown two times

integration_tests/tests/tf_bedrock_builder.rs
```diff
@@ -0,0 +1,29 @@
...
+    let builder = DeploymentBuilder::new(TopologyConfig::with_node_numbers(2))
+        .with_deployment_seed(DeploymentSeed::new([0x4c; 32]))
+        .with_test_context("lez-tf-bedrock-builder");
+
+    let bedrock = deployment
+        .deploy_and_expose(BedrockApp::from_builder(builder))
+        .await?;
+
+    bedrock.cryptarchia_info().await?;
+    shutdown_lez_deployment(&deployment).await?;
```
Arjentix: Btw, maybe it will be better to have `shutdown()` as a method? This types has a lot of them so I see no reason to make it a separate function

integration_tests/cucumber_tests/cucumber.rs
```diff
@@ -0,0 +1,226 @@
...
+use tracing::{info, warn};
+use wallet::SUPPRESS_VERBOSE_PRINTS;
+
+pub const TARGET: &str = "cucumber_main";
+type ScenarioAttempts = Arc<Mutex<HashMap<String, u8>>>;
+
+#[tokio::main]
+#[expect(clippy::print_stdout, reason = "Cucumber logs test code")]
+async fn main() {
+    logos_blockchain_testing_framework::env::set_default_env(SUPPRESS_VERBOSE_PRINTS, "true");
```
Arjentix, line 32: Why we set env vars for our wallet using logos library...?

Btw AFAIK real [`std::env::set_var()`](https://doc.rust-lang.org/std/env/fn.set_var.html) is unsafe

integration_tests/cucumber_tests/cucumber.rs
```diff
@@ -0,0 +1,226 @@
...
+    integration_tests::cucumber::default::init_tracing();
+    reap_all_stale_port_blocks();
```
Arjentix, line 45: Do we really need this? Why would ports become stale?

integration_tests/cucumber_tests/cucumber.rs
```diff
@@ -0,0 +1,226 @@
...
+    let scenario_attempts: ScenarioAttempts = Arc::new(Mutex::new(HashMap::new()));
+    let teardown_failed = Arc::new(AtomicBool::new(false));
+
+    let output_dir = create_scenario_output_dir();
+    let junit_xml_file = OpenOptions::new()
+        .append(true)
+        .create(true)
+        .open(output_dir.join("cucumber-output-junit.xml"))
+        .inspect_err(|err| warn!(target: TARGET, "Failed to open output file: {err}"))
+        .expect("should create or open output file");
```
Arjentix: So maybe just `context()` and return `anyhow::Result<()>`?

integration_tests/cucumber_tests/cucumber.rs
```diff
@@ -0,0 +1,226 @@
...
+    logos_blockchain_testing_framework::env::set_default_env(SUPPRESS_VERBOSE_PRINTS, "true");
+    integration_tests::cucumber::default::init_tracing();
+    reap_all_stale_port_blocks();
+    println!("args: {:?}", std::env::args());
+
+    let scenario_attempts: ScenarioAttempts = Arc::new(Mutex::new(HashMap::new()));
+    let teardown_failed = Arc::new(AtomicBool::new(false));
+
+    let output_dir = create_scenario_output_dir();
+    let junit_xml_file = OpenOptions::new()
```
Arjentix, line 52: Can you explain why do we need this file? I don't quite get this

integration_tests/cucumber_tests/cucumber.rs
```diff
@@ -0,0 +1,226 @@
...
+        })
+        // Runs Cucumber. Features sourced from a Parser are fed to a Runner, which
+        // produces events handled by a Writer.
+        .run(get_feature_path())
+        .await;
+
+    // Clean up manually reserved handshake port block files for this process
+    release_reserved_port_block();
+
+    if failed.execution_has_failed() || teardown_failed.load(std::sync::atomic::Ordering::Relaxed) {
```
Arjentix, line 159: Maybe consider [`Notify`](https://docs.rs/tokio/latest/tokio/sync/struct.Notify.html) instead of atomic?

integration_tests/cucumber_tests/cucumber.rs
```diff
@@ -0,0 +1,226 @@
...
+        .join(ARTEFACTS)
+        .join(feature_name)
+        .join(scenario_name.trim().replace(' ', "_"))
+        .join(run_attempt)
+}
+
+// Increment and return the attempt count for the given scenario. Counts
+// are tracked per-scenario, and keyed by a combination of feature and
+// scenario name.
+#[expect(clippy::significant_drop_tightening, reason = "Compiler weirdness")]
```
Arjentix: Hah, I think ~compiler~ clippy is fine here, most probably it wants something like this:

```rust
let key = format!("{feature}::{scenario}");

scenario_attempts
    .lock()
    .unwrap_or_else(std::sync::PoisonError::into_inner)
    .entry(key)
    .or_insert(0)
    .wrapping_add(1);

format!("attempt_{}", *entry)
```

That way guard lives exactly as much as we need it

integration_tests/cucumber_tests/features/authenticated_transfers.feature
```diff
@@ -0,0 +1,80 @@
...
+  @auth_transfer_ci
+  # Mirrors integration_tests/tests/auth_transfer/public.rs::successful_transfer_to_existing_account.
+  # Coverage is equivalent for the transfer behavior and stronger for lifecycle coverage:
+  # Cucumber also verifies indexer convergence and explicit runtime teardown. It uses balance
+  # deltas instead of the legacy test's fixed 9900/20100 values.
+  Scenario: Transfer funds between configured public accounts
+    Given a LEZ stack with configured public accounts
+    When I transfer 100 from the first configured public account to the second as "PUBLIC_TRANSFER"
+    Then the sender balance for transfer "PUBLIC_TRANSFER" decreases by 100
+    And the receiver balance for transfer "PUBLIC_TRANSFER" increases by 100
```
Arjentix, line 12: So `PUBLIC_TRANSFER` works like operation id here, right?

integration_tests/cucumber_tests/features/authenticated_transfers.feature
```diff
@@ -0,0 +1,80 @@
...
+    Then I stop the runtime
+
+  @auth_transfer_ci
+  # Mirrors integration_tests/tests/auth_transfer/public.rs::failed_transfer_with_insufficient_balance.
+  # Coverage is equivalent or stronger: it uses 10001 rather than 1_000_000, explicitly checks
+  # block absence, and also verifies indexer convergence and explicit runtime teardown.
+  Scenario: Reject a public transfer with insufficient sender balance
+    Given a LEZ stack with configured public accounts
+    When I attempt to transfer 10001 from the first configured public account to the second
+    Then the transfer is rejected
```
Arjentix, line 25: But why here it works without a concrete name for transfer...?

integration_tests/src/cucumber/error.rs
Arjentix, line 6: Nice descriptive error type. But I can see there are a lot of `message` fields and no underlying errors. So I suspect some of this `message` fields can actually be something like `#[source] std::io::Error` or `#[source] anyhow::Error` (this is probably from our test harness)

integration_tests/src/cucumber/default.rs
```diff
@@ -0,0 +1,86 @@
...
+        panic!("Feature path does not exist: {}", feature_path.display());
+    }
+    feature_path
+}
+
+/// Creates the output directory for the current scenario and returns its path.
+#[must_use]
+#[expect(clippy::print_stdout, reason = "Cucumber logs test code")]
+pub fn create_scenario_output_dir() -> PathBuf {
+    logos_blockchain_testing_framework::env::set_default_env(TF_KEEP_LOGS, "true");
```
Arjentix: Is it something to configure bedrock harness?

integration_tests/src/cucumber/default.rs
```diff
@@ -0,0 +1,86 @@
...
+    let output_dir = current_dir.join(SCENARIO_OUTPUT_DIR_REL);
+    fs::create_dir_all(output_dir.clone()).expect("should succeed");
+    println!("Output directory: {}", output_dir.display());
+    output_dir
+}
+
+/// Get the number of retries for failed scenarios from the `CUCUMBER_RETRIES`
+/// environment variable. Retries are opt-in: an unset variable and an explicit
+/// zero both disable retries.
+pub fn get_retries() -> Result<Option<usize>, String> {
```
Arjentix: Let's not make `String` errors, better to use `anyhow` at least

integration_tests/src/testing_framework/apps/bedrock.rs
```diff
@@ -0,0 +1,262 @@
...
+/// A TF-managed Logos blockchain cluster used as LEZ's Bedrock layer.
+///
+/// Logos owns node configuration generation through [`TopologyConfig`] and
+/// [`DeploymentBuilder`]. Callers can use the default node-count constructor or
+/// provide a configured builder without duplicating Logos deployment logic.
+#[derive(Clone)]
+pub struct BedrockApp {
+    builder: DeploymentBuilder,
+    scenario_base_dir: Option<PathBuf>,
+}
```
Arjentix, line 25: So do I understand right that this lets us run bedrock natively instead of docker setup we use in integration tests?

integration_tests/src/testing_framework/apps/bedrock.rs
```diff
@@ -0,0 +1,262 @@
...
+        }
+        Ok(())
+    }
+
+    /// Waits until the Bedrock node has produced its first post-genesis block.
+    ///
+    /// The node wallet indexes genesis funding while processing the live chain;
+    /// callers that immediately submit a wallet-funded transaction must wait for
+    /// this transition first.
+    pub async fn wait_for_first_block(&self) -> Result<(), DynError> {
```
Arjentix, line 179: One thing I still don't understand... So blockchain team provides us a lot of tools, but we still need to write stuff like that manually... Why? Isn't there some setup we can just import and use?

integration_tests/src/tf/apps/bedrock.rs
```diff
@@ -0,0 +1,262 @@
...
+            match self.cluster.first_client() {
+                Some(client) => match client.consensus_info().await {
+                    Ok(info) if info.cryptarchia_info.height > 0 => return Ok(()),
+                    Ok(_) => {}
+                    Err(error) => last_error = Some(error.to_string()),
+                },
+                None => last_error = Some("Bedrock cluster has no node clients".to_owned()),
+            }
+            sleep(POLL_INTERVAL).await;
+        }
```
Arjentix: Let's use tokio's [`timeout()`](https://docs.rs/tokio/latest/tokio/time/fn.timeout.html)

integration_tests/src/cucumber/steps/mod.rs
```diff
@@ -0,0 +1,13 @@
...
+pub mod accounts;
+pub mod committee;
+pub mod environment;
+pub mod indexer;
+pub mod transfers;
+
+pub const TARGET: &str = "cucumber_steps";
+
+pub(super) fn log_step(step: &Step) {
+    tracing::info!(target: TARGET, "Executing Cucumber step: {}", step.value);
```
Arjentix, line 12: I appreciate the usage of tracing but let's also write it in a more structured way:


```suggestion
    tracing::info!(target: TARGET, step = %step.value, "Executing Cucumber step");
```

To my understanding this makes the log more structural and grepable

integration_tests/src/testing_framework/apps/wallet.rs
```diff
@@ -0,0 +1,734 @@
...
+    },
+    WalletPassword {
+        response: oneshot::Sender<Result<String, String>>,
+    },
+    Shutdown {
+        response: oneshot::Sender<Result<(), String>>,
+    },
+}
+
+struct WalletActor {
```
Arjentix, line 107: Considering the fact that we already started using [`kameo`](https://crates.io/crates/kameo) as our actor framework maybe it makes sense to rewrite `WalletActor` in the same style? Just to be consistent.

It's actually quite handy, I think you saw examples in `sequencer/actors`.

integration_tests/src/tf/wait.rs
```diff
@@ -0,0 +1,173 @@
...
+        /// Sequencer block height the indexer was expected to reach.
+        target: u64,
+        /// Highest indexer block height observed while polling.
+        last_observed: u64,
+        /// Duration spent polling before timing out.
+        elapsed: Duration,
+    },
+}
+
+impl Display for IndexerCatchUpError {
```
Arjentix: Why not with `thiserror::Error`?

Arjentix (commented):
Huge work, thanks. I'm happy with the result, however left some comments I'd like to fix/discuss

Arjentix:
I think this also requires some changes in our CI to properly run it

integration_tests/src/utils.rs
hansieodendaal: Comments addressed in 84196f3f699455da2e84df8509dddb2c8bb16e92.

integration_tests/src/cucumber/world.rs
hansieodendaal: Removed. Comments addressed in https://github.com/logos-blockchain/logos-execution-zone/commit/84196f3f699455da2e84df8509dddb2c8bb16e92.

integration_tests/Cargo.toml
hansieodendaal: Comments addressed in https://github.com/logos-blockchain/logos-execution-zone/commit/84196f3f699455da2e84df8509dddb2c8bb16e92.

lez/wallet-ffi/wallet_ffi.h
```diff
@@ -193,7 +200,7 @@ typedef struct FfiU128 {
  * byte arrays since C doesn't have native u128 support.
  */
 typedef struct FfiAccount {
-  struct FfiBytes32 program_owner;
+  struct FfiProgramId program_owner;
```
moudyellaz: `types.rs:86` is still `FfiBytes32` and dev's header matched it: revert this file?

integration_tests/src/lib.rs
```diff
@@ -3,256 +3,12 @@
 //! non-test consumers (e.g. `integration_bench`) can depend on them without
 //! pulling in the test files.
 
-use std::time::Duration;
+/// Cucumber world, configuration, and step support for integration tests.
+pub mod cucumber;
```
moudyellaz, line 8: The CI archive builds `integration_tests` with no features, so this module pulls cucumber + quick-xml into every matrix job: gate it behind the `cucumber` feature?

integration_tests/cucumber_tests/cucumber.rs
```diff
@@ -0,0 +1,226 @@
...
+    integration_tests::cucumber::default::init_tracing();
+    reap_all_stale_port_blocks();
+    println!("args: {:?}", std::env::args());
+
+    let scenario_attempts: ScenarioAttempts = Arc::new(Mutex::new(HashMap::new()));
+    let teardown_failed = Arc::new(AtomicBool::new(false));
+
+    let output_dir = create_scenario_output_dir();
+    let junit_xml_file = OpenOptions::new()
+        .append(true)
```
moudyellaz, line 53: `append` on a fixed path means run two writes a second `<testsuites>` root into the same file: `truncate` instead?

lez/wallet-ffi/wallet_ffi.h
hansieodendaal: Yeah, it should not be changed, it keep on creeping into the commit set whenever I compile.

integration_tests/cucumber_tests/cucumber.rs
hansieodendaal: We only create a single junit xml summary file for all cucumber runs. We specifically do this so we can split the runs in CI but then get a single summary report.

integration_tests/cucumber_tests/features/multi_sequencer.feature
```diff
@@ -0,0 +1,35 @@
+Feature: LEZ multi-sequencer committee convergence
+
+  @multi_sequencer_ci
+# Mirrors the committee convergence, cross-sequencer transfer, and indexer checks in
+# integration_tests/tests/multi_sequencer.rs::multi_sequencer_committee_converges.
+# Cucumber preserves the legacy lifecycle: configured committee members are staked in
+# shared genesis, A creates the channel, B joins later, and the scenario waits for B's
+# posting turn. Cucumber additionally uses parameterized aliases and explicit runtime
+# teardown.
+  @multi_sequencer
```
erhant: maybe we remove this and keep `@multi_sequencer_ci`, as is the case for other features?

hansieodendaal: Good catch, that was an old marker

integration_tests/src/tf/apps/wallet.rs
```diff
@@ -0,0 +1,734 @@
...
+            public_accounts,
+            private_accounts,
+            state_dir: configured_state_dir,
+            initialize_private_accounts: initialize_private_account_funding,
+        } = self;
+        // WalletCore initialization currently exposes a non-general borrowed
+        // lifetime in its async API. Keep that implementation detail inside a
+        // dedicated blocking thread/runtime; scenario operations use the
+        // actor below and never create nested runtimes.
+        let setup_public_accounts = public_accounts.clone();
```
erhant: do we need all these `clone`s here? if I see it correctly, we could maybe move it directly within the closure

.gitignore
hansieodendaal: This is generated by the underlying circuit library during integration runs, so it is a project-generated artifact rather than a local editor/tool artifact. Renaming it would require changing the upstream library. It is also present in the `logos-blockchain` repo runs.

erhant: @Arjentix I was also irked by this :D 

maybe we can open an issue upstream

lez/wallet/src/lib.rs
hansieodendaal: Can we make an issue and fix it in a consecutive PR to use `-v` cli arg? This exclusion is important for now, otherwise no sense can be made of any cucumber log on-screen outputs.

lez/indexer/service/src/service.rs
hansieodendaal: Cucumber redeploys services repeatedly and needs teardown to be deterministic. The RPC server handle only lets us await the server; the subscription loop was previously cancelled/aborted during drop with no way to await it. This handle lets `IndexerHandle::shutdown()` wait for both pieces to terminate before the deployment is released, rather than racing redeployment.

integration_tests/Cargo.toml
hansieodendaal: Yes, agreed. The Cucumber instrumentation is deliberately using structured `tracing`; it doesn’t require us to migrate the existing `log` call sites as part of this PR.

integration_tests/src/testing_framework/mod.rs
hansieodendaal: It should stay a sibling module. These are LEZ adapters to the testing framework, and they already have direct non-Cucumber tests such as `tf_app_deployments.rs` and `tf_bedrock_builder.rs`. Moving them under Cucumber would create exactly the coupling this architecture is trying to avoid.

erhant (approved):
LGTM, thank you! It is indeed a lot of work 🙏🏻 

in the end I really like how the `.feature` files read, and I feel like this will pay-off for the future tests, even though it is an effort to rewrite stuff like this as it's technically verbose

thanks for addressing the comments swiftly as well

integration_tests/tests/tf_app_deployments.rs
hansieodendaal: The indexer is exposed into `DeployContext`, and `shutdown_lez_deployment()` retrieves the exposed `LezIndexerClient` itself, so a local variable is not needed.

integration_tests/tests/tf_bedrock_builder.rs
hansieodendaal: That call could be removed here, but the separate helper remains useful for complete LEZ stacks where dependency-ordered shutdown matters.

integration_tests/cucumber_tests/cucumber.rs
hansieodendaal: It reaps stale reservation files, not sockets. A killed test process can leave its port-block claim behind; the helper checks the recorded PID and only removes claims for dead processes.

This was specifically added to avoid race conditions and conflicts in CI runners. and it also helps if heavy concurrency is enabled on a local machine, possibly running form two different cucumber test instantiations.

integration_tests/cucumber_tests/cucumber.rs
hansieodendaal: This is a summary file for all cucumber runs, very useful as a reference when many scenarios were run, and also in CI where we typically run focused scenarios as single test processes.

integration_tests/cucumber_tests/cucumber.rs
hansieodendaal: There isn’t a waiter here; the hooks only accumulate a failure bit which is inspected after Cucumber returns. `Notify` would add event semantics we don’t need.

integration_tests/cucumber_tests/features/authenticated_transfers.feature
hansieodendaal: Yes - it is a scenario-local alias/key for the transfer artifact. It lets later steps refer to exactly that submitted transaction; it is not an on-chain/protocol operation ID.

integration_tests/cucumber_tests/features/authenticated_transfers.feature
hansieodendaal: Names are used for successful transfers because later steps may refer to multiple successful transactions. A rejected submission produces no successful artifact/hash/inclusion lifecycle, and these scenarios only track the latest rejected attempt. If we add scenarios with multiple rejected attempts, naming those would make sense

integration_tests/src/cucumber/default.rs
hansieodendaal: `TF_KEEP_LOGS` preserves per-node working directories that would otherwise be deleted during teardown. In CI specifically, we would like to keep the data around for failing scenarios, but also very useful for local debugging and analysis.

I would change this to set the default before Tokio threads start as part prior comments.

integration_tests/src/testing_framework/apps/bedrock.rs
hansieodendaal: Yes, this path is TF-managed local process deployment rather than Cucumber managing a Docker stack itself.

`BedrockApp` is the LEZ adapter around the Logos testing framework’s local-cluster deployment path. It builds the blockchain-owned `DeploymentBuilder`/`TopologyConfig`, then deploys it through `DeployContext::deploy_local_cluster::<LbcEnv>`.

integration_tests/src/testing_framework/apps/bedrock.rs
hansieodendaal: @andrussal and I discussed this a bit. Currently we do not have reusable stuff like that in crates; if it becomes available in future, we can look at this again. In the meantime, doing these low-level assertions is the way to go.

integration_tests/src/cucumber/steps/mod.rs
hansieodendaal: This is really just preference; one is more human readable and the other more grepable; in the logos-blockchain project we were more in favour of the former. I would call this a follow-up task.

integration_tests/src/testing_framework/apps/wallet.rs
hansieodendaal: I think this could be a follow-up PR if required; it would not be a drop-in consistency refactor. This wrapper exists specifically to keep one `WalletCore` owner on a dedicated OS thread/current-thread runtime.

hansieodendaal:
> I think this also requires some changes in our CI to properly run it

Yes. CI integration is intentionally the immediate follow-up to this initial framework PR; this PR establishes and validates the runner/scenario architecture locally first.

integration_tests/src/tf/apps/wallet.rs
hansieodendaal: Please see changes implemented in 8c2a8d4d21f9a0de1896d6e2cf0e6e477027f3f7.

integration_tests/cucumber_tests/features/multi_sequencer.feature
hansieodendaal: Please see changes implemented in 8c2a8d4d21f9a0de1896d6e2cf0e6e477027f3f7.

integration_tests/src/lib.rs
hansieodendaal: Please see changes implemented in 8c2a8d4d21f9a0de1896d6e2cf0e6e477027f3f7.

lez/wallet-ffi/wallet_ffi.h
hansieodendaal: Please see changes implemented in 8c2a8d4d21f9a0de1896d6e2cf0e6e477027f3f7.

integration_tests/src/tf/wait.rs
hansieodendaal: Please see changes implemented in 8c2a8d4d21f9a0de1896d6e2cf0e6e477027f3f7.

integration_tests/src/tf/apps/bedrock.rs
hansieodendaal: Please see changes implemented in 8c2a8d4d21f9a0de1896d6e2cf0e6e477027f3f7.

lez/sequencer/core/src/block_publisher.rs
hansieodendaal: Please see changes implemented in 8c2a8d4d21f9a0de1896d6e2cf0e6e477027f3f7.

test_fixtures/src/setup.rs
hansieodendaal: Please see changes implemented in 8c2a8d4d21f9a0de1896d6e2cf0e6e477027f3f7.

integration_tests/Cargo.toml
hansieodendaal: Please see changes implemented in 8c2a8d4d21f9a0de1896d6e2cf0e6e477027f3f7.

integration_tests/src/lib.rs
hansieodendaal: Please see changes implemented in 8c2a8d4d21f9a0de1896d6e2cf0e6e477027f3f7.

integration_tests/tests/tf_app_deployments.rs
hansieodendaal: Please see changes implemented in 8c2a8d4d21f9a0de1896d6e2cf0e6e477027f3f7.

integration_tests/tests/tf_app_deployments.rs
hansieodendaal: Please see changes implemented in 8c2a8d4d21f9a0de1896d6e2cf0e6e477027f3f7.

integration_tests/tests/tf_app_deployments.rs
hansieodendaal: Please see changes implemented in 8c2a8d4d21f9a0de1896d6e2cf0e6e477027f3f7.

integration_tests/tests/tf_bedrock_builder.rs
hansieodendaal: Please see changes implemented in 8c2a8d4d21f9a0de1896d6e2cf0e6e477027f3f7.

integration_tests/cucumber_tests/cucumber.rs
hansieodendaal: Please see changes implemented in 8c2a8d4d21f9a0de1896d6e2cf0e6e477027f3f7.

integration_tests/cucumber_tests/cucumber.rs
hansieodendaal: Please see changes implemented in 8c2a8d4d21f9a0de1896d6e2cf0e6e477027f3f7.

integration_tests/cucumber_tests/cucumber.rs
hansieodendaal: Please see changes implemented in 8c2a8d4d21f9a0de1896d6e2cf0e6e477027f3f7.

integration_tests/src/cucumber/error.rs
hansieodendaal: Please see changes implemented in 8c2a8d4d21f9a0de1896d6e2cf0e6e477027f3f7.

integration_tests/src/cucumber/default.rs
hansieodendaal: Please see changes implemented in 8c2a8d4d21f9a0de1896d6e2cf0e6e477027f3f7.

integration_tests/src/cucumber/default.rs
hansieodendaal: Please see changes implemented in 8c2a8d4d21f9a0de1896d6e2cf0e6e477027f3f7.

integration_tests/src/cucumber/steps/transfers/assertions.rs
```diff
@@ -0,0 +1,523 @@
...
+}
+
+#[then("no transfer is included in a block")]
+#[expect(
+    clippy::needless_pass_by_ref_mut,
+    reason = "Cucumber step functions require `&mut World` as the first parameter"
+)]
+fn assert_no_transfer_is_included(world: &mut CucumberWorld, step: &Step) -> StepResult {
+    log_step(step);
+    if world.environment.transfers.rejected.is_none() {
```
moudyellaz: Same condition as `the transfer is rejected` above and no block is queried: drop this step and the "explicitly checks block absence" claim on the scenario?

test_fixtures/src/setup.rs
```diff
@@ -108,15 +108,22 @@ impl SequencerSetup {
...
-        let sequencer_handle = self.setup_at(temp_sequencer_dir.path()).await?;
+        let sequencer_handle = self
+            .setup_owned(temp_sequencer_dir.path().to_owned())
+            .await?;
 
         Ok((sequencer_handle, temp_sequencer_dir))
     }
 
     /// Set up the sequencer in an explicit `home` directory owned by the caller.
     ///
```
moudyellaz, line 117: `setup_in` has no callers and the same body as `setup_at`, same for `initialize_wallet_public_accounts` and the `testing-framework-runner-local` workspace entry: drop the three?

integration_tests/src/cucumber/steps/transfers/public.rs
```diff
@@ -0,0 +1,279 @@
...
+    ensure_transfer_name_available(world, &transfer_name)?;
+    let context = world.lez()?;
+    let accounts = context.existing_public_accounts().await?;
+    let sender = accounts
+        .first()
+        .copied()
+        .ok_or(StepError::MissingSelectedAccount)?;
+    let receiver = accounts
+        .get(1)
+        .copied()
```
moudyellaz, line 35: This balance+nonce snapshot is repeated verbatim in all four steps: one helper in `helpers.rs` next to `transfer_details`?

integration_tests/src/cucumber/steps/transfers/assertions.rs
hansieodendaal: Please see  changes in 124c8fc01af97826ec2d9e4ca60c24caf11a1c01.

test_fixtures/src/setup.rs
hansieodendaal: Please see  changes in 124c8fc01af97826ec2d9e4ca60c24caf11a1c01.

integration_tests/src/cucumber/steps/transfers/public.rs
hansieodendaal: Please see changes in 124c8fc01af97826ec2d9e4ca60c24caf11a1c01.

.gitignore
Arjentix: Wow, okay, funny. Opening an issue won't hurt, I think

lez/wallet/src/lib.rs
Arjentix: Created #772

integration_tests/src/testing_framework/apps/bedrock.rs
Arjentix: I now wonder what should we do about it in LEZ. The problem is that we currently have:
1. Bedrock docker setup which is used in integration tests and local scripts like `just run-bedrock` (from `Justfile`). This let's us ensure our local setup is the same setup being tested, because all docker configuration is the same. Our local setup also works through `docker compose up` in repo root, which run our lez containers together with bedrock container.
2. Native setup for cucumber tests. This may use different Bedrock version than we use in docker, so one day we may find our self in strange situation where integration test works but the same cucumber test don't.

Currently I have to idea how to resolve this. I'd like to have local setup to be deterministic and run the same thing we run in tests.

Do you have any ideas how to unify this?

integration_tests/src/cucumber/steps/mod.rs
Arjentix: Well, actually it's not just preference, here the summary from Claude:

**Why `info!(new_value = value, "Value updated")` beats `info!("Value updated: {value}")`**

1. **Queryable, not greppable.** JSON output gives `{"new_value": 42}` — you can filter `new_value > 10` in Loki/Elastic/Datadog. A string forces regex extraction on every query.

2. **Stable message = groupable events.** `"Value updated"` is one constant, so all occurrences aggregate into a single bucket for counts, dashboards, and alerts. Interpolation makes every line a unique string, which breaks grouping.

3. **`EnvFilter` can target fields, not substrings.** `RUST_LOG='my_app[{new_value}]=debug'` works; there is no filter syntax for "message contains".

4. **Types survive the trip.** Numbers stay numbers, bools stay bools. Once interpolated, everything is text, and text doesn't compare or sum.

5. **Fields propagate; messages don't.** Put a field on a `#[instrument]` span and every event inside it inherits the context automatically — no repeating `{request_id}` in twenty format strings.

6. **Log schema decouples from local variables.** `new_value = v` lets you rename the variable without changing the log contract, and vice versa.

7. **It's the authors' stated idiom.** "The more idiomatic style… is to use these macros to record structured data rather than unstructured messages" — the [Tokio announcement post](https://tokio.rs/blog/2019-08-tracing).

8. **It costs nothing.** No allocation by default, no formatting unless a subscriber is actually interested. And you can keep the readable sentence too: `info!(new_value = value, "Value updated to {value}")`.

**The one-liner for the review comment:** if a human is the only consumer, interpolate; if a machine will ever need it, make it a field — and you rarely know in advance which one it'll be.

Arjentix: But yeah, anyway we'll have to move from log to tracing and address this formatting as well, so follow-up task works

integration_tests/src/testing_framework/apps/wallet.rs
Arjentix: Okay, maybe indeed doesn't worth it

lez/sequencer/core/src/block_publisher.rs
```diff
@@ -291,93 +285,93 @@ impl BlockPublisherTrait for ZoneSdkPublisher {
...
-                                let withdraw_count = withdrawals.len();
-                                let published = if withdrawals.is_empty() {
-                                    sequencer.handle()
-                                        .publish(data_bounded)
-                                        .await
-                                        .context("Failed to publish block")
-                                } else {
-                                    sequencer.handle()
-                                        .publish_atomic_withdraw(data_bounded, withdrawals)
+                            // Drain external commands by calling the borrowing
```
Arjentix, line 288: Is it just another tabulation in this block of code? Probably accidental change?

integration_tests/src/cucumber/error.rs
```diff
@@ -0,0 +1,121 @@
...
+    #[error("Duplicate transfer artifact '{name}'")]
+    DuplicateTransferArtifact { name: String },
+    /// A named transfer has not yet been observed in a block.
+    #[error("Transfer artifact '{name}' has no recorded inclusion block")]
+    MissingTransferInclusion { name: String },
+    /// LEZ application deployment failed.
+    #[error("Deployment failed: {message}")]
+    DeploymentFailed { message: String },
+    /// An operational deployment error with its original source preserved.
+    #[error("Deployment failed: {source}")]
```
Arjentix, line 49: This is not obvious, but actually you don't need to specify source in message string, it will work better with:


```suggestion
    #[error("Deployment failed")]
```

Source error is already marked as `#[source]` and is persisted in the chain. Pretty printing (with `{:#}`) this error wrapped in `anyhow::Error` would give a good-looking trace.

Same for other error variants here with same problem.

Arjentix (approved):
LGTM, just a couple of remarks

moudyellaz (commented):
lgtm, thanks for addressing my comments.

moudyellaz (approved):

integration_tests/src/testing_framework/apps/bedrock.rs
hansieodendaal: I agree we should ultimately have one deterministic Bedrock version for both the TF tests and the Docker/local setup.

For this PR I suggest we keep the scope smaller: LEZ already pins the Logos SDK/TF crates to `97d5e104...` (_without a corresponding release and no binaries to download_), so I'll make the native TF path build logos-blockchain-node from that same pinned revision. That guarantees the Cucumber tests use a node matching the SDK they compile against.

The existing Docker setup (`0.2.1-lssa`) will remain unchanged here. As a follow-up, I think we should move LEZ onto an official Logos release and align both the SDK pin and Docker/node artifact to that release, so we have one version contract rather than two independently maintained revisions. That will let TF download the pinned pre-built binary.

lez/sequencer/core/src/block_publisher.rs
hansieodendaal: Thanks yes, to be fixed in the consecutive PR that adds this to CI.

integration_tests/src/cucumber/error.rs
hansieodendaal: Thanks, will look at this in the consecutive PR that adds this to CI.

integration_tests/src/testing_framework/apps/bedrock.rs
hansieodendaal: See #773

lez/sequencer/core/src/block_publisher.rs
hansieodendaal: See issue #774

integration_tests/src/cucumber/error.rs
hansieodendaal: See issue #774

PR #743, opened 2026-08-16
feat(indexer): bound the peer-block cache behind a hash-certified refetch

moudyellaz:
## 🎯 Purpose

The verifier's peer-block cache holds every block ever read from a peer channel, so a long-lived peer grows the indexer's memory without bound. Bounding it naively would break the rule the forgery test rests on: inside the verified run, absence must mean the peer never produced the block, and an evicted block is not a forged one.

## ⚙️ Approach

- [x] Keep full block bodies only in a window behind the verified run tip (`peer_block_cache_window`, default 1024; zero rejected, `u32::MAX` for effectively unbounded), demoting older ids to an append-only index of `(slot, hash)` at 48 bytes per block, so index absence keeps meaning forgery
- [x] Split the inside-run lookup: a demoted id resolves as `Evicted` and is refetched from its recorded slot, served only if its recomputed hash matches the index and, with a pinned key, its signature verifies, so a same-content twin with a corrupted signature can never brick verification
- [x] Keep every refetch failure retriable in the lag class, logged on the shared alert cadence; escalation to a permanent verdict stays reserved for true Behind lookups, so an evicted body never becomes absence evidence
- [x] Leave run extension, equivocation and displacement rules untouched; the equivocation check also consults the index, so a differing re-read at an evicted id is still refused

## 🧪 How to Test

`cargo test -p indexer_core` and the `cross_zone_verified`, `cross_zone_ping`, `indexer_stall` e2e.

## 🔗 Dependencies

#731

## 🔜 Future Work

None

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests

### DIFF
--- a/lez/indexer/core/src/cross_zone_verifier.rs
+++ b/lez/indexer/core/src/cross_zone_verifier.rs
@@ -1,6 +1,7 @@
 use std::{
     collections::{HashMap, HashSet},
     fmt::{self, Display, Formatter},
+    num::{NonZeroU32, NonZeroU64},
     sync::{
         Arc, Mutex, MutexGuard, PoisonError,
         atomic::{AtomicU64, Ordering},
@@ -51,10 +52,6 @@ const PEER_BLOCK_WAIT_TIMEOUT: Duration = Duration::from_secs(300);
 /// advanced by, so `waited` counts sleeps rather than wall time and, since a
 /// sleep can overshoot, understates it.
 const PEER_BLOCK_POLL_INTERVAL: Duration = Duration::from_secs(1);
-
-/// Peer-block bodies kept behind each verified run tip when
-/// [`IndexerConfig::peer_block_cache_window`] is not configured.
-const DEFAULT_BODY_WINDOW: u64 = 1024;
 
 /// Why a cross-zone dispatch could not be verified.
 ///
@@ -246,17 +243,17 @@ struct PeerBlocks {
 
 impl Default for PeerBlocks {
     fn default() -> Self {
-        Self::new(None)
+        Self::new(crate::config::default_peer_block_cache_window())
     }
 }
 
 impl PeerBlocks {
-    /// `configured` is [`IndexerConfig::peer_block_cache_window`]; zero never
-    /// reaches here, [`CrossZoneVerifier::start`] rejects it.
-    fn new(configured: Option<u32>) -> Self {
+    /// `configured` is [`IndexerConfig::peer_block_cache_window`]; zero is
+    /// unrepresentable there.
+    fn new(configured: NonZeroU32) -> Self {
         Self {
             chains: Arc::default(),
-            window: configured.map_or(DEFAULT_BODY_WINDOW, u64::from),
+            window: NonZeroU64::from(configured).get(),
         }
     }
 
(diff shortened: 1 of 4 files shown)


lez/indexer/core/src/cross_zone_verifier.rs
```diff
@@ -47,6 +52,10 @@ const PEER_BLOCK_WAIT_TIMEOUT: Duration = Duration::from_secs(300);
 /// sleep can overshoot, understates it.
 const PEER_BLOCK_POLL_INTERVAL: Duration = Duration::from_secs(1);
 
+/// Peer-block bodies kept behind each verified run tip when
+/// [`IndexerConfig::peer_block_cache_window`] is not configured.
+const DEFAULT_BODY_WINDOW: u64 = 1024;
```
erhant: I feel like `peer_block_cache_window` should be `u32` instead of `Option<u32>`, defaulting to 1024?
We can still reject 0 in doing so.

erhant: actually a better idea: `NonZeroU32`

erhant (approved):
LGTM, one small nit 🙏🏻

lez/indexer/core/src/cross_zone_verifier.rs
moudyellaz: Took the `NonZeroU32`: zero is unrepresentable and rejected at config parse rather than at start, the `Option` is gone, omitted still means 1024. In af090f3c.

lez/indexer/core/src/cross_zone_verifier.rs
```diff
@@ -112,9 +118,23 @@ pub type SeenKey = (MessageKey, [u8; 32]);
 
 /// One peer zone's cached blocks, plus how far this reader has read them as an
```
agureev, line 119: Possibly worth expanding the structs description in the future given new fields?

lez/indexer/core/src/cross_zone_verifier.rs
```diff
@@ -172,20 +222,41 @@ impl PeerChain {
...
 }
 
 /// Cache of finalized peer-zone blocks, filled by per-peer reader tasks and read
 /// by the verifier to re-derive cross-zone dispatch transactions.
-#[derive(Clone, Default)]
+#[derive(Clone)]
 struct PeerBlocks {
     chains: Arc<RwLock<HashMap<ZoneId, PeerChain>>>,
+    /// Bodies kept behind each peer's run tip.
+    window: u64,
```
agureev, line 241: This seems to act just a default value to be filled for `insert` command for new `PeerChain`. Do you think it may be better to instead have it fed as an explicit argument or an `Option` so that upon a miss the default peer window is read?

agureev (approved):
looks good, left a two small comments

lez/indexer/core/src/cross_zone_verifier.rs
moudyellaz: The two-tier shape is documented on the fields right below; happy to lift a summary line up if you find it unclear.

lez/indexer/core/src/cross_zone_verifier.rs
moudyellaz: It is a per-construction copy of one config value, identical for every chain and immutable after start, so nothing can drift; keeping it on the chain keeps eviction self-contained, but happy to pass it as an argument if you prefer.

lez/indexer/core/src/cross_zone_verifier.rs
agureev: That's fine, not a principal point given the comments on the fields themselves.

lez/indexer/core/src/cross_zone_verifier.rs
agureev: We can leave it, there is no objective argument to be had here I think other than subjective code cleanliness argument, I agree that currently there is no drift.

PR #744, opened 2026-08-16
perf(lez)!: borrow account data in the guests instead of cloning

moudyellaz:
## 🎯 Purpose

Guests cloned whole account data buffers where the consumer only reads them, `clone().into_inner()` feeding functions that take `&[u8]`, and each clone costs risc0 cycles per delivery. One batch, because every guest edit moves image ids.

## ⚙️ Approach

- [x] Replace all 20 clone sites with borrows: 17 across six built-in guests (ping_sender, ping_receiver, wrapped_token, cross_zone_inbox, bridge_lock, clock) and 3 in the test guests; `Data` already carries `Deref` and `AsRef` to `[u8]`, so call sites deref-coerce
- [x] Two asserts additionally dropped a `to_vec`, comparing borrowed bytes directly

Six guest image ids move, so this rides the same genesis window as the stack below it.

## 🧪 How to Test

`RISC0_DEV_MODE=1 cargo test --release -p integration_tests --test cross_zone_state_machine`, the sequencer suite with `--features mock`, and the five cross-zone e2e.

## 🔗 Dependencies

#712

## 🔜 Future Work

None

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


erhant (approved):
LGTM ty!

agureev (approved):

PR #746, opened 2026-08-17
feat(fees): fee program skeleton with zero fees

erhant:
## 🎯 Purpose

First PR of the fee subsystem train (**#746** → #747 → #750 → #752 → #753 → #754 → #758 → #762). It introduces the structural skeleton the rest of the train builds on: a `FeeProgram` Risc0 guest, its three genesis-registered accounts, and a forced per-block fee transaction that every block-consuming path validates.

**This PR is behaviorally inert.** The fee transaction carries an all-zero `BlockFeeSummary`, the fee program echoes its accounts unchanged, no fee is ever charged, and no balance moves. What changes is block *structure*: every block now ends `…user txs, fee tx, clock tx`, and that shape is enforced everywhere blocks are built or replayed.

Interim policy until the train's later PRs land (and pending team decisions, e.g. the private-payer question): **all transaction classes remain fee-exempt**.

> [!NOTE]
>
> I've also added **`.claude/lez-fees/`** which are design/spec reference material for the whole PR stack, committed once here so later PRs can point at it. You can check them out if curious but I will remove them before merging this back to `dev`. I kept it in-code instead of Notion because it is easier to use a local scratchpad like that during development.

## ⚙️ Approach

The fee program is a system program in the same mold as the clock: sequencer-injected, never user-invocable, with its accounts protected from outside modification.

- [x] Add the `fee_core` crate (`lez/programs/fee/core`): `BlockFeeSummary` (`gas_used_exec`, `gas_used_stor`, `revenue_base`, `revenue_tip` — all zero for now) as the program's `Instruction`, plus faucet-style PDA seeds `/LEZ/v0.3/FeeSeed/{State,Escrow,Inbox}/` and the account-ID derivations.
- [x] Add the `fee` Risc0 guest (`lez/programs/fee/src/main.rs`): verifies it receives exactly the three fee accounts (state, escrow, inbox), verifies all are owned by the fee program, and re-emits them unchanged. Registered in `lez/programs` (`FEE_ELF`/`FEE_ID`, `programs::fee()`).
- [x] Register the three fee accounts at genesis, owned by the fee program with zero balance, in both `testnet_initial_state` and the sequencer's `build_genesis_state`; expose them via `system_accounts::{fee_state_account_id, fee_escrow_account_id, fee_inbox_account_id, fee_account_ids, fee_account}`.
[...]

PR #747, opened 2026-08-17
feat(fees): fee market computations under zero load

erhant:
## 🎯 Purpose

Second PR of the fee subsystem train (#746 → **#747** → #750 → #752 → #753 → #754 → #758 → #762). It adds the fee market's arithmetic and persistent state, and runs the market live in every block — but under zero load. Since metering and charging don't exist yet, every block's fee summary is all-zero, so the only observable change is what is *tracked*: the fee-state account now carries real market state that advances with the chain. Nothing is *charged* — base fees hold their floor, no balance moves, and the forced fee tx summary remains all-zero byte-for-byte.

## ⚙️ Approach

**`fee_core::market`** (`lez/programs/fee/core/src/market.rs`) — the dual EIP-1559-style market, one controller per resource:

- Protocol constants: execution `TARGET_GAS_EXEC = 5M` / `MAX_GAS_EXEC = 10M`, storage `TARGET_GAS_STOR = 500K` / `MAX_GAS_STOR = 1M`, damping `d = 8` for both, minimum base fee `8` for both, `SMOOTHING_WINDOW = 50`. Compile-time asserts pin `MAX = 2·TARGET` for both resources (the ±12.5% bound and elasticity framing depend on it). `BASE_FEE_*_MAX = u64::MAX / MAX_GAS_*` guarantees `gas · price` fits `u64`.
- `next_base_fee(b, g, target, d, lo, hi)`: delta-form update with the deviation clamped to `target` (so no usage value can move the price more than `b/d`), a guaranteed `max(1, ·)` up-step (the price can always rise off the floor, even where the proportional delta rounds to 0), products widened to `u128` (no overflow at the price cap), and saturation into `[lo, hi]`.

**`fee_core::state::FeeState`** (`lez/programs/fee/core/src/state.rs`) — the persistent market state, Borsh-serialized into the fee-state account's `data`: `{base_fee_exec, base_fee_stor, window: [u128; 50], payout_carry, height}`. The revenue window is a height-keyed ring (`slot = height % 50`); the escrow balance is deliberately *not* a field — it lives in the escrow account. `apply_block(summary)` advances the height, records the block's base revenue, computes the smoothed payout `⌊(carry + Σ window) / 50⌋` with an exact carry (nothing stranded, carry always < 50), and steps both base fees.

**Fee guest** (`lez/programs/fee/src/main.rs`) — no longer echoes accounts unchanged: it rejects summaries above the per-block gas caps, deserializes `FeeState` from the state account, applies the market update, and writes the new state back. Since the transition still pins the summary all-zero, it asserts the payout is 0 and passes escrow/inbox through untouched.

**Genesis wiring** — `system_accounts::fee_state_account()` seeds the state account with `FeeState::genesis()` (fees at floor, empty window, height 0); `testnet_initial_state` and the storage-indexer test fixture distinguish the state account (with data) from escrow/inbox (plain empty fee accounts).

- [x] Add `fee_core::market`: dual-market constants and the `next_base_fee` controller, with compile-time `MAX = 2·TARGET` asserts
- [x] Add `fee_core::state::FeeState`: Borsh-serialized market state with the 50-block revenue ring and payout carry
- [x] Change the fee guest to apply the per-block market update (summary still validated all-zero by the transition)
- [x] Change genesis to seed the fee-state account with `FeeState::genesis()` (`system_accounts`, `testnet_initial_state`, storage test fixture)
- [x] Add `lez/programs/fee/core` to workspace members — deliberate: unlike the other `*_core` program crates (path-deps only), this puts fee_core's unit tests in CI
- [x] Add a chain_state test that fee state advances with the chain under zero load (height tracks blocks, fees hold the floor, escrow/inbox balances stay 0)

Note on the diff: the 17 `artifacts/lez/programs/*.bin` guest binaries and `test_fixtures/fixtures/prebuilt_sequencer_db.dump` are expected rebuild churn (the fee guest changed, and all guests re-link) — no need to diff them.

## 🧪 How to Test

```sh
```

fee_core's unit tests cover the market math directly: floor holding under zero load, the guaranteed +1 up-step off the floor, exact ±12.5% at full deviation, the deviation clamp under extreme usage, saturation at both bounds (including the u128-widening case at the price cap), a bounded-adjustment grid over price × usage extremes, exact amortization of a revenue pulse over the 50-block window, and payout ≤ cumulative revenue at every block. `chain_state::apply` and `testnet_initial_state` tests verify the end-to-end zero-load invariants.

Plus the standard lints: `cargo +nightly fmt`, `cargo clippy --workspace --all-targets --all-features -- -D warnings`, `taplo fmt`, `cargo machete`, `cargo deny check`.

## 🔗 Dependencies

- #746 (base branch `erhant/fees/initial-fee-program`) — the fee program skeleton, forced fee tx, and fee system accounts this builds on.

## 🔜 Future Work

The rest of the train, in order:

- #750 — transaction wire format (fee fields on transactions)
- #752 — metering (real gas measurements feeding the summary)
- #753 — assessment formulas and producer identity
- #754 — actual charging (balances move; the zero-payout assert and all-zero summary pin are lifted)
- #758 — admission control
- #762 — fee-aware integration suite + LGO-scale fixture (closes the train)

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


lez/chain_state/src/apply.rs
```diff
@@ -285,6 +285,44 @@ mod tests {
         assert!(matches!(err, BlockIngestError::EmptyBlock));
     }
 
+    #[test]
+    fn fee_state_advances_with_the_chain() {
```
jonesmarvin8, line 289: Just to clarify, this is a framework test that will be updated with actual fee changes in future PRs?

lez/chain_state/src/apply.rs
```diff
@@ -285,6 +285,44 @@ mod tests {
...
+                .balance,
+            0
+        );
+        assert_eq!(
+            state
+                .get_account_by_id(system_accounts::fee_inbox_account_id())
+                .balance,
+            0
+        );
+    }
```
jonesmarvin8, line 324: Since the specs require escrow withdraws for block production to be handled over 50 blocks. Could we have a test that "demonstrates" this? This can be saved for a future PR. Mentioning it here since the framework could be initialized here.

lez/programs/fee/core/src/market.rs
```diff
@@ -0,0 +1,211 @@
...
+/// One base-fee update.
+///
+/// The deviation clamp bounds the move to `max(1, b / d)`. Products are widened
+/// to u128 to keep the function total over any u64 input; under the
+/// `b <= BASE_FEE_*_MAX` invariant the u64 products already fit
+/// (`b * deviation <= u64::MAX / 2`), so the widening is robustness, not
+/// overflow avoidance. A moving branch saturates into `[lo, hi]`; at target `b`
+/// is returned unchanged (callers keep `b` within `[lo, hi]`).
+#[must_use]
+pub fn next_base_fee(b: u64, g: u64, target: u64, d: u64, lo: u64, hi: u64) -> u64 {
```
jonesmarvin8: Is it necessary to retain the 1 character variable names from the specs?

lez/programs/fee/core/src/state.rs
```diff
@@ -0,0 +1,216 @@
+//! Persistent fee market state, stored in the fee-state account's data.
+
+#![expect(
+    clippy::arithmetic_side_effects,
```
Arjentix: Hm, are you sure about this lint? I'd better place it in concrete places because this is the important code

lez/programs/fee/core/src/state.rs
```diff
@@ -0,0 +1,216 @@
...
+              carry, and additions are checked"
+)]
+
+use borsh::{BorshDeserialize, BorshSerialize};
+
+use crate::{BlockFeeSummary, market};
+
+/// The fee market's persistent state, Borsh-serialized into the fee-state
+/// account's `data`. The escrow is the escrow *account balance*, deliberately
+/// not a field here.
```
Arjentix:
```suggestion
/// The fee market's persistent state.
```

I think other information is redundant

lez/programs/fee/core/src/market.rs
```diff
@@ -0,0 +1,211 @@
...
+            ),
+            100
+        );
+    }
+
+    #[test]
+    fn live_upward_from_the_minimum() {
+        // One unit above target must move the price up by at least 1, even at
+        // b = 8 where the proportional delta rounds to 0.
+        let next = next_base_fee(
```
jonesmarvin8, line 129: For readability, could we have `b=8` defined above? Similar to the next test.

lez/programs/fee/core/src/state.rs
```diff
@@ -0,0 +1,216 @@
...
+pub struct FeeState {
+    /// Execution base fee for the current block.
+    pub base_fee_exec: u64,
+    /// Storage base fee for the current block.
+    pub base_fee_stor: u64,
+    /// Base revenue of the last [`market::SMOOTHING_WINDOW`] blocks; slot
+    /// `height % SMOOTHING_WINDOW` is the most recent.
+    pub window: [u128; market::SMOOTHING_WINDOW],
+    /// Payout division remainder, always < [`market::SMOOTHING_WINDOW`].
+    pub payout_carry: u128,
```
jonesmarvin8, line 20: Should this be `Balance` instead of hardcoded as `u128`?

lez/programs/fee/core/src/state.rs
```diff
@@ -0,0 +1,216 @@
...
+
+/// The fee market's persistent state, Borsh-serialized into the fee-state
+/// account's `data`. The escrow is the escrow *account balance*, deliberately
+/// not a field here.
+#[derive(Debug, Clone, PartialEq, Eq, BorshSerialize, BorshDeserialize)]
+pub struct FeeState {
+    /// Execution base fee for the current block.
+    pub base_fee_exec: u64,
+    /// Storage base fee for the current block.
+    pub base_fee_stor: u64,
```
jonesmarvin8: Perhaps a `Fee` alias should be defined for u64?

lez/programs/fee/core/src/state.rs
```diff
@@ -0,0 +1,216 @@
...
+    fn serialization_round_trips() {
+        let mut state = FeeState::genesis();
+        state.apply_block(&summary_with_revenue(12_345));
+        assert_eq!(FeeState::from_bytes(&state.to_bytes()), state);
+    }
+
+    #[test]
+    fn serialized_layout_is_pinned() {
+        // The state lives in consensus account data, so its byte layout is part
+        // of the protocol: 8 (base_fee_exec) + 8 (base_fee_stor) + 50·16
```
jonesmarvin8, line 135: For some of these constants `BASE_FEE_EXEC_MIN` and etc, can we define these so that the equation for `EXPECTED_LEN` is given in terms of these constants instead of "magical" numbers?

lez/chain_state/src/apply.rs
jonesmarvin8: Discovered this test in fee::core::state.

Arjentix (approved):
LGTM, but I'd summon fee designers here to check calculations

jonesmarvin8 (approved):
Lgtm. Left a few minor comments.

lez/programs/fee/core/src/state.rs
Arjentix: Isn't it strange that Fee is u64, but Balance is u128?

schouhy (approved):
I'm not sure about the need of tracking the FeeState in the LEZ state. But I'll follow along.

moudyellaz:
> I'm not sure about the need of tracking the FeeState in the LEZ state. But I'll follow along.

@schouhy the payout is computed inside the fee guest, so the market state has to be an account the guest can read and write. Deriving it node-locally from block summaries would work but would put it outside anything the proof covers.

moudyellaz (approved):
lgtm, thanks!

lez/programs/fee/core/src/state.rs
erhant: yet again spec stuff

> Base fees, gas amounts, and per-resource fee products MUST fit in 64 bits; the parameter caps below guarantee this for every valid block. Fee totals, balances, revenue totals, and every intermediate product MUST be computed in at least 128-bit width.

I agree (along with @jonesmarvin8 comments on #746) that I can add alias'es to these at elast, e.g. `Fee` and `Balance` or such; but let me read all reviews together to come up with an all-encompassing solution that can address all at once, e.g. `fix here, and apply the fixes as I rebase to PRs up the stack`

lez/programs/fee/core/src/state.rs
erhant: this is actually a tiny number, in the specs its in the range `0..49` because the smoothing window is 50
keeping it same type eases the computations I think

lez/programs/fee/core/src/state.rs
erhant: Shrunk in https://github.com/logos-blockchain/logos-execution-zone/compare/4227c38d08f6e1aac8f930f7aceeca0734ac96e9..8e14df06e3f532c8cecd7703abfc9ee84d1d14e1

lez/programs/fee/core/src/state.rs
erhant: Sure thing, placed where it is needed https://github.com/logos-blockchain/logos-execution-zone/compare/4227c38d08f6e1aac8f930f7aceeca0734ac96e9..8e14df06e3f532c8cecd7703abfc9ee84d1d14e1

lez/programs/fee/core/src/state.rs
erhant: Done in https://github.com/logos-blockchain/logos-execution-zone/compare/4227c38d08f6e1aac8f930f7aceeca0734ac96e9..8e14df06e3f532c8cecd7703abfc9ee84d1d14e1

lez/programs/fee/core/src/state.rs
erhant: Done in https://github.com/logos-blockchain/logos-execution-zone/compare/4227c38d08f6e1aac8f930f7aceeca0734ac96e9..8e14df06e3f532c8cecd7703abfc9ee84d1d14e1

lez/programs/fee/core/src/state.rs
erhant: I think this can stay as `u128`; it's definitely not Balance-related so no type aliasing for this one; and it's harmless for it to stay w.r.t spec as is

lez/programs/fee/core/src/market.rs
erhant: Of course, we can be verbose;

Done in https://github.com/logos-blockchain/logos-execution-zone/compare/4227c38d08f6e1aac8f930f7aceeca0734ac96e9..8e14df06e3f532c8cecd7703abfc9ee84d1d14e1

lez/programs/fee/core/src/market.rs
erhant: Done in https://github.com/logos-blockchain/logos-execution-zone/compare/4227c38d08f6e1aac8f930f7aceeca0734ac96e9..8e14df06e3f532c8cecd7703abfc9ee84d1d14e1

lez/chain_state/src/apply.rs
erhant: yep, zero-load coverage for what this PR wires up (height advances, floor holds, escrow/inbox empty).

the charged-path tests (revenue, congestion, payouts) come with the metering/charging PRs later

lez/programs/fee/core/src/market.rs
jonesmarvin8: Thanks!

lez/programs/fee/core/src/state.rs
jonesmarvin8: You're right. Since it's a small number though. Does it make sense to use u8 instead of u128? For efficient storage. Though, this would result in conversion between u8 to u128 for arithmetic later on.

lez/programs/fee/core/src/state.rs
jonesmarvin8: Thanks!

lez/programs/fee/core/src/state.rs
jonesmarvin8: Thanks!

lez/programs/fee/core/src/state.rs
erhant: Besides extra conversions, I would in particular prefer to keep `u128` because this is a _remainder_ of `numerator = payout * window_len + payout_carry`, i.e.

```rs
let payout = numerator / window_len;
let payout_carry = numerator % window_len;
```

Here, `numerator` is the sum of revenues (u128), and window length is `SMOOTHING_WINDOW`; which is technically at most `u64` as per the block height. In any case, keeping `u128` is the most sensible option, and `u8` does not save anything really compared to all the other stuff going on.

lez/programs/fee/src/main.rs
```diff
@@ -48,8 +49,32 @@ fn main() {
...
+    // pinned all-zero this cannot bite, but once charging lifts the pin an
+    // attacker-supplied `revenue_base` near u128::MAX would reach the smoothing
+    // window and trip its checked-add (a consensus fault). Bound it to
+    // `gas_used_exec·base_fee_exec + gas_used_stor·base_fee_stor` then.
+    if summary.gas_used_exec > market::MAX_GAS_EXEC || summary.gas_used_stor > market::MAX_GAS_STOR
+    {
+        panic!("Block fee summary exceeds per-block gas caps");
+    }
+
+    let mut fee_state = FeeState::from_bytes(&pre_state.account.data.clone().into_inner());
```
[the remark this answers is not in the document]
erhant: Fixed in https://github.com/logos-blockchain/logos-execution-zone/pull/747/commits/44f7f5fde396cfd9b139577a170e99145732688e

lez/programs/fee/core/src/state.rs
```diff
@@ -0,0 +1,223 @@
...
+            .expect("height increment at u64::MAX is a consensus fault");
+
+        let window_len = u64::try_from(market::SMOOTHING_WINDOW).expect("window length fits u64");
+        let slot = usize::try_from(self.height % window_len).expect("slot index fits usize");
+        self.window[slot] = summary.revenue_base;
+
+        let numerator = self.window.iter().fold(self.payout_carry, |acc, revenue| {
+            acc.checked_add(*revenue)
+                .expect("window sum of 50 u128 revenues cannot overflow in practice")
+        });
```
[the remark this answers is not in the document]
erhant: Keeping `u128` as is

erhant:
Closing in favor of #801

PR #750, opened 2026-08-17
feat(fees): fee fields and payer authorization on the wire format

erhant:
## 🎯 Purpose

Third PR of the fees train (#746 → #747 → **#750** → #752 → #753 → #754 → #758 → #762). It puts the fee subsystem's data on the wire: every public transaction message now carries signed fee fields (`payer`, `gas_limit`, `tip`, `max_fee`), and the witness set can carry a sponsor's fee authorization.

> [!IMPORTANT]
> **This PR is consensus-relevant: it changes the public message hash layout.** The four fee fields sit flat on `Message` after `instruction_data` and are covered by the Borsh-serialized message hash, so every witness signature authorizes them — tampering with a fee field after signing invalidates the whole witness set. The pinned-hash unit test was updated accordingly (+64 zero bytes for `FeeFields::ZERO`). This is the part to review most carefully.

Design ruling baked in here (fees team, Q1): **the payer is always explicitly designated on the message, never inferred from the witness set.** A payer is fee-authorized iff it is one of the ordinary signers *or* it supplies the dedicated fee witness — which is what makes sponsored/third-party fee payment expressible at the wire level.

Nothing is enforced yet: fee fields are carried, hashed, and displayed, but no code charges, meters, or rejects based on them. Beyond the serialization change, this PR is behaviorally inert.

## ⚙️ Approach

**`lee`: wire format + authorization rules**

- [x] Add fee fields to the public `Message` (`lee/state_machine/src/public_transaction/message.rs`), flat after `instruction_data`, covered by the message hash.
  - `Message::try_new` keeps its arity and fills in `FeeFields::ZERO` — correct for exempt/system transactions and while the zero-fee policy holds. Charged construction goes through the new `try_new_with_fees`; `new_preserialized` gains the `fees` parameter. This wrapper strategy is deliberate churn control: dozens of existing call sites stay untouched until the enforcement PRs flip them over.
- [x] Add an optional `fee_witness` slot to `WitnessSet` (`witness_set.rs`): `with_fee_signer` attaches a sponsor's signature over the same message hash, `is_valid_for` chains it into verification (an invalid fee witness invalidates the set).
- [x] New `lee::fees` module (`lee/state_machine/src/fees.rs`): `FeeFields` (+ `ZERO`), `SYSTEM_PAYER = [0; 32]` (the fee-exempt payer recorded on system-generated transactions), the `SignedMessage` trait, and the authorization rules `fee_authorized_account_ids` / `is_fee_authorized`.

**Wallet wiring** (folded in so the new fields are exercised end to end)

- [x] The wallet designates the transaction's **first public signing account** as payer (`AccountManager::fee_payer_account_id`); its ordinary signature already fee-authorizes it, so no separate fee witness is needed. `send_pub_tx`'s signature is unchanged.
- [x] Defaults (`lez/wallet/src/lib.rs`): `DEFAULT_GAS_LIMIT = 2_000_000` (~3× the widest measured program call, 1/5 of the per-block cap), tip `0`, and `DEFAULT_MAX_FEE = (2M gas + 100k data bytes) × 64 assumed base fee = 134,400,000` — sized against 8× the genesis-minimum base fee so defaults survive early congestion without re-signing.

**Read path + fallout**

- [x] Indexer protocol carries the fee fields (`lez/indexer/service/protocol/src/{lib,convert}.rs`) and the explorer displays payer / gas limit / tip / max fee on transaction details.
- [x] The indexer **FFI surface is deliberately lossy for now**: fee fields are dropped on the way out and reconstructed as zeros on the way in, both sites FIXME'd (`lez/indexer/ffi/src/api/types/transaction.rs`).
- [x] The 64-byte message growth tripped `clippy::large_enum_variant` on the sequencer's gossip `TxEvaluation`; suppressed with a reasoned `#[expect]` (boxing would allocate on the hot validation path).
- [x] `test_fixtures/fixtures/prebuilt_sequencer_db.dump` regenerated (message layout change).

**Deliberately not changed**: deployment messages did **not** gain fee fields — the deployment message is unsigned (no witness set), so its fee fields await the deployment signing / replay-protection design (open spec question). No guest binary changes in this PR (`artifacts/` untouched).

## 🧪 How to Test

Mirror CI:

```sh
RISC0_DEV_MODE=1 cargo nextest run -p integration_tests -p test_fixtures
cargo +nightly fmt --check && taplo fmt --check && cargo machete && cargo deny check
```

Unit tests in `lee` cover the new surface directly:

- `fees.rs`: self-payer authorized by its ordinary signature; sponsor payer rejected without and accepted with the fee witness; tampered fee fields invalidate every signature (proving the fields are inside the signed hash); a fee witness over the wrong message invalidates the set; zero-fields Borsh round trip.
- `message.rs`: the pinned byte-level hash preimage test, updated for the 64 appended fee bytes.
- `account_manager.rs`: payer selection (first public account; `None` when there is no public account).

## 🔗 Dependencies

- #747 (fee computations) — base of this branch.

## 🔜 Future Work

- Gas metering in the VM (#752), fee assessment + block producer integration (#753), enforcement in the state machine (#754), and mempool admission checks (#758), with the fee-aware integration-suite conversion closing the train (#762) — until then the fields are carried but never enforced.
- Fee fields on deployment messages, pending the deployment signing / replay-protection spec ruling.
- Carry fee fields (and the fee witness) across the indexer FFI instead of zeroing them.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


lee/state_machine/src/public_transaction/message.rs
```diff
@@ -16,6 +16,10 @@ pub struct Message {
     pub account_ids: Vec<AccountId>,
     pub nonces: Vec<Nonce>,
     pub instruction_data: InstructionData,
+    pub payer: AccountId,
+    pub gas_limit: u64,
+    pub tip: u64,
```
jonesmarvin8: Can we introduce aliases for these u64's to avoid hardcoding types (and to match with previous conventions of using aliases.

lee/state_machine/src/public_transaction/message.rs
```diff
@@ -16,6 +16,10 @@ pub struct Message {
...
+    pub max_fee: u128,
```
Arjentix: So I get why we need `gas_limit` and `max_fee`, but I don't get why we need `payer` and `tip`...

lee/state_machine/src/public_transaction/message.rs
Arjentix, line 23: Let's apply destructuring pattern here to not to forget to update this impl in future if needed

lee/state_machine/src/public_transaction/message.rs
```diff
@@ -58,15 +86,25 @@ impl Message {
...
+            gas_limit: fees.gas_limit,
+            tip: fees.tip,
+            max_fee: fees.max_fee,
         }
     }
 
+    #[must_use]
+    pub const fn fees(&self) -> FeeFields {
+        FeeFields::new(self.payer, self.gas_limit, self.tip, self.max_fee)
+    }
```
Arjentix: Maybe better to name it `FeeInfo` and store it as `fee_info` field?

lee/state_machine/src/public_transaction/witness_set.rs
```diff
@@ -22,13 +23,35 @@ impl WitnessSet {
...
         Self {
             signatures_and_public_keys,
+            fee_witness: None,
         }
     }
 
+    /// Adds a sponsor's fee authorization: a signature over the same message
+    /// hash by an account outside the ordinary witness set.
+    #[must_use]
+    pub fn with_fee_signer(mut self, message: &Message, payer_key: &PrivateKey) -> Self {
```
Arjentix: This creates the possibility to first create witness for one message and then add fee signer for another message...

I understand that this is just simpler for backward-compatibility, but I suggest to fix it somehow. Shouldn't fees be obligatory? Then we can move payer to the `for_message()` constructor

lee/state_machine/src/public_transaction/witness_set.rs
```diff
@@ -50,6 +73,23 @@ impl WitnessSet {
     pub const fn from_raw_parts(signatures_and_public_keys: Vec<(Signature, PublicKey)>) -> Self {
         Self {
             signatures_and_public_keys,
+            fee_witness: None,
+        }
+    }
```
Arjentix: Not so `from_raw_parts()` anymore... As well as `into_raw_parts()`

lee/state_machine/src/public_transaction/witness_set.rs
```diff
@@ -50,6 +73,23 @@ impl WitnessSet {
...
+    #[must_use]
+    pub const fn from_raw_parts_with_fee_witness(
+        signatures_and_public_keys: Vec<(Signature, PublicKey)>,
+        fee_witness: Option<(Signature, PublicKey)>,
+    ) -> Self {
+        Self {
+            signatures_and_public_keys,
+            fee_witness,
         }
     }
```
Arjentix: I'd remove this in favor of true `from_raw_parts()`

lee/state_machine/src/fees.rs
```diff
@@ -0,0 +1,165 @@
...
+    fn payer(&self) -> AccountId;
+}
+
+/// Accounts whose valid signature over `message` accompanies the transaction:
+/// the ordinary witnesses plus the fee witness, if any.
+#[must_use]
+pub fn fee_authorized_account_ids<M: SignedMessage>(
+    message: &M,
+    witness_set: &WitnessSet,
+) -> Vec<AccountId> {
```
Arjentix: I guess you abstracted from public and private messages with this `SignedMessage`. But that means you also need to abstract from `WitnessSet` which is public here, while private `WitnessSet` is a different type

jonesmarvin8 (approved):
Lgtm. Left some minor comments.

lez/wallet/src/account_manager.rs
```diff
@@ -447,6 +447,26 @@ impl AccountManager {
...
 
+    /// The account that pays this transaction's fee: the first public signing
+    /// account. Its ordinary signature covers the message, so it is always
+    /// fee-authorized without a separate fee witness.
+    ///
+    /// A tracked but unsignable public account (`sk: None`, e.g. an AMM pool or
+    /// token-definition PDA the wallet passes as a non-signing input) must be
+    /// skipped: designating it payer writes an account nothing can authorize
+    /// into the signed `payer` field, and the transaction is then rejected for
+    /// a missing fee authorization once fees are enforced.
```
Arjentix:
```suggestion
    /// The account that pays this transaction's fee: the first public signing
    /// account.
```

lee/state_machine/src/public_transaction/message.rs
```diff
@@ -16,6 +16,10 @@ pub struct Message {
     pub account_ids: Vec<AccountId>,
     pub nonces: Vec<Nonce>,
     pub instruction_data: InstructionData,
+    pub payer: AccountId,
```
schouhy: `fee_payer` would be more descriptive I think

lee/state_machine/src/public_transaction/witness_set.rs
schouhy: I second this comment

lee/state_machine/src/fees.rs
```diff
@@ -0,0 +1,165 @@
...
+
+use crate::{AccountId, public_transaction::WitnessSet};
+
+/// The payer recorded on system-generated transactions, which are fee-exempt.
+pub const SYSTEM_PAYER: AccountId = AccountId::new([0_u8; 32]);
+
+/// The signed fee fields of a transaction message. Stored flat on each message
+/// kind and covered by its hash, so every witness signature authorizes them.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, BorshSerialize, BorshDeserialize)]
+pub struct FeeFields {
```
schouhy: I'd name this `FeeDeclaration`

lez/sequencer/core/src/gossip/validation.rs
```diff
@@ -9,6 +9,10 @@ use common::transaction::LeeTransaction;
 const BLOCK_HEADER_OVERHEAD: u64 = 200;
 
 #[derive(Debug)]
+#[expect(
+    clippy::large_enum_variant,
+    reason = "Accept dominates and the value is transient per gossiped message; boxing would add an allocation on the hot validation path"
```
schouhy: no idea what this means

schouhy (approved):
LGTM left some comments

moudyellaz (approved):
lgtm, thanks!

lee/state_machine/src/public_transaction/message.rs
erhant: `tip` is an extra payment by the `payer` (as per the spec) that incentivizes the sequencer to pick the ones with the most fees

from the specs:
> The producer MAY use tips to choose an ordering; the protocol does not constrain that choice, but the order encoded in the block is, like all block contents, consensus data.

we do have such an ordering, but not in this PR but a later one; I could've been more pedantic and add those later when they are needed but I wanted to keep the wire format PR more complete 🙏🏻

lee/state_machine/src/public_transaction/message.rs
erhant: following the spec on this, we would have to escalate it to the tokenomics team

lee/state_machine/src/public_transaction/message.rs
erhant: I think I will go with https://github.com/logos-blockchain/logos-execution-zone/pull/750#discussion_r3818172103 `FeeDeclaration`

lez/sequencer/core/src/gossip/validation.rs
erhant: Without the `allow`, it will get mad for [`large_enum_variant`](https://rust-lang.github.io/rust-clippy/rust-1.94.0/index.html#large_enum_variant) like such:

```rs
error: large size difference between variants
  --> lez/sequencer/core/src/gossip/validation.rs:16:1
   |
16 | / pub enum TxEvaluation {
17 | |     /// Structurally valid and authenticated; forward and admit.
18 | |     Accept(LeeTransaction),
   | |     ---------------------- the largest variant contains at least 304 bytes
19 | |     /// Malformed / forbidden; do not forward. `GossipSub` peer scoring is not
20 | |     /// configured, so this does not currently penalize the propagating peer.
21 | |     Reject(String),
   | |     -------------- the second-largest variant contains at least 24 bytes
22 | | }
   | |_^ the entire enum is at least 304 bytes
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/rust-1.94.0/index.html#large_enum_variant
   = note: `-D clippy::large-enum-variant` implied by `-D clippy::all`
   = help: to override `-D clippy::all` add `#[allow(clippy::large_enum_variant)]`
help: consider boxing the large fields or introducing indirection in some other way to reduce the total size of the enum
   |
18 -     Accept(LeeTransaction),
18 +     Accept(Box<LeeTransaction>),
   |
```

The reason string _"Accept dominates and the value is transient per gossiped message; boxing would add an allocation on the hot validation path"_ basically says that its not worth Box'ing this guy, as its the majority of hits on this enum. The lint also says:

> This lint obviously cannot take the distribution of variants in your running program into account. It is possible that the smaller variants make up less than 1% of all instances, in which case the overhead is negligible and the boxing is counter-productive. Always measure the change this lint suggests.

I do agree the reason string can be made simpler though, will do so 🙏🏻

erhant: Simplified in https://github.com/logos-blockchain/logos-execution-zone/pull/750/commits/adb3077c8e988ea6f34b8c92d77aa1ca34581960

lee/state_machine/src/fees.rs
erhant: Done in https://github.com/logos-blockchain/logos-execution-zone/pull/750/commits/adb3077c8e988ea6f34b8c92d77aa1ca34581960

lez/wallet/src/account_manager.rs
erhant: Simplified in https://github.com/logos-blockchain/logos-execution-zone/pull/750/commits/adb3077c8e988ea6f34b8c92d77aa1ca34581960

lee/state_machine/src/public_transaction/message.rs
erhant: Done as `FeeDeclaration` in https://github.com/logos-blockchain/logos-execution-zone/pull/750/commits/adb3077c8e988ea6f34b8c92d77aa1ca34581960

lee/state_machine/src/public_transaction/witness_set.rs
erhant: I would rather do this in a later PR on top of all these I think, with some FIXME comment
it will GREATLY increase the changed lines, `for_message` has like 100 calls that would require this change

lee/state_machine/src/public_transaction/witness_set.rs
erhant: Similar to `for_message` (https://github.com/logos-blockchain/logos-execution-zone/pull/750#discussion_r3818156300) this has quite a lot of calls; I can change both of them in a single PR that touches a lot of files with the simple mechanical change

lee/state_machine/src/public_transaction/witness_set.rs
erhant: In line with https://github.com/logos-blockchain/logos-execution-zone/pull/750#discussion_r3818156300 comment, I can handle this in a new PR that touches a lot of parts.

lee/state_machine/src/public_transaction/message.rs
erhant: Done in https://github.com/logos-blockchain/logos-execution-zone/pull/750/commits/a0a8e3b805317241f61a4bb7b84b5f1fa99d5ead

lee/state_machine/src/public_transaction/message.rs
erhant: Done in https://github.com/logos-blockchain/logos-execution-zone/pull/750/commits/dc471f0e2df9b1e514115254998d5145b0490c99

lee/state_machine/src/fees.rs
erhant: I think we can handle this when we deal with pptx fees, if thats ok?

lee/state_machine/src/public_transaction/witness_set.rs
erhant: https://github.com/logos-blockchain/logos-execution-zone/issues/781

lee/state_machine/src/public_transaction/witness_set.rs
erhant: https://github.com/logos-blockchain/logos-execution-zone/issues/781

lee/state_machine/src/public_transaction/witness_set.rs
erhant: https://github.com/logos-blockchain/logos-execution-zone/issues/780

lee/state_machine/src/public_transaction/message.rs
jonesmarvin8: I think this is a "minor" enough relabeling that we shouldn't need to escalate. Just need to make note so that their document can be updated?

erhant: I'm moving all these fields under `fee: Option<FeeDeclaration>`, so now `payer` will be more well-read; i.e.` message.fee.payer` instead of `message.payer`

lee/state_machine/src/public_transaction/witness_set.rs
erhant: With the new logic where fee logic is part of `Message`, this is no longer an issue.

lee/state_machine/src/public_transaction/witness_set.rs
erhant: With the new logic where fee logic is part of `Message`, this is no longer an issue.

lee/state_machine/src/public_transaction/witness_set.rs
erhant: With the new logic where fee logic is part of `Message`, this is no longer an issue.

lee/state_machine/src/fees.rs
```diff
@@ -0,0 +1,169 @@
...
+}
+
+/// Whether the transaction's fee is authorized: a fee-exempt message always is;
+/// a charged message requires its designated payer to be among the signers.
+#[must_use]
+pub fn is_fee_authorized<M: SignedMessage>(message: &M, witness_set: &WitnessSet) -> bool {
+    message
+        .payer()
+        .is_none_or(|payer| fee_authorized_account_ids(message, witness_set).contains(&payer))
+}
```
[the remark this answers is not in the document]
erhant: Fixed in https://github.com/logos-blockchain/logos-execution-zone/pull/750/commits/0c6c9b54bef7c8945c1743278c3cdbd3f4e09a02

lee/state_machine/src/fees.rs
Arjentix: Then why we need `SignedMessage` now?

lez/sequencer/core/src/gossip/validation.rs
```diff
@@ -9,6 +9,9 @@ use common::transaction::LeeTransaction;
 const BLOCK_HEADER_OVERHEAD: u64 = 200;
 
 #[derive(Debug)]
+// `Accept` is intentionally left unboxed: it is the common outcome and the enum
+// is short-lived per gossiped message, so boxing it would only add a heap
+// allocation on the hot validation path.
```
Arjentix, line 14: So I think this comment is no longer needed as there are no clippy warning. Otherwise it should have been `#[expect]`

lez/wallet/src/lib.rs
```diff
@@ -57,6 +57,25 @@ pub const SUPPRESS_VERBOSE_PRINTS: &str = "SUPPRESS_VERBOSE_PRINTS";
...
+/// Serialized-size allowance the default `max_fee` is sized against.
+const ASSUMED_DATA_BYTES: u128 = 100_000;
+
+/// Default cap on the fee reservation for wallet-built public transactions.
+#[expect(
+    clippy::as_conversions,
+    reason = "u128::from is not const; the widening is lossless"
+)]
+pub const DEFAULT_MAX_FEE: u128 =
+    (DEFAULT_GAS_LIMIT as u128 + ASSUMED_DATA_BYTES) * ASSUMED_BASE_FEE;
```
Arjentix, line 77: I don't understand the meaning here... How can we sum gas limit and bytes?

Arjentix (approved):
LGTM, some small questions

moudyellaz (approved):
lgtm, thanks!

moudyellaz (approved):
lgtm, thanks!

moudyellaz (approved):
lgtm, thanks!

erhant:
Closing in favor of #801

PR #752, opened 2026-08-17
feat(fees): cycle-count metering for public execution

erhant:
## 🎯 Purpose

Fourth PR of the fees train (#746 → #747 → #750 → **#752** → #753 → #754 → #758 → #762).

To charge for public execution we first have to be able to *measure* it. This PR makes public execution metered: every risc0 executor session now reports how many cycles it consumed and runs under an explicit cycle budget, with budget exhaustion surfacing as a typed `LeeError::OutOfGas` instead of an opaque execution failure.

This PR only adds the instrument — **zero behavior change, zero ripple**. All existing entry points keep their signatures (as thin wrappers over the budgeted forms), the default budget equals the previous hardcoded 32M-cycle session limit, and no callers outside `lee/state_machine` are touched. Actually deriving budgets from `gas_limit` and charging for cycles lands later in the train (#753/#754).

## ⚙️ Approach

- Change `Program::execute` to take a `cycle_budget: u64` and return `(ProgramOutput, u64)`, where the cycle count comes from risc0's `SessionInfo::cycles()`. The budget is enforced by passing it as the executor's `session_limit` (`lee/state_machine/src/program/mod.rs`).
- Add `Program::execute_session` as the **single recognition point** that translates the executor's session-limit bail into `LeeError::OutOfGas { budget }`. This is a string match on `"Session limit exceeded"` — admittedly brittle, and marked with a `FIXME` in the code; risc0 doesn't expose a typed error for this today, and confining the match to one function keeps the blast radius minimal.
- Promote the old `MAX_NUM_CYCLES_PUBLIC_EXECUTION` constant to `pub const DEFAULT_PUBLIC_CYCLE_BUDGET` (same 32M value), used by every path that doesn't yet carry a transaction-specific budget — preserving existing behavior exactly.
- Add `ValidatedStateDiff::from_public_transaction_with_budget` returning `(Self, ExecutionOutcome)`. The budget is **shared across the whole call chain**: each nested/chained program session gets `cycle_budget.saturating_sub(cycles_used)` as its limit, so a chain cannot exceed the top-level budget in aggregate (`lee/state_machine/src/validated_state_diff/mod.rs`). The existing `from_public_transaction` becomes a wrapper that applies the default budget and discards the outcome.
- Add `ExecutionOutcome { cycles }` with an `ExecutionOutcome::FREE` constant (0 cycles) for transaction kinds that meter nothing, re-exported from the crate root.
- Add `LeeError::OutOfGas { budget }` (`lee/state_machine/src/error.rs`).
- No guest binary changes — everything here is host-side, so the committed Risc0 `artifacts/` are untouched.

**Key finding (documented in code and tests):** risc0's `SessionInfo::cycles()` reports **po2-padded** per-segment totals, while the session limit gates *actual* executed cycles. A boundary budget (reported cycles − 1) therefore does *not* reliably trip the limit — enforcement granularity is coarser than the reported number. The tests account for this by using decisively insufficient budgets (e.g. a quarter of the measured chain total, not `full_cycles - 1`). The consequence for the charging PR (deferred there): the charged amount must be clamped to `cycles.min(gas_limit)`, and a metered overshoot past the budget must never be treated as a consensus fault.

- [x] Change `Program::execute` to accept a cycle budget and return the metered cycle count
- [x] Add `execute_session` as the single `OutOfGas` recognition point (with `FIXME` on the string match)
- [x] Add `ValidatedStateDiff::from_public_transaction_with_budget` with chain-shared budget accounting and `ExecutionOutcome`
- [x] Add `DEFAULT_PUBLIC_CYCLE_BUDGET` (= previous hardcoded limit) and keep old signatures as wrappers — no behavior change

## 🧪 How to Test

```sh
RISC0_DEV_MODE=1 cargo nextest run -p lee --all-features
```

Covers the new tests:
- `program::tests`: cycles are reported and within budget; a tiny budget yields `OutOfGas { budget }`.
- `validated_state_diff::tests`: the budgeted entry point produces an identical diff to the unbudgeted one with a nonzero cycle outcome; an exhausted budget surfaces `OutOfGas`; a chain-calling transaction that fits per-call still exhausts a starved shared budget (`chained_calls_share_one_budget`, which also documents the po2-padding caveat); `ExecutionOutcome::FREE` is zero.

(`--all-features` matters generally in this workspace — `sequencer_core` tests are silently skipped without it, and CI runs with it.)

Then the CI mirror:

```sh
cargo +nightly fmt --check
taplo fmt --check
cargo machete
cargo deny check
```

## 🔗 Dependencies

- #750 (fee wire format) — this branch is stacked on `erhant/fees/fee-wire-format`.

## 🔜 Future Work

Deferred to #753/#754 by design:

- The settlement-shaped `(outcome, Result)` execution form and `MalformedProgramOutput` handling.
- Deriving cycle budgets from the transaction's `gas_limit`, and the `cycles.min(gas_limit)` clamp required by the po2-padding finding above.
- Migrating consumers (sequencer/executor paths) onto the budgeted entry points and actually charging fees.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests

### DIFF
--- a/lee/state_machine/src/error.rs
+++ b/lee/state_machine/src/error.rs
@@ -1,7 +1,7 @@
 use std::io;
 
 use lee_core::{
-    account::{Account, AccountId},
+    account::{Account, AccountId, Cycles},
     program::ProgramId,
 };
 use thiserror::Error;
@@ -21,7 +21,7 @@ pub enum LeeError {
     InvalidInput(String),
 
     #[error("Execution exceeded its cycle budget of {budget} cycles")]
-    OutOfGas { budget: u64 },
+    OutOfGas { budget: Cycles },
 
     #[error("Program violated execution rules")]
     InvalidProgramBehavior(#[from] InvalidProgramBehaviorError),
--- a/lee/state_machine/src/error.rs
+++ b/lee/state_machine/src/error.rs
@@ -20,6 +20,9 @@ pub enum LeeError {
     #[error("Invalid input: {0}")]
     InvalidInput(String),
 
+    #[error("Execution exceeded its cycle budget of {budget}")]
+    OutOfGas { budget: u64 },
+
     #[error("Program violated execution rules")]
     InvalidProgramBehavior(#[from] InvalidProgramBehaviorError),
 
--- a/lee/state_machine/src/error.rs
+++ b/lee/state_machine/src/error.rs
@@ -20,7 +20,7 @@ pub enum LeeError {
     #[error("Invalid input: {0}")]
     InvalidInput(String),
 
-    #[error("Execution exceeded its cycle budget of {budget}")]
+    #[error("Execution exceeded its cycle budget of {budget} cycles")]
     OutOfGas { budget: u64 },
 
     #[error("Program violated execution rules")]
--- a/lee/state_machine/src/program/mod.rs
+++ b/lee/state_machine/src/program/mod.rs
@@ -2,7 +2,7 @@ use std::borrow::Cow;
 
 use borsh::{BorshDeserialize, BorshSerialize};
 use lee_core::{
-    account::AccountWithMetadata,
+    account::{AccountWithMetadata, Cycles},
     from_frame,
     program::{InstructionData, ProgramId, ProgramInput, ProgramOutput},
     to_frame,
@@ -14,7 +14,7 @@ use crate::error::LeeError;
 /// The cycle budget applied to public execution paths that do not yet carry
 /// a transaction-specific budget. Enforcement of the fee spec's per-block cap
 /// replaces this in the charging transition.
-pub const DEFAULT_PUBLIC_CYCLE_BUDGET: u64 = 1024 * 1024 * 32; // 32M cycles
+pub const DEFAULT_PUBLIC_CYCLE_BUDGET: Cycles = 1024 * 1024 * 32; // 32M cycles
 
 #[derive(Clone, Debug, PartialEq, Eq, BorshSerialize, BorshDeserialize)]
 pub struct Program {
@@ -60,8 +60,8 @@ impl Program {
         caller_program_id: Option<ProgramId>,
         pre_states: &[AccountWithMetadata],
         instruction_data: &InstructionData,
-        cycle_budget: u64,
-    ) -> Result<(ProgramOutput, u64), LeeError> {
+        cycle_budget: Cycles,
+    ) -> Result<(ProgramOutput, Cycles), LeeError> {
         // Write inputs to the program
         let mut env_builder = ExecutorEnv::builder();
         env_builder.session_limit(Some(cycle_budget));
@@ -95,7 +95,7 @@ impl Program {
     fn execute_session(
         env: ExecutorEnv<'_>,
         elf: &[u8],
-        cycle_budget: u64,
+        cycle_budget: Cycles,
     ) -> Result<risc0_zkvm::SessionInfo, LeeError> {
         default_executor().execute(env, elf).map_err(|e| {
             // check for "Guest panicked" to prevent spoofing
--- a/lee/state_machine/src/program/mod.rs
+++ b/lee/state_machine/src/program/mod.rs
@@ -91,14 +91,17 @@ impl Program {
     /// typed [`LeeError::OutOfGas`]. The only place that error string is
     /// recognized.
     ///
-    /// FIXME: This is a brittle string match; the executor should provide a typed error
+    /// FIXME: This is a brittle string match; the executor should provide a typed error.
     fn execute_session(
         env: ExecutorEnv<'_>,
         elf: &[u8],
         cycle_budget: u64,
     ) -> Result<risc0_zkvm::SessionInfo, LeeError> {
         default_executor().execute(env, elf).map_err(|e| {
-            if format!("{e:#}").contains("Session limit exceeded") {
+            // check for "Guest panicked" to prevent spoofing
+            // via `panic!("Session limit exceeded")` cases
+            let message = format!("{e:#}");
+            if message.contains("Session limit exceeded") && !message.contains("Guest panicked") {
                 LeeError::OutOfGas {
                     budget: cycle_budget,
                 }
--- a/lee/state_machine/src/program/mod.rs
+++ b/lee/state_machine/src/program/mod.rs
@@ -11,9 +11,10 @@ use risc0_zkvm::{ExecutorEnv, ExecutorEnvBuilder, default_executor};
 
 use crate::error::LeeError;
 
-/// Maximum number of cycles for a public execution.
-/// TODO: Make this variable when fees are implemented.
-const MAX_NUM_CYCLES_PUBLIC_EXECUTION: u64 = 1024 * 1024 * 32; // 32M cycles
+/// The cycle budget applied to public execution paths that do not yet carry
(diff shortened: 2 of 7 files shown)


lee/state_machine/src/program/mod.rs
```diff
@@ -71,18 +73,40 @@ impl Program {
...
+            // via `panic!("Session limit exceeded")` cases
+            let message = format!("{e:#}");
+            if message.contains("Session limit exceeded") && !message.contains("Guest panicked") {
+                LeeError::OutOfGas {
+                    budget: cycle_budget,
+                }
+            } else {
+                LeeError::ProgramExecutionFailed(e.to_string())
+            }
+        })
```
schouhy, line 111: This string match looks very bad. Is it fixed in the next PRs in the stack?

lee/state_machine/src/validated_state_diff/mod.rs
```diff
@@ -53,13 +53,47 @@ impl ValidatedStateDiff {
...
+            timestamp,
+            crate::program::DEFAULT_PUBLIC_CYCLE_BUDGET,
+        )
+        .map(|(diff, _)| diff)
+    }
+
+    /// Validates and executes `tx` under `cycle_budget`, shared by every call
+    /// in the chain: each nested session is limited to the remaining budget, so
+    /// the chain cannot exceed the budget in aggregate.
+    pub fn from_public_transaction_with_budget(
```
schouhy:
```suggestion
    pub fn from_public_transaction_with_cycle_budget(
```

schouhy (approved):
LGTM. Left some comments

moudyellaz (approved):
lgtm, thanks!

lee/state_machine/src/program/mod.rs
erhant: It is ugly indeed, but nothing we can do about it I believe:

https://docs.rs/risc0-zkvm/latest/risc0_zkvm/enum.ExitCode.html#variant.SessionLimit

> NOTE: The current version of the RISC Zero zkVM will never exit with an exit code of SessionLimit. This is because the system cannot currently prove that the session limit as been reached.

If you look at the code of risc0 as well from where we call it, it has this:

```rs
                    CycleLimit::Hard(max_cycles) => {
                        if self.cycles.user >= max_cycles {
                            bail!(
                                "Session limit exceeded: {} >= {max_cycles}",
                                self.cycles.user
                            );
                        }
                    }
```

so it literally `anyhow::bail`s with that string.

Note that we need an extra check to rule out "Guest panicked" because the error message can be spoofed.

lee/state_machine/src/validated_state_diff/mod.rs
erhant: Done in https://github.com/logos-blockchain/logos-execution-zone/pull/752/commits/b30fe0503dbeac99e5b9fbf57e2cdd3048de6f42

lee/state_machine/src/program/mod.rs
```diff
@@ -57,10 +58,11 @@ impl Program {
         caller_program_id: Option<ProgramId>,
         pre_states: &[AccountWithMetadata],
         instruction_data: &InstructionData,
-    ) -> Result<ProgramOutput, LeeError> {
+        cycle_budget: u64,
```
jonesmarvin8: Since we have other u64's running around. Could we use an alias for this?

lee/state_machine/src/validated_state_diff/mod.rs
```diff
@@ -53,13 +53,47 @@ impl ValidatedStateDiff {
     }
 }
 
+/// The metered result of a public execution: the cycle count accumulated
+/// across every call in the chain.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct ExecutionOutcome {
+    pub cycles: u64,
```
jonesmarvin8: As mentioned earlier, this `u64` would be nice to use an alias; same as earlier's.

lee/state_machine/src/validated_state_diff/tests.rs
```diff
@@ -539,3 +539,113 @@ fn privacy_garbage_proof_is_rejected() {
...
+    let from = AccountId::from(&PublicKey::new_from_private_key(&from_key));
+    let to = AccountId::new([2_u8; 32]);
+    let state = V03State::new()
+        .with_public_accounts(public_state_from_balances(&[(from, 1_000), (to, 0)]))
+        .with_test_programs();
+    let instruction: (
+        u128,
+        lee_core::program::ProgramId,
+        u32,
+        Option<lee_core::program::PdaSeed>,
```
jonesmarvin8, line 602: If possible, the u128, u32 here in `instruction`, could we specify these with aliases? For readability. If not possible, that's fine.

jonesmarvin8 (approved):
Lgtm. Just minor comments about aliases...

lee/state_machine/src/validated_state_diff/mod.rs
```diff
@@ -120,11 +155,15 @@ impl ValidatedStateDiff {
...
-            let mut program_output = program.execute(
+            let (mut program_output, call_cycles) = program.execute(
                 caller_data.program_id,
                 &chained_call.pre_states,
                 &chained_call.instruction_data,
+                cycle_budget.saturating_sub(cycles_used),
             )?;
+            cycles_used = cycles_used
+                .checked_add(call_cycles)
+                .expect("cycle sums fit u64: overflow would need ~2^64 executed cycles");
```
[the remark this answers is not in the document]
erhant: Done in https://github.com/logos-blockchain/logos-execution-zone/pull/752/commits/8ad1ae61ad50d97456bf48af239c83c1f914a127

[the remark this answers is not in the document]
erhant: Using saturating add because its cant fail at all

lee/state_machine/src/error.rs
```diff
@@ -20,6 +20,9 @@ pub enum LeeError {
     #[error("Invalid input: {0}")]
     InvalidInput(String),
 
+    #[error("Execution exceeded its cycle budget of {budget}")]
+    OutOfGas { budget: u64 },
```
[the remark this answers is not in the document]
erhant: Done in https://github.com/logos-blockchain/logos-execution-zone/pull/752/commits/198e5d43587dedfda0621cad8510b4ee0456e751

lee/state_machine/src/validated_state_diff/mod.rs
erhant: Added in https://github.com/logos-blockchain/logos-execution-zone/pull/752/commits/852eb04b9d9914b1da8c65561aec6cd7efc2fb3f

lee/state_machine/src/program/mod.rs
erhant: Added in https://github.com/logos-blockchain/logos-execution-zone/pull/752/commits/852eb04b9d9914b1da8c65561aec6cd7efc2fb3f

lee/state_machine/src/validated_state_diff/tests.rs
erhant: This is just a tiny test `instruction` that doesn't even need types really, except due to `Option` being `None` here

So I would rather keep it

moudyellaz (approved):
lgtm, thanks!

erhant:
Closing in favor of #801

PR #753, opened 2026-08-17
feat(fees): fee arithmetic and producer identity

erhant:
## 🎯 Purpose

Fifth PR of the fee-subsystem train (#746 → #747 → #750 → #752 → **#753** → #754 → #758 → #762). It lands the two remaining prerequisites for actually charging fees, while itself staying **inert with respect to money movement**:

1. **Fee arithmetic** — the pure reserve/actual-fee formulas and static validity rules from the spec, as library code in `fee_core` with the spec's worked examples pinned as golden-vector tests.
2. **Producer identity** (spec Q4) — blocks now declare *who produced them*: the `BlockHeader` carries the producing sequencer's public key, covered by the block hash and enforced by a new consensus rule. Fee payouts in #754 will credit this key's account.

This branch also absorbed its former parent (the small `money-flow` branch), hence the two-topic name.

## ⚙️ Approach

**Fee arithmetic** (`lez/programs/fee/core/src/assess.rs`, `validity.rs`):

- `FeeTxView` — the fee-relevant view of a transaction. `Public` carries `payer`, `gas_limit`, `data_bytes`, `tip`, `max_fee`; `Private` carries only `payer`, because every private tx prices identically: gas_exec is the fixed STARK-verification cost `PRIVATE_VERIFY_GAS` and gas_stor is the canonical padded wire size `PRIVATE_GAS_STOR` (so private reserve == private actual, regardless of reported cycles — asserted in tests).
- `fee_reserve` = `gas_limit·base_fee_exec + gas_stor·base_fee_stor + tip` — the amount held from the payer before execution, at the block's opening fee state.
- `fee_actual_base` = `gas_exec·base_fee_exec + gas_stor·base_fee_stor` — the fee actually owed, where `gas_exec` is the charged (clamped) cycle count from #752 for public txs.
- `validate_static_tx` — pre-execution checks: `data_bytes` in `1..=MAX_GAS_STOR`, `gas_limit ≤ MAX_GAS_EXEC`, `max_fee ≥ fee_reserve` (boundary is exact: `max_fee == reserve` passes, `reserve − 1` fails). Private views are statically valid here — their wire size/padding is enforced by wire-format and proof rules, not fee code.
- `accumulate_exec_gas` / `accumulate_stor_gas` — block gas-cap accumulators, widened to u128 and checked against the caps (never wrapped); reaching the cap exactly is allowed.
- All products are widened to u128; the const caps from `market.rs` (`MAX_GAS·BASE_FEE_MAX` fits u64 per resource) keep sums far from overflow — exercised by a test at the extreme caps.
- **Golden vectors from the spec**: at genesis fees (both base fees = 8), a 50,000-cycle / 200-byte public tx prices at **401,600**; every private tx prices at **5,070,616** (= 409,764·8 + 224,063·8). The private constants (`PRIVATE_VERIFY_GAS = 409,764`, `PRIVATE_GAS_STOR = 224,063 = PROOF_BYTES 223,551 + PRIVATE_PAD_BYTES 512`) are **provisional** and FIXME-marked pending the final LEZ wire-format numbers.

**Producer identity** (`lez/common/src/block.rs`, `lez/chain_state/src/apply.rs`):

- `BlockHeader` gains `producer: lee::PublicKey` — the block-producing sequencer's signing key.
- It is hash-covered by making it a parameter: `compute_hash(&producer)` hashes `SHA256(PREFIX || borsh(data) || borsh(producer))`. Passing it as a parameter (rather than a `HashableBlockData` field) was deliberate: the ~25 existing `HashableBlockData` literals across the codebase stay untouched, since every hash site funnels through `into_pending_block` / `recompute_hash`, and no site can compute a hash without deciding the producer.
- `into_pending_block` derives the producer from the signing key, so producing a block cannot declare a key other than the one that signs.
- **One new consensus rule**: `BlockIngestError::InvalidProducerSignature` — `validate_against_tip` now requires the header signature to verify against the *declared* producer (`has_valid_producer_signature`). Test `producer_signature_must_verify` forges the producer and re-hashes, so the hash check passes but the signature check rejects.
- The field is carried through the indexer protocol (`indexer_service_protocol::BlockHeader` + error variant + `PublicKey` `Display`), the indexer FFI (`FfiBlockHeader.producer`, regenerated `indexer_ffi.h`), the mock service, and the explorer (Producer row on the block page).

Changes:

- [x] Add `fee_core::assess`: `FeeTxView`, `fee_reserve`, `fee_actual_base`, with spec golden vectors pinned in tests
- [x] Add `fee_core::validity`: `validate_static_tx`, checked block gas-cap accumulators, `FeeError`
- [x] Add provisional private-pricing constants to `fee_core::market` (FIXME-marked)
- [x] Add `producer` to `BlockHeader`, hash-covered via `compute_hash(&producer)`
- [x] Add the `InvalidProducerSignature` consensus rule in `chain_state::apply`
- [x] Carry `producer` through indexer protocol, FFI, mock service, and explorer UI
- [x] Regenerate the committed sequencer-db test fixture (header format changed)

## 🧪 How to Test

```sh
```

Review focus for the new tests:

- `fee_core::assess` / `fee_core::validity` — the golden vectors (`spec_worked_example_public` = 401,600, `spec_worked_example_private` = 5,070,616), the exact `max_fee` boundary, and the cap accumulators (exact at the cap, checked beyond it, no wrap at `u64::MAX`).
- `chain_state` — `producer_signature_must_verify` (forged producer with a recomputed hash is rejected).

Standard lints: `cargo +nightly fmt`, `cargo clippy --workspace --all-targets --all-features -- -D warnings`, `taplo fmt`, `cargo machete`, `cargo deny check`.

## 🔗 Dependencies

- #752 (cycle-count metering) — base of this branch; supplies the charged-cycle counts that `fee_actual_base` prices.

## 🔜 Future Work

- #754 puts this arithmetic in motion: reserve/settle in the guest, revert-keeps-fee semantics, actual money movement to escrow and the producer's account.
- #758 adds fee-based admission on top; #762 converts the integration suites to the fee world and closes the train.
- Re-pin the provisional private constants (`PRIVATE_VERIFY_GAS`, `PROOF_BYTES`, `PRIVATE_PAD_BYTES`, `PRIVATE_GAS_STOR`) once the final LEZ wire-format numbers land (FIXME in `market.rs`).
- Consider caching the producer key instead of re-deriving it from the signing key per block (TODO in `into_pending_block`); derive `MAX = 2·TARGET` instead of asserting it (FIXME in `market.rs`).

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


lez/common/src/block.rs
```diff
@@ -76,7 +79,21 @@ impl Block {
...
+
+    /// Whether the header signature verifies against the embedded producer
+    /// key. Every valid block must satisfy this.
+    ///
+    /// This attests only that the producer signed the *declared* `header.hash`,
+    /// not that the hash matches the block contents — so it is not an
+    /// authenticity check on its own. Pair it with a `recompute_hash` check (as
+    /// `validate_against_tip` does) before trusting it.
+    #[must_use]
+    pub fn has_valid_producer_signature(&self) -> bool {
```
moudyellaz, line 105: the signature is checked against the producer key carried in the same header, both supplied by whoever built the block. With #754 this field decides who receives the block's tips and payout, so what binds it to an authorized local sequencer? Peer zones pin the key via `is_signed_by`, the local chain has no equivalent.

moudyellaz (approved):
lgtm, thanks! Left one comment.

lez/common/src/block.rs
erhant: indeed, on this PR the producer key is self-asserted and nothing depends on which key it is yet; `has_valid_producer_signature` only proves the block is internally consistent (signature matches the embedded key), but here #753 just introduces the field.

This actually is a problem in #754 as you have also [commented](https://github.com/logos-blockchain/logos-execution-zone/pull/754#discussion_r3826593245), where `header.producer` becomes the tips/payout recipient.

I will fix this together with the payout: alongside the new _claim_ route for crediting the producer, the sequencer will verify `header.producer` against its own configured signing key (the local analogue of a peer zone's pinned `is_signed_by` key) before paying out, rather than trusting the self-declared producer.

So, this will be fixed within #754 so it ships with the code that relies on it as well; would make reasoning about it much easier I think.

lez/common/src/block.rs
```diff
@@ -111,34 +128,42 @@ pub struct HashableBlockData {
 }
 
 impl HashableBlockData {
-    /// Domain-separated hash of the block contents: `SHA256(PREFIX || borsh(self))`.
-    /// The single source of truth for both producing and verifying a block hash.
+    /// Domain-separated hash of the block contents and its producer:
+    /// `SHA256(PREFIX || borsh(self) || borsh(producer))`. The single source of
+    /// truth for both producing and verifying a block hash; no site can hash
```
jonesmarvin8, line 145: site?

lez/programs/fee/core/src/assess.rs
```diff
@@ -0,0 +1,178 @@
...
+/// The fee-relevant view of a transaction (spec *Transactions*).
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum FeeTxView {
+    Public {
+        payer: AccountId,
+        gas_limit: u64,
+        data_bytes: u64,
+        tip: u64,
+        max_fee: u128,
+    },
```
jonesmarvin8, line 22: We should use aliases introduced in previous PRs.

Though, I'm confused. fees in previous PRs were u64. But here it's u128 (like `Balance`). Is this intentional?

lez/programs/fee/core/src/assess.rs
```diff
@@ -0,0 +1,178 @@
...
+/// The amount held from the payer before execution, at the block's opening
+/// base fees: `gas_limit·base_fee_exec + gas_stor·base_fee_stor + tip`.
+///
+/// The caller must pass a cap-validated view (gas within `MAX_GAS_*`) and a
+/// well-formed `FeeState` whose base fees stay in `[MIN, BASE_FEE_*_MAX]` — the
+/// bounds `next_base_fee` maintains. Under those preconditions every product
+/// fits u64 and the u128 sum cannot overflow; on an unchecked view against a
+/// corrupt state it could wrap silently in release.
+#[must_use]
+pub fn fee_reserve(view: &FeeTxView, fee_state: &FeeState) -> u128 {
```
jonesmarvin8: Should this output `Balance` or `Fee`?

jonesmarvin8: This is a bit unclear to me from the specs.

jonesmarvin8 (commented):
There's some hardcoded `u128`'s that I'm unsure whether these should be `Fee`(u64) `Balance` (u128). This is my main concern.

schouhy (changes requested):
hey thanks! 

I think we can remove the producer identity machinery
the block signature and the producer field are redundant because writing to Bedrock already requires being in the channel writers, and each inscription comes signed by the writer. So the LEZ header signature just repeats what the mantle tx already provides 

the payout doesn't need identity either. We can make the 4th account of the fee tx a free beneficiary. Tips come from the block's own txs and the smoothed payout comes from FeeState, so validators can still check how much without checking who was the recipient. The byte compare of the fee tx could just skip the 4th account id. I mean if a producer sends rewards to a burn address, that's not something the protocol needs to prevent I think 

I mean we can delete the header signature, the producer field (and has_valid_producer_signature)

lez/programs/fee/core/src/assess.rs
erhant: Yes, spec tells us that fees are `u64` but computations over them are `u128` (..yeah i know 😢)

> Base fees and gas are u64; fee totals, balances, and every intermediate product are u128

lez/programs/fee/core/src/assess.rs
erhant: > Base fees and gas are u64; fee totals, balances, and every intermediate product are u128

So this is a computation over fees, meaning that it is `u128`. Using an alias here is actually bad because its neither a fee nor a balance, its an intermediate result; but I will still go with using Balance because a computed fee is something that the user should be able to cover with their Balance.

lez/programs/fee/core/src/assess.rs
jonesmarvin8: I'll let this go for now. But can we make this an issue so that we can check with TKE? It doesn't change logic either way...

lez/programs/fee/core/src/assess.rs
jonesmarvin8: Hurts me a little bit, but okay.

lez/programs/fee/core/src/market.rs
```diff
@@ -31,8 +31,20 @@ pub const BASE_FEE_STOR_MAX: Fee = u64::MAX / MAX_GAS_STOR;
...
+// FIXME: Provisional: re-pin with the LEZ wire-format numbers (spec Parameters TODO).
+/// Execution gas charged to every private transaction (STARK receipt
+/// verification, RISC Zero 3.0.5).
+pub const PRIVATE_VERIFY_GAS: u64 = 409_764;
+/// Proof bytes inside every private transaction.
+pub const PROOF_BYTES: u64 = 223_551;
+/// Payload size every private transaction is padded to.
+pub const PRIVATE_PAD_BYTES: u64 = 512;
+/// Canonical serialized size of every private transaction.
+pub const PRIVATE_GAS_STOR: u64 = 224_063;
```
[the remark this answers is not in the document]
erhant: Done in https://github.com/logos-blockchain/logos-execution-zone/pull/753/commits/3c1e33740f453df8b125545dd95a40842fc799f3

lez/chain_state/src/apply.rs
```diff
@@ -260,6 +263,21 @@ mod tests {
...
+        let mut state = initial_state();
+        let genesis = produce_dummy_block(1, None, vec![]);
+        // Forge the producer: replace it with a different key and re-hash so
+        // the hash check passes but the signature no longer verifies.
+        let mut forged = genesis;
+        forged.header.producer = lee::PublicKey::new_from_private_key(
+            &lee::PrivateKey::try_new([9_u8; 32]).expect("valid key"),
+        );
+        forged.header.hash = forged.recompute_hash();
+        let err = apply_block(None, &forged, &mut state).expect_err("should reject");
```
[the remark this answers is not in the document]
erhant: This will be affected by the new producer logic (see schouhy's review)

lez/programs/fee/core/src/assess.rs
erhant: created at https://github.com/logos-blockchain/logos-execution-zone/issues/794

erhant:
Closing in favor of #801

PR #754, opened 2026-08-17
feat(fees): wire all the fee-charging logic

erhant:
## 🎯 Purpose

Sixth PR of the fees train (#746 → #747 → #750 → #752 → #753 → **#754** → #758 → #762), and its centerpiece: everything before this built the machinery (system accounts, wire format, metering, arithmetic, forced fee transaction) — **this PR is where the money actually moves**. Charged transactions now reserve, execute under their gas limit, and settle a real fee; the block's revenue flows through the fee inbox into escrow and the producer; and the forced fee transaction carries the real byte-derived `BlockFeeSummary` instead of the all-zero placeholder.

**Recommend reviewing commit-by-commit.** The first commit (`feat(fees): implement settlement and fee movements`, ~910 lines net) is the consensus core; the second (`test(fees): fee-aware block building and assertions across the suites`, ~420 lines) is the mechanical conversion of the test world to fee-aware block building. The 17 rebuilt guest binaries under `artifacts/` and the regenerated `prebuilt_sequencer_db.dump` fixture are expected churn.

### Classification (`lez/chain_state/src/classify.rs`, new)

Every user transaction is classified at its turn into `FeeClass::{Exempt, Charged(FeeTxView)}`. Charged transactions get a byte-derived `FeeTxView` (borsh size → `data_bytes`, plus the message's payer/gas_limit/tip/max_fee). Interim exemptions, each documented in the module:

- **Private transactions and program deployments** — pending the private-payer (Q3) and deployment replay-protection decisions; excluded from metering so free traffic cannot move the public base fee.
- **System injections** — identified by shape (empty witness set invoking bridge `Deposit` or cross-zone inbox `Dispatch`); the recognizer `is_system_injection` lives in `lez/common/src/transaction.rs`, shared by builder and transition so they can never disagree.
- **Genesis transactions** — config/supply setup.
- **Full-sweep vault claims** (`is_full_vault_sweep`) — the bootstrap case: all funding lands in vaults while fees debit account balances, so a charged first claim could never pay. Marked `FIXME` for removal with Vault.

### The charged settlement cycle (`lez/chain_state/src/apply.rs`)

`settle_charged_transaction` is the reserve → execute → settle cycle, and it exists **exactly once**: the validator's `apply_block_to_state`, the sequencer's block builder, and the block-building test helpers all call the same `settle_transaction`; storage's replay (`lez/storage/src/indexer/mod.rs`) now delegates wholesale to `chain_state::apply::apply_block_to_state`, deleting its parallel hand-rolled replay. The steps, in order:

1. `validate_static_tx` against the block's opening `FeeState` (affordability at gas_limit, max_fee, per-tx caps — from #753).
2. **Must have a signer** — a fee-witness-only transaction would burn no nonce and be replayable forever.
3. The designated payer's authorization must be present (`is_fee_authorized`).
4. **Authenticate signers** (signatures + replay nonces, via the now-`pub` `authenticate_public_transaction_signers`) *before any money moves*, so a revert-with-fee can only ever charge an authorized transaction.
5. Debit the payer the **full reserve** (`fee_reserve`, priced at gas_limit).
6. Metered execution with `gas_limit` as the cycle budget, via the new settlement-shaped `ValidatedStateDiff::from_public_transaction_metered` which returns `(ExecutionOutcome, Result<…>)` — the metered outcome survives a failed execution; out-of-gas is metered at the whole budget.
7. Clamp `charged = cycles.min(gas_limit)` — risc0's po2-padded reporting can overshoot the session limit (the #752 finding), so overshoot is clamped, never a consensus fault.
8. Checked accumulation into the block's exec/stor gas totals; breaching a cap is the new `BlockIngestError::GasCapExceeded`.
9. Apply the diff — or **on execution failure, revert**: discard the effects but keep the fee and advance the signers' nonces (`V03State::advance_nonces`).
10. Release `reserve − actual` back to the payer; credit `fee_base + tip` to the fee inbox; accumulate `revenue_base`/`revenue_tip` into the summary.

Any *settlement* failure (steps 1–5, 8) invalidates the whole block on the apply path, and drops the transaction on the builder path.

> ⚠️ A charged transaction whose *execution* fails is now an **included revert** — fee kept, nonces burned, effects discarded. Previously any execution failure invalidated the block (validator) or dropped the transaction (builder). This is what makes fees DoS-meaningful: you pay for the cycles you consume whether or not your transaction succeeds. We can think of preserving error message etc. too (perhaps an Event-like logic).

### The forced fee transaction carries the real summary

- `fee_invocation(summary, producer)` now takes the settled `BlockFeeSummary` and the producer account (derived from the block header's producer key), which rides as the **4th account** so the guest can pay it (`lez/common/src/transaction.rs`).
- The validator byte-compares the fee transaction **after** settling the user transactions — its summary is derived from their execution. `derive_block_summary` (settlement on a scratch clone) is the builder/test helper that produces it.

> ⚠️ The producer account is **auto-initialized** (`initialize_producer_account`: assign `authenticated_transfer` ownership iff still default-owned) immediately before the fee transaction, as a consensus rule, at the same point in the builder and in `apply_block_to_state`. Program output may not modify an unclaimed default account, so without this the fee guest could not credit a never-seen producer. It is the same initialization genesis applies to user accounts.

### The fee guest moves money (`lez/programs/fee/src/main.rs`)

Now takes 4 accounts (state, escrow, inbox, producer) and, on top of #747's market update:

- **Conservation pin:** asserts the inbox balance equals the block's `revenue_base + revenue_tip` byte-for-byte with the summary — the transition credited exactly the block's revenue, anything else is an invalid block.
- **Drains the inbox to zero:** base revenue → escrow, tips → producer.
- **Pays the smoothed payout** (from `FeeState::apply_block`) from escrow to the producer. The `payout == 0` placeholder assert from #747 is gone.

### Sequencer builder (`lez/sequencer/core/src/lib.rs`)

- User/gossip transactions go through the same `settle_transaction` on a **scratch clone**: a transaction that cannot pay, breaches a cap, or fails settlement is dropped, never included; only success commits the scratch state and summary.
- The old `validate_on_state` guard pass (restricted-system-account rules) is kept *in front of* settlement — see Future Work for the double-execution FIXME.
- The in-loop size probe uses a default-summary placeholder fee tx, which sizes identically since `BlockFeeSummary` is fixed-size; the real `fee_invocation(summary, producer)` is built after the loop.
- Genesis and block production both initialize the producer account at the exact point the validator replay does.

### Supporting changes

- `lee/state_machine`: transition-side ledger primitives `V03State::{credit_balance, debit_balance, initialize_account_owner, advance_nonces}`, the new `LeeError::InsufficientBalance`, and the `from_public_transaction_metered` split described above.
- New `BlockIngestError::{InvalidFeeClass, GasCapExceeded}` variants, mirrored through the indexer protocol (`lez/indexer/service/protocol`).
- **Testnet initial state rescaled to LGO scale** (10^13-class balances, `lez/testnet_initial_state/src/lib.rs`): once transfers pay real fees, the old toy 10 000/20 000 balances could not afford a single reservation.
- Test suites (chain_state, storage, sequencer, indexer block_store) converted to fee-aware `settled_block`-style helpers built on `derive_block_summary`, with fee-aware assertions: recipient balances exact, sender balances relative (`< initial − transferred`), and full conservation checks — `fees_paid == escrow + producer`, inbox drains to zero every block.

- [x] Add the fee classifier with the interim exemption policy (`chain_state::classify`)
- [x] Implement the single shared reserve → execute → settle cycle (`settle_charged_transaction`) and wire it into validator apply, sequencer builder, and storage replay
- [x] Change execution failure of a charged transaction to an included revert-with-fee that burns nonces
- [x] Add consensus-rule producer-account auto-initialization before the fee transaction
- [x] Change the forced fee transaction to carry the real byte-derived summary + producer as 4th account, byte-compared after settlement
- [x] Change the fee guest to drain the inbox (base → escrow, tips → producer), pin conservation, and pay the smoothed payout
- [x] Rescale testnet balances to LGO scale and convert all suites to fee-aware building/assertions
- [x] Rebuild guest artifacts and regenerate the prebuilt sequencer DB fixture

## 🧪 How to Test

```sh
RISC0_DEV_MODE=1 cargo nextest run --workspace --all-features
```

The heart of it is the **conservation assertions** across the converted suites:

- `chain_state`: `apply::tests` walks 10 charged transfers and asserts the recipient gained exactly the transfers, the sender paid a nonzero fee, the inbox drained every block, and `fees_paid == escrow + producer` with a started smoothed payout; `chain.rs` re-runs the reorg/finality matrix on fee-settled blocks. `lee` covers the metered-failure/ledger-helper/nonce primitives.
- `storage::indexer::tests`: replay-through-`chain_state` breakpoint/batch tests with fee-aware balances.
- `sequencer_core` (needs `--all-features` — silently skipped without it; CI runs with it): block-tail shape via `assert_block_tail`, follow/orphan/restart flows on settled peer blocks.
- `indexer_core::block_store`: accept/replay/breakpoint tests on settled blocks, plus the retryable-failure test reworked around a system-shaped bogus deposit (a charged overdraft no longer produces a block-level failure — it reverts-with-fee inside a valid block).

Then the CI mirror:

```sh
cargo +nightly fmt --check
taplo fmt --check
cargo machete
cargo deny check
```

## 🔗 Dependencies

- #753 (fee arithmetic + forced fee transaction) — this branch is stacked on `erhant/fees/fee-arithmetic-and-producer`.

## 🔜 Future Work

- **#758** adds the admission/quote RPC surface (fee estimation for wallets, mempool admission against the fee state); **#762** converts the integration suites to the fee world and closes the train.
- **Builder guard-pass double execution** (`FIXME` in `apply_mempool_transaction`): user/gossip transactions currently execute once for the old restricted-system-account guard and once for metered settlement. Folding the guards into the settlement halves the per-tx cost — but note the current split also shapes semantics: a failure the guard catches is **dropped uncharged**, while only failures inside metered settlement become revert-with-fee. Unifying them means deciding which failures deserve which treatment.
- `is_full_vault_sweep` is scaffolding to be deleted with the Vault program.
- Private-payer and deployment replay-protection decisions will shrink the exemption list in `classify`.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


lee/state_machine/src/state/mod.rs
```diff
@@ -199,6 +199,59 @@ impl V03State {
...
+
+    /// Assigns `program_owner` to the account if it is still default-owned.
+    /// Used to initialize the producer account before the fee program credits
+    /// it: program output may not modify an unclaimed default account.
+    pub fn initialize_account_owner(&mut self, account_id: AccountId, program_owner: ProgramId) {
+        let account = self.get_account_by_id_mut(account_id);
+        if account.program_owner == lee_core::program::DEFAULT_PROGRAM_ID {
+            account.program_owner = program_owner;
+        }
+    }
```
schouhy, line 215: Oh this is not good IMO. This breaks a core property of the lee state: it should only mutate through valid transactions. With this method the lee state stops being a function of the genesis state and the list of transactions applied. I think this should be refactored to put this in transaction so that we can keep the determinism of the lee state as a function of the applied transactions.

lez/sequencer/core/src/lib.rs
```diff
@@ -991,10 +1043,12 @@ impl<BP: BlockPublisherTrait> SequencerCore<BP> {
             }
         }
 
+        let fee_tx = fee_invocation(summary, producer_account);
+        common::transaction::initialize_producer_account(&mut working_state, producer_account);
```
schouhy, line 1300: This mutates the state without a transaction

schouhy (changes requested):
Thanks for the PR. I'm worried about the new method to mutate the LEE state outside transaction executions.

lez/sequencer/core/src/lib.rs
```diff
@@ -682,45 +684,76 @@ impl<BP: BlockPublisherTrait> SequencerCore<BP> {
...
-                        } else {
-                            error!(
-                                "Transaction with hash {tx_hash} failed execution check with error: {err:#?}, skipping it",
-                            );
+                // Guard pass: the restricted-system-account rules, validated
+                // exactly as before charging existed.
+                // FIXME: this executes the transaction once for the guard and
+                // once for the metered settlement below; fold the guards into
+                // the settlement to halve the per-tx execution cost.
+                if let Err(err) = tx.validate_on_state(state, block_height, timestamp) {
```
moudyellaz, line 912: Guard pass runs at the 32M default budget regardless of `gas_limit`, and `validate_static_tx` has no floor on `gas_limit`, so `gas_limit = 0` buys 32M cycles for the storage fee alone and adds nothing to `gas_used_exec`. Should the guard pass run under the transaction's own `gas_limit`?

lez/common/src/transaction.rs
```diff
@@ -245,15 +236,106 @@ pub fn clock_invocation(timestamp: clock_core::Instruction) -> lee::PublicTransa
     )
 }
 
+/// Whether `tx` is a sequencer-injected system transaction.
+///
+/// Identified by shape: an empty witness set invoking the bridge deposit or
+/// the cross-zone inbox dispatch. Fee- and cap-exempt. Shared by the sequencer
+/// (build) and the transition (replay) so the two can never disagree.
+#[must_use]
+pub fn is_system_injection(tx: &LeeTransaction) -> bool {
```
moudyellaz, line 245: Anything the builder injects that this does not recognise classifies as `Charged`, fails `max_fee >= reserve` on `FeeFields::ZERO`, and invalidates a block the sequencer itself produced. maybe a test pinning the two lists together?

lez/chain_state/src/apply.rs
```diff
@@ -193,6 +231,194 @@ pub fn apply_block_to_state(block: &Block, state: &mut V03State) -> Result<(), B
...
+        ));
+    }
+    // Authenticate (signatures + replay nonces) before any money moves, so a
+    // revert-with-fee can only ever charge an authorized transaction.
+    let signers = lee::authenticate_public_transaction_signers(public_tx, state)
+        .map_err(|err| fee_validity(err.to_string()))?;
+
+    let reserve = fee_reserve(view, opening);
+    state
+        .debit_balance(view.payer(), reserve)
```
moudyellaz: An unaffordable reserve kills the block; an execution failure one step later only reverts. Intentional that these sit on opposite sides of the line?

lee/state_machine/src/state/mod.rs
moudyellaz: hmm i also agree on this .. the guest cannot absorb this -> empty witness set rules out `Claim::Authorized`, the producer account is not a fee-program PDA so `Claim::Pda` is out, and a still-default account cannot be modified at all. I think here we need a new claim route.

moudyellaz (commented):
hey thanks for the PR. I left some comments, would like to re-review later, so please once it is fixed, ask for a review again!

lez/chain_state/src/classify.rs
```diff
@@ -0,0 +1,69 @@
...
+        return Ok(FeeClass::Exempt);
+    }
+
+    // A public transaction that survives the shape-based exemptions above is
+    // charged iff it carries a fee declaration. A missing declaration is exempt
+    // by construction (nothing to charge); admission rejects user-submitted
+    // exempt transactions, only system-shaped ones legitimately omit the fee.
+    let Some(fee) = public_tx.message().fee else {
+        return Ok(FeeClass::Exempt);
+    };
```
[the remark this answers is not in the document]
erhant: Yes I had changed this right after I clicked on review, see https://github.com/logos-blockchain/logos-execution-zone/pull/754/commits/4d1b345e5bd80bc84fbb3b72cf0df7a2da58414f

lez/sequencer/core/src/lib.rs
erhant: I'm taking a look at this still*

schouhy (commented):
hey thanks! 

I think we can avoid exposing the new raw state mutator `advance_nonces`. There's a nice invariant I'd like to keep: the only way to mutate `V03State` is `apply_state_diff`, and there are few constructors of those diffs.

Do you think `from_public_transaction_metered` could return the final diff of the tx in both cases? On success the effects plus the signer nonces . And on failure a diff with only the signer nonce bumps. Then `chain_state` applies whatever comes back and maybe `advance_nonces` doesn't need to exist? 

what do you think?

Also I think `initialize_account_owner` can be removed as proposed [here](https://github.com/logos-blockchain/logos-execution-zone/pull/753#pullrequestreview-5025508858)

erhant:
> hey thanks!
> 
> I think we can avoid exposing the new raw state mutator `advance_nonces`. There's a nice invariant I'd like to keep: the only way to mutate `V03State` is `apply_state_diff`, and there are few constructors of those diffs.
> 
> Do you think `from_public_transaction_metered` could return the final diff of the tx in both cases? On success the effects plus the signer nonces . And on failure a diff with only the signer nonce bumps. Then `chain_state` applies whatever comes back and maybe `advance_nonces` doesn't need to exist?
> 
> what do you think?
> 
> Also I think `initialize_account_owner` can be removed as proposed [here](https://github.com/logos-blockchain/logos-execution-zone/pull/753#pullrequestreview-5025508858)

Hmm I like the idea of burying the revert case as a succesful return too, makes it cleaner really.
I'm currently rebasing from 746 up and beyond, I will handle the revert case like that 🙏🏻

lee/state_machine/src/validated_state_diff/mod.rs
```diff
@@ -94,28 +94,147 @@ impl ValidatedStateDiff {
...
+            account_ids,
+            instruction_data,
+            authorized,
+            Vec::new(), // no nonces to advance!
+            state,
+            block_id,
+            timestamp,
+            crate::program::DEFAULT_PUBLIC_CYCLE_BUDGET,
+            &mut 0, // dont care
+        )
```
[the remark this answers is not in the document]
erhant: Fixed in https://github.com/logos-blockchain/logos-execution-zone/pull/754/commits/351b6acaaf9264406bdfbadd833095f4c43d909e

lez/chain_state/src/apply.rs
```diff
@@ -193,6 +233,225 @@ pub fn apply_block_to_state(block: &Block, state: &mut V03State) -> Result<(), B
...
+    if !lee::is_fee_authorized(public_tx.message(), public_tx.witness_set()) {
+        return Err(fee_validity(
+            "designated payer's authorization is missing".into(),
+        ));
+    }
+    // Capture the signer set up front so a reverted action can still burn the
+    // replay nonces.
+    let signers = lee::authenticate_public_transaction_signers(public_tx, state)
+        .map_err(|err| fee_validity(err.to_string()))?;
+
```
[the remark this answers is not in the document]
erhant: This is no longer possible, as we do require at least one signer (previously we had `fee_witness` stuff, now no more)

lez/chain_state/src/apply.rs
erhant: An unaffordable tx should not have been included anyways, so thats on sequencer

this is more well-handled now that we don't have `.debit_balance` (or such raw mutations)

erhant:
I had a plan about producer account initialization, where the sequencer was supposed to auto-inject an initialization tx on the condition that it was not initialized at all; but spoke with @schouhy and with the new staked join/leave stuff and some changes from cross-zones, we can remove the producer logic here

lee/state_machine/src/state/mod.rs
erhant: https://github.com/logos-blockchain/logos-execution-zone/pull/754#issuecomment-5423027616

lez/sequencer/core/src/lib.rs
erhant: https://github.com/logos-blockchain/logos-execution-zone/pull/754#issuecomment-5423027616

lez/sequencer/core/src/lib.rs
erhant: Opened an issue on this: https://github.com/logos-blockchain/logos-execution-zone/issues/796

erhant: It requires me to do a bit of core additions with helper functions, I would prefer to do it in a different context.

lez/chain_state/src/apply.rs
```diff
@@ -193,6 +233,225 @@ pub fn apply_block_to_state(block: &Block, state: &mut V03State) -> Result<(), B
...
+                })?;
+            // The builder guards user transactions off the restricted accounts;
+            // the apply path must too, or a block author could drain them.
+            validate_no_restricted_account_modification(state, &diff).map_err(|err| {
+                BlockIngestError::RestrictedAccountModification {
+                    tx_index,
+                    reason: err.to_string(),
+                }
+            })?;
+            state.apply_state_diff(diff);
```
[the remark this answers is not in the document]
erhant: https://github.com/logos-blockchain/logos-execution-zone/issues/796 refers to this too

lez/sequencer/core/src/lib.rs
erhant: Fixed in https://github.com/logos-blockchain/logos-execution-zone/pull/801/commits/13b6841ebabbe2acb975ce6cdfe71a6e0763e0fa on #801

lez/chain_state/src/apply.rs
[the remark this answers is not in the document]
erhant: Fixed in https://github.com/logos-blockchain/logos-execution-zone/pull/801/commits/13b6841ebabbe2acb975ce6cdfe71a6e0763e0fa of #801

lee/state_machine/src/validated_state_diff/mod.rs
```diff
@@ -95,28 +95,180 @@ impl ValidatedStateDiff {
...
+    pub fn from_fee_settlement_invocation(
+        program_id: ProgramId,
+        account_ids: &[AccountId],
+        instruction_data: &[u8],
+        authorized: &HashSet<AccountId>,
+        state: &V03State,
+        block_id: BlockId,
+        timestamp: Timestamp,
+    ) -> Result<Self, LeeError> {
+        let mut cycles_used = 0; // dont care
```
moudyellaz, line 193: reserve and refund are full guest runs at 32M with cycles discarded, so a charged tx runs four executions and bills one. could those two diffs be built arithmetically inside `lee` instead, keeping the `apply_state_diff`-only invariant without the guest runs?

erhant:
Closing in favor of #801

PR #758, opened 2026-08-17
feat(fees): admission screen, getFeeState RPC, and mempool cap budgeting

erhant:
# feat(fees): admission screen, getFeeState RPC, and mempool cap budgeting

## 🎯 Purpose

Seventh PR of the fee-subsystem train (#746 → #747 → #750 → #752 → #753 → #754 → **#758** → #762). With #754 the block transition charges fees authoritatively; what was still missing was everything *around* that authority:

1. A fee-invalid transaction (zero `max_fee`, unfundable payer, oversized gas limit, …) was accepted at `sendTransaction`, gossiped to peers, and sat in the mempool forever while the client watched it silently never land.
2. Wallets had no way to learn the current base fees or the block gas caps, so they could not size `max_fee` at all.
3. The block builder discovered the per-block gas caps only *after* executing a transaction (`ExecGasCapExceeded` at settlement), wasting the execution and dropping the transaction.

This PR closes all three: a **fee-admission door** at RPC ingest, a **`getFeeState`** quote RPC, and **declared-gas budgeting** in the block builder. Admission is deliberately **anti-spam, not consensus** — the block transition remains the sole authority; the door only turns away transactions no block could ever include, and tells the client which rule they broke.

## ⚙️ Approach

**Wire types** (`lez/sequencer/service/protocol/src/fees.rs`, new):

- `AdmissionRejection` — 7 arms (`DataBytesOutOfRange`, `GasLimitExceedsMax`, `MaxFeeBelowReserve`, `FeeWitnessOnly`, `UnauthorizedPayer`, `PayerCannotFund`, `OtherFeeValidity`), each carrying the values that decided it and each mapped to a stable JSON-RPC error code in the reserved `-31_980..=-31_986` band. The structured rejection rides in the error's `data` field so clients read it structurally instead of parsing prose.
- `FeeStateQuote` — the current base fees plus a floor/ceiling **band** for the next block's fees (the block being filled is unobservable at query time, so the market is stepped once at an empty block and once at a cap-full block; every possible next-block base fee lies between), the settled `height` for staleness checks, and the block gas caps — exactly what a wallet needs to size `max_fee`.

**The screen** (`lez/sequencer/core/src/fees.rs`, new): `screen(tx, state)` mirrors the settlement's check order (`validate_static_tx` for data-bytes range, gas-limit cap, `max_fee >= fee_reserve`), then the signer/payer-authorization checks, then the one check only a live state can answer: payer balance vs. the fee reserve. Fee-exempt classes (private transactions, deployments, full vault sweeps) pass unscreened. Two verdicts are advisory by nature (base fees and balances move) and documented as such. `fee_quote(state)` prices the band through the same `fee_core::market` arithmetic the block transition moves the real fee state with.

**Wiring**: the screen runs in the executor actor's `Transaction` handler against the head state, and the verdict returns as a typed reply — `SubmitOutcome::{Admitted, Rejected(AdmissionRejection)}` — rather than an actor failure. The RPC service renders a rejection as `rejection_error()` (own code + rendered reason + structured `data`). The gossip publish was **moved after the admission verdict**, so refused spam is never propagated to peers (and a full-mempool error has already returned by then, so no publish can be lost to it either).

**Mempool cap budgeting** (`lez/sequencer/core/src/lib.rs`): the builder tracks a `DeclaredGasBudget` — the running sum of included charged transactions' declared `gas_limit`/`data_bytes` against `MAX_GAS_EXEC`/`MAX_GAS_STOR` — **before** executing. A charged transaction that doesn't fit is deferred (`push_front` + stop filling, mirroring the existing size-deferral precedent) with nothing executed. Since actual gas ≤ declared gas per transaction, a block built under this budget can never breach the consensus caps: the settle-path cap error becomes unreachable in the builder and no execution is wasted. Static validity bounds every charged tx by the caps, so a lone tx always fits an empty budget and deferral never loops forever.

**Deliberate divergences** (documented at the code sites): there is no `ExceedsBlockCap` rejection arm and no private-fee quote — under the interim policy exempt transactions contribute no gas and pay nothing, and static validity already bounds charged transactions by the caps. Oversized deployments are admitted (exempt *and* uncapped; bounded only by the ingest size check).

- [x] Add `AdmissionRejection` (7 arms, stable codes -31_980..=-31_986) and `FeeStateQuote` wire types to `sequencer_service_protocol`
- [x] Add `sequencer_core::fees::{screen, fee_quote}` mirroring the settlement's check order plus the live payer-affordability check
- [x] Wire the screen into the executor actor's `Transaction` handler with a typed `SubmitOutcome` reply; render rejections as structured JSON-RPC errors
- [x] Move the gossip publish after the admission verdict so refused transactions are not propagated
- [x] Add the `getFeeState` RPC method (client generated for the wallet automatically by the jsonrpsee macro)
- [x] Add declared-gas budgeting to the block builder: defer-not-drop when a charged tx's declared gas would exceed the block caps
- [x] Extract `chain_state::apply::opening_fee_state` and reuse it at all three fee-state read sites

## 🧪 How to Test

Mirror CI locally:

```sh
cargo +nightly fmt --check
taplo fmt --check && cargo machete && cargo deny check

# unit suites (fees screen/quote, executor admission, builder deferral, wire-type roundtrips)
```

Highlights to look at:

- `sequencer_core::fees::tests` — every rejection arm has a positive/negative test, and each rejected case cross-checks `settle_verdict` so admission is provably no *looser* than the block transition for the shared checks; `a_reserve_computed_from_the_quote_matches_the_one_admission_uses` pins that a wallet pricing off the quote lands exactly on the admission threshold.
- `a_block_full_of_declared_gas_defers_the_rest` (`sequencer_core::tests`) — six 2M-limit transfers: five fill the 10M budget exactly, the sixth leads the *next* block (deferred, not dropped).
- `handle_transaction_rejects_a_fee_invalid_submission` (executor actor) — the rejection comes back as a reply, not an error.

**Note:** the integration suites still assert pre-fee balances and submit zero-fee raw transactions, so this door breaks several of them by design; their fee-aware conversion (fixture rescale included) lands in #762, stacked on this PR. Run them there.

## 🔗 Dependencies

- #754 (`erhant/fees/actually-charge-txes`) — the settlement/charging layer this door fronts. Merge the train in order: #746 → #747 → #750 → #752 → #753 → #754 → #758.

## 🔜 Future Work

- **Fee-aware integration suite** (#762, stacked on this PR): fixture balance rescale to LGO scale, fee-aware assertions, admission-rejection tests, and the account-less-invocation findings it surfaces.
- **Wallet-side gas estimation.** The default wallet `gas_limit` (2M) means a block carries at most 5 charged wallet transactions against the 10M cap. Wallets should size `gas_limit` from `getFeeState` plus per-program estimation instead of a flat default.
- Revisit the divergences when the interim policy ends: if exempt classes ever contribute gas or private transactions are charged, admission needs an `ExceedsBlockCap`-style arm and the quote needs private pricing.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests

### DIFF
--- a/lez/chain_state/src/apply.rs
+++ b/lez/chain_state/src/apply.rs
@@ -163,15 +163,7 @@ pub fn apply_block_to_state(block: &Block, state: &mut V03State) -> Result<(), B
     // The fee tx is byte-compared only after the user transactions have been
     // settled: its summary is derived from their execution.
 
-    // A malformed fee-state account is consensus-critical corruption (only the
-    // fee guest may write it), so the panic inside `from_bytes` is the
-    // halt-the-node behavior the spec prescribes for consensus faults.
-    let opening = FeeState::from_bytes(
-        &state
-            .get_account_by_id(system_accounts::fee_state_account_id())
-            .data
-            .into_inner(),
-    );
+    let opening = opening_fee_state(state);
     let mut summary = BlockFeeSummary::default();
 
     let is_genesis = block.header.block_id == GENESIS_BLOCK_ID;
@@ -233,6 +225,21 @@ pub fn apply_block_to_state(block: &Block, state: &mut V03State) -> Result<(), B
     Ok(())
 }
 
+/// Reads the fee state a block opening on `state` prices against.
+///
+/// A malformed fee-state account is consensus-critical corruption (only the
+/// fee guest may write it), so the panic inside `from_bytes` is the
+/// halt-the-node behavior the spec prescribes for consensus faults.
+#[must_use]
+pub fn opening_fee_state(state: &V03State) -> FeeState {
+    FeeState::from_bytes(
+        &state
+            .get_account_by_id(system_accounts::fee_state_account_id())
+            .data
+            .into_inner(),
+    )
+}
+
 /// Derives the block fee summary the given user transactions settle to.
 ///
 /// Runs the settlement on a scratch clone of `state`. What block builders
@@ -244,12 +251,7 @@ pub fn derive_block_summary(
     timestamp: Timestamp,
 ) -> Result<BlockFeeSummary, BlockIngestError> {
     let mut scratch = state.clone();
-    let opening = FeeState::from_bytes(
-        &scratch
-            .get_account_by_id(system_accounts::fee_state_account_id())
-            .data
-            .into_inner(),
-    );
+    let opening = opening_fee_state(state);
     let mut summary = BlockFeeSummary::default();
     for (tx_index, transaction) in transactions.iter().enumerate() {
         settle_transaction(
@@ -761,7 +763,7 @@ mod tests {
             fee_core::Instruction::Refund {
                 amount: inbox_revenue,
             },
-            common::test_utils::test_fee_fields(attacker),
+            common::test_utils::test_fee_declaration(attacker),
         )
         .expect("drain message builds");
         let witness = lee::public_transaction::WitnessSet::for_message(&message, &[&attacker_key]);
--- a/lez/chain_state/src/apply.rs
+++ b/lez/chain_state/src/apply.rs
@@ -251,7 +251,7 @@ pub fn derive_block_summary(
     timestamp: Timestamp,
 ) -> Result<BlockFeeSummary, BlockIngestError> {
     let mut scratch = state.clone();
-    let opening = opening_fee_state(state);
+    let opening = opening_fee_state(&scratch);
     let mut summary = BlockFeeSummary::default();
     for (tx_index, transaction) in transactions.iter().enumerate() {
         settle_transaction(
--- a/lez/sequencer/actors/executor/src/actor.rs
+++ b/lez/sequencer/actors/executor/src/actor.rs
@@ -305,7 +305,10 @@ impl<BP: BlockPublisherTrait + Send + Sync + 'static, S: StorageActorTrait> Mess
         _ctx: &mut Context<Self, Self::Reply>,
     ) -> Self::Reply {
         GetFeeQuoteReply {
-            quote: self.sequencer.with_state(sequencer_core::fees::fee_quote).await,
+            quote: self
+                .sequencer
+                .with_state(sequencer_core::fees::fee_quote)
+                .await,
         }
     }
 }
--- a/lez/sequencer/actors/executor/src/actor.rs
+++ b/lez/sequencer/actors/executor/src/actor.rs
@@ -32,7 +32,8 @@ use crate::{
     protocol::{
         GetAccount, GetAccountBalance, GetAccountNonces, GetAccountReply, GetBlock, GetBlockRange,
         GetChannelId, GetChannelIdReply, GetCrossZoneDeadLetters, GetCrossZoneDeadLettersReply,
-        GetLastBlockId, GetProofsAndRoot, GetTransaction, ProduceBlock, Transaction,
+        GetFeeQuote, GetFeeQuoteReply, GetLastBlockId, GetProofsAndRoot, GetTransaction,
+        ProduceBlock, SubmitOutcome, Transaction,
     },
 };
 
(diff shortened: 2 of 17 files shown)


lez/sequencer/core/src/lib.rs
```diff
@@ -75,6 +76,43 @@ const RETIRE_DISPATCH_AFTER_FAILURES: u32 = 3;
 /// block; nothing is dropped.
 const MAX_DISPATCHES_PER_BLOCK: usize = 16;
 
+/// The block's declared-gas budget: the sum of included charged transactions'
+/// signed limits, tracked against the caps the transition enforces on actuals.
+/// Each transaction's actual gas is bounded by its declared gas, so a block
+/// built under this budget can never breach the consensus caps — and no
+/// execution is wasted discovering that.
+#[derive(Clone, Copy, Debug, Default)]
+struct DeclaredGasBudget {
```
moudyellaz, line 107: At the wallet's 2M default this fills a block at 5 transfers, ~0.4 TPS, against a measured cost of ~82k cycles each. Is gas estimation the whole answer, or should the budget fill on an estimate rather than the declared limit?

lez/sequencer/actors/executor/src/actor.rs
```diff
@@ -162,16 +163,27 @@ impl<BP: BlockPublisherTrait + Send + 'static> Message<ProduceBlock> for Executo
...
         &mut self,
         Transaction { transaction }: Transaction,
         _ctx: &mut Context<Self, Self::Reply>,
     ) -> Self::Reply {
+        // Fee admission against the head state, before the mempool sees it.
+        // Advisory (base fees and balances move), but everything it turns
+        // away would have been refused by the block builder anyway.
+        if let Err(rejection) = self
+            .sequencer
+            .with_state(|state| sequencer_core::fees::screen(&transaction, state))
```
moudyellaz, line 208: Only the RPC path is screened. Gossip goes `evaluate_transaction` -> mempool with no fee checks, so every arm here is bypassable by peering. Are you planning to screen gossip too once that handler has a state handle?

moudyellaz (approved):
lgtm, thanks! Left a couple of questions/comments

lez/sequencer/actors/executor/src/actor.rs
erhant: Opened a follow up issue https://github.com/logos-blockchain/logos-execution-zone/issues/797

lez/sequencer/service/protocol/src/fees.rs
```diff
@@ -0,0 +1,147 @@
...
+    #[error("payer {payer:?} holds {balance} but the fee reserve is {fee_reserve}")]
+    PayerCannotFund {
+        payer: AccountId,
+        balance: u128,
+        fee_reserve: u128,
+    },
+    #[error("a public transaction must declare a fee")]
+    MissingFeeDeclaration,
+    #[error("fee-invalid: {reason}")]
+    OtherFeeValidity { reason: String },
```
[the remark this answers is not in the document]
erhant: no such variant exists in code; it's a stale PR-description artifact

lez/chain_state/src/apply.rs
```diff
@@ -244,12 +251,7 @@ pub fn derive_block_summary(
...
 ) -> Result<BlockFeeSummary, BlockIngestError> {
     let mut scratch = state.clone();
-    let opening = FeeState::from_bytes(
-        &scratch
-            .get_account_by_id(system_accounts::fee_state_account_id())
-            .data
-            .into_inner(),
-    );
+    let opening = opening_fee_state(state);
     let mut summary = BlockFeeSummary::default();
```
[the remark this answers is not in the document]
erhant: `derive_block_summary` now reads the opening fee state from `scratch`, not `state` (https://github.com/logos-blockchain/logos-execution-zone/pull/758/commits/165891d33dc3f2c0eaa0933d4d1e7e309c8e88fb)

erhant:
Closing in favor of #801

PR #762, opened 2026-08-18
test(fees): fee-aware integration suite and LGO-scale fixture

erhant:
## 🎯 Purpose

_Last_ PR of the fee-subsystem train (#746 → #747 → #750 → #752 → #753 → #754 → #758 → **#762**): once #754 charges transactions and #758 screens them at the door, the integration suite — written in the pre-fee world — no longer describes reality:

1. The prebuilt fixture funds the wallet's public accounts with 10,000/20,000, roughly three orders of magnitude below a single charged transfer's fee **reserve** (~16.6M at wallet defaults), so every charged public transaction in the suite is unaffordable.
2. Balance assertions hardcode pre-fee arithmetic (`9900`/`20100`-style), which charging invalidates on the sender side.
3. Several tests submit raw zero-fee transactions that #758's admission door now refuses at `sendTransaction` — including the adversarial faucet-impersonation tests, whose enforcement got strictly *stronger* but whose success path changed shape.

This PR converts the integration world to the fee world, and pins the new user-visible semantics as tests while doing so. It also surfaces one genuine protocol gap (account-less invocations cannot pay fees — see Future Work).

## ⚙️ Approach

**Fixture rescale**: `INITIAL_PUBLIC_BALANCES_FOR_WALLET` → `[10^13, 2×10^13]` (the `testnet_initial_state` LGO-scale precedent) and the committed sequencer-db fixture regenerated. Private balances deliberately stay at 10k/20k: privacy-preserving transactions are fee-exempt under the interim policy, so every private-side assertion stays exact and untouched.

**Assertion conversion** (auth_transfer public/private, vault, 4× indexer-state-consistency, wallet-ffi, account): recipients gain exactly the amount; charged senders assert "amount plus a positive fee" (`assert_sender_paid_fee`); fixture-end balances reference the config constant instead of literals. Shielded/deshielded transfers are exempt PP transactions, so their public legs move exactly against a captured before-balance.

**Semantics pinned as tests**:

- Faucet-impersonation tests (3) now assert the `-31_982` (`MaxFeeBelowReserve`) admission rejection at submission — the attack is refused at the door instead of accepted-then-dropped.
- `transfer_beyond_balance_is_dropped_uncharged` (was `failed_transfer_with_insufficient_balance`): the builder's guard pass catches execution failures *before* fee settlement, so the transaction is dropped — never included, nothing charged, no nonce burned. Only failures that first surface inside metered settlement (e.g. running out of declared gas) become included-but-reverted, fee-paying transactions. This test will flag the semantics flip if the guard-pass double-execution FIXME is ever resolved by folding the guards into settlement.
- `fresh_account_cannot_pay_to_initialize_itself` (was `initialize_public_account`): the free-Init bootstrap is dead by design — a fresh account cannot pay to claim itself, and funding it via a transfer auto-claims it (while `Initialize` requires a pristine `Account::default()`). Bootstrap coverage lives in `successful_transfer_to_new_account` (claiming transfer) and the exempt full-vault-sweep path.
- `program_deployment`'s claimer invocation now exercises the **sponsored-fee path** end to end: the new account signs (burning its replay nonce), a funded account is designated payer and authorizes via `with_fee_signer` — the first integration-level use of #750's sponsor wire format.

**Per-flow fixes**:

- `cross_zone_bridge`: `SupplyBridgeLockHolding` credits the holder's spendable balance directly, so the lock transaction just gains fee fields (payer = holder, whose signature fee-authorizes) and an LGO-scale seed; the debit assertion becomes fee-aware.
- `tps`: vault claims stay zero-fee by claiming the **full** per-account supply (fee-exempt full sweep); transfers carry fee fields with a deliberately tight declared `gas_limit` of 100k, sized from a **measured** native transfer (81,858 exec cycles / 310 storage bytes / 657,344 fee at genesis prices) so the builder can still pack ~100 transfers per 10M-cap block.
- Unaffected by design, verified: bridge forged-deposit tests (system-shaped, exempt at the door, still never land), the ingress-guard test (the sequencer-only rule fires before the fee screen), the cross-zone state-machine tests (direct-state, no RPC), and gossip/multi-sequencer (already using the fee-carrying test helper against LGO-scale testnet accounts).

- [x] Rescale fixture public balances to LGO scale and regenerate the committed fixture
- [x] Convert balance assertions to fee-aware form across the affected suites
- [x] Assert the admission-door rejection in the three faucet-impersonation tests
- [x] Pin dropped-uncharged vs revert-with-fee semantics (`transfer_beyond_balance_is_dropped_uncharged`)
- [x] Pin the account-bootstrap decision (`fresh_account_cannot_pay_to_initialize_itself`)
- [x] Exercise the sponsored-fee path in `program_deployment`
- [x] Fee-fix `cross_zone_bridge` and `tps` with measured gas costs
- [x] `#[ignore]` three cross-zone tests blocked on account-less-invocation fee support, with reason strings

## 🧪 How to Test

```sh
# the point of this PR (default features, as CI runs them; needs Docker for bedrock)
RISC0_DEV_MODE=1 cargo nextest run -p integration_tests

# unaffected but verifies no collateral damage
```

Note: the TPS test's 8 TPS target now sits near the consensus ceiling the fee caps imply (~122 transfers per 10M-cap block at the measured ~82k cycles, ≈10 TPS at 12s blocks). It passes, but has little headroom — flag it for retuning if it flakes.

## 🔗 Dependencies

- #758 (`erhant/fees/fee-admission`) — the door these tests assert against. Merge the train in order; this PR closes it.

## 🔜 Future Work

- **Account-less invocations cannot pay fees (protocol gap, needs a spec decision).** Three cross-zone tests (`cross_zone_ping`, `cross_zone_verified`, `cross_zone_watcher_restart`) are `#[ignore]`d: the unsigned, account-less `ping_sender` Send carries no payer account to charge, and the guest's fixed two-account arity leaves no room to add one — the wallet cannot build such a transaction either (no signing account means no payer). The zero-signer sponsored shape is deliberately banned as replayable (`FeeWitnessOnly`), so this needs either payer-tolerant guest arities or a spec ruling, in the same family as deployment replay protection.
- TPS-test target retuning against the fee-implied throughput ceiling (see above).
- Wallet-side gas estimation from `getFeeState` (tracked on #758) would let these tests drop hand-tuned gas limits.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


integration_tests/tests/cross_zone_ping.rs
```diff
@@ -34,6 +34,9 @@ const DELIVERY_TIMEOUT: Duration = Duration::from_secs(480);
 const PING_PAYLOAD: &[u8] = b"hello-cross-zone";
 
 #[test]
+#[ignore = "blocked on fee support for account-less invocations: the unsigned ping_sender Send \
```
moudyellaz, line 37: Noting `ping_sender` is a registered built-in, not a test fixture, so this is a real gap rather than test debt. Any idea on how cross-zone fees will work?

integration_tests/tests/tps.rs
```diff
@@ -134,11 +148,12 @@ impl TpsTestManager {
             .windows(2)
             .map(|pair| {
                 let amount: u128 = 1;
-                let message = putx::Message::try_new(
+                let message = putx::Message::try_new_with_fees(
                     program.id(),
                     [pair[0].1, pair[1].1].to_vec(),
                     [Nonce(1_u128)].to_vec(),
                     authenticated_transfer_core::Instruction::Transfer { amount },
+                    lee::FeeFields::new(pair[0].1, TPS_TRANSFER_GAS_LIMIT, 0, 2_000_000),
```
moudyellaz: Does the hardcoded 2M `max_fee` survive the base-fee rise this test's own load causes? ~8.2M actual per block against a 5M target is ~+8%/block, and the reserve crosses 2M at base_fee 20.

moudyellaz (commented):
hey thanks for the PR. I left some comments, would like to re-review later, so please once it is fixed, ask for a review again!

integration_tests/tests/wallet_ffi.rs
```diff
@@ -676,7 +678,7 @@ fn test_wallet_ffi_get_account_private() -> Result<()> {
         account.program_owner,
         programs::authenticated_transfer().id()
     );
-    assert_eq!(account.balance, 10000);
+    assert_eq!(account.balance, INITIAL_PUBLIC_BALANCES_FOR_WALLET[0]);
```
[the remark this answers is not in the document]
erhant: Fixed in https://github.com/logos-blockchain/logos-execution-zone/pull/762/commits/4cc2bb46e289e51ddd0c9be8ff0299a8f2e9e387

integration_tests/tests/tps.rs
erhant: Fixed in https://github.com/logos-blockchain/logos-execution-zone/pull/762/commits/4cc2bb46e289e51ddd0c9be8ff0299a8f2e9e387, now uses `wallet:: DEFAULT_MAX_FEE`

integration_tests/tests/cross_zone_ping.rs
erhant: re-enabled this test by keeping cross-zone ping's exempt, but we should find a better solution to this (#799 tracks this)

erhant:
Closing in favor of #801

PR #763, opened 2026-08-18
feat(cross_zone)!: accept peer blocks signed by any key in a configured set

### LINKED ISSUE #684: Peer block-signing key is a single pinned key, so a decentralized peer zone's blocks are rejected

moudyellaz:
## 🎯 Purpose

A peer zone is identified by exactly one pinned block-signing key, so a peer running more than one sequencer would have its honest blocks rejected, and on the watcher that ends delivery from the peer permanently. Closes #684.

## ⚙️ Approach

Peer identity becomes a key set: a block is acceptable when signed by any configured key, and a single-sequencer peer is the one-element case.

- [x] Rename `CrossZonePeer.expected_block_signing_pubkey` (optional single key) to `expected_block_signing_pubkeys` (a list, empty = unchecked)
- [x] Add `signed_by_any` to the shared acceptance policy and route all three enforcement sites through it: `screen_peer_block`, the verifier's re-derivation Forged gate, and the evicted-block refetch
- [x] Parse configured keys through one shared `pinned_keys` helper on both the watcher and the verifier
- [x] Refuse unknown fields in cross-zone config, so a stale or misspelled key fails startup instead of silently pinning nothing

## 🧪 How to Test

Unit tests cover acceptance by any listed key (the signer deliberately not first) and refusal when no listed key signed, at the shared policy, the watcher, and the verifier.

## 🔗 Dependencies

Stacked on #743.

## 🔜 Future Work

Key rotation rules arrive with decentralized sequencing. `CrossZonePeer` is host-only but lives in the guest-linked inbox core crate, so config changes move image ids; moving it out is a candidate follow-up when no genesis window is available to ride.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


agureev (approved):

erhant (approved):
LGTM!

PR #770, opened 2026-08-19
feat!: i/o changes and borsh codec

agureev:
## 🎯 Purpose

This PR introduces an optimization on i/o Risc0 operations as well as the introduction of the universal borsh codec to be used across the rust-based code.

As a result of this PR, standard guest program cycles have decreased by 5-10x depending on the number of i/o operations and data size. For comparisson tables, check [this thread](https://discord.com/channels/973324189794697286/1539680320440377417).

To illustrate the changes taken, notice that only 0.4% of `auth_transfer` cycles were program's core logic. The rest was occupied by Risc0's i/o operations and serde.

Some of the load bearing operation examples for `auth_transfer`:

1. In-guest base58 transcoding of every AccountId - 64% of all user cycles

As `AccountID` derived display-oriented `SerializeDisplay, DeserializeFromStr)` with 

```
impl Display for AccountId {
    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
        write!(f, "{}", self.value.to_base58())
    }
}
```

Every serde backend - r0's serde in particular - serialized IDs as base58 strings everywhere. Decoded as such on input and encoded as such on output. Bit-intensive operations are very costy.

The rest of the points are general:

2. Separate reads involved one syscall per 4-byte word

```
let self_program_id: ProgramId = env::read();
let caller_program_id: Option<ProgramId> = env::read();
let pre_states: Vec<AccountWithMetadata> = env::read();
let instruction_words: InstructionData = env::read();
```
separate explicit `read`s introduced quite an extensive syscall overhead, as R0 also documents, instead, we can use `read_slice` decoding the whole input structure at once with only 2 syscall made.

3. Word-inflated encodings

Without custom packing R0s serde encoded each `u8` of a `Vec` into one separate 4-byte word. In order to minimize this cost, every user would have to be wary of hand-packing all the data R0 gives it because this also inflates the `commit` cost.

Using a default codec like borsh, already in use, which also provides dense packing by default kills such packing concerns and hence minimizes packing size of data.

4. Paging cost

Serde as dependency and impls occupy the compiled ELFs more than borsh.

## ⚙️ Approach

- [x] Remove serde codec in favor of borsh
- [x] Remove `read` and `commit` in favor of `_slice` variants
- [x] Unite `read`s of program inputs as one command

## 🧪 How to Test

Run all tests to check that format changes persist the semantics.

## 🔗 Dependencies

#756 

## 🔜 Future Work

1. Estimate whether the borsh encoding is fine for program deployments.

2. Check whether pre-states can be committed to rather than copies to journal output per program.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


lee/state_machine/core/src/frame.rs
```diff
@@ -0,0 +1,85 @@
...
+//!
+//! A frame is a 4-byte little-endian length prefix followed by the payload bytes. The same
+//! [`to_frame`] layout is used by the guest journal commit, the circuit's `env::verify`
+//! reconstruction, and the host input write.
+
+/// Frames `payload`'s borsh serialization as [`to_frame`] would, but serializes directly into
+/// one exact-sized buffer instead of copying an intermediate payload vector.
+#[must_use]
+pub fn to_borsh_frame<T: borsh::BorshSerialize>(payload: &T) -> Vec<u8> {
+    let len = borsh::object_length(payload).expect("borsh serialization is infallible");
```
erhant, line 11: one nit: we do a double serialization here to prevent the const of vector reallocations
once in `object_length`, and then within `to_writer`

I wonder if "single serialization + vec reallocs" is faster than "double serialization"

Claude suggest a code as below if we want to do so:

```rs
  let mut framed = vec![0u8; 4];            // reserve the length slot
  borsh::to_writer(&mut framed, payload).expect("borsh serialization is infallible");
  let len = u32::try_from(framed.len() - 4).expect("frame payload length must fit in u32");
  framed[..4].copy_from_slice(&len.to_le_bytes());
  framed
```

erhant (approved):
LGTM; again, great find to save A LOT of cycles 🙏🏻 

one small comment

Pravdyvy (approved):
lgtm

lee/state_machine/core/src/frame.rs
agureev: when i saw your suggestion that sounded believable but turns out that it introduces ~22% cycle increase on empirical measurement

apparently reallocation are the most costy things ere, while `object_length` does not re-serialize the bulk, it checks the buffer length only

jonesmarvin8 (approved):
Lgtm.

Issue #773, opened 2026-08-20
Create one deterministic Bedrock version for TF tests and the Docker/local setup


hansieodendaal:
I agree we should ultimately have one deterministic Bedrock version for both the TF tests and the Docker/local setup.

For this PR I suggest we keep the scope smaller: LEZ already pins the Logos SDK/TF crates to `97d5e104...` (_without a corresponding release and no binaries to download_), so I'll make the native TF path build logos-blockchain-node from that same pinned revision. That guarantees the Cucumber tests use a node matching the SDK they compile against.

The existing Docker setup (`0.2.1-lssa`) will remain unchanged here. As a follow-up, I think we should move LEZ onto an official Logos release and align both the SDK pin and Docker/node artifact to that release, so we have one version contract rather than two independently maintained revisions. That will let TF download the pinned pre-built binary.

_Originally posted by @hansieodendaal in https://github.com/logos-blockchain/logos-execution-zone/pull/741#discussion_r3821980185_

PR #779, opened 2026-08-21
test: adds cucumber integration tests to CI

### LINKED ISSUE #774: We don't need to specify source in message string
This is not obvious, but actually you don't need to specify source in message string, it will work better with:

```suggestion
    #[error("Deployment failed")]
```

Source error is already marked as `#[source]` and is persisted in the chain. Pretty printing (with `{:#}`) this error wrapped in `anyhow::Error` would give a good-looking trace.

Same for other error variants here with same problem.

_Originally posted by @Arjentix in https://github.com/logos-blockchain/logos-execution-zone/pull/741#discussion_r3821560127_

Also https://github.com/logos-blockchain/logos-execution-zone/pull/741#discussion_r3821518043

hansieodendaal:
## 🎯 Purpose

Add CI coverage for the Cucumber integration tests introduced in #741.

Fixes #774.

image

## ⚙️ Approach

- [x] Add a dedicated Cucumber integration-test workflow for PRs and pushes to `dev`/`main`
- [x] Reuse the existing LEZ CI image and `run-in-ci-image` execution model
- [x] Build the Cucumber test target separately from scenario execution
- [x] Run the four initial Cucumber CI suites sequentially with CI concurrency set to 1
- [x] Add JUnit reporting and failure diagnostics
- [x] Fix source-bearing `thiserror` display messages from #774
- [x] Fix the accidental `block_publisher.rs` indentation change

## 🧪 How to Test

Open the PR and verify the new `Cucumber integration tests` workflow:

- builds the Cucumber test target successfully;
- runs `@smoke_ci`, `@auth_transfer_ci`, `@indexer_state_ci`, and `@multi_sequencer_ci`;
- attempts later suites even if an earlier suite fails;
- uploads the JUnit report and failure diagnostics as appropriate.

## 🔗 Dependencies

None. #741 is already merged into `dev`.

## 🔜 Future Work

#773 will align Bedrock/node binary provisioning so the testing framework can use a deterministic downloadable binary when available and fall back to a source build when required.

Further legacy integration tests can be migrated to Cucumber incrementally.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


hansieodendaal:
General question (@moudyellaz, @Arjentix, @erhant): Should we retire the legacy integration tests that are covered by this initial cucumber CI?

Arjentix:
> Should we retire the legacy integration tests that are covered by this initial cucumber CI?

I don't mind leaving them for now. Maybe once we port all tests we can remove rust integration tests.

Arjentix:
@hansieodendaal I can see [this job](https://github.com/logos-blockchain/logos-execution-zone/actions/runs/32655213421/job/97232778215?pr=779) took pretty much an hour. While we made efforts to make our CI as fast as possible (see other jobs in this PR). Can we speed up this job?

integration_tests/src/cucumber/default.rs
```diff
@@ -57,11 +57,11 @@ pub fn create_scenario_output_dir() -> Result<PathBuf> {
 }
 
 /// Get the number of retries for failed scenarios from the `CUCUMBER_RETRIES`
-/// environment variable. Retries are opt-in: an unset variable and an explicit
-/// zero both disable retries.
+/// environment variable. If the variable is not set, defaults to 2 retries. If
+/// the variable is set to 0, returns None.
 pub fn get_retries() -> Result<Option<usize>> {
```
Arjentix, line 62: Technically this comment implies that return value is either None or Some with non zero, so I propose to use `Result<Option<NonZeroUsize>>`

.github/workflows/cucumber-integration-tests.yml
```diff
@@ -0,0 +1,122 @@
...
+
+      # Build the Cucumber test target first so compile time is separate from
+      # the individual scenario suite timings below.
+      - name: Build cucumber tests
+        id: build_cucumber
+        uses: ./.github/actions/run-in-ci-image
+        with:
+          image: ${{ needs.ci-image.outputs.image }}
+          env: |
+            RUSTFLAGS=-A single-use-lifetimes
```
Arjentix, line 64: Hm, why is this needed?

.github/workflows/cucumber-integration-tests.yml
```diff
@@ -0,0 +1,122 @@
...
+            RISC0_DEV_MODE=1
+          run: >-
+            env -u GITHUB_ACTIONS -u CI
+            cargo test -p integration_tests --features cucumber
```
Arjentix, line 68: I can see it uses integration tests package (makes sense). We also have a job called `integration-tests-prebuild` which builds this package once and after that all integration tests start reusing the binary. So maybe we can reuse it here instead of building from scratch? AFAIK `cucumber` feature doesn't disable any functionality so we can add this feature to `integration-tests-prebuild`

.github/workflows/cucumber-integration-tests.yml
```diff
@@ -0,0 +1,122 @@
...
+        uses: ./.github/actions/run-cucumber-suite
+        with:
+          suite_tag: "@smoke_ci"
+          suite_name: smoke
+          artifact_subdir: "LEZ integration environment"
+          max_concurrent_scenarios: "1"
+          image: ${{ needs.ci-image.outputs.image }}
+
+      # Feature: Authenticated transfers (@auth_transfer_ci)
+      - name: Run cucumber suite - authenticated transfers
```
Arjentix, line 84: These (and other) features looks like that can be run in parallel. Do we run them in parallel here?

hansieodendaal: Each suite runs sequentially. We have the option to set `max_concurrent_scenarios: "1"` to something else than `1`, for example `3`, however, node block progression becomes increasingly non-deterministic in a constrained environment such as CI. I would leave that as an optimization for later on.

.github/workflows/cucumber-integration-tests.yml
hansieodendaal: `RUSTFLAGS=-A single-use-lifetimes` suppresses build warnings because the `logos-blockchain` build environment is different form the LEZ one - what passes there creates warnings here, specifically the issue listed.

hansieodendaal:
> @hansieodendaal I can see [this job](https://github.com/logos-blockchain/logos-execution-zone/actions/runs/32655213421/job/97232778215?pr=779) took pretty much an hour. While we made efforts to make our CI as fast as possible (see other jobs in this PR). Can we speed up this job?

Tackling issue #773 should speed up this entire process build-wise. After that it will come down to test optimization and possible concurrency usage (https://github.com/logos-blockchain/logos-execution-zone/pull/779#discussion_r3852557122).

GitHub provided runners are also not the fastest under load.

.github/workflows/cucumber-integration-tests.yml
hansieodendaal: Good thought. Cucumber uses its own test runner and cannot be executed from the `nextest` archive produced by `integration-tests-prebuild`, so adding the `cucumber` feature there would not let this workflow reuse that archive. 

The Cargo target directory is shared across the disposable CI-image containers, so the Cucumber binary build is reused by the subsequent suite steps within this job. The next PR, tackling #773, should also improve deterministic node provisioning/build caching. Sharing the Rust build across the separate General and Cucumber workflow runs would need a different artifact/cache design, which I suggest we handle in the next PR.

integration_tests/src/cucumber/default.rs
hansieodendaal: Fixed in c400685925ab1fc9597efe2af709dabee45af020.

Arjentix (approved):
LGTM, thanks, looking forward to possible optimizations

integration_tests/src/cucumber/steps/accounts/helpers.rs
```diff
@@ -1,11 +1,25 @@
...
                 lee::AccountId::from(&lee::PublicKey::new_from_private_key(&private_key));
-            (configured_account == account).then_some(balance)
+            if configured_account != account {
+                return None;
+            }
+
+            // The public Cucumber fixture deliberately skips private-account
+            // initialization, but genesis still adds the private pool total
+            // to its public funder account.
+            if index == 0 {
```
moudyellaz: `index == 0` encodes "the first configured public account is the genesis private funder"; can this key off the funder identity instead of position, so a fixture reorder fails rather than silently computing a wrong expected balance?

moudyellaz (approved):
lgtm, thanks. Left a minor comment!

integration_tests/src/cucumber/steps/accounts/helpers.rs
hansieodendaal: Fixed in 066137f7228ff11ca848ef4023bce17106a62ea0.

PR #783, opened 2026-08-22
feat(lee):  program updates - add private execution of permanently-immutable programs

jonesmarvin8:
## 🎯 Purpose

Extends program updates. Programs as accounts (due to mutability of programs) removed function privacy from LEZ. E.g., privacy transactions must disclose what programs were used within its execution. This is necessary for program updatability to enforce program versions across all program owned accounts.

In PR extends the program model by allowing certain programs to be executable without disclosure. Specifically, once a program's `update_auth` is set to `None` (e.g., program can no longer be updated) the program can be executed in a privacy transaction with function privacy. That is, the privacy transaction does not disclose these programs.

## ⚙️ Approach

- [X] Update `RotateUpdateAuth` logic to enforce "finalized" version before `update_auth` can be withdrawn.
- [X] Special logic to have sequencer emit private PDA commitment directly through public transaction once `update_auth = None`.
- [X] Branching paths for function privacy execution in privacy preserving circuit.

## 🧪 How to Test

All previous tests should work; new test:

`lez::programs::loader::core` (loader_core unit tests):
- `rotate_update_auth_to_none_requires_a_finalized_program`
- `rotate_update_auth_to_none_emits_immutable_mirror_commitment`
- `execute_deploy_single_batch_finalizes_atomically` (updated)
- `execute_deploy_single_batch_finalizes_atomically_even_with_real_update_auth`
 
`lee::state_machine::validated_state_diff`  (full state-machine tests):
- `private_claim_with_no_matching_commitment_is_rejected`
- `private_claim_matching_a_real_commitment_passes_verification`

## 🔗 Dependencies

Depends on the long stack of programs as accounts stack.

## 🔜 Future Work

- Enforce version control at the program level before execution. (depends on changes from incremental updates for program structure).
- Add support for "shadow programs"
- Clean up.

## 📋 PR Completion Checklist

- [X] Complete PR description
- [X] Implement the core functionality
- [X] Add/update tests
- [X] Add/update documentation and inline comments

PR #784, opened 2026-08-22
feat(lee):  program updates - add shadow programs

jonesmarvin8:
## 🎯 Purpose

This PR finishes program upgradability stack. Prior to program upgradability, programs could be executed in privacy transactions that were not known to the public state. E.g., the program has never been deployed on chain. This PR reintroduces this notion.

Programs that have never been deployed on chain and can only be executed in the private state are referred to as shadow programs. This PR provides the wiring for shadow programs execution in privacy execution.

## ⚙️ Approach

- [X] `AccountId::for_shadow_program(image_id)` derives a dispatch address as a pure function of `image_id` alone, domain-separated from public/private PDA derivation so it can never collide with either even if the raw bytes happen to coincide.
- [X] Unlike the existing `Public`/`Private` claim paths, nothing about a shadow program's identity is ever committed to, in any form. Its identity is established fresh, in-circuit, on every single dispatch: the circuit decodes and hashes the real elf supplied as a private `ShadowProgramWitness`, derives the address, and feeds the result into the same `program_owner` claim-checking mechanism every other program uses.
- [X] That's a real, unamortized cost paid every time, with no cheaper path possible without disclosing the elf — the same cost regime privacy-preserving transactions already lived in before image-id claims existed at all.
- [X] Identical bytecode from different provers intentionally collides on the same address; this is accepted as natural deduplication rather than a concern, since shadow programs can only ever operate on private accounts and normal account authorization already governs access.
- [X] `ProgramWithDependencies::as_shadow_program`/`with_shadow_dependency` mark a program or dependency as shadow-resolved; `execute_and_prove_with_padded_inputs` splits the call graph into ordinary `ProgramImageClaim::Public` claims and `ShadowProgramWitness`es accordingly.

## 🧪 How to Test

All previous tests should pass;

7 new tests were added:

`lee/state_machine/core/src/program/tests.rs` 
- `for_shadow_program_matches_pinned_value`
- `for_shadow_program_differs_for_different_image_id`
- `for_shadow_program_is_deterministic`
- `for_shadow_program_differs_from_public_pda`

`lee/state_machine/src/privacy_preserving_transaction/circuit/tests.rs`
- `as_shadow_program_is_deterministic_across_independent_callers`
- `as_shadow_program_differs_for_different_programs`

`lee/state_machine/src/state/tests/circuit.rs`
- `shadow_program_claims_a_private_pda_it_legitimately_owns`

## 🔗 Dependencies
Builds off of [PR 783](https://github.com/logos-blockchain/logos-execution-zone/pull/783).

## 🔜 Future Work

- General clean up for removing `ProgramDeploymentTransaction`, `ProgramId`.
- Program upgradability is finished.

## 📋 PR Completion Checklist

- [X] Complete PR description
- [X] Add/update tests


jonesmarvin8:
Stale PR; this is along the program deployment array approach.

PR #785, opened 2026-08-22
feat(indexer): introduce event-filters

agureev:
## 🎯 Purpose

#709 stores all emitted events per tx. This PR introduces an event-filtering system, where each specific indexer can declare the filters-to-watch in the config, allowing to keep track of events emitted by specific programs and specific selectors or to keep watch of all the events.

"Keeping watch" here means persistence. Hence, this is a natural way of only tracking events which matter for your specific application.

## ⚙️ Approach

- [x] Introduce `EventFilter` and `SelectorFilter` enums
- [x] Introduce event filter argument to `open_db`
- [x] Filter event persistence based on the filters starting up

## 🧪 How to Test

See `config.rs` and `block_store.rs` tests

## 🔗 Dependencies

#709 

## 🔜 Future Work

Continue event integration

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


jonesmarvin8 (approved):
Lgtm.

schouhy (approved):

lez/indexer/core/src/block_store.rs
```diff
@@ -300,6 +325,55 @@ impl IndexerStore {
...
+        "persisted event-filter segments start beyond the next block to ingest"
+    );
+    let change = match segments.last_mut() {
+        Some((filter, _)) if *filter == configured => None,
+        Some((filter, from)) if *from == seam => {
+            *filter = configured;
+            Some("replaced the last")
+        }
+        _ => {
+            segments.push((configured, seam));
```
moudyellaz, line 360: Worth a test for an existing db with blocks but no segments? That's the arm every upgrading deployment takes, and the fresh-open test only covers seam 0.

lez/indexer/core/src/config.rs
```diff
@@ -54,6 +60,9 @@ pub struct IndexerConfig {
     /// Defaults to `false`: on mismatch the indexer refuses to start.
     #[serde(default)]
     pub allow_chain_reset: bool,
+    /// Which emitted events this indexer persists.
+    #[serde(default)]
```
moudyellaz, line 64: Should a config that omits this warn at startup? The default keeps nothing, so an upgraded indexer captures no events and every event query returns `UncoveredEventQuery`.

moudyellaz (approved):
lgtm, thanks! Left 2 questions

lez/indexer/core/src/block_store.rs
agureev: something like [this](https://github.com/logos-blockchain/logos-execution-zone/commit/32a0640ef8db4cbed890a5068cda9675a1fb1fd2)?

pushed onto the integration branch

lez/indexer/core/src/config.rs
agureev: something like [this](https://github.com/logos-blockchain/logos-execution-zone/commit/fecf014883b5b3d327d42f4f471f62cea0e3e697)?

this is similarly on the integration branch

PR #790, opened 2026-08-25
refactor(lee): replace ChainedCall.pre_states with pre_state_refs

jonesmarvin8:
## 🎯 Purpose

This PR refactors `ChainedCalls` to replace `pre_states: Vec<AccountWithMetadata>` with `pre_state_refs: Vec<AccountId>`. This removes the potential of account states becoming stale from consecutive chaincalls, and prefers programs from setting `is_authorized`.

## ⚙️ Approach

- [X] Replace `ChainedCall.pre_states` with `pre_state_refs: Vec<AccountId>`, resolving each call's real pre-states from tracked state in both the public-transaction and privacy-preserving execution paths
- [X] Update every chained-call construction site (built-in and test programs) to pass account ids instead of full account values, dropping the manual `is_authorized`/account-value bookkeeping programs used to thread through by hand
- [X] Remove `malicious_authorization_changer`, `malicious_injector`, `malicious_launderer`, and `undeclaring_pda_delegator`, along with the tests exercising the forgery/stale-declaration attacks they demonstrated — that attack class is no longer expressible
- [X] Add `initialize_then_fund` and tests covering a recipient claimed and funded via two separate chained calls

## 🧪 How to Test

New tests:
- `separate_initialize_and_fund_chain_calls_allow_unauthorized_private_recipient`
- `separate_initialize_and_fund_chain_calls_succeed_publicly_for_public_recipient`
- `separate_initialize_and_fund_chain_calls_for_public_recipient_privately`

## 🔗 Dependencies

N/A

## 🔜 Future Work

N/A

## 📋 PR Completion Checklist

- [X] Complete PR description
- [X] Add/update tests


lee/state_machine/core/src/program/mod.rs
```diff
@@ -250,7 +250,10 @@ pub struct CallerData {
 pub struct ChainedCall {
     /// The program ID of the program to execute.
     pub program_id: ProgramId,
-    pub pre_states: Vec<AccountWithMetadata>,
+    /// The accounts the callee should receive as `pre_states`, named by id only. The protocol
+    /// resolves each account's real value and `is_authorized` from its own tracked state — never
+    /// supplied by the calling program.
+    pub pre_state_refs: Vec<AccountId>,
```
agureev: `pre_state_refs` sounds a bit more complex than what it is, no? why not just name it `accounts` or something?

agureev (approved):

lee/state_machine/core/src/program/mod.rs
jonesmarvin8: Changed

lee/state_machine/src/state/tests/circuit.rs
erhant, line 1: the two tests below tested negative-paths that seem to missing now:

- `delegated_pda_is_not_authorized_in_sibling_call`
- `public_pda_first_sight_grant_does_not_extend_to_sibling_calls`

could be worth visiting those in the new model within a test?

erhant (approved):
LGTM, one comment!

lee/state_machine/src/state/tests/circuit.rs
jonesmarvin8: `delegated_pda_is_not_authorized_in_sibling_call` was a test to detect a program forging authorization through sending `pre_state: AccountWithMetadata`. Since we are only passing `AccountId`s, programs can no longer exploit this attack anymore. Making this test stale.

`public_pda_first_sight_grant_does_not_extend_to_sibling_calls` tested the same type of exploit but for PDA accounts. Since this attack is no longer possible by the restricted design, these tests are no longer necessary imo.

agureev: But that's a good point, what these tests were testing was that the PDA authorization is subtree-based. Do we have a test that asserts the authorization in the new refactor is subtree-based for PDAs?

jonesmarvin8: Good point. I've reworked the logic for these two tests and added them back in. I mistakenly thought they were an in program grift check. Thanks!

lee/state_machine/src/privacy_preserving_transaction/circuit/mod.rs
```diff
@@ -125,6 +204,30 @@ pub fn execute_and_prove_with_padded_inputs(
...
+                post.account().program_owner
+            };
+            materialized_state.insert(
+                pre.account_id,
+                Account {
+                    program_owner,
+                    ..post.account().clone()
+                },
+            );
+            if pre.is_authorized {
```
moudyellaz, line 262: `host_authorized_accounts` is tx-wide but the circuit scopes PDA grants to the callee subtree, so a delegated PDA is fed as authorized to a later sibling call with no `pda_seeds` and the circuit then rejects. Should this set be carried per chained call instead?

lee/state_machine/src/privacy_preserving_transaction/circuit/mod.rs
```diff
@@ -96,10 +104,35 @@ pub fn execute_and_prove_with_padded_inputs(
...
+    // authorization may come from a witnessed `ask` with no public-derivable equivalent.
+    let mut host_authorized_accounts: HashSet<AccountId> = pre_states
+        .iter()
+        .filter(|pre| pre.is_authorized)
+        .map(|pre| pre.account_id)
+        .collect();
+
+    // First-sighting position in the circuit's own traversal order, needed to look up a
+    // private-PDA account's witnessed `(npk, vpk, identifier)`.
+    let mut position_by_account: HashMap<AccountId, usize> = pre_states
```
moudyellaz: This seeds positions from top-level input order while the circuit assigns them at first sight in `program_output.pre_states` order; with two or more accounts, does a top-level program that omits one of its pre_states desync the `account_identities` index?

lee/state_machine/src/state/tests/circuit.rs
```diff
@@ -560,33 +561,66 @@ fn sibling_declaring_delegated_pda(pda_is_authorized: bool) -> Result<(), LeeErr
         [(callee_id, callee), (sibling_id, sibling)].into(),
     );
 
-    execute_and_prove(
+    // `callee` gets the PDA's account_id *and* the matching `pda_seeds` — real delegation.
+    // `sibling` gets only the account_id (via `include_pda = true`), no `pda_seeds` — it
+    // should see `is_authorized == false` and panic on it.
```
moudyellaz: The sibling actually receives `is_authorized == true` from the host and does not panic, so this fails in the circuit rather than the guest; was `sibling_call_may_declare_delegated_pda_unauthorized` intentionally dropped as a capability?

lee/state_machine/src/state/tests/circuit.rs
```diff
@@ -732,212 +766,53 @@ fn inherited_scope_passes_through_intermediate_calls() {
...
-        crate::test_methods::auth_asserting_noop(),
-        false,
-    )
-    .expect("a caller's pda_seeds must authorize a public PDA it delegates at first sight");
-
-    // The callee ran with the PDA authorized (auth_asserting_noop did not panic), while
-    // the journal exports the credential view: a seed grant is not a signer-backed claim.
-    assert_eq!(output.public_actions.len(), 1);
-    assert_eq!(output.public_actions[0].pre.account_id, pda);
-    assert!(!output.public_actions[0].pre.is_authorized);
```
moudyellaz, line 872: This was the only assertion on #683's public-PDA journal mask and nothing asserts it now; can this and the other caller-seed first-sight tests come back, given delegation is still expressible via `accounts` + `pda_seeds`?

lee/state_machine/src/validated_state_diff/mod.rs
```diff
@@ -120,30 +107,44 @@ impl ValidatedStateDiff {
...
+            // The caller only names which accounts to call with (`accounts`); resolve their
+            // actual values from the protocol's own tracked state, not from anything it asserts.
+            let real_pre_states: Vec<AccountWithMetadata> = chained_call
+                .accounts
+                .iter()
+                .map(|account_id| {
+                    let account = state_diff
+                        .get(account_id)
+                        .cloned()
+                        .unwrap_or_else(|| state.get_account_by_id(*account_id));
```
moudyellaz: The public path materializes a default account for an id nobody declared while the private host path errors `UnknownChainedCallAccount`; is that asymmetry intentional?

moudyellaz (changes requested):

lee/state_machine/src/state/tests/circuit.rs
jonesmarvin8: Fixed; reworked `sibling_call_may_declare_delegated_pda_unauthorized` with new logic.

lee/state_machine/src/state/tests/circuit.rs
jonesmarvin8: Yes. I removed more tests than I meant by mistake. I've reworked all of the tests that are still applicable with this design.

lee/state_machine/src/privacy_preserving_transaction/circuit/mod.rs
jonesmarvin8: You're right. I've fixed this. Thanks.

lee/state_machine/src/privacy_preserving_transaction/circuit/mod.rs
jonesmarvin8: it does.

I've added a fix that helps with first level of chain calls; plausibly, came up with a solution for subsequent, but this requires a more extensive change. I would prefer to think on this a bit longer and add it in after RC. This would only affect back ends stuff, not programs.

Tests include: `globally_authorized` and `top_level_reordering_through_a_passthrough_is_still_provable`

lee/state_machine/src/validated_state_diff/mod.rs
jonesmarvin8: Not intentional, fixed. Thanks!

lee/state_machine/core/src/program/mod.rs
```diff
@@ -250,7 +250,10 @@ pub struct CallerData {
 pub struct ChainedCall {
     /// The program ID of the program to execute.
     pub program_id: ProgramId,
-    pub pre_states: Vec<AccountWithMetadata>,
+    /// The accounts the callee should receive as `pre_states`, named by id only. The protocol
+    /// resolves each account's real value and `is_authorized` from its own tracked state — never
+    /// supplied by the calling program.
+    pub accounts: Vec<AccountId>,
```
```suggestion
    pub pre_state_ids: Vec<AccountId>,
```

lee/state_machine/core/src/program/mod.rs
```diff
@@ -250,7 +250,10 @@ pub struct CallerData {
 pub struct ChainedCall {
     /// The program ID of the program to execute.
     pub program_id: ProgramId,
-    pub pre_states: Vec<AccountWithMetadata>,
+    /// The accounts the callee should receive as `pre_states`, named by id only. The protocol
```
```suggestion
    /// The ids of the accounts the callee should receive as `pre_states`. The protocol
```

schouhy (approved):
lgtm.

Heads-up that there's a subtle reduction in protocol checks with this. Previously a program could chain another program (or itself) by passing a specific `pre_state` of an account, and the whole execution would fail if that exact pre state was not the actual value of that account the moment that chained call execution kicks in. If there were programs relying on that, happily switching to pass the account id instead of the full pre_state could silently introduce a bug.

lee/state_machine/core/src/program/mod.rs
jonesmarvin8: This actually is a revert back to close what I had originally. But, think is more readable than `accounts`.

lee/state_machine/core/src/program/mod.rs
jonesmarvin8: Applied.

moudyellaz (approved):

PR #749, opened 2026-08-17
feat(lee_core): incremental updates - add AccountDiff core types

jonesmarvin8:
## 🎯 Purpose

First PR for incremental updates. This PR adds types introduced in [incremental updates](https://app.notion.com/p/Incremental-update-proposal-3b28f96fb65c808db8f4c693a8fd6365), but does not incorporate these into the wiring/functionality. Main focus is on `BalanceDiff` and tests. Future PR will begin to incorporate the usage of these new types with programs.

## ⚙️ Approach

- [X] `BalanceDiff` and unit tests
- [X] `AccountDiff`/`AccountDiffOutput` with simplic unused logic
- [X] `ProgramCall`, scoped down to just `Execute`; `CallKind` kept as a hook for a future protocol upgrade

## 🧪 How to Test

All previous tests should work; new unit tests added:
- `apply_balance_diff_add_succeeds`
- `apply_balance_diff_add_zero_is_noop`
- `apply_balance_diff_add_overflow_is_rejected`
- `apply_balance_diff_sub_succeeds`
- `apply_balance_diff_sub_zero_is_noop`
- `apply_balance_diff_sub_down_to_exactly_zero_succeeds`
- `apply_balance_diff_sub_insufficient_balance_is_rejected`
- `serde_roundtrip_for_balance_diff`
- `borsh_roundtrip_for_balance_diff`

## 🔗 Dependencies

N/A

## 🔜 Future Work

First step of [implementation strategy](https://app.notion.com/p/Incremental-update-proposal-3b28f96fb65c808db8f4c693a8fd6365?source=copy_link#3b48f96fb65c80c6ada0e2bf1fd5387c) for incremental updates.
- Next PR rewires programs to produce `AccountDiffOutput`. The circuit/sequencer immediately apply `AccountDiff` to `Account`.
- Later PR updates privacy transactions to emit `AccountDiff`s instead of `post_states: vec<Account>`.

## 📋 PR Completion Checklist

- [X] Complete PR description
- [X] Implement the core functionality
- [X] Add/update tests
- [X] Add/update documentation and inline comments

PR #791, opened 2026-08-25
feat(lee)!: incremental updates - wire programs to diff-native execution

jonesmarvin8:
## 🎯 Purpose

Second PR for incremental updates. This PR rewire programs to output `AccountDiffOutput` (wrapper of `AccountDiff` and `claim`) rather than `AccountPostState`. This change restricts what programs can concretely affect. These resulting diffs from a program's execution is immediately applied afterwards to produce the `AccountPostState`. This restricts the PR's updates to program wiring.

Breaking change: The renames of `AccountDiff`, etc as requested by @schouhy is a breaking change from the previous PR 749.

## ⚙️ Approach

- [X] Rewire every guest program (`lez/programs`, `test_methods`, `test_programs`, `examples`) to emit `AccountDiffOutput`/`AccountDiff` instead of `AccountPostState`, via `read_lee_call`'s `CallKind`-prefixed frame.
- [X] `validate_execution` drops the `ModifiedNonce`/`ModifiedProgramOwner`/`NonDefaultAccountWithDefaultOwner` checks — `AccountDiff` has no `nonce`/`program_owner` field, so a program can no longer express those changes at all, and ownership is governed exclusively by the claim mechanism. Total balance conservation is now checked as `sum(Add) == sum(Sub)` across a call's diffs, rather than comparing pre/post totals.
- [X] Sequencer (`validated_state_diff`) and privacy-preserving circuit (`execution_state`) both apply `apply_balance_diff` and copy `diff_data` to materialize each account's post-state.
- [X] Remove `nonce_changer`, `program_owner_changer`, and `modified_transfer` (and their negative tests) — the class of bug they exercised is now structurally impossible.
- [X] Drop guest-side balance-sufficiency checks (`checked_add`/`checked_sub` with panics); `apply_balance_diff`'s single, centralized checked arithmetic now enforces this at materialization time.
- [X] Privacy transactions still do not append `AccountDiff`s to the circuit's own journal — `PrivacyPreservingCircuitOutput`'s `public_actions`/`private_actions` are untouched; diffs are strictly internal to how the circuit computes each account's post-state.

## 🧪 How to Test

All previous tests should pass.

## 🔗 Dependencies

[PR 749](https://github.com/logos-blockchain/logos-execution-zone/pull/749): introduces the core `AccountDiff`/`BalanceDiff`/`AccountDiffOutput` types this PR wires in.

## 🔜 Future Work

- General clean up; remove `AccountPostState`.

## 📋 PR Completion Checklist

- [X] Complete PR description
- [X] Add/update tests


examples/program_deployment/methods/guest/src/bin/hello_world_with_move_function.rs
```diff
@@ -24,63 +24,83 @@ const MOVE_DATA_FUNCTION_ID: u8 = 1;
...
-            .expect("Data should fit within the allowed limits");
-        this
+            .expect("Data should fit within the allowed limits")
     };
 
-    AccountPostState::new_claimed_if_default(post_account, Claim::Authorized)
+    AccountDiffOutput::new_claimed_if_default(
+        AccountDiff {
+            id: pre_state.account_id,
+            diff_balance: BalanceDiff::Add(0),
```
agureev: Should we maybe make diff balance an `Option`? So that instead of adding 0 you can specify that there is `None` diff?

test_programs/guest/src/bin/simple_balance_transfer.rs
```diff
@@ -32,25 +38,33 @@ fn main() {
...
+        diff_balance: BalanceDiff::Add(balance),
+        diff_data: None,
+    };
 
     ProgramOutput::new(
         self_program_id,
         caller_program_id,
         instruction_data,
-        vec![sender_pre, receiver_pre],
+        vec![sender_pre.clone(), receiver_pre.clone()],
```
agureev: These vectors being misaligned also I think should be at some point united per-slot.

agureev: There should be like an AccountOutput{prestate, state_diff }

lee/state_machine/core/src/program/mod.rs
```diff
@@ -888,36 +789,35 @@ pub fn validate_execution(
...
-                },
-            );
-        }
     }
 
-    // 8. Total balance is preserved
-    let Some(total_balance_pre_states) =
-        WrappedBalanceSum::from_balances(pre_states.iter().map(|pre| pre.account.balance))
+    // 5. Total balance is preserved
+    let Some(total_added) =
```
agureev, line 964: Feels like this should be just a `reduce` with an accumulator, this seems a bit of an overkill

lee/state_machine/core/src/program/mod.rs
```diff
@@ -710,20 +634,15 @@ pub enum ExecutionValidationError {
...
     BalanceSumOverflow,
 
     #[error(
-        "Total balance across accounts is not preserved: total balance in pre-states {total_balance_pre_states}, total balance in post-states {total_balance_post_states}"
+        "Total balance across accounts is not preserved: total added {total_added}, total subtracted {total_subbed}"
     )]
     MismatchedTotalBalance {
-        total_balance_pre_states: WrappedBalanceSum,
-        total_balance_post_states: WrappedBalanceSum,
+        total_added: WrappedBalanceSum,
```
agureev, line 708: Hm but that would make this error less explicit...

Maybe just file an issue to consider this?

agureev (approved):

examples/program_deployment/methods/guest/src/bin/hello_world_with_move_function.rs
jonesmarvin8: Addressed in 749, and rebased this branch on it.

Note: this issue was originally mentioned in conversation, and I thought the request was referring to 749 instead of here.

lee/state_machine/core/src/program/mod.rs
jonesmarvin8: Created [Issue 803](https://github.com/logos-blockchain/logos-execution-zone/issues/803)

lee/state_machine/core/src/program/mod.rs
jonesmarvin8: Created [Issue 803](https://github.com/logos-blockchain/logos-execution-zone/issues/803)

test_programs/guest/src/bin/simple_balance_transfer.rs
jonesmarvin8: I think this change might be too large for the scope of this PR.

lee/state_machine/core/src/account.rs
schouhy: not introduced in this pr, but I think `AccountDiff` is a misnomer. the data in it is not a diff, only the balance. Maybe `PostStateOutcome` or `PostStateEffects` with `balance_diff` and `new_data` fields

jonesmarvin8: I agree. The names are weird given thee limited scope we're going for now.
`AccountDiff` -> `PostStateEffects`
`AccountDiffOutput` -> `PostStateEffectsOutput`
`diff_data` -> `new_data`

lee/privacy_preserving_circuit/src/execution_state.rs
```diff
@@ -407,11 +409,22 @@ impl ExecutionState {
                 authorized_output_accounts.push(pre_account_id);
             }
 
-            if let Some(claim) = post.required_claim() {
+            let diff = diff_output.diff();
+
+            let balance = apply_balance_diff(pre_account.balance, diff.diff_balance)
+                .expect("balance diff must be valid; validate_execution already checked it");
```
erhant: `validate_execution checks` conservation `sum(Add) == sum(Sub)` and that decreases are owner-authorized, but it does not seem to check per-account sufficiency (`Sub <= pre.balance`) or per-account `Add` overflow. So this expect can actually fire.

now that guests drop their own `checked_sub` guards (e.g. `simple_balance_transfer` just emits `Sub(balance)`), an over-spend of more than the sender holds passes `validate_execution` (since `Sub(b) == Add(b)`), and then `apply_balance_diff(sender.balance, Sub(b))` returns `InsufficientBalance` here, and will panic. 

`Add` overflow is very unlikely, but `Sub` case may be considered!

erhant (approved):
LGTM, one comment!

lee/privacy_preserving_circuit/src/execution_state.rs
jonesmarvin8: Thanks. I've added an additional constraint to check the balance are usable. Additionally, two tests: `validate_execution_rejects_insufficient_balance_even_if_globally_conserved` and `validate_execution_rejects_add_overflow`.

lee/state_machine/core/src/program/mod.rs
```diff
@@ -840,34 +764,21 @@ pub fn validate_execution(
...
-        }
-
-        // 4. Program ownership changes are not allowed
-        if pre.account.program_owner != post.account.program_owner {
-            return Err(ExecutionValidationError::ModifiedProgramOwner {
-                account_id: pre.account_id,
-            });
-        }
-
+    for (pre, diff_output) in pre_states.iter().zip(post_diff) {
```
moudyellaz: `PostStateEffects.id` is never compared to `pre.account_id` here or in either materializer, and with `ModifiedNonce`/`ModifiedProgramOwner` gone nothing detects a `pre_states`/`post_diffs` misalignment any more; should this loop check the ids line up?

lee/state_machine/core/src/program/mod.rs
```diff
@@ -542,8 +475,8 @@ pub struct ProgramOutput {
     pub instruction_data: InstructionData,
     /// The account pre states the program received to produce this output.
     pub pre_states: Vec<AccountWithMetadata>,
-    /// The account post states the program execution produced.
-    pub post_states: Vec<AccountPostState>,
+    /// The account diffs the program execution produced.
+    pub post_diffs: Vec<PostStateEffectsOutput>,
```
moudyellaz: This renames the guest-facing output type and field (and `read_lee_inputs` to `read_lee_call`), which breaks every guest in `lez-programs` and anything on `lez-framework`'s `#[lez_program]` macro, but no commit carries `!` or a `BREAKING CHANGE:` footer; can those be added?

lez/programs/sequencer_stake/src/main.rs
```diff
@@ -355,21 +380,22 @@ fn finalize_unstake(
     );
 
     // no signature check: already authorized back in UnstakeRequest
-    let mut ownership_account_new = ownership_account.account.clone();
-    ownership_account_new.balance = ownership_account_new
-        .balance
-        .checked_sub(pending.amount)
-        .expect("insufficient staked balance");
```
moudyellaz, line 488: Dropping the guest-side guards replaces domain messages like this one with a generic `InvalidBalanceDiff` from the protocol; is losing the per-program wording intended for the wallet-facing error surface?

lee/state_machine/src/state/tests/claiming.rs
```diff
@@ -529,95 +529,3 @@ fn claiming_mechanism_cannot_claim_initialied_accounts() {
         )) if err_account_id == account_id
     ));
 }
-
-/// This test ensures that even if a malicious program tries to perform overflow of balances
-/// it will not be able to break the balance validation.
-#[test]
-fn malicious_program_cannot_break_balance_validation_if_not_in_genesis() {
```
moudyellaz, line 744: This also asserted state was left untouched after the rejection, which the two new `validate_execution` unit tests do not cover; can an end-to-end replacement come back?

lee/state_machine/core/src/program/mod.rs
jonesmarvin8: Added.

lez/programs/sequencer_stake/src/main.rs
jonesmarvin8: Yes. This shift does lose per-program wording errors. This is necessary for the (eventual) full scope of incremental updates. Ultimately, to address the protocol-level race condition for privacy transactions with public accounts, the balance updates would need to be handled after the fact by the sequencer.

lee/state_machine/src/state/tests/claiming.rs
jonesmarvin8: Added test `insufficient_balance_transfer_leaves_state_untouched`.

lee/state_machine/core/src/program/mod.rs
jonesmarvin8: Thanks! I added this. And a test program (`reordering_transfer`) and test (`reordered_post_diffs_are_rejected`) to demonstrate this.

erhant (approved):

lee/state_machine/core/src/account.rs
```diff
@@ -106,10 +106,22 @@ pub enum BalanceDiffError {
 }
 
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, BorshSerialize, BorshDeserialize)]
-pub struct AccountDiff {
+pub struct PostStateEffects {
```
schouhy, line 116: why does the post state effects need an `id`? isnt the id inferred by the position in the post state vec and the same position in the pre state?

lee/state_machine/core/src/account.rs
```diff
@@ -106,10 +106,22 @@ pub enum BalanceDiffError {
...
     pub id: AccountId,
     pub diff_balance: Option<BalanceDiff>,
-    pub diff_data: Option<Data>,
+    pub new_data: Option<Data>,
+}
+
+impl PostStateEffects {
+    /// A diff that leaves `id`'s balance and data untouched.
+    #[must_use]
+    pub const fn unchanged(id: AccountId) -> Self {
```
schouhy:
```suggestion
    pub const fn new_unchanged(id: AccountId) -> Self {
```

lee/state_machine/core/src/program/mod.rs
```diff
@@ -302,105 +293,47 @@ pub struct AccountPostState {
...
-        self.account
-    }
-}
-
 #[derive(Debug, Clone, Serialize, Deserialize, BorshSerialize, BorshDeserialize)]
 #[cfg_attr(any(feature = "host", test), derive(PartialEq, Eq))]
-pub struct AccountDiffOutput {
-    diff: AccountDiff,
+pub struct PostStateEffectsOutput {
+    diff: PostStateEffects,
```
```suggestion
    effects: PostStateEffects,
```

lee/state_machine/core/src/program/mod.rs
```diff
@@ -302,105 +293,47 @@ pub struct AccountPostState {
...
-    pub fn into_account(self) -> Account {
-        self.account
-    }
-}
-
 #[derive(Debug, Clone, Serialize, Deserialize, BorshSerialize, BorshDeserialize)]
 #[cfg_attr(any(feature = "host", test), derive(PartialEq, Eq))]
-pub struct AccountDiffOutput {
-    diff: AccountDiff,
+pub struct PostStateEffectsOutput {
```
schouhy: The naming does not make much sense to me. `PostStateEffectOutput` is a wrapper around `PostStateEffect` with a claiming flag. Why `Output`? Maybe better `PostStateEffectWithClaiming`. But that's a mouthful. Why not just move the `claim` field inside the `PostStateEffect` struct and avoid the mamushka?

lee/state_machine/core/src/program/mod.rs
jonesmarvin8: Changed.

lee/state_machine/core/src/account.rs
jonesmarvin8: Addressed.

lee/state_machine/core/src/account.rs
jonesmarvin8: This is a hold over for incremental updates. `PostStateEffects.id` guarantees that the wrong account `PostStateEffects` cannot be passed for a different `Account`.

This is worth keeping since we're still investigating incremental updates, and we want to avoid breaking changes on the program shape.

lee/state_machine/core/src/program/mod.rs
jonesmarvin8: I agree that right now `PostStateEffects` with claiming makes sense due to the limited usage of `PostStateEffects`.

Ultimately, we want `PostStateEffects` to be used with incremental updates. This means that the updates will "live" longer than claiming. E.g., sequencer receives the `PostStateEffects` for each public account used in a privacy transaction. To this end, the `claim` field would be awkward to keep passing.

We can rename `PostStateEffectsWithClaiming` if you really want...

lee/state_machine/core/src/program/mod.rs
moudyellaz: Title and body flag it now, but no branch commit carries `!` or the footer, and this repo merges with merge commits so only branch commits land on dev (compare 4bda70f02): could the rename commit carry the footer?

moudyellaz (approved):
lgtm, thanks! Left one nit

lee/state_machine/core/src/program/mod.rs
schouhy: no, I'm not convinced about that naming either. It's just that by looking at the name I get no information of what that struct is with `PostStateEffectsOutput`. I'd put `Output` to a struct that's the result of a process, but in this case that's already in the `Effects` part

lee/state_machine/core/src/account.rs
schouhy: I think it's better to have the circuit produce this pairing between effects and ids instead of the program itself

schouhy (approved):

lee/state_machine/core/src/program/mod.rs
```diff
@@ -807,97 +792,94 @@ pub fn read_input_frame() -> Vec<u8> {
...
-        ProgramInput {
-            self_program_id,
-            caller_program_id,
-            pre_states,
-            instruction,
-        },
-        instruction_data,
+/// Responds to a call kind this program doesn't implement with a no-op — a deliberate skip,
+/// not a failure.
+pub fn respond_unsupported_call<T>(call: ProgramCall<T>) -> ! {
```
moudyellaz, line 848: `ProgramOutput` records no call kind, and the no-op echoes `instruction_data`, `self_program_id` and `caller_program_id`, so it clears `env::verify` and all three asserts at `execution_state.rs:156/170/179`. `program_outputs` is prover-supplied on the private path, so `CallKind::Unknown(n)` on `flash_swap_initiator`'s `InvariantCheck` self-call skips the only check that the callback repaid the vault, while chained calls 1 and 2 still commit. Pin the call kind into `ProgramOutput`, or keep `Unsupported` fatal until a second kind exists?

lee/state_machine/src/state/tests/claiming.rs
```diff
@@ -62,6 +62,84 @@ fn unauthorized_public_account_claiming_fails() {
...
+/// program listing its diffs in a different order than it received the corresponding pre-states
+/// still validates and applies correctly.
+#[test]
+fn reordered_state_diffs_still_succeed() {
+    let program = crate::test_methods::reordering_transfer();
+    let from_key = PrivateKey::try_new([23; 32]).unwrap();
+    let from = AccountId::from(&PublicKey::new_from_private_key(&from_key));
+    let initial_balance = 10;
+    let mut state = V03State::new()
+        .with_public_accounts(public_state_from_balances(&[(from, initial_balance)]))
```
moudyellaz: This fails on rule 2, not on ordering: `public_state_from_balances` (`state/tests/mod.rs:194`) seeds `program_owner = simple_balance_transfer`, but the executing program is `reordering_transfer`, so the nonzero `Sub` is rejected as `UnauthorizedBalanceDecrease` whatever order the diffs come in. Seed the owner as the executing program so the test exercises what its name claims?

lee/state_machine/core/src/program/mod.rs
```diff
@@ -302,131 +293,88 @@ pub struct AccountPostState {
...
-            claim: None,
-        }
-    }
+/// A single account's full pre-state paired with the diff a program's execution applies to it.
+#[derive(Debug, Clone, Serialize, Deserialize, BorshSerialize, BorshDeserialize)]
+#[cfg_attr(any(feature = "host", test), derive(PartialEq, Eq))]
+pub struct AccountStateDiff {
+    pub pre_state: AccountWithMetadata,
+    pub post_balance_diff: BalanceDiff,
+    pub post_data: Data,
```
moudyellaz: `post_data` is `Data`, not `Option<Data>`, so every account's data now travels twice per output, once inside `pre_state` and again as `post_data`, where the previous shape omitted it when unchanged. Output size goes up in the PR whose point is diff-native execution. Any reason not to keep it optional?

lee/state_machine/core/src/program/mod.rs
```diff
@@ -919,11 +901,21 @@ pub fn validate_execution(
...
+
+        // 5. A non-default account left with the default owner must be a claimless byte-identical
         //    echo: `Claim` is applied after this check, so a claimed echo would seize the account.
-        if post.account.program_owner == DEFAULT_PROGRAM_OWNER && pre.account != Account::default()
-        {
-            let claimless_echo = post.account == pre.account && post.required_claim().is_none();
+        let post_owner_is_default =
+            diff.post_claim.is_none() && account_program_owner == DEFAULT_PROGRAM_OWNER;
+        if post_owner_is_default && pre.account != Account::default() {
+            let claimless_echo =
```
moudyellaz: `claimless_echo` requires `post_balance_diff == BalanceDiff::Add(0)`, while rule 2 exempts `Sub(0)`, so `Sub(0)` on a non-default account with the default owner is rejected as `NonDefaultAccountWithDefaultOwner` while the identical `Add(0)` no-op passes. Treat both zero diffs as an echo?

lee/state_machine/src/state/tests/privacy_preserving.rs
```diff
@@ -379,93 +356,6 @@ fn data_changer_program_should_fail_for_too_large_data_in_privacy_preserving_cir
...
-    );
-
-    assert_circuit_proving_failure(
-        &result,
-        "Unallowed modification of program owner for account",
-    );
-}
-
 #[test]
 fn transfer_from_non_owned_account_should_fail_in_privacy_preserving_circuit() {
```
moudyellaz, line 360: Both length tests were correctly dropped here since the mismatch is unrepresentable now, but the public path got `program_should_fail_with_missing_output_accounts` re-pointed at `DeclaredAccountMissingFromOutput` and the circuit path got no replacement. I could not find an equivalent completeness check in `derive_from_outputs`; the private-PDA loop covers dropped private PDAs by position, nothing seems to cover a dropped public account. Is there one I am missing, or should the circuit-side test come back?

moudyellaz (changes requested):

lee/state_machine/core/src/program/mod.rs

lee/state_machine/core/src/program/mod.rs
jonesmarvin8: Thanks! Oversight on my part while trying to get these changes in before RC. Fixed.

lee/state_machine/core/src/program/mod.rs
jonesmarvin8: Pinned call kind to `ProgramOutput`; currently anything other than `Execute` will have fatal `Unsupported` until a second kind exists.

Will add a unit test when working with incremental updates further.

lee/state_machine/src/state/tests/claiming.rs
jonesmarvin8: Fixed the test `program_should_fail_with_missing_output_accounts`. Silly error. Thanks.

lee/state_machine/src/state/tests/privacy_preserving.rs
jonesmarvin8: This was actually a gap in `dev` as well.

First layer of executions did not check whether the initial pre states (used for the execution call) matched with the program outputs. Chaincalls has this check. Added, and fixed the test.

moudyellaz (approved):
lgtm, thanks!

schouhy (commented):
I think BalanceDiff variants should be mutually exclusive: Add(balance > 0), Sub(balance > 0) and Zero.

PR #793, opened 2026-08-26
test: one deterministic Bedrock version

### LINKED ISSUE #773: Create one deterministic Bedrock version for TF tests and the Docker/local setup

hansieodendaal:
## 🎯 Purpose

Resolve #773 by making the Logos Blockchain revision resolved by LEZ's Cargo dependencies the single source of truth for Bedrock node provisioning.

Previously, LEZ could use different Bedrock versions depending on the execution path:

- Rust/TF dependencies were pinned to a `logos-blockchain` Git revision;
- TF could build a node from that revision;
- the Docker setup independently used `0.2.1-lssa`.

This PR removes that independent Bedrock version. In CI, the Linux Bedrock node is resolved once and shared across General integration tests, Docker-backed tests, TF consumers, and Cucumber.

Fixes #773.

**🔒 Version Contract**

The Cargo-resolved Logos revision is authoritative.

```mermaid
flowchart TD
    A["Cargo.toml / Cargo.lock<br/>Logos dependencies"] --> B["cargo metadata --locked<br/>exact Logos commit"]

    B --> C{"Matching published<br/>release asset?"}
    C -->|Yes| D["Download + verify SHA-256"]
    C -->|No| E["Build exact Cargo checkout"]

    D --> F["Platform-specific resolved node"]
    E --> F

    F --> G["Linux CI artifact"]
    G --> H["General / TF"]
    G --> I["Cucumber"]
    G --> J["Docker Bedrock"]

    F --> K["Host-native TF<br/>when requested"]
```

Platform-specific binaries may differ, but they must all correspond to the same Cargo-resolved revision.

## ⚙️ Approach

- [x] Derive the exact `logos-blockchain` revision from `cargo metadata --locked`
- [x] Reject Cargo graphs containing multiple Logos revisions
- [x] Use an exact matching release asset when available and verify its SHA-256
- [x] Otherwise build `logos-blockchain-node` from the exact Cargo checkout
- [x] Resolve/build the Linux Bedrock node once per CI run and share it as an artifact
- [x] Reuse that artifact across TF, Docker-backed tests, integration tests, and Cucumber
- [x] Make Cucumber a reusable workflow while keeping it separate from nextest
- [x] Remove the independently versioned `0.2.1-lssa` node image
- [x] Support Docker/Linux and host-native local resolution
- [x] Exclude integration-test binaries with no runnable nextest tests from the CI matrix
- [x] Add resolver and target-discovery coverage

**Resolver**

The shared resolver:

1. derives the exact Logos revision from Cargo;
2. checks for a published release resolving to that commit;
3. selects only the exact expected platform asset;
4. verifies its published SHA-256;
5. otherwise builds:

```bash
cargo build --locked --release \
  -p logos-blockchain-node \
  --features testing
```

The source checkout is verified before building.

Outputs:

```text
bedrock/.resolved/
├── logos-blockchain-node
└── metadata.json
```

Host-native resolution uses:

```text
bedrock/.resolved-host/
├── logos-blockchain-node
└── metadata.json
```

Metadata records the revision, target platform, resolution method, release information when applicable, and binary SHA-256.

HTTP reads use bounded retries for transient failures; ordinary client errors such as `404` are not retried.

**Single CI producer**

`ci.yml` is the top-level CI orchestrator.

The CI image and Bedrock node are each produced once. Unrelated checks remain parallel, while only actual Bedrock consumers wait for the node artifact:

```mermaid
flowchart TD
    A["CI image"] --> B["Bedrock node resolver"]
    A --> C["lint / unit / fmt / other CI"]
    A --> D["integration-tests-prebuild"]

    B --> E["bedrock-node artifact"]
    D --> F["nextest archive"]

    E --> G["Integration-test matrix"]
    F --> G

    E --> H["test_fixtures / Docker"]
    E --> I["TF consumers"]
    E --> J["Reusable Cucumber workflow"]
```

TF consumers receive:

```text
LOGOS_BLOCKCHAIN_NODE_BIN=<workspace>/bedrock/.resolved/logos-blockchain-node
```

This removes the previous need to recreate Cargo Git state or unset `CI` / `GITHUB_ACTIONS` to force TF onto its source-build provider.

**Cucumber CI**

Cucumber no longer resolves or builds another Bedrock node.

The reusable workflow receives the shared CI image, downloads the same `bedrock-node` artifact, builds the Cucumber target, then runs:

```text
@smoke_ci
    ↓
@auth_transfer_ci
    ↓
@indexer_state_ci
    ↓
@multi_sequencer_ci
```

Each suite receives the shared node through `LOGOS_BLOCKCHAIN_NODE_BIN`.

Cucumber remains a separate execution path from nextest.

**Integration-test target discovery**

The integration-test matrix is now derived from the full nextest test inventory rather than just the binary list.

Only binaries containing at least one runnable test are scheduled. This avoids downloading and extracting the large nextest archive for binaries that would simply report `0 tests run`.

`tps` remains intentionally excluded.

**Temporary TF test quarantine**

The three `tf_app_deployments` tests are temporarily ignored because concurrent execution exposed a separate port-allocation race in the upstream Logos testing infrastructure.

The observed `AddrInUse` node startup failures break the Bedrock network and cause downstream cluster/indexer failures.

Fixing that cross-process port reservation mechanism is outside this PR. Once fixed upstream, these tests should be re-enabled.

**Docker-backed Bedrock**

The independently versioned `0.2.1-lssa` image is removed.

Docker now builds a small runtime image containing the same Cargo-authoritative Linux node used by CI:

```text
Cargo-resolved Logos revision
            │
            ▼
       Linux node
        ┌───┴────┐
        ▼        ▼
       CI      Docker
```

The existing Bedrock configuration, ports, and launch script remain unchanged.

**Local development**

Docker-backed Bedrock:

```bash
just resolve-bedrock-node
just run-bedrock
```

Host-native TF:

```bash
just resolve-host-bedrock-node

LOGOS_BLOCKCHAIN_NODE_BIN="$PWD/bedrock/.resolved-host/logos-blockchain-node" \
  cargo test -p integration_tests ...
```

Docker and host-native binaries may differ by platform, but both correspond to the same Cargo-resolved Logos revision.

Docker-backed tests also validate stored resolver metadata against the current `Cargo.lock`.

**Coverage**

Resolver tests cover:

- exact release/commit matching;
- asset and digest validation;
- duplicate/unverifiable asset rejection;
- archive extraction;
- multiple Cargo revision rejection;
- source checkout verification;
- transient HTTP retries;
- live published-release lookup.

Target-discovery tests cover:

- excluding binaries with no runnable tests;
- preserving runnable binaries;
- `tps` exclusion;
- rejecting unsuitable binary-only nextest listings.

## 🧪 How to Test

Verify that CI demonstrates:

- one CI-image producer;
- one Cargo-authoritative Bedrock resolution/build;
- one `bedrock-node` artifact;
- General and Cucumber both consuming that artifact;
- unrelated CI work remaining parallel with the Bedrock build;
- no integration-test matrix jobs for binaries without runnable tests;
- all expected General/TF/Docker-backed tests passing;
- all four Cucumber suites passing.

Resolver logs should report:

- Cargo-resolved Logos revision;
- target platform;
- release download or source-build resolution;
- resulting binary SHA-256.

**Expected result**

```text
                 LEZ Cargo dependencies
                          │
                          ▼
                  exact Logos revision
                          │
              ┌───────────┴───────────┐
              ▼                       ▼
       Linux node binary       Host-native binary
              │                       │
      ┌───────┼────────┐              ▼
      ▼       ▼        ▼         local native TF
   General Cucumber  Docker
     CI              Bedrock
```

Updating LEZ's Logos dependency is sufficient to change the Bedrock node revision used throughout the project.

If an exact release asset exists, it is downloaded and verified. Otherwise that exact revision is built from source.

## 🔗 Dependencies

Depends on #779.

## 🔜 Future Work

This PR deliberately does not:

- change the authoritative Logos dependency revision;
- route Cucumber through nextest;
- share the nextest/Cucumber test build products;
- change Cucumber scenario concurrency;
- fix the upstream port-allocation race exposed by `tf_app_deployments`;
- introduce `workflow_run` or cross-workflow-run artifact plumbing.

Once the upstream allocator is fixed, the temporarily ignored `tf_app_deployments` tests should be re-enabled.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


.github/scripts/resolve_bedrock_node.py
```diff
@@ -0,0 +1,558 @@
...
+) -> dict[str, str] | None:
+    for release in releases:
+        if release.get("draft") or not release.get("published_at"):
+            continue
+
+        tag = release.get("tag_name")
+        if not isinstance(tag, str):
+            continue
+        try:
+            commit = commit_resolver(tag).lower()
```
moudyellaz: `matching_release` resolves `GET /commits/{tag}` for every published release before checking whether that release even carries the expected asset, so with 41 releases and no match that is 41 API calls per run; can the asset-name filter run first?

.github/scripts/resolve_bedrock_node.py
```diff
@@ -0,0 +1,558 @@
...
+
+
+def asset_digest(asset: dict[str, Any], assets: list[dict[str, Any]]) -> str | None:
+    digest = asset.get("digest")
+    if isinstance(digest, str) and digest.startswith("sha256:"):
+        digest = digest.removeprefix("sha256:").lower()
+        if SHA256_PATTERN.fullmatch(digest):
+            return digest
+
+    checksum_names = {
```
moudyellaz: The `SHA256SUMS` fallback reads the digest from the same release that published the asset, so unlike the GitHub-computed `digest` field it only proves integrity, not authenticity; is the fallback worth keeping?

integration_tests/tests/tf_app_deployments.rs
```diff
@@ -21,6 +21,7 @@ use testing_framework_app::{AppHostEnv, AppHostTopology, DeployContext};
 use testing_framework_core::scenario::NodeClients;
 
 #[tokio::test]
+#[ignore = "Flaky when multiple integration tests run concurrently in CI"]
```
moudyellaz: With the new discovery script dropping binaries that have no runnable tests, this target disappears from the matrix entirely, so there is no "0 tests run" job left as a reminder; can the tracking issue number go in the `#[ignore]` reason?

moudyellaz (approved):
lgtm, thanks. Left some minor comments!

Justfile
```diff
@@ -80,14 +81,24 @@ run-bedrock *args:
...
+        docker compose up --build
     else
         mkdir -p "$(dirname "$log")"
         printf '\n=== %s  bedrock ===\n' "$(date -Is)" >>"$log"
-        docker compose up 2>&1 | tee -a "$log"
+        docker compose up --build 2>&1 | tee -a "$log"
     fi
 
+resolve-bedrock-node:
+    @bash .github/scripts/resolve_bedrock_node_in_docker.sh
```
Arjentix: Maybe better place this scripts into `bedrock` directory then?

test_fixtures/src/setup.rs
```diff
@@ -217,13 +219,84 @@ pub async fn indexer_client(addr: SocketAddr) -> Result<IndexerClient> {
...
+}
+
 pub async fn setup_bedrock_node() -> Result<(DockerCompose, SocketAddr)> {
     let manifest_dir = env!("CARGO_MANIFEST_DIR");
     let bedrock_compose_path = PathBuf::from(manifest_dir).join("../bedrock/docker-compose.yml");
+    let resolved_binary =
+        PathBuf::from(manifest_dir).join("../bedrock/.resolved/logos-blockchain-node");
+    if !resolved_binary.is_file() {
+        bail!(
+            "Resolved Bedrock node is missing at {}. Run `just resolve-bedrock-node` before starting Docker-backed tests.",
```
Arjentix, line 286: I wonder if it makes sense now to keep docker-backed bedrock in tests now. If we're building host-native binary anyway maybe it makes sense to port tests to native bedrock run?

Although I'm not sure. E.g. docker compose provides us a unified setup for tests and local docker running.

integration_tests/tests/tf_app_deployments.rs
Arjentix: Looks a bit strange that this became flaky... Why?

integration_tests/src/cucumber/steps/transfers/public.rs
```diff
@@ -170,11 +170,10 @@ async fn transfer_to_new_public_account(
     Ok(())
 }
 
-#[when(expr = "I attempt to transfer {int} from the first configured public account to the second")]
+#[when("I attempt to transfer twice the first configured public account balance to the second")]
```
Arjentix: Isn't it too specific now?

.github/workflows/ci.yml
```diff
@@ -26,6 +26,62 @@ jobs:
...
+    with:
+      image: ${{ needs.ci-image.outputs.image }}
+    permissions:
+      contents: read
+      packages: read
+    concurrency:
+      group: cucumber-${{ github.ref }}
+      cancel-in-progress: true
+
+  bedrock-resolver-tests:
```
Arjentix, line 77: Can you explain what it does any why we need that?

Arjentix (approved):
LGTM, thanks. Tbh I'm a bit worried how complex this became and already forgot why we can't just always use pinned docker version

Justfile
hansieodendaal: Moved the Bedrock-specific resolver, wrapper and tests to `bedrock/tools/` and kept `bedrock/scripts/` for scripts used by the Bedrock runtime itself, and left generic CI helpers under `.github/scripts/`.

Fixed in e8274b26155f6e58b1091e821331c0738c2c76f6.

integration_tests/src/cucumber/steps/transfers/public.rs
hansieodendaal: Agreed. Changed the step to express the actual invariant - transferring more than the available balance - and now use `balance + 1` rather than an arbitrary `2x` multiplier. The full Cucumber suite also passes locally at concurrency 6 with this change.

Fixed in e8274b26155f6e58b1091e821331c0738c2c76f6.

.github/scripts/resolve_bedrock_node.py
hansieodendaal: Changed the resolver to first check that the release contains the expected platform asset before resolving its tag to a commit. Added a regression test verifying that the commit resolver is not called when the asset is absent.

Fixed in e8274b26155f6e58b1091e821331c0738c2c76f6.

.github/scripts/resolve_bedrock_node.py
hansieodendaal: Removed the `SHA256SUMS` fallback and now only accept a valid GitHub `asset.digest`. If a matching release does not provide one, we treat that release as unusable and fall back to building the exact Cargo-resolved revision from source.

Fixed in e8274b26155f6e58b1091e821331c0738c2c76f6.

integration_tests/tests/tf_app_deployments.rs
hansieodendaal: Added the upstream references to the ignore reason. The race is tracked in logos-blockchain#3431 and fixed by logos-blockchain#3467; LEZ's currently pinned Logos revision predates that fix, so these can be re-enabled once the dependency advances past it.

Fixed in e8274b26155f6e58b1091e821331c0738c2c76f6.

test_fixtures/src/setup.rs
hansieodendaal: The native and Docker paths exercise different consumers. TF needs a host-native Bedrock executable, while the Docker-backed tests exercise the Compose/runtime setup also used by `just run-bedrock`. The important invariant here is that both consume a node resolved from the same Cargo-selected Logos revision; Docker is no longer an independent version source.

integration_tests/tests/tf_app_deployments.rs
hansieodendaal: Added the upstream references to the ignore reason. The race is tracked in logos-blockchain#3431 and fixed by logos-blockchain#3467; LEZ's currently pinned Logos revision predates that fix, so these can be re-enabled once the dependency advances past it.

Fixed in e8274b26155f6e58b1091e821331c0738c2c76f6.

.github/workflows/ci.yml
hansieodendaal: It covers the resolver logic that enforces the core invariant of this PR: the Bedrock node must correspond to the Cargo-resolved Logos revision. The unit tests cover matching/fallback behavior, while the small live-release check catches changes in the real GitHub release/asset contract that mocks would not. 

Removed the unnecessary dependency on the CI image, so this remains a lightweight standalone check.

hansieodendaal:
> LGTM, thanks. Tbh I'm a bit worried how complex this became and already forgot why we can't just always use pinned docker version

Please see https://github.com/logos-blockchain/logos-execution-zone/pull/793#discussion_r3967359744.

hansieodendaal:
@Arjentix and @moudyellaz, I know you have approved, thanks, but I fixed/answered the minor comments. Happy to resolve those?

PR #798, opened 2026-08-26
refactor(sequencer, storage): move rocksdb communication to `StorageActor`

Arjentix:
## 🎯 Purpose

This is a big refactor of our storage. This PR reimplements rocksdb communication from scratch and moves it to `StorageActor`. Also some protocol types where rethinked.

## ⚙️ Approach

- Added `actor` feature to actor crates which helps dependent actors not to build extra actor and it's dependencies when they need just trait and protocol
- Transaction index is moved from `SequencerCore` to `StorageActor`
- Rocksdb communication is split into different parts:
    - `db.rs`: agnostic wrapper around `rocksdb` crate
    - `entities.rs`: encoding entities (ex. cells)
    - `encoding.rs`: encoding utilities
    - `actor.rs`: high-level management of db
- Some protocol types for `StorageActor` where refactored and cleaned
- Completely removed `storage/src/sequencer/` module
- All tests were ported and added new ones

## 🧪 How to Test

```bash
cargo nextest run --all-features -p sequencer_storage_actor
```

## 🔗 Dependencies

None

## 🔜 Future Work

- Decide what we do with Indexer
- And then refactor indexer part of `storage` crate
- After that remove `storage` crate entirely

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


lez/sequencer/actors/storage/src/actor/db.rs
```diff
@@ -0,0 +1,590 @@
...
+    /// Key the value is stored under.
+    ///
+    /// Key bytes are compared lexicographically to determine order in the column family.
+    type Key: AsRef<[u8]>;
+
+    const COLUMN_FAMILY: C;
+
+    /// Name telling this type apart from the others sharing its column family.
+    #[must_use]
+    fn type_name() -> &'static str {
```
moudyellaz: Could `Storable` carry an explicit `const TYPE_NAME`? `std::any::type_name` isn't stable across compiler versions and silently changes these keys on any crate or module rename.

lez/sequencer/actors/storage/src/actor.rs
```diff
@@ -400,127 +672,130 @@ impl Message<DumpDb> for StorageActor {
...
+            &mut batch,
+            new_withdraw_intents,
+            &consumed_withdrawals,
+        )?;
+
+        // Cross-zone dispatches
+        self.stage_settled_dispatches(&mut batch, finalized_dispatch_records)?;
+
+        // State
+        self.db().put_batch(
```
moudyellaz: dev gated this on `chain_changed`, so a checkpoint-only update did not rewrite the state. Now every idle event serializes the whole `V03State`. The ported test also dropped the `stored_balance` assertion that covered it.

.deny.toml
```diff
@@ -33,6 +33,7 @@ allow = [
   "Apache-2.0 WITH LLVM-exception",
   "Apache-2.0",
   "BlueOak-1.0.0",
+  "0BSD",
```
moudyellaz: Is `0BSD` here for `enum-iterator`? There's a scoped exception for `managed` at line 56, would another one fit better than a blanket allow?

lez/sequencer/actors/storage/src/actor/db.rs
Arjentix: Indeed, this was a real issue. I was too happy with how `std::any::type_name()` is ergonomic. I replaced it now with `TYPE_NAME` and `type_name!()` macro on impl sides which helps a bit to ensure a real type name provided

.deny.toml
Arjentix: Moved to exceptions now, same as `managed`. But AFAU 0BSD should be good for us. But anyway let the legal team deal with that

lez/sequencer/actors/storage/src/actor.rs
Arjentix: Ah, I see now, thanks, fixed

moudyellaz (approved):

lez/sequencer/actors/storage/src/actor/db.rs
```diff
@@ -0,0 +1,627 @@
...
+    key: Vec<u8>,
+    value: Vec<u8>,
+}
+
+impl Dump {
+    /// The single compressed blob a dump ships as.
+    pub fn to_bytes(&self) -> Result<Vec<u8>> {
+        /// Keeps a committed dump small, at a decompression cost that stays
+        /// negligible.
+        const ZSTD_LEVEL: i32 = 19;
```
Pravdyvy: I think there is no reason to store const in this scope. Better to move it into the module scope.

lez/sequencer/actors/storage/src/actor/db.rs
```diff
@@ -0,0 +1,627 @@
...
+    #[test]
+    fn iter_yields_items_in_ascending_key_order() {
+        let dir = tempfile::tempdir().expect("Failed to create temp dir");
+        let db = database_with(&dir, &[277, 0, 500, 1, 2, 256]);
+
+        assert_eq!(
+            collect_ids(db.iter::<Number>()),
+            vec![0, 1, 2, 256, 277, 500]
+        );
+    }
```
Pravdyvy: This test looks a little misleading for me. I mean that we receive sorted list here only because keys support BigEndian encoding. It feels like we need more restriction on keys, something like
```
type Key: AsRef[u8] + Ord + BytewiseOrd;
```
where `BytewiseOrd` is some trait that ensures that `AsRef[u8]` cast is consistent with `Ord`.

lez/sequencer/actors/storage/src/actor/entities.rs
```diff
@@ -0,0 +1,359 @@
...
+
+impl db::ColumnFamily for ColumnFamily {
+    fn options(&self) -> rocksdb::Options {
+        /// Memtable size for the families holding small records, well above
+        /// what they take between flushes.
+        const SMALL_WRITE_BUFFER_SIZE: usize = 4 * 1024 * 1024;
+
+        /// Values above this are large enough to be worth keeping out of the
+        /// LSM tree.
+        const MIN_BLOB_SIZE: u64 = 4 * 1024;
```
Pravdyvy: Again, not sure what meaning constants in this scope have.

lez/sequencer/actors/storage/src/actor/db.rs
```diff
@@ -0,0 +1,627 @@
...
+}
+
+/// Column families affects how data is stored on disk and are used to group data which requires
+/// common configuration.
+pub trait ColumnFamily: Into<&'static str> + enum_iterator::Sequence {
+    /// Options the family is created and opened with.
+    fn options(&self) -> rocksdb::Options {
+        rocksdb::Options::default()
+    }
+}
```
Pravdyvy: Naming a little awkward for RocksDB. By ColumnFamily RocksDB means not a family of columns, family of cells in a column. So this trait, in facs should be named `ColumnFamilies` or `ColumnFamilyKinds`.

lez/sequencer/actors/storage/src/actor/db.rs
Pravdyvy, line 566: This test looks a little misleading for me. I mean that we receive sorted list here only because keys support BigEndian encoding. It feels like we need more restriction on keys in general case if we want to support ordered iteration, something like
```
type Key: AsRef[u8] + Ord + BytewiseOrd;
```
where `BytewiseOrd` is some trait that ensures that `AsRef[u8]` cast is consistent with `Ord`.
Otherwise we can just mark/rename `BigEndian` as sorted key and explicitly state that it is mandatory for ordered iteration.

Pravdyvy (approved):
lgtm. Some design comments.

lez/sequencer/actors/storage/src/actor/db.rs
Arjentix: I've got what you mean now. Yeah, let's name it `ColumnFamilies` since the main idea is to be able to iterate over column families via `enum_iteator::Sequence` impl

lez/sequencer/actors/storage/src/actor/db.rs
```diff
@@ -0,0 +1,627 @@
...
+
+pub trait Storable<C: ColumnFamily>: BorshSerialize + BorshDeserialize {
+    /// Key the value is stored under.
+    ///
+    /// Key bytes are compared lexicographically to determine order in the column family.
+    type Key: AsRef<[u8]>;
```
Arjentix: #

lez/sequencer/actors/storage/src/actor/db.rs
Arjentix: Ah, yes, actually I have a doc about key ordering in `Storable` trait docs:

```rust
    /// Key the value is stored under.
    ///
    /// Key bytes are compared lexicographically to determine order in the column family.
    type Key: AsRef<[u8]>;
```

And in `BigEndian`:

```rust
/// Big Endian numbers encoding.
///
/// Useful for preserving the numeric order of numbers with lexicographical byte order.
```

I'll reword this to `Useful for preserving ascending order`. And also add note about ordering to `iter()` and `Iter`. And also will rework this test to be less inconvenient.

But in general yes, it's expected for user to either iterate in meaningless order or provide a good key which will form an order they want

Issue #799, opened 2026-08-26
fees: cross-zone dispatch fee model (account-less invocations have no payer)


erhant:
Cross-zone dispatches carried by account-less built-in invocations have no fee model. The concrete blocker (moudy, #762 `cross_zone_ping.rs:37`): `ping_sender`'s `Send` is an **unsigned, account-less** invocation — there is no payer account to charge, and the fee guest's fixed two-account arity leaves no room to add one. `ping_sender` is a registered built-in (not a test fixture), so this is a real gap, not test debt.

The integration test `ping_crosses_from_zone_a_to_zone_b` is currently `#[ignore]`d for this reason.

> [!CAUTION]
>
> Under the interim fee policy, system-injected / cross-zone dispatch transactions are fee-exempt (like bridge deposits and other injections), so nothing charges for cross-zone traffic today. That's fine as a placeholder but leaves cross-zone dispatch unpriced.

## Open design questions

- Who pays for a cross-zone dispatch — the emitting user, the sequencer, a zone-level account?
- For account-less invocations (`ping_sender::Send`), where does the payer come from, given the guest's fixed account arity? Does the emitter carry a fee declaration, or is cross-zone traffic priced at a different layer (e.g. the Bedrock inscription)?
- How does this interact with the producer-reward / stake-ownership routing being settled separately?

Needs a spec decision before the ignored cross-zone tests can assert real fee behavior.

erhant:
Interim mitigation (not a fix): as of commit `6417d75db` within #801, empty-witness `ping_sender.Send` is recognized as a system injection in `is_system_injection` and treated **fee-exempt**, purely to keep the cross-zone example tests enabled (`cross_zone_ping, cross_zone_verified, cross_zone_watcher_restart`).

This only exempts the one known `ping_sender.Send` shape — it does not solve this issue. Outbound cross-zone sends are account-less (no payer) and can be initiated by arbitrary programs, which a shape-based classifier can't enumerate.

erhant:
Spoke with @moudyellaz today, we decided to keep cross-zone stuff fee-exempt for now (w.r.t #801) until we come up with a better fix later. So, this issue may persist for a bit.

Issue #797, opened 2026-08-26
fees: screen gossip-ingested transactions through fee admission


erhant:
Fee admission (`sequencer_core::fees::screen`) runs only on the **RPC submission** path (the executor's `Transaction` handler). Gossip-ingested transactions take a different path — `on_gossip_message` in `lez/sequencer/core/src/gossip/network.rs` runs only the **stateless** `evaluate_transaction` (size/parse) and pushes straight to the mempool with `mempool.try_push((TransactionOrigin::Gossip, tx))`, bypassing the fee-admission screen entirely (moudy, #758 `actor.rs:178`).

So every admission arm (`MaxFeeBelowReserve`, `PayerCannotFund`, `MissingFeeDeclaration`, …) is bypassable by peering.

## Not a consensus gap

Admission is advisory anti-spam, not consensus. A gossip-bypassed transaction cannot enter a block invalidly:

- the block builder's `settle_transaction` still reserves/settles each charged tx and **drops** any that cannot pay or that breach a cap;
- the builder's declared-gas pre-screen drops over-cap transactions before execution.

The cost of the bypass is **mempool hygiene**: spam that admission would have turned away at the door can occupy mempool slots until the builder discards it.

## Blocked on: a state handle in the gossip actor

`on_gossip_message` holds only `self.mempool` and a `self.seen` set — it has **no head-state handle**, so it cannot call `screen(&tx, state)` today. Closing this needs a head-state snapshot/handle wired into the gossip network actor (kept current), then a `screen` call gating the `mempool.try_push` on first sight.

## Refs

PR #758 (moudy `actor.rs:178`).

PR #801, opened 2026-08-27
feat: fees

### LINKED ISSUE #796: fees: fold the block-building guard pass into the metered settlement (double execution + gas_limit-unbounded guard)
In the sequencer block builder ([`apply_mempool_transaction`](https://github.com/logos-blockchain/logos-execution-zone/blob/main/lez/sequencer/core/src/lib.rs)), every user/gossip transaction is executed **twice**:

1. a **guard pass** — `tx.validate_on_state(...)`, which executes the program at the **default 32M cycle budget** (`DEFAULT_PUBLIC_CYCLE_BUDGET`) and checks the restricted (clock/faucet/fee) and bridge account rules;
2. the **metered settlement** — `chain_state::apply::settle_transaction`, which executes the action again, capped at the transaction's declared `gas_limit`.

This is already flagged by the `FIXME` above the guard. 

Two things fall out of it:

### 1. The guard ignores `gas_limit` (@moudyellaz also commented this on #754)

The guard runs at the 32M default regardless of the tx's `gas_limit`, so a `gas_limit = 0` transaction buys up to **32M cycles of guard execution while paying only the storage fee** — cheap builder-CPU griefing.

This is **builder-only, not a consensus divergence**: followers verify blocks via `settle_transaction`, which has no guard pass, and the metered `gas_used_exec` only ever accumulates the settlement's cycles (so `gas_limit = 0` contributes nothing to exec gas everywhere). The cost is wasted builder work, not a consensus fault.

### 2. Double execution

Every user/gossip tx executes twice (guard + settlement), doubling per-tx execution cost on the builder.

## Why they're coupled (prerequisite)

The clean fix is to **fold the guard into the single metered settlement** — one execution at `gas_limit`, which also makes the guard naturally honor `gas_limit` and closes (1).

But the guard pass is currently the **only** place the **bridge increase-only rule** (`validate_bridge_account_modification`) runs. The consensus apply path (`settle_transaction`, both the exempt and charged branches) enforces `validate_no_restricted_account_modification` (clock/faucet/fee) but **not** the bridge rule (Copilot, #754 `apply.rs:308`/`405`). 

That is itself a real consensus gap: honest followers don't independently enforce the bridge increase-only rule, so a malicious producer could include a user tx that decreases the bridge balance and followers would accept it. (Exploitability is bounded by the guest's program-ownership rules, so this is defense-in-depth today, but the whole point of the guard is to not rely on the bridge program being bug-free.)

So the guard can't just be deleted — its bridge check has to move into settlement first.

erhant:
## 🎯 Purpose

Introduces the LEZ transaction-fee subsystem end-to-end.

Every non-exempt public transaction now **reserves** its fee up front, is metered under its declared gas limit, and is **refunded** the unspent reserve; per-block revenue is distributed through a dedicated fee program, and a dual base-fee market moves the price each block.

A non-consensus RPC admission door turns away transactions no block could include, and a `getFeeState` quote lets wallets size `max_fee`.

Squashed from the 8-PR stack (#746 → #747 → #750 → #752 → #753 → #754 → #758 → #762) into one change on top of `dev`.

Closes #796 with a single-pass check instead of double-pass.

> [!NOTE]
>
> **Extras:** git-ignore `sequencer_stake_signing_key`, makes kameo log-levels `warn` and above in Justfile and CI (otherwise its very hard to read logs)

### Fee market & types (`lez/programs/fee/core`)

- [x] Add a **dual base-fee market** (`market.rs`): separate exec and storage prices, an EIP-1559-style `next_base_fee` controller bounded to ±1/8 per block, with per-resource targets/caps.
- [x] Add `FeeState` (`state.rs`): the two base fees plus a smoothing window that spreads producer payouts; advanced once per block by `apply_block`.
- [x] Add the reserve/actual formulas (`assess.rs`): `fee_reserve` = `gas_limit·base_fee_exec + gas_stor·base_fee_stor + tip`; `fee_actual_base` clamps executed cycles to the declared limit so `actual + tip ≤ reserve`.
- [x] Add static validity + gas-cap accumulators (`validity.rs`), `BlockFeeSummary`, and the fee program `Instruction` (`Distribute` / `Refund`).

### Fee on the wire (`lee/state_machine`)

- [x] Add `FeeDeclaration { payer, gas_limit, tip, max_fee }` carried as `Message.fee: Option<FeeDeclaration>` (`None` = fee-exempt), covered by the message hash so an ordinary witness signature authorizes it; `is_fee_authorized` scans the witness set for the designated payer.

### Metering & the settlement primitive (`lee/state_machine`)

- [x] Meter public execution against a cycle budget = the transaction's declared `gas_limit` (`program/mod.rs`).
- [x] Add `from_public_transaction_metered`, returning an always-applicable diff: 
    - full effects on **success**
    - or a **nonce-only diff** on **revert** (the charged-but-reverted transaction still burns its replay protection) 
    - `Err` is reserved for a malformed transaction. 

### The fee guest (`lez/programs/fee/src/main.rs`)

- [x] Add `distribute`: market update, drain inbox: base → escrow, tips → producer, pay smoothed payout escrow → producer, summary pinned byte-for-byte
- [x] Add `refund`: return the unspent reserve to the payer.

### Consensus settlement (`lez/chain_state`)

- [x] Add `classify.rs`: charged vs exempt; a non-exempt public transaction that omits its fee is rejected with `MissingFeeDeclaration`.
- [x] Add `settle_charged_transaction`: **reserve → metered action → refund** and have `apply_block_to_state` re-derive the block summary and byte-compare the forced fee transaction. Shared by the sequencer and indexer.

### Sequencer: block building + admission (`lez/sequencer`)

- [x] Settle each mempool transaction in the block builder, budget a block by declared gas (defer-if-doesn't-fit, drop-if-unfittable), append the fee + clock tail.
- [x] Add RPC admission `screen` + `fee_quote`, the `AdmissionRejection` / `FeeStateQuote` wire types, the executor's `Transaction` gate and `GetFeeQuote`, and the `getFeeState` RPC.

### Genesis, block header, and read-side

- [x] Add a `producer` public key to the block header; payouts credit its account, and the header signature verifies against it. Genesis seeds the fee accounts and initializes the producer.
- [x] Rescale public fixture balances to LGO scale (a charged transaction reserves `gas_limit·base_fee` up front). Designate the wallet's first public *signing* account as fee payer, sizing `max_fee` off `DEFAULT_MAX_FEE` / the quote. Propagate fee fields through the indexer protocol, FFI, and explorer.

> [!CAUTION]
>
> **Some tx'es fee-exempt:** 
>
> - Private transactions will be reworked (until a solution is found)
> - Program deployments will be reworked (until this can be merged to Public tx'es)
> - Producer will be reworked
> - System injections (bridge deposits, cross-zone dispatches), genesis transactions.
> - Full vault sweeps pay nothing and contribute nothing to the block gas totals.

### Unit tests

```bash
# Everything CI runs (fee_core, lee, chain_state, common, storage, sequencer):

# Fee-focused subset:
RISC0_DEV_MODE=1 cargo nextest run -p fee_core -p lee -p chain_state -p common -p sequencer_core --all-features
```

### Integration tests

```bash
 # fee-aware transfers, charged-revert e2e
RISC0_DEV_MODE=1 cargo nextest run -p integration_tests -E 'binary(auth_transfer)' 
# fee-exempt staking lifecycle
RISC0_DEV_MODE=1 cargo nextest run -p integration_tests -E 'binary(sequencer_stake_demo)' 
# stakeless reconstruct genesis
RISC0_DEV_MODE=1 cargo nextest run -p integration_tests -E 'binary(sequencer_bootstrap)'  
# fee-exempt cross-zone lock
RISC0_DEV_MODE=1 cargo nextest run -p integration_tests -E 'binary(cross_zone_bridge)'    
```

Committed `fee.bin` and the prebuilt sequencer DB fixture were regenerated with `just build-artifacts`.

### Live demo

One terminal each:

```bash
just run-bedrock      # L1 node (Docker), port 18080
just run-indexer      # port 8779
just run-sequencer    # port 3040
```

Then, in a fourth terminal:

```bash
# Import the two genesis test accounts and claim their vaults:
just wallet-import-test-accounts

# Balances before:
just run-wallet account get --account-id Public/CbgR6tj5kWx5oziiFptM7jMvrQeYY3Mzaao6ciuhSr2r
just run-wallet account get --account-id Public/2RHZhw9h534Zr3eq2RGhQete2Hh667foECzXPmSkGni2

# Charged public transfer of 1000:
just run-wallet auth-transfer send \
  --from Public/CbgR6tj5kWx5oziiFptM7jMvrQeYY3Mzaao6ciuhSr2r \
  --to   Public/2RHZhw9h534Zr3eq2RGhQete2Hh667foECzXPmSkGni2 \
  --amount 1000

# Balances after (wait a block, ~few seconds):
just run-wallet account get --account-id Public/CbgR6tj5kWx5oziiFptM7jMvrQeYY3Mzaao6ciuhSr2r
just run-wallet account get --account-id Public/2RHZhw9h534Zr3eq2RGhQete2Hh667foECzXPmSkGni2
```

Expected: the recipient gains exactly `1000`; the sender is debited `1000 + fee` with `0 < fee ≤ DEFAULT_MAX_FEE`. The current fee quote is visible via the sequencer's `get_fee_state` RPC. (`just clean` resets DBs/keys between runs.)

## 🔗 Dependencies

None.

## 🔜 Future Work

- ~~**#796** — fold the builder's guard pass into the single metered settlement (removes double execution; makes the guard honor `gas_limit`); requires first porting the bridge increase-only guard into the apply path.~~
- **#797** — screen gossip-ingested transactions through admission (needs a head-state handle in the gossip actor; not a consensus gap — the builder still drops invalid gossip txs).
- **#799** — cross-zone dispatch fee model (account-less invocations have no payer).
- ~~**Producer identity** — the header `producer` field / signature is slated for removal in favor of routing rewards to the sequencer's stake-ownership account; being settled separately.~~ #816 implemented this
- Spec TBDs: private-transaction fees, deployment replay protection.
- Implement fee logic in FFI & module-facing parts.

### Migration: public-account registration under fees (FFI consumers)

Registering an unfunded public account via a bare `Initialize` is no longer possible: it is a charged transaction whose only signer (and therefore fee payer) is the fresh account itself, which holds nothing and cannot receive funds before it is claimed. Fee admission refuses it with `PayerCannotFund`, and `wallet_ffi_register_public_account` surfaces that as a non-`Success` error with `transfer_result.success == false`.

**The supported path:** a public account is claimed by its **first funded transfer** — the recipient co-signs, the sender pays the amount plus the fee, and the account is created, funded, and claimed atomically (the `send_claiming_new_account` pattern). 

> [!NOTE]
>
> **Private** registration is unaffected: privacy-preserving transactions are fee-exempt under the interim policy.

**Follow-up:** rework the wallet-FFI surface:
- deprecate/remove `wallet_ffi_register_public_account`'s public-account path
- expose a claiming-transfer entry point over the FFI (sender wallet funds and claims a recipient account in one call), so downstream consumers (e.g. the basecamp wallet module) have a first-class replacement;
- map `PayerCannotFund` to a dedicated FFI error code instead of the generic `NetworkError`, so callers can distinguish "unfundable payer" from transport failures.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


erhant:
# Review guide

Read bottom-up: the **vocabulary** first, then the **consensus rule** that must be correct, then the **builders** that produce blocks to that rule, then the periphery, and the **tests** last.

Most of the line count is tests/fixtures; the consensus-critical surface is a few hundred lines in the middle.

## Skim Through

1. `lez/programs/fee/core/src/assess.rs` — the reserve/actual formulas.
2. `lee/state_machine/src/validated_state_diff/mod.rs` — metering + the revert-as-nonce-only-diff settlement primitive.
3. `lez/chain_state/src/apply.rs` — `settle_charged_transaction` (reserve → action → refund) and `apply_block_to_state`. **This is the heart of the PR.**
4. `lez/programs/fee/src/main.rs` — the guest that moves the money.
5. `lez/programs/fee/core/src/market.rs` — `next_base_fee` (the price controller).

Everything else supports these five.

## Full Review

### 1 — Fee primitives (the vocabulary) · `lez/programs/fee/core/src/`
- **`lib.rs`** — `BlockFeeSummary`, `Instruction` (`Distribute`/`Refund`), fee account PDAs. Start here for the types.
- **`market.rs`** — protocol constants + `next_base_fee`. Check: ±1/8 clamp, the `MAX_GAS·BASE_FEE_MAX` fits-u64 sizing, min/max saturation.
- **`state.rs`** — `FeeState::apply_block` and the payout smoothing window.
- **`assess.rs`** — `fee_reserve` / `fee_actual_base`. Check the clamp keeps `actual + tip ≤ reserve` (a payer is never billed past `gas_limit`).
- **`validity.rs`** — `validate_static_tx`, `FeeError`, gas-cap accumulators.

### 2 — Fee on the wire · `lee/state_machine/src/`
- **`fees.rs`** — `FeeDeclaration`, `is_fee_authorized` (payer's signature only, id-checked before verify), `SignedMessage`.
- **`public_transaction/message.rs`** — `fee: Option<FeeDeclaration>`, `try_new_with_fees`, and the hash-pinning tests (any layout change trips them).

### 3 — Metering & the settlement primitive · `lee/state_machine/src/`
- **`program/mod.rs`** — cycle budget = declared `gas_limit`; the `Cycles` alias.
- **`validated_state_diff/mod.rs`** — the key file: `from_public_transaction_metered`
  (Ok = effects-or-nonce-only diff, Err = malformed), `from_fee_settlement_invocation`,
  `execute_authorized`. Check: revert still bumps the signer nonces; auth failure
  is a hard reject; `apply_state_diff` remains the only mutator.

### 4 — The fee guest (ZK program) · `lez/programs/fee/src/main.rs`
- `distribute` (market update + inbox drain + smoothed payout, summary pinned
  byte-for-byte, conservation `inbox == revenue_total`) and `refund`.

### 5 — Consensus settlement (the heart) · `lez/chain_state/src/`
- **`classify.rs`** — charged/exempt + `MissingFeeDeclaration` (no fee-evasion by
  omission).
- **`apply.rs`** — `settle_charged_transaction` (reserve → metered action →
  refund; drop vs revert on opposite sides of the line), `apply_block_to_state`
  (re-derives the summary, byte-compares the forced fee tx), `derive_block_summary`.
  Shared by sequencer and indexer — that's what keeps build and replay in lockstep.
- **`ingest_error.rs`** — the rejection taxonomy.

### 6 — Genesis / header / seeding glue
- **`lez/common/src/transaction.rs`** — `fee_invocation`,
  `fee_reserve_invocation`, `fee_refund_invocation`, `initialize_producer_account`,
  the restricted/bridge guards, `is_system_injection`.
- **`lez/common/src/block.rs`** — the header `producer` field +
  `has_valid_producer_signature`.
- **`lez/system_accounts/src/lib.rs`**, **`lez/testnet_initial_state/src/lib.rs`** —
  fee account ids; LGO-scale genesis balances.

### 7 — Sequencer: block building + admission · `lez/sequencer/`
- **`core/src/lib.rs`** — the block-building loop: opening fee state, declared-gas
  budgeting (defer/drop), the producer + fee/clock tail.
- **`core/src/fees.rs`** — admission `screen` (runs the same checks as settlement,
  plus payer affordability) and `fee_quote`. Non-consensus / anti-spam.
- **`service/protocol/src/fees.rs`** — `AdmissionRejection` (stable error codes),
  `FeeStateQuote`.
- **`actors/executor/src/actor.rs`** — the `Transaction` admission gate +
  `GetFeeQuote`. **`actors/rpc_server/.../service.rs`** — `getFeeState` RPC +
  rejection rendering.

### 8 — Read-side (skim) · indexer / explorer / wallet
- `lez/indexer/service/protocol/src/convert.rs`, `indexer/core/src/block_store.rs`, `indexer/ffi/*` — fee fields through the indexer protocol + FFI.
- `lez/explorer_service/*` — fee display.
- `lez/wallet/src/account_manager.rs` (`fee_payer_account_id`) + `lib.rs` (`DEFAULT_MAX_FEE`) — the wallet builds fee-aware transactions.

### 9 — Tests & fixtures (the bulk; verify last)
- Unit tests colocated with each module above.
- `integration_tests/tests/*` (~14 files) — fee-aware assertions; the notable pins are admission rejections (`auth_transfer/public.rs`), dropped-vs-reverted, and the tps/wallet_ffi fixes.
- `test_fixtures/` + committed `fee.bin` / `prebuilt_sequencer_db.dump` — regenerated artifacts.

## What to scrutinize hardest

- **`apply.rs` settlement** — the reserve/action/refund conservation and the drop-vs-revert split (a precondition failure invalidates the block; an action failure reverts and keeps the fee).
- **`from_public_transaction_metered`** — the nonce-only revert diff is what makes a charged-but-reverted transaction non-replayable.
- **build vs replay parity** — both go through `chain_state::apply`, but the builder additionally runs a guard pass (see #796) and the declared-gas budget.
- **admission is advisory** — nothing consensus rests on it; the apply path is authoritative (relevant to the gossip-bypass follow-up #797).

lee/state_machine/src/validated_state_diff/mod.rs
```diff
@@ -54,35 +54,221 @@ impl ValidatedStateDiff {
...
+            Self(StateDiff {
+                signer_account_ids: signers,
+                public_diff: HashMap::new(),
+                new_commitments: Vec::new(),
+                new_nullifiers: Vec::new(),
+                program: None,
+            })
+        });
+        (ExecutionOutcome { cycles }, Ok(diff))
+    }
```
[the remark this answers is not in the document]
erhant: First part of the comment is obviously wrong, `result` is just "used" there in a pattern-match, the macro does not move it.

As for the error handling & charging, I will add `is_chargeable` function to `LeeError` to fix it

lez/common/src/test_utils.rs
```diff
@@ -63,6 +63,12 @@ pub fn produce_dummy_block(
...
 ) -> Block {
+    transactions.push(LeeTransaction::Public(fee_invocation(
+        fee_core::BlockFeeSummary::default(),
+        lee::AccountId::from(&lee::PublicKey::new_from_private_key(
+            &sequencer_sign_key_for_testing(),
+        )),
+    )));
     transactions.push(LeeTransaction::Public(clock_invocation(
         id.saturating_mul(100),
     )));
```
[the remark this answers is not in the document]
erhant: Done in https://github.com/logos-blockchain/logos-execution-zone/pull/801/commits/357d85a163ca246d669d6eacf64613520efe544c

lee/state_machine/src/validated_state_diff/mod.rs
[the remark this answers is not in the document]
erhant: Done in https://github.com/logos-blockchain/logos-execution-zone/pull/801/commits/357d85a163ca246d669d6eacf64613520efe544c

lee/state_machine/src/error.rs
```diff
@@ -81,6 +84,17 @@ pub enum LeeError {
...
 }
 
+impl LeeError {
+    /// Whether this error should be charged for rather than reject the block.
+    #[must_use]
+    pub const fn is_chargeable(&self) -> bool {
+        // Only structural defects a correct proposer would never include reject
+        // the block; every genuine action failure is charged and reverted.
+        // Auth/nonce failures never reach here — they bail before execution.
+        !matches!(self, Self::InvalidInput(_) | Self::OutOfValidityWindow)
```
moudyellaz: `OutOfValidityWindow` is raised on `program_output` after the guest runs, so a program that emits an expired window gets free execution: dropped uncharged, no nonce burned, replayable. should this split on "detectable before execution" instead?

lez/chain_state/src/apply.rs
```diff
@@ -153,21 +185,274 @@ pub fn apply_block_to_state(block: &Block, state: &mut V03State) -> Result<(), B
...
+        FeeClass::Exempt => {
+            let diff = transaction
+                .compute_state_diff(state, block_id, timestamp)
+                .map_err(|err| BlockIngestError::StateTransition {
+                    tx_index,
+                    reason: format!("{:#}", anyhow::Error::from(err)),
+                })?;
+            // The builder guards user transactions off the restricted accounts;
+            // the apply path must too, or a block author could drain them.
+            validate_no_restricted_account_modification(state, &diff).map_err(|err| {
```
moudyellaz, line 350: exempt txs no longer get the bridge increase-only guard on either path, but dev applies it to every user tx via `validate_on_state` in the builder. revert looks right for injected deposits (they legitimately debit the escrow), so scope it to private/deployment rather than drop it?

lez/common/src/transaction.rs
```diff
@@ -88,22 +88,14 @@ impl LeeTransaction {
...
-            .into_iter()
-            .chain(std::iter::once(system_accounts::faucet_account_id()));
-        for account_id in restricted_modification_accounts {
-            validate_doesnt_modify_account(state, &diff, account_id)?;
-        }
-
-        self.validate_bridge_account_modification(state, &diff)?;
-
+        validate_no_restricted_account_modification(state, &diff)?;
+        validate_bridge_account_modification(state, &diff, matches!(self, Self::Public(_)))?;
```
moudyellaz, line 92: `validate_on_state` is test-only now, so those two tests assert guards no live path calls. the restricted half is re-implemented in settlement, the bridge half isn't. point the tests at settlement instead?

lez/wallet/src/account_manager.rs
```diff
@@ -447,6 +447,21 @@ impl AccountManager {
             .collect()
     }
 
+    /// The account that pays this transaction's fee: the first public signing
+    /// account. Its ordinary signature covers the message, so it is fee-authorized
+    /// without a separate fee witness. Non-signing public accounts (`sk: None`)
+    /// are skipped — an unsignable account could never authorize the fee.
+    pub fn fee_payer_account_id(&self) -> Option<AccountId> {
+        self.states.iter().find_map(|state| match state {
```
moudyellaz: payer is picked by position with no balance check, so a wallet whose first signing account is unfunded can't transact at all even with funds in a later one. pick the first that can fund the reserve?

integration_tests/tests/auth_transfer/public.rs
```diff
@@ -100,41 +119,65 @@ pub async fn successful_transfer_to_new_account() -> Result<()> {
...
 
-    let failed_send = wallet::cli::execute_subcommand(ctx.wallet_mut(), command).await;
-    assert!(failed_send.is_err());
+    let sender = ctx.existing_public_accounts()[0];
+    let receiver = ctx.existing_public_accounts()[1];
+    let sender_before = account_balance(&ctx, sender).await?;
+    let receiver_before = account_balance(&ctx, receiver).await?;
+    let sender_nonce_before = get_account(&ctx, sender).await?.nonce.0;
+
+    // Submit through the program facade, which (unlike the CLI's `Send`) runs
```
moudyellaz: ran this locally: fails at line 155 with `InsufficientFundsError`, not at the assertions — the facade does have a balance pre-check (`native_token_transfer/mod.rs:31`), so it never reaches the sequencer. combined with no settlement-level test for a charged revert (`is_chargeable` appears in no test outside `error.rs`), revert-with-fee has no working end-to-end coverage.

lez/chain_state/src/classify.rs
```diff
@@ -0,0 +1,92 @@
...
+/// `is_genesis` covers the genesis block's config/supply transactions. The
+/// forced fee and clock transactions never reach this classifier: the
+/// transition strips them positionally before the user-transaction loop.
+///
+/// # Errors
+///
+/// [`ClassifyError::MissingFeeDeclaration`] if a user public transaction that is
+/// not an exempt shape omits its fee, and [`ClassifyError::Unserializable`] if a
+/// charged transaction cannot be serialized to price its storage gas.
+pub fn classify(
```
schouhy, line 52: if a user submits a bridge tx with an empty witness set. won't it get through? Looks like it will be classified as fee exempt and those dont guard the bridge

lee/state_machine/src/error.rs
erhant: decided to charge validity window error as well in https://github.com/logos-blockchain/logos-execution-zone/pull/801/commits/752ef180ae8a591401d800fffe3dbdaed1605238

integration_tests/tests/auth_transfer/public.rs
erhant: added an e2e test in https://github.com/logos-blockchain/logos-execution-zone/pull/801/commits/752ef180ae8a591401d800fffe3dbdaed1605238 called `a_charged_action_that_reverts_is_charged_not_block_rejected`

lez/chain_state/src/apply.rs
erhant: Scoped it in https://github.com/logos-blockchain/logos-execution-zone/pull/801/commits/5c4bd7ed0ec1aa8ce03932e6ec846bef64674279

lez/chain_state/src/classify.rs
erhant: turns out this is a pre-existing bridge-feature gap, not something from this PR.

there's no L1 verification of a deposit's `l1_deposit_op_id`: the receipt PDA only enforces exactly-once, and the empty-witness deposit self-authorizes via the bridge PDA seed (no signature). 

meaning that a forged deposit is byte-identical to a real one, and the fee classification is neutral to it.

an increase-only guard can't distinguish them — legit deposits debit the escrow too — so I scoped the exempt-branch guard to **private/deployment** (which never legitimately touch the bridge) as a partial mitigation (https://github.com/logos-blockchain/logos-execution-zone/pull/801#discussion_r3879378505), and left deposit authenticity to L1 verification, which is the real fix.

filed this one at #809

lez/wallet/src/account_manager.rs
erhant: fixed in https://github.com/logos-blockchain/logos-execution-zone/pull/801/commits/faa66a64a6e71ea48876d89ba9229c7d44ec9192

indeed it will now return the first one that has some balance, otherwise `None`

lez/common/src/transaction.rs
erhant: implemented in https://github.com/logos-blockchain/logos-execution-zone/pull/801/commits/faa66a64a6e71ea48876d89ba9229c7d44ec9192

lez/chain_state/src/classify.rs
erhant: implemented more fixes in https://github.com/logos-blockchain/logos-execution-zone/pull/801/commits/faa66a64a6e71ea48876d89ba9229c7d44ec9192 too

lee/state_machine/src/fees.rs
```diff
@@ -0,0 +1,165 @@
...
+    fn signing_hash(&self) -> [u8; 32];
+    /// The designated fee payer, or `None` for a fee-exempt message.
+    fn payer(&self) -> Option<AccountId>;
+}
+
+/// Whether the transaction's fee is authorized: a fee-exempt message always is;
+/// a charged message requires a valid signature by its designated payer.
+///
+/// Scans for the payer's signature directly — checking the account id before
+/// verifying the signature, so only the payer's signature is verified, not
```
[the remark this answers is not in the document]
erhant: Don't care about sponsored payments really

lez/wallet/src/account_manager.rs
erhant: relaxed a bit in https://github.com/logos-blockchain/logos-execution-zone/pull/801/commits/d3a866e25c5269b8498185fc2fc25acd44c61b65

it failed some existing tests, so now if none are funded this returns the first account regardless

integration_tests/tests/auth_transfer/public.rs
```diff
@@ -10,19 +10,37 @@ use lee::{PublicKey, public_transaction};
...
     account::Label,
     cli::{
         CliAccountMention, Command, SubcommandReturnValue, account::AccountSubcommand,
         programs::native_token_transfer::AuthTransferSubcommand,
     },
+    program_facades::native_token_transfer::NativeTokenTransfer,
 };
 
+/// The sender's post-transfer balance is `before - amount - fee` where the
+/// fee depends on metered cycles, so sender assertions are relative (a
```
jonesmarvin8: Is it necessary to mention metered cycles in this function? This function just asserts that positive fees were paid.

lez/indexer/ffi/src/api/types/transaction.rs
```diff
@@ -65,6 +65,9 @@ impl From<Box<FfiPublicTransactionBody>> for PublicTransaction {
                     std_vec.into_iter().map(Into::into).collect()
                 },
                 instruction_data: value.message.instruction_data.into(),
+                // FIXME: The FFI surface does not carry the fee yet; reconstruct
```
jonesmarvin8, line 68: It doesn't appear that the description of the PR notes that `ffi` exempts public transactions from fees.

lee/state_machine/src/validated_state_diff/mod.rs
```diff
@@ -55,35 +55,221 @@ impl ValidatedStateDiff {
...
 
+    /// Shared execution core: validates and executes one program invocation
+    /// (with its chained calls), producing a diff. `authorized` is the guest's
+    /// `is_authorized` set; `nonce_bearers` become the diff's `signer_account_ids`
+    /// (their nonces advance on apply).
+    #[expect(
+        clippy::too_many_arguments,
+        reason = "the execution core threads the full invocation context"
+    )]
+    fn execute_authorized(
```
jonesmarvin8, line 243: Is this preferable to receive `message` components separately rather than bundled as a `message`?

lee/state_machine/src/validated_state_diff/mod.rs
```diff
@@ -55,35 +55,221 @@ impl ValidatedStateDiff {
...
+        block_id: BlockId,
+        timestamp: Timestamp,
+        cycle_budget: u64,
+        cycles_used: &mut u64,
     ) -> Result<Self, LeeError> {
         let signer_account_ids = authenticate_public_transaction_signers(tx, state)?;
         let message = tx.message();
+        // Signers both authorize the execution and advance their replay nonces.
+        let authorized: HashSet<AccountId> = signer_account_ids.iter().copied().collect();
+        Self::execute_authorized(
```
jonesmarvin8, line 221: It seems to me that we could pass `tx` to `execute_authorized` and then unpack `signer_account_ids` and `message` there?

jonesmarvin8: Appears that from_fee_settlement_invocation is the only call of this function that doesn't have the `tx`. So, that is likely why, but looks strange.

lee/state_machine/src/validated_state_diff/mod.rs
```diff
@@ -55,35 +55,221 @@ impl ValidatedStateDiff {
...
+        program_id: ProgramId,
+        account_ids: &[AccountId],
+        instruction_data: &[u8],
+        authorized: &HashSet<AccountId>,
+        state: &V03State,
+        block_id: BlockId,
+        timestamp: Timestamp,
+    ) -> Result<Self, LeeError> {
+        let mut cycles_used = 0; // dont care
+        Self::execute_authorized(
```
jonesmarvin8, line 195: Perhaps we could set up `tx` here to make the code look slightly cleaner on `execute_authorized` call?

integration_tests/tests/auth_transfer/public.rs
```diff
@@ -189,37 +235,41 @@ async fn two_consecutive_successful_transfers() -> Result<()> {
...
-        account.program_owner,
-        programs::authenticated_transfer().id().into()
+    let refused = wallet::cli::execute_subcommand(ctx.wallet_mut(), command).await;
+    let err = refused
+        .expect_err("an unfunded account must not be able to pay for its own initialization");
+    // Pin the specific rejection (PayerCannotFund, code -31_984): a bare
+    // `is_err()` would pass equally on a network or wallet-build failure.
+    assert!(
+        err.to_string().contains("but the fee reserve is"),
+        "expected a PayerCannotFund rejection, got: {err}",
```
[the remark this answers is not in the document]
erhant: Fixed in https://github.com/logos-blockchain/logos-execution-zone/pull/801/commits/0dcae85910dd2ee8ff14f9c1c82c135bc4250509

integration_tests/tests/auth_transfer/public.rs
erhant: I think it wouldn't hurt to mention that we can't know the fee ahead of time exactly, otherwise we could have like `before, after, amount_sent, fees` as parameters and check that `before - amount_sent - fee == after`

reworded slightly in https://github.com/logos-blockchain/logos-execution-zone/pull/801/commits/6c5f4e5419dc9544d04e5368710c972f067b5712

lez/indexer/ffi/src/api/types/transaction.rs
erhant: Added to future work in PR desc

lee/state_machine/src/validated_state_diff/mod.rs
erhant: > Appears that from_fee_settlement_invocation is the only call of this function that doesn't have the tx.

this is the heart of this PR really, within the `feeReserve - action - feeRefund` triplet from the user, we either have 2 separate tx'es that should always be bundled together (and have separate "own worlds", tx hashes etc.) or we embed the `reserve` as a core logic within state transition, as is done here

I think we should not treat it as a bogus-tx to be unpacked for the sake of preserving the old code, it is a transition logic on its own and should be treated with the fine-granularity as done here

erhant: normally we could have baked the fee logic into the protocol itself to not resort to tricks we do here, but time constraints..

lee/state_machine/src/validated_state_diff/mod.rs
erhant: would like to keep it as is, as per https://github.com/logos-blockchain/logos-execution-zone/pull/801/changes#r3903746485

lee/state_machine/src/validated_state_diff/mod.rs
erhant: would like to keep as is, as per https://github.com/logos-blockchain/logos-execution-zone/pull/801/changes#r3903746485

lez/storage/src/indexer/mod.rs
```diff
@@ -223,71 +220,12 @@ impl RocksDBIO {
...
-        })?;
-
-    Ok(())
+fn apply_block_transactions(block: &Block, state: &mut V03State) -> DbResult<()> {
+    // The indexer replays through the same transition the sequencer and
+    // validators use, so the fee settlement arithmetic exists exactly once.
+    chain_state::apply::apply_block_to_state(block, state)
+        .map(drop)
+        .map_err(|err| DbError::db_interaction_error(format!("block replay failed: {err}")))
 }
```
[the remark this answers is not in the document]
erhant: consensus invariants are already enforced at ingestion, not skipped. `apply_block_transactions` is only reached from state reconstruction, which replays blocks already in the DB; they are already validated.

indeed they can be corrupted, but RocksDB has its own block checksums to guard against that; and it would not be a "defense-in-depth" that is worth doing to do all the hash recomputation & signature checks for block replays from db

will keep as is

lee/state_machine/src/validated_state_diff/mod.rs
```diff
@@ -55,35 +55,221 @@ impl ValidatedStateDiff {
...
+            timestamp,
+            cycle_budget,
+            &mut cycles_used,
+        );
+        let cycles = if matches!(result, Err(LeeError::OutOfGas { .. })) {
+            // if it ran out of gas, we can charge the entire usage
+            cycle_budget
+        } else {
+            cycles_used
+        };
```
schouhy: if `result` returns a different `Err`, this would fallback to use `cycles_used`. Shouldn't an execution error pay the full reserve?

lez/common/src/transaction.rs
```diff
@@ -257,6 +215,232 @@ pub fn clock_invocation(timestamp: clock_core::Instruction) -> lee::PublicTransa
...
+#[must_use]
+pub fn fee_reserve_invocation(payer: AccountId, amount: u128) -> lee::public_transaction::Message {
+    lee::public_transaction::Message::try_new(
+        programs::authenticated_transfer().id(),
+        vec![payer, system_accounts::fee_inbox_account_id()],
+        vec![],
+        authenticated_transfer_core::Instruction::Transfer { amount },
+    )
+    .expect("Fee reserve message should always be constructable")
+}
```
schouhy, line 406: This ties the fee to the specific auth-transfer program. A pattern like in the stake program could be implemented, where the tx carries the program id and the instruction that will move the funds. The stake program has a builtin check to assert the funds arrive. That can be avoided here, the lee state checking that itself.

schouhy: not a blocker for this PR though

lee/state_machine/src/error.rs
```diff
@@ -81,6 +84,21 @@ pub enum LeeError {
...
+    /// rather than rejecting the block.
+    ///
+    /// - A malformed transaction is caught before the program runs, costs no cycles, and rejects
+    ///   the block.
+    /// - Auth and nonce failures never reach here; they bail before execution.
+    /// - Every other failure that surfaces during or after execution, so it is charged and
+    ///   reverted: the payer pays and the nonce advances.
+    #[must_use]
+    pub const fn is_chargeable(&self) -> bool {
+        !matches!(self, Self::InvalidInput(_))
```
schouhy: an invalid program id found during execution on a chained call produces `InvalidInput`. I think those should be chargable

lez/sequencer/core/src/lib.rs
```diff
@@ -99,6 +100,43 @@ type FoundingStake = (
...
+    /// never be included in any block, so it must be dropped rather than
+    /// deferred (a deferral would repeat for ever).
+    const fn fits_empty(view: &fee_core::assess::FeeTxView) -> bool {
+        view.gas_limit() <= fee_core::market::MAX_GAS_EXEC
+            && view.gas_stor() <= fee_core::market::MAX_GAS_STOR
+    }
+
+    /// Adds a contribution [`Self::fits`] has already cleared.
+    const fn add(&mut self, view: &fee_core::assess::FeeTxView) {
+        self.exec = self.exec.saturating_add(view.gas_limit());
```
schouhy: shouldn't this accumulate the used gas (the gas that was charged) instead of the user declared gas limit? A user sending a tx with a high gas limit for a tiny execution that succeeds could end up taking the full capacity of the block no?

schouhy (commented):
Hey, thanks for this huge effort! I'm leaving some comments here while I finish reviewing

lez/sequencer/core/src/lib.rs
```diff
@@ -923,45 +961,78 @@ impl<BP: BlockPublisherTrait, S: StorageActorTrait> SequencerCore<BP, S> {
...
             // Gossiped transactions arrive from untrusted peers, same as
             // user-submitted ones, so they get the same full state validation.
             TransactionOrigin::User | TransactionOrigin::Gossip => {
-                let validated_diff = match tx.validate_on_state(state, block_height, timestamp) {
-                    Ok(diff) => diff,
+                // Settle on a scratch clone: settlement is the single validation
+                // and charging pass (restricted- and bridge-account guards
+                // included), so a transaction that cannot pay, breaches a cap, or
+                // touches a restricted account is dropped, never included.
+                let mut scratch = state.clone();
```
schouhy, line 1009: expensive clone here

lez/sequencer/actors/rpc_server/src/actor/service.rs
schouhy: we probably need to raise this

lee/state_machine/src/error.rs
schouhy: what about host-side errors? could those end up being chargable? (sequencer runs out of memory to run the guest or stuff like that)

lez/sequencer/core/src/lib.rs
erhant: we sadly have this clone in multiple places, in particular in `IndexerStore.accept_block` we have it with a TODO comment too: 

```rs
// TODO: we use scratch state to be atomic, but need to revisit how expensive a clone is
let mut scratch = self.current_state.read().await.clone();
```

we shall revisit the entire `pre/post` usage to get rid of the clones perhaps, in a different PR

lez/common/src/transaction.rs
erhant: Will add a note in the code there to see if we can do it like the stake program, but the actual change can come with a later PR if that is ok 🙏🏻

lez/sequencer/actors/rpc_server/src/actor/service.rs
erhant: raised here https://github.com/logos-blockchain/logos-execution-zone/pull/801/commits/744d2a6c0f6cbfea9e966d400d4e13cc5f06c02e

also found places where the constant was not used correctly; now I grouped them at a common place & also respects the mantle max block size

lee/state_machine/src/error.rs
erhant: good catch, added an extra catch about the invalid program id (https://github.com/logos-blockchain/logos-execution-zone/pull/801/commits/744d2a6c0f6cbfea9e966d400d4e13cc5f06c02e)

lee/state_machine/src/validated_state_diff/mod.rs
erhant: actually yeah, it makes the most sense to charge the budget on any error (done in https://github.com/logos-blockchain/logos-execution-zone/commit/744d2a6c0f6cbfea9e966d400d4e13cc5f06c02e)

lez/common/src/transaction.rs
schouhy: sure! :pray:

lez/sequencer/core/src/lib.rs
schouhy: sounds good

lez/sequencer/core/src/lib.rs
erhant: good point!

fixed in https://github.com/logos-blockchain/logos-execution-zone/pull/801/commits/cee8b69c7690068bff78a3be554334cd91d96134

the admission pre-check still adds the candidate's declared limits on top (to check the worst-case), but after each settlement the budget syncs to the actually charged totals

now a HUGE declaration only occupies the block during its own settlement, not for the rest of the packing pass. (which was your observation)

lez/common/src/transaction.rs
erhant: added a small comment in https://github.com/logos-blockchain/logos-execution-zone/pull/801/commits/cee8b69c7690068bff78a3be554334cd91d96134

lee/state_machine/src/error.rs
schouhy: thanks! what about the host side errors? does that make sense?

erhant: ah yes, i was going to open an issue about it; looking at it a bit more

the only host-side error that can charge something is when r0vm fails during execution but is caught like a "program execution failed"; kind of like a guest panic

this will be charged (because right now we can't differentiate such an error), but it will cause a problem on consensus when a followed sees this "failing" tx; but it actually passes

the fix is about differentiating r0-side errors, and this issue goes back all the way back when we were implementing recovery for the indexer; where a mis-timed CTRL+C caused r0 to fail during block ingestion, and made it seem like guest panicked, and wrote a faulty state to disk

we have since fixed that graceful termination issue, and other types of issues like OOM kill the process anyways so they wont bite us I think, the only edge case is r0vm failing for some reason (we havent seen that happen)

erhant:
```rs
impl BlockIngestError {
    /// Whether the failure may be transient rather than a property of the block.
    ///
    /// FIXME: `StateTransition` is too coarse — its `reason` string mixes genuine
    /// state-transition rejections with infra failures (risc0 executor teardown,
    /// storage errors). Once it carries a structured cause, narrow this so only
    /// infra failures retry.
    #[must_use]
    pub const fn is_retryable(&self) -> bool {
        matches!(self, Self::StateTransition { .. })
    }
}
```

this was the thing we added for indexer, so that a r0vm failure can be recovered from

erhant: opened https://github.com/logos-blockchain/logos-execution-zone/issues/820

schouhy: thanks!

schouhy (approved):
LGTM 🚀

lee/state_machine/src/validated_state_diff/mod.rs
jonesmarvin8: Reasonable.

integration_tests/tests/auth_transfer/public.rs
jonesmarvin8: Looks good.

moudyellaz (approved):

PR #805, opened 2026-08-27
feat(cross-zone)!: lifetime mint caps per peer source, dead-letter requeue

moudyellaz:
## 🎯 Purpose

A compromised or buggy peer zone can currently mint unbounded wrapped value: the wrapped token authorizes sources but nothing bounds how much each source may mint. This PR adds a per-source lifetime mint cap enforced in the wrapped-token guest, and gives operators a way back for deliveries that failed while a cap (or any transient cause) was in the way: dead letters now retain the transaction and can be requeued over RPC.

## ⚙️ Approach

- `WrappedTokenConfig.sources` becomes `Vec<SourceEntry>` carrying `mint_cap: Option<Balance>` and a guest-owned `minted` counter; `UpdateSources` takes `SourcePolicy` entries (no counter), and the guest carries counters over on update so spent allowance cannot be reset. A source removed and re-added starts at zero; `mint_cap: Some(0)` is the pause idiom.
- The guest enforces the cap on `Mint` with a checked counter written back as a config post-state; a breach fails the whole delivery, so the message stays unseen and redeliverable after a cap raise.
- `CrossZoneRoute` gains `mint_cap` (`serde(default)`, `deny_unknown_fields`); genesis refuses a cap on a non-minting target, a cap with no `source_authority` configured, and duplicated source pairs.
- `DeadLetterDispatchRecord` retains the encoded transaction (bounded by `MAX_REQUEUEABLE_DISPATCH_BYTES`, 64 KiB; larger deliveries stay listed without bytes). New `requeueCrossZoneDeadLetter` RPC restores a dead letter to the pending list with a clean attempt count, returning `requeued`, `already_pending`, `not_found`, or `not_retained`. Stores holding the old record layout drop the retained list on open, keeping the lifetime count.

Breaking changes (also in commit footers):
- `WrappedTokenConfig` gains per-source `mint_cap` and `minted` fields and `UpdateSources` now carries `SourcePolicy` entries, changing both encodings (existing zone state needs regenesis).
- The stored dead-letter cell layout (migrated on open by dropping old-layout lists).
- `CrossZoneRoute` config keys are now strict; unknown keys fail startup.

Known limitation: `requeueCrossZoneDeadLetter` is unauthenticated like the rest of the RPC surface; the blast radius is bounded to re-attempting deliveries this node already accepted from a peer.

## 🧪 How to Test

- `RISC0_DEV_MODE=1 cargo test --release -p integration_tests --test cross_zone_state_machine` covers the cap paths: at-cap, over-cap, accumulation, uncapped, replay, counter carry-over, re-add, refused-then-redelivered, cross-source independence.
- `RISC0_DEV_MODE=1 cargo test --release -p storage` covers requeue outcomes, the cap-full refusal, byte-bound retention, and the layout migration.
- `RISC0_DEV_MODE=1 cargo test --release -p cross_zone -p sequencer_core --features mock` covers the genesis guards and route config parsing.

## 🔜 Future Work

None.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


lez/programs/wrapped_token/core/src/lib.rs
```diff
@@ -80,9 +86,37 @@ pub struct WrappedTokenConfig {
...
+    /// The peer zone the emission came from.
+    pub src_zone: ZoneId,
+    /// The program on that zone that emitted it.
+    pub src_program_id: ProgramId,
+    /// Lifetime mint allowance for this source; `None` is uncapped.
+    ///
+    /// Lifetime rather than windowed: the guest has no clock in the dispatch
+    /// account list, and the authority can raise the cap as honest volume
+    /// grows, which serves the same purpose without new plumbing.
+    pub mint_cap: Option<Balance>,
```
agureev, line 107: Why doesn't SourceEntry just carry a SourcePolicy inside? Seems repetitive

integration_tests/tests/cross_zone_state_machine.rs
```diff
@@ -91,7 +124,17 @@ fn seed_wrapped_config_with_governance(
         minter: programs::cross_zone_inbox().id(),
         governance,
         authority,
-        sources,
+        sources: sources
```
agureev: Can't this just `uncapped_entries`?

agureev (changes requested):

lez/programs/wrapped_token/core/src/lib.rs
moudyellaz: Done in https://github.com/logos-blockchain/logos-execution-zone/commit/1ce6eae051b020197b71501186f82d2e37c52685

integration_tests/tests/cross_zone_state_machine.rs
moudyellaz: Done in https://github.com/logos-blockchain/logos-execution-zone/commit/ea79bb1cdd0778530960d23fa81b3e6bf49137e8

agureev (approved):

lez/storage/src/sequencer/mod.rs
```diff
@@ -385,9 +412,28 @@ impl RocksDBIO {
...
 
+    /// Drops a retained dead-letter list whose layout predates the retained
+    /// transaction bytes.
+    ///
+    /// Old records carried only the transaction size, so there is nothing to
+    /// requeue and no lossless rewrite; the lifetime give-up count lives in its
+    /// own cell and survives. Without this, the first read of the old blob
+    /// errors, and one of those sits on the follow sink's settlement path,
+    /// where a persist failure halts the node.
+    fn drop_undecodable_dead_letters(&self) -> DbResult<()> {
```
erhant, line 427: this technically drops when it can't read, not just decode

I doubt it matters much because:
- its unlikely to error here at start
- its unlikely to be able to `del` as well if you can't even `get_opt`

but we maybe more strict to delete only on decode errors if we want to

erhant (approved):
lgtm, one small nit!

lez/storage/src/sequencer/mod.rs
moudyellaz: Done in https://github.com/logos-blockchain/logos-execution-zone/commit/92b50b2a0ecc967bed39372960dee6674847e426

PR #806, opened 2026-08-27
feat(cross-zone): committee floor, suspend a peer below its minimum accredited committee

moudyellaz:
## 🎯 Purpose

A peer zone whose sequencer committee has shrunk is cheaper to attack: fewer keys need to collude to sign a forged peer block. Nothing today bounds how small a peer's committee may get before this zone keeps consuming it. This PR lets each zone declare, per peer, the minimum live accredited committee it will accept, and stops consuming a peer that falls below it until the committee recovers.

## ⚙️ Approach

- A shared `CommitteeFloorState` latch in the `cross_zone` crate, folded once per pass by both consumers: fail closed before the first successful read, keep the last known size for a bounded run of read failures then fail closed again, a floor of 0 never suspends.
- Sequencer watcher: reads the peer's live accredited key set before each pass and suspends recording when it is below the floor, with a per-peer gauge `cross_zone_peer_committee_suspended` and error-level transition logs on the alert cadence. Deliveries already recorded still drain into blocks, since production drains the store independently of the watcher.
- Indexer verifier: the same gate on `read_peer`, reporting the peer `Suspended` and clearing the absence-evidence run so a stale verdict can never escalate. Health ordering becomes Halted, Suspended, Holed, Live, Lagging.
- Config: `CrossZonePeer.min_committee_size: u32` with `serde(default)`.

Breaking change (also in the commit footer): `getStatus`'s `cross_zone_peers[].health` can now be `Suspended`. Protocol clients older than this release fail to decode a snapshot carrying it; clients from this release on decode unknown future health strings as `Unknown`.

Known limitations: the committee size read is the live adopted set, so a short L1 reorg can briefly flap a peer in and out of Suspended (deliberate: suspension is reversible and the delivery floor holds, so nothing is lost, and it closes the finalization-lag window a finalized-only read would leave open). A reader that dies while suspended leaves the peer showing `Suspended` in `getStatus` indefinitely, the same staleness class as a dead reader stuck at `Live`. Sequencer-side suspension is surfaced by gauge and log only; there is no sequencer status string.

## 🧪 How to Test

- `RISC0_DEV_MODE=1 cargo test --release -p cross_zone` covers the latch fold table (fail-closed first read, bounded keep-last-known, resume edge, re-dip).
- `RISC0_DEV_MODE=1 cargo test --release -p sequencer_core --features mock` covers the watcher gate and the drain-continues-while-suspended guarantee.
- `RISC0_DEV_MODE=1 cargo test --release -p indexer_core -p indexer_service_protocol` covers the verifier suspension, the stale-escalation guard, the health precedence, and the Suspended and Unknown status strings.

## 🔜 Future Work

None.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


lez/indexer/core/src/status.rs
```diff
@@ -68,6 +68,9 @@ pub enum PeerHealth {
     Lagging,
     /// Stuck on a slot it cannot read.
     Holed,
+    /// The peer's live committee is below the configured floor; reading is
+    /// suspended until it recovers.
+    Suspended,
     /// A verified-absence verdict was issued against this peer's chain.
     Halted,
```
agureev: Besides this PR, but why is this enum duplicated over from lib.rs?

agureev: Well, now the lib.rs also has Unknown....

agureev (approved):

erhant (approved):
LGTM!

Issue #644, opened 2026-07-24
Reconstruct bridge-lock holdings from the genesis block instead of direct-seeding


moudyellaz:
Follow-up from PR #615 (cross-zone genesis-as-config).

## Current state

Bridge-lock holder balances are seeded directly into genesis state by both the sequencer (`build_initial_state`) and the indexer (from `IndexerConfig.bridge_lock_holdings`), rather than emitted as a genesis transaction. So the indexer still needs out-of-band config to know which holdings exist; unlike the rest of the #615 work (cross-zone programs are builtins, inbox and wrapped-token config are `InitConfig` genesis txs), holdings are not reconstructable from the genesis block alone.

## Why they are direct-seeded

The holder must be a bridge_lock-owned account at the holder's own pubkey: `build_holding_account` sets `program_owner = bridge_lock`, and `bridge_lock::Lock` asserts that and debits the native balance directly (the substitution pin that closes the mint-from-nothing hole). `Claim::Authorized` needs the account signer-authorized in the tx and the genesis builder has the holder's pubkey but not its key, and `Claim::Pda` forces a bridge_lock PDA id rather than the holder's pubkey.

## Goal

Make holdings reconstructable from the finalized genesis block so no zone needs out-of-band holding config, consistent with the decentralized-sequencing direction (a peer reconstructs a zone's genesis by replaying its block 1).

PR #807, opened 2026-08-28
feat(cross-zone)!: guardrails riders, bridge-lock holding PDAs, builtin gating, seed normalization

moudyellaz:
## 🎯 Purpose

The remaining guardrails riders in one genesis window. The main one is #644: bridge-lock holder balances move into per-holder holding PDAs, pinning a lock to a genuine holding and letting the indexer reconstruct holdings from the genesis block instead of a mirrored config field. The rest: normalize the holding seed domain, gate the cross-zone builtins on cross_zone presence, and type genesis supply amounts as Balance.

## ⚙️ Approach

- #644: bridge-lock gains holding PDAs and an appended `InitHolding` instruction (idempotent, never resets a funded holding). `Lock` takes five accounts and debits the holding; the holder only signs and is echoed untouched. Genesis emits an `InitHolding` plus a faucet credit per `SupplyBridgeLockHolding`, both replayable, so the sequencer's directly-seeded holdings and the indexer's `bridge_lock_holdings` config field are both deleted.
- Execution rule 7 is relaxed so a claimless byte-identical echo of an unowned account with history validates, which the holder echo needs. The exemption is claimless only: a claim rides an echo and is gated on the owner being default, so a claimed echo stays refused.
- The six cross-zone builtins and their four InitConfig genesis transactions are gated on `cross_zone` presence, both sides. Program ids are content-derived, so a zone that declares cross_zone keeps byte-identical genesis; a source-only zone declares `cross_zone` with an empty peer list.
- The wrapped-token holding seed domain is normalized to the slash-terminated house form, and the two genesis supply builders take `Balance`.

Known limitations: cross_zone presence is a one-way genesis decision and cannot change on an existing chain, so this is a chain-reset release for a configless deployment that wants the builtins. Sequencer and indexer must agree on cross_zone presence or their genesis states diverge; the paired genesis fingerprint logs are the operator's diagnostic.

## 🧪 How to Test

- `RISC0_DEV_MODE=1 cargo test --release -p lee_core --features host` covers the rule 7 change: claimless echo passes, mutated copy and claimed echo of a non-fresh account both refused, first claim of a fresh account passes.
- `RISC0_DEV_MODE=1 cargo test --release -p integration_tests --test cross_zone_state_machine` covers the holding debit, `InitHolding` idempotency, the wrong-holding and zero-amount refusals, and the gating shape.
- `RISC0_DEV_MODE=1 cargo test --release -p integration_tests --test cross_zone_bridge` reconstructs the source-zone holding on the indexer from the genesis block with the config field gone.
- `RISC0_DEV_MODE=1 cargo test --release -p sequencer_core --features mock -p testnet_initial_state` covers the genesis gating and the registry membership.

## 🔜 Future Work

None.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


lee/state_machine/core/src/program/mod.rs
```diff
@@ -825,9 +825,17 @@ pub fn validate_execution(
...
-        //    account
-        if post.account.program_owner == DEFAULT_PROGRAM_OWNER && pre.account != Account::default()
+        // 7. An account left with the default program owner must either have been a default account
+        //    (a fresh claimless write) or be a claimless byte-identical echo. Modification is what
+        //    needs ownership, and a claimless echo changes nothing, so refusing it would only bar a
+        //    signed account with history from appearing in an account list unclaimed (a signer's
+        //    nonce alone makes it non-default). The claimless condition is load-bearing: a `Claim`
+        //    is applied after this check and gated only on the owner being default, so exempting a
+        //    claimed echo would let any authorized program seize a non-fresh account it merely
+        //    echoes. A first claim of a genuinely fresh account still passes: its pre is default.
```
agureev: Maybe minimize the comment a bit?

lee/state_machine/core/src/program/mod.rs
```diff
@@ -825,9 +825,17 @@ pub fn validate_execution(
...
+        if post.account.program_owner == DEFAULT_PROGRAM_OWNER
+            && pre.account != Account::default()
+            && !(post.account == pre.account && post.required_claim().is_none())
```
agureev: This seems correct but the conjunction is a bit hard to pass, especially given that it is not in  normal form now

maybe opt in for something more explicit like:

```
if post.account.program_owner == DEFAULT_PROGRAM_OWNER && pre.account != Account::default() {
    let claimless_echo = post.account == pre.account && post.required_claim().is_none();
    if !claimless_echo {
        return Err(ExecutionValidationError::NonDefaultAccountWithDefaultOwner {
            account_id: pre.account_id,
        });
    }
}
```

agureev (approved):
looks good!

left a couple of comments for readibility

lee/state_machine/core/src/program/mod.rs
moudyellaz: Done in https://github.com/logos-blockchain/logos-execution-zone/commit/bd2b381711528db027681793334e217bb20c9fb4

lee/state_machine/core/src/program/mod.rs
moudyellaz: Done in https://github.com/logos-blockchain/logos-execution-zone/commit/bd2b381711528db027681793334e217bb20c9fb4

lez/chain_state/src/chain.rs
erhant, line 1: perhaps an `initial_state()` with `initial_state_with_crosszone()` differentiation would save us from these diffs? a single boolean for the sake of cross-zone seems to far-reaching just for an initial state constructor

erhant (approved):
LGTM, small nit

lez/chain_state/src/chain.rs
moudyellaz: The sequencer and indexer both feed a runtime `config.cross_zone.is_some()` into it, so the two-constructor split would just move the branch into every production caller. The test diffs are one-time.

PR #813, opened 2026-08-29
fix(lee): check fed pre-states against chained call refs

agureev:
## 🎯 Purpose

Previously, there was no check in the PPC asserting that the pre-states run were exactly the ones fed by the chained call. Instead, we blindly trusted the `ProgramOutput`s that were fed by the prover.

Likewise, to make sure the public state matched, we make sure that the account ids in the public side match the chained calls exactly. The pre-states are already checked.

## ⚙️ Approach

- [x] Add an `ensure` checking account ID match for public path
- [x] Add an `assert` checking pre-state match between PPC input and chained call.

## 🔗 Dependencies

#790

## 🔜 Future Work

Refactor to get rid of all the repetitive footguns. Users don't need to send in so much data.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


schouhy (approved):

jonesmarvin8 (approved):
Lgtm. Thanks

moudyellaz (approved):

PR #815, opened 2026-08-30
feat(lez/wallet): add program_loader wallet cli/ffi support

jonesmarvin8:
## 🎯 Purpose

Adds the wallet cli/ffi support PR #812 left as Future Work, and fixes the wallet's dead
`DeployProgram` command — it still built `ProgramDeploymentTransaction`, whose state-transition
handler was removed earlier on that branch.

## ⚙️ Approach

- [x] Add a `program_loader` wallet facade: `NewSegment`/`UploadHeader`/`UpdateHeader`
      primitives, plus `Deploy`/`Update` orchestrators that chunk an ELF and drive the
      tail-to-head segment + header transaction sequence
- [x] Add a `program_loader` CLI subcommand group. Every target account (header, each segment) is
      caller-supplied — no hidden key generation, matching every other program-facing command
- [x] Add matching wallet-ffi bindings; regenerate `wallet_ffi.h`
- [x] Remove the dead `DeployProgram`/`send_program_deployment_transaction` path

## 🧪 How to Test

New tests in `integration_tests/tests/program_deployment.rs`:
- `deploy_and_execute_program` (fixed — asserted against the wrong bijected address before)
- `program_loader_new_segment_and_upload_header_deploys_program`
- `program_loader_update_header_repoints_program`
- `program_loader_deploy_rejects_segment_count_mismatch`
- `deploy_invalid_program_fails` (fixed)

## 🔗 Dependencies

[PR #812](https://github.com/logos-blockchain/logos-execution-zone/pull/812)

## 🔜 Future Work

- Support for functional privacy for immutable programs (and shadow programs).

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


lez/wallet/src/program_facades/program_loader.rs
```diff
@@ -0,0 +1,193 @@
...
+    }
+
+    /// Walks a segment chain from `first_segment` via the network, following `next_segment`
+    /// until `None`. Used by the standalone `UploadHeader`/`UpdateHeader` entry points, which
+    /// only take `first_segment` from the caller rather than the whole chain (unlike
+    /// [`Self::deploy`]/[`Self::update`], which already have it from their own `segments` arg).
+    pub async fn resolve_chain(&self, first_segment: AccountId) -> Result<Vec<AccountId>> {
+        let mut chain = Vec::new();
+        let mut next = Some(first_segment);
+        while let Some(id) = next {
```
moudyellaz, line 209: This walks the chain over the network with no cap or visited set, so a malformed or hostile chain hangs the wallet in an RPC loop; should it bound the walk rather than trust whatever the node returns?

lez/wallet/src/program_facades/program_loader.rs
```diff
@@ -0,0 +1,193 @@
...
+        let chunks: Vec<&[u8]> = bytecode.chunks(MAX_SEGMENT_DATA_LEN).collect();
+        if chunks.len() != segments.len() {
+            return Err(ExecutionFailureKind::SegmentCountMismatch {
+                expected: chunks.len(),
+                actual: segments.len(),
+            }
+            .into());
+        }
+
+        for i in (0..chunks.len()).rev() {
```
moudyellaz, line 172: If any segment fails partway, the ones that already landed stay claimed and write-once, so retrying with the same `segments` list trips `execute_new_segment`'s default-account assert on the first one and the user has to supply a fresh set; worth documenting or making resumable?

lez/wallet/src/program_facades/program_loader.rs
```diff
@@ -0,0 +1,193 @@
...
+    ) -> Result<AccountId> {
+        self.upload_segments_and_header(header, segments, bytecode, immutable, false)
+            .await?;
+        Ok(header)
+    }
+
+    /// Like [`Self::deploy`], but signs the final step with `UpdateHeader` against an existing
+    /// header account instead of claiming a new one. Segments are always freshly uploaded —
+    /// segments are write-once, so there's no reuse of a prior chain.
+    pub async fn update(
```
moudyellaz, line 138: Every update uploads a whole fresh chain and nothing reclaims the previous one, so a program's segment accounts accumulate per update with no way to release them; is that intended for now?

lez/wallet/src/program_facades/program_loader.rs
jonesmarvin8: Adding a FIXME so that it's documented, and so we can come back and fix it after RC.

lez/wallet/src/program_facades/program_loader.rs
jonesmarvin8: It is currently intended, though it can be improved upon. Adding a FIXME so that it's documented, and so we can come back and fix it after RC.

lez/wallet/src/program_facades/program_loader.rs
jonesmarvin8: Use the cap introduced in PR 812 (shifted to `program_loader` for access reasons), and added test `program_loader_resolve_chain_rejects_overlong_chain`

lez/wallet/src/lib.rs
```diff
@@ -837,6 +841,41 @@ impl WalletCore {
...
         tx_pre_check: impl FnOnce(&[&Account]) -> Result<(), ExecutionFailureKind>,
+    ) -> Result<HashType, ExecutionFailureKind> {
+        self.send_pub_tx_to_account_with_pre_check(
+            accounts,
+            instruction_data,
+            program_loader_core::immutable_deploy_account_id(program_id),
+            tx_pre_check,
+        )
+        .await
+    }
```
schouhy: let's have this function take a `program_account_id` and have callers that want the conversion from program id to convert it to an account id on call site

lez/wallet/src/lib.rs
```diff
@@ -837,6 +841,41 @@ impl WalletCore {
...
+
+    pub async fn send_pub_tx_to_account_with_pre_check(
+        &self,
+        accounts: Vec<AccountIdentity>,
+        instruction_data: InstructionData,
+        program_account_id: AccountId,
+        tx_pre_check: impl FnOnce(&[&Account]) -> Result<(), ExecutionFailureKind>,
```
schouhy: IMO these are the two function to keep, but with the old names (without `_to_account` which I find cofunsing)

lez/wallet/src/program_facades/program_loader.rs
```diff
@@ -0,0 +1,205 @@
...
+/// Every account (segment, header) is caller-supplied — no key generation happens here. Callers
+/// create accounts first via the ordinary `account new public` flow, the same way every other
+/// program-facing command takes accounts as `AccountId`s rather than conjuring them.
+pub struct ProgramLoader<'wallet>(pub &'wallet WalletCore);
+
+impl ProgramLoader<'_> {
+    /// Writes one bytecode segment at `target` (must already be a default/unclaimed account,
+    /// signed for by `target`'s own key). `next_segment`, if present, must already hold a valid
+    /// segment — chains are always linked tail-to-head.
+    pub async fn new_segment(
```
schouhy, line 20: this could check that the next segment already exists and is owned by the program loader, before submitting the tx.

lez/wallet/src/program_facades/program_loader.rs
```diff
@@ -0,0 +1,205 @@
...
+        self.0
+            .send_pub_tx_to_account(accounts, instruction_data, PROGRAM_LOADER_ACCOUNT_ID)
+            .await
+    }
+
+    /// Creates a new program header at `target` (must already be a default/unclaimed account,
+    /// signed for by `target`'s own key) pointing at `chain` — the entire segment chain, in link
+    /// order, starting at `first_segment`. `image_id` is always recomputed from the chain by
+    /// `program_loader`, never trusted from the caller.
+    pub async fn upload_header(
```
schouhy:
```suggestion
    pub async fn create_header(
```

lez/wallet/src/program_facades/program_loader.rs
```diff
@@ -0,0 +1,205 @@
...
+        let mut accounts = vec![AccountIdentity::Public(target)];
+        accounts.extend(next_segment.map(AccountIdentity::PublicNoSign));
+
+        self.0
+            .send_pub_tx_to_account(accounts, instruction_data, PROGRAM_LOADER_ACCOUNT_ID)
+            .await
+    }
+
+    /// Creates a new program header at `target` (must already be a default/unclaimed account,
+    /// signed for by `target`'s own key) pointing at `chain` — the entire segment chain, in link
```
schouhy: it points to the first_segment, no? the chain is the second to last segments

lez/wallet/src/program_facades/program_loader.rs
```diff
@@ -0,0 +1,205 @@
...
+    /// order, starting at `first_segment`. `image_id` is always recomputed from the chain by
+    /// `program_loader`, never trusted from the caller.
+    pub async fn upload_header(
+        &self,
+        target: AccountId,
+        first_segment: AccountId,
+        chain: &[AccountId],
```
schouhy: I think the variable name `chain` is confusing, specially on the blockchain context. I'd prefer something more verbose but more explicit: `segment_tail_ids`, or `second_to_last_segment_ids`, or something like that

lez/wallet/src/program_facades/program_loader.rs
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    /// Rewrites an existing header at `header` — an ordinary `is_authorized`-gated data
+    /// mutation, so `header`'s own (still-authorized) key must sign. Same chain/`image_id`
+    /// handling as [`Self::upload_header`].
+    pub async fn update_header(
+        &self,
+        header: AccountId,
+        first_segment: AccountId,
+        chain: &[AccountId],
```
schouhy: same here with the `chain` variable name

moudyellaz (approved):

lez/wallet/src/program_facades/program_loader.rs
```diff
@@ -0,0 +1,205 @@
...
+        self.0
+            .poll_and_finalize_public_transaction(tx_hash)
+            .await
+            .context("header transaction did not finalize")?;
+
+        Ok(())
+    }
+
+    /// Walks a segment chain from `first_segment` via the network, following `next_segment`
+    /// until `None`. Used by the standalone `UploadHeader`/`UpdateHeader` entry points, which
```
schouhy:
```suggestion
    /// until `None`. Used by the standalone `CreateHeader`/`UpdateHeader` entry points, which
```

lez/wallet/src/cli/programs/program_loader.rs
```diff
@@ -0,0 +1,253 @@
...
+            .new_segment(target_id, bytecode, next_segment)
+            .await?;
+
+        println!("Segment uploaded at {target_id}");
+        wallet_core
+            .poll_and_finalize_public_transaction(tx_hash)
+            .await
+    }
+
+    async fn handle_upload_header(
```
schouhy:
```suggestion
    async fn handle_create_header(
```

schouhy (approved):
lgtm thanks! left some comments

lez/wallet/src/program_facades/program_loader.rs
jonesmarvin8: Addressed

lez/wallet/src/program_facades/program_loader.rs
jonesmarvin8: Yes. I was thinking about things as 0, 1...N-1. So, I fixed the phrasing to hopefully be more clear.

lez/wallet/src/program_facades/program_loader.rs
jonesmarvin8: Fixed. Thanks.

lez/wallet/src/cli/programs/program_loader.rs
jonesmarvin8: Fixed. Thanks.

lez/wallet/src/program_facades/program_loader.rs
jonesmarvin8: Fixed.

lez/wallet/src/program_facades/program_loader.rs
jonesmarvin8: Renamed, but we do send the entire chain (header included). So, renamed to `chain_segment_ids`.

The header is needed to make sure the rest of the chain matches with the claimed `image_id`. This will be especially crucial if we let segments to be re-writable (as Moudy suggested).

lez/wallet/src/program_facades/program_loader.rs
jonesmarvin8: Added. Now it checks 1) not default account, `program_owner` is `program_loader` and the data stored corresponds with a `ProgramSegment` (not `ProgramHeader`).

lez/wallet/src/lib.rs
jonesmarvin8: Fixed.

lez/wallet/src/lib.rs
jonesmarvin8: Fixed.

PR #816, opened 2026-08-31
feat(fees): check L1 inscription for `producer`

erhant:
This fixes the auto-init issues with `producer`, instead using the L1 inscription signer owner as the producer account.


lez/chain_state/src/chain.rs
```diff
@@ -198,6 +198,20 @@ impl ChainState {
...
+    ///
+    /// When `producer_key` is the block's L1 inscription signer,
+    /// its reward target must match the signer's stake entry
+    /// (see [`verify_producer`]). `None` skips the check, e.g. via
+    /// paths without an L1 signer such as tests, standalone, mocks.
+    fn apply_adopted_with_producer(
+        &mut self,
+        block: &Block,
+        producer_key: Option<sequencer_stake_core::SequencerKey>,
+    ) -> AcceptOutcome {
```
schouhy: why do we need to force the reward target to match something? I'm still confused about that. Can't a producer choose freely it's own account for receiving funds? The stake account seems like a good default since it must exist and is in control of the producer. But I still think enforcing rewards to land there is too much

erhant: you are right, and actually changing it to allow any account instead of a specific producer made things much easier, code went from 300 lines to 130

erhant: i'm merging this to the main PR, since its 100 lines, and will do the final fixes there (there are other reviews)

PR #817, opened 2026-09-01
feat!: authorized balance decrease/permissionless claims/pinata removal (WARNING!  UNSAFE!)

agureev:
## 🎯 Purpose

This PR serves unites three large-scale issues:

1) Remove pinata logic
2) Introduce a protocol-wide check only allowing authorized accounts to decrease balance
3) Remove `Claim` instructions, making any program that changes its data its owner (if default initially)

The PR also changes up the testnet documentation to reflect how the flow now works. However, it is currently in the 0.1 folder. Let me know if that needs changing.

WARNING!

This PR is *not* safe. It has an inherent rent condition where any default-owned program ID can be claimed by a malicious program. Depending on the circumstance, it can squat and permanently deny the address, freezing any balance already sitting there.

This was agreed-to as an RC-level PR that would allow [namespaced account state](https://github.com/logos-blockchain/logos-execution-zone/pull/789).

## ⚙️ Approach
In [1e7aa1f](https://github.com/logos-blockchain/logos-execution-zone/pull/817/commits/1e7aa1f58cf336d6b9aa3ada8d3cb825055d1bd4):
- [x] Delete the pinata programs
- [x] Change the funding documentation to use native transfer programs

In [c3a366c](https://github.com/logos-blockchain/logos-execution-zone/pull/817/commits/c3a366cc7b34aac5f4894111995c837bd24ea3ac):
- [x] Make balance decreases athorization-gated
- [x] Remove the pinata-based clock program in favor of a native one

In [2811d06](https://github.com/logos-blockchain/logos-execution-zone/pull/817/commits/2811d0601279a95c8619ec7096bffb7bf8c6e80d):
- [x] Remove all initialization instructions for programs

In [7d11e7b](https://github.com/logos-blockchain/logos-execution-zone/pull/817/commits/7d11e7b1a08c2d05a5d41a8c367011116efeb23e)
- [x] Make any default-owned program be claimed by any program modifying its data

In [3f3d4a0](https://github.com/logos-blockchain/logos-execution-zone/pull/817/commits/3f3d4a01c53ad1d68b1ee4dfba1bb144c2adab72):
- [x] Make a tool pinning auth transfer's ID to use in bridge and stake

In [4e2ecb1](https://github.com/logos-blockchain/logos-execution-zone/pull/817/commits/4e2ecb1003b2dbcf609ef790887c479b9f0e9d06):
- [x] Lock debit now chains to auth-transfer through PDA to allow for balance decrease

In [44ab6b1](https://github.com/logos-blockchain/logos-execution-zone/pull/817/commits/44ab6b11a59c8f1c48e1932392b7810c2682e021):
- [x] Similar for staking

In [74a152a](https://github.com/logos-blockchain/logos-execution-zone/pull/817/commits/74a152a8be99567045b48de3c2ed797a0547f967):
- [x] Remove vaults
- [x] Make accounts directly funded

Rest of the commits are stylistic

## 🔜 Future Work

Namespaced account state.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


lez/programs/authenticated_transfer/image_id.rs
```diff
@@ -0,0 +1,11 @@
...
+// `artifacts/lez/programs/authenticated_transfer.bin` — do not edit by hand.
+//
+// `include!`d by guests that chain calls into `authenticated_transfer`, never
+// by `authenticated_transfer` itself: its own ELF must not depend on this
+// value, or refreshing it would move the very image id it pins.
+#[expect(
+    clippy::unreadable_literal,
+    reason = "generated image id from risc0, a cryptographic hash as a u32 array"
+)]
+pub const AUTHENTICATED_TRANSFER_IMAGE_ID: [u32; 8] = [4264830291, 3778066642, 2382944248, 3177366655, 2604850701, 2386984500, 1210597961, 3690780271];
```
Arjentix: IMO this can be achieved with a `build.rs` and without a need for a separate tool. Let's consider this option

Justfile
```diff
@@ -17,13 +17,25 @@ build-artifacts:
     @rm -rf {{ARTIFACTS}}
     @just build-artifact lee/privacy_preserving_circuit
     @just build-artifact lez/programs programs
+    @just refresh-image-id-pin
```
Arjentix: See my comment above, but if for some reason we'll go with a separate tool invocation I think we need to adjust CI check to also check `image_id.rs` file

test_fixtures/src/config.rs
```diff
@@ -19,6 +19,13 @@ use wallet::config::{MultiSequencerClientConfig, SequencerConnectionData, Wallet
 pub const INITIAL_PUBLIC_BALANCES_FOR_WALLET: [u128; 2] = [10_000, 20_000];
 pub const INITIAL_PRIVATE_BALANCES_FOR_WALLET: [u128; 2] = [10_000, 20_000];
 
+/// The public account genesis over-supplies so it can seed the private accounts.
+///
+/// Only the circuit writes a commitment, so genesis — a run of public transactions — cannot
+/// create a private account. Their balances are supplied here instead and moved into
+/// commitments by `fund_private_accounts`, which leaves this account at its configured balance.
```
Arjentix:
```suggestion
```

In general I like to remove second paragraph of Claude's code

test_fixtures/src/config.rs
```diff
@@ -211,26 +218,31 @@ pub fn genesis_from_accounts(
...
+                }
+            });
 
     let supply_bridge_account = GenesisAction::SupplyBridgeAccount { balance: 1_000_000 };
 
     public_genesis
-        .chain(private_genesis)
         .chain(std::iter::once(supply_bridge_account))
         .collect()
 }
```
Arjentix, line 253: Tbh I think now this function signature does not reflect what happens inside. I mean it's not obvious that in fact private accounts are needed to sum their balances and put it into the first public account. I'd reconsider code in this module and maybe restructure something

integration_tests/tests/bridge.rs
```diff
@@ -207,406 +195,22 @@ async fn private_bridge_deposit_invocation_is_dropped() -> anyhow::Result<()> {
...
+    assert_eq!(recipient_balance_after, recipient_balance_before);
     assert!(
         tx_on_chain.is_none(),
         "Privacy-preserving bridge::Deposit invocation should be rejected"
     );
 
     Ok(())
 }
-
-// async fn submit_bedrock_deposit(
```
Arjentix, line 230: Hm, why removing that btw? I think this was commented because withdrawals are not reenabled yet.

lez/programs/faucet/core/src/lib.rs
Arjentix: This can be renamed to just `GenesisTransfer`

lez/programs/faucet/core/src/lib.rs
```diff
@@ -6,19 +6,6 @@ const FAUCET_SEED_DOMAIN_SEPARATOR: [u8; 32] = *b"/LEZ/v0.3/FaucetSeed/000000000
 
 #[derive(BorshSerialize, BorshDeserialize)]
 pub enum Instruction {
```
Arjentix, line 8: IMO now that we don't have Pinata we don't have a way for a random user to mint tokens to experiment with in testnet. This may be a problem for program devs especially once we have fees. I think we need to address this somehow. Maybe introduce pinata functionality in `faucet` program?

lez/programs/authenticated_transfer/image_id.rs
Arjentix: And maybe not only for `authenticated_transfer` but for all programs in this crate

lez/sequencer/core/src/tests/reconstruction.rs
```diff
@@ -536,231 +536,6 @@ fn deposit_event_record(
     }
 }
 
-// /// Builds a signed public bridge `Withdraw` transaction (the normal user path).
```
Arjentix, line 539: Also not worth removing. This is commented because withdrawals are disabled for now, but will be enabled in [future](https://github.com/logos-blockchain/logos-execution-zone/pull/745).

Arjentix (commented):
Did a very high-level review, but have some comments for now

test_fixtures/src/config.rs
agureev: yeah, i'll work on improving the docs today

lez/programs/authenticated_transfer/image_id.rs
agureev: > And maybe not only for authenticated_transfer but for all programs in this crate

Let's keep this minimal for now. We only need auth transfer's ID to hardcode it in the staking and bridging program, everything else would be dead code for now and we can do it later.

> IMO this can be achieved with a build.rs and without a need for a separate tool. Let's consider this option

Could you elaborate? We need this to write to a dependency crate so that the stake and bridge programs can hardcode the ID. I am unsure how the usual R0 output helps us here, maybe I'm just being stupid.

Justfile
agureev: addressed

integration_tests/tests/bridge.rs
agureev: good catch! thanks

lez/programs/faucet/core/src/lib.rs
agureev: good catch!

lez/programs/faucet/core/src/lib.rs
agureev: Thread-addressed. Tokens in testnet gotten through Bedrock faucet and then deposited

lez/sequencer/core/src/tests/reconstruction.rs
agureev: Good catch

lez/programs/authenticated_transfer/image_id.rs
Arjentix: Ah, I see now. I spent some time trying to implement this in another way but the end result didn't look any better.

My problem is that we already build program ids: it's `prorgrams` crate with `artifacts` feature enabled. And now we kind of doing the same again but in another way.

Arjentix: So as I can see in fact you rebuild programs twice for any change (e.g. lee_core changed). That's not so desirable...

I propose to rollback this change and maybe pass the id the same way as before. We've already discussed that maybe authenticated transfer should not exist as it is now. Maybe we can implement it as just lib crate functions and use in all programs.

lez/indexer/core/Cargo.toml
```diff
@@ -9,7 +9,6 @@ workspace = true
 
 [features]
 default = []
-testnet = []
```
Arjentix, line 12: Nice, finally we made it

lez/sequencer/core/src/lib.rs
```diff
@@ -1140,10 +1140,19 @@ impl<BP: BlockPublisherTrait, S: StorageActorTrait> SequencerCore<BP, S> {
         // the very task production needs). Draining here also subsumes the old
         // startup replay.
         //
-        // Skip any deposit whose receipt PDA already exists in the state we
+        // Skip any deposit whose receipt PDA the bridge owns in the state we
         // build on — it was minted by us or by a peer whose block we adopted.
         // An orphan reverts the receipt with the block, so the next turn
         // re-mints without any bookkeeping of our own.
+        //
+        // TODO(squatting): a receipt owned by anyone else — a program that
```
moudyellaz, line 1240: This TODO says a squatted receipt is re-executed every block for ever, but not that each denied deposit adds another permanent per-block execution, so the cost accumulates rather than being one stuck deposit. Could the failure budget or the deposit dead letter it proposes land alongside, so a mint that can never succeed retires instead of taxing every future block?

lez/wallet-ffi/wallet_ffi.h
```diff
@@ -950,112 +949,6 @@ enum WalletFfiError wallet_ffi_account_id_for_private_pda(struct FfiProgramId pr
                                                           struct FfiU128 identifier,
                                                           struct FfiBytes32 *account_id);
 
-/**
```
moudyellaz, line 957: `wallet_ffi.h` drops eight symbols that `logos-execution-zone-module/src/lez_core_module.cpp` still calls, so that repo won't build once it picks this up. Can a module PR land alongside?

moudyellaz (approved):
lgtm, thanks. Left minor comments!

The WARNING says a squatter can "steal whatever balance was initially there", but `a_squatter_acquires_a_funded_account_but_still_cannot_spend_it` pins the opposite: a debit needs `is_authorized`, which a squatter cannot forge, so the outcome is a freeze rather than theft. Could you reword to "squat and permanently deny the address, freezing any balance already sitting there"?

lez/sequencer/core/src/lib.rs
agureev: That's a fair point, but I am wary of introducing temporary quickfix code here that can just bloat and possibly be forgotten.

All these squatting problems will be gone after namespaced design, so I am uncertain that there is any benefit in adding extra code that will be gone in a week. If anything, the open holes are good incentive to finalize proper fixes ASAP :)

What do you think?

lez/wallet-ffi/wallet_ffi.h
agureev: Yes, for sure! Will start working on it!

lez/programs/authenticated_transfer/image_id.rs
agureev: can we do something like [804db07](https://github.com/logos-blockchain/logos-execution-zone/pull/817/commits/804db074fb4b63ff98873cd7967f4fe789fd5dcf) instead?

lez/wallet-ffi/wallet_ffi.h
agureev: UPD: Will probably wait until the rebase (esp with fees) to fully understand what will be gone and build on Erhan's PR

lez/sequencer/core/src/lib.rs
moudyellaz: it is fine then since they will be gone in the next PRs. thanks

lez/programs/authenticated_transfer/image_id.rs
Arjentix: After discussion we decided to leave as is and go with https://github.com/logos-blockchain/logos-execution-zone/issues/828 for future

lez/programs/authenticated_transfer/core/build.rs
```diff
@@ -0,0 +1,4 @@
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    build_utils::include_image_id("lez/programs", "authenticated_transfer")?;
```
Arjentix, line 3: Hm, why is it needed twice: here and in `programs/build.rs`?

lee/state_machine/core/src/program/mod.rs
```diff
@@ -685,11 +588,6 @@ pub enum ExecutionValidationError {
         executing_program_id: ProgramId,
     },
 
-    #[error(
```
Arjentix: While I understand now why it's not an error in namespaced accounts approach, I still don't understand why it's not an error in this PR

lee/state_machine/src/state/tests/implicit_claiming.rs
Arjentix, line 1: So this is basically a demonstration of squatting?

Arjentix (approved):
LGTM, thanks

lez/programs/authenticated_transfer/core/build.rs
Arjentix: Looks like `programs/build.rs` doesn't need it

lee/state_machine/core/src/program/mod.rs
agureev: Because unowned accounts can now have non-zero balance, which is also the semantics we want in the namespaced approach (which is a bad name btw, I should use state-shared)

This, e.g. allows to fund PDAs directly sans vaults

lee/state_machine/src/state/tests/implicit_claiming.rs
agureev: in some sense, yes

this file is mostly for demonstrating semantics

this entire file will get erased when namespacing comes in

lez/programs/authenticated_transfer/core/build.rs
agureev: Good catch, i'll try to address this during the rebase

schouhy (approved):
lgtm thanks!

lez/programs/authenticated_transfer/core/build.rs
agureev: check: [32f4d32](https://github.com/logos-blockchain/logos-execution-zone/pull/817/commits/32f4d325f2bf72e74d20c4a5c3b712b2ba2520ec)

PR #818, opened 2026-09-02
refactor(lez): implement 2/3 signature thresholds for slashing

schouhy:
## 🎯 Purpose

Slashing currently supports only 1 of N thresholds: any accredited sequencer could burn a peer's stake on its own signature. This PR raises that to 2/3 , so a slash needs agreement from most of the committee.

It also moves slashing out of the sequencer core into its own actor.

Same single offence as before: inscribing a payload that is not a block.

## ⚙️ Approach

The threshold is two thirds of the accredited committee, capped at `committee_size - 1` and floored at 1. Approvals are gossiped on their own gossip topic.

- [x] Move slashing into a `sequencer_slasher_actor` crate: records offences, collects approvals, proposes the `Slash`
- [x] Add `slash_approval_threshold(committee_size)` and compute it from the live accredited count, replacing the `SLASH_APPROVAL_THRESHOLD` constant
- [x] Gossip approvals on `/lez/<channel>/v1/slash-approvals`, verifying the signature before forwarding
- [x] Modify the slashing integration test from two sequencers to three.

## 🧪 How to Test

```
cargo test -p sequencer_slasher_actor
RISC0_DEV_MODE=1 cargo test -p integration_tests --test slashing
```

## 🔗 Dependencies

None

## 🔜 Future Work

- More offences.
- The cap makes a two key committee 1 of 2, so with exactly two sequencers either one can still slash the other alone.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


lez/sequencer/core/src/block_publisher.rs
```diff
@@ -470,15 +470,16 @@ impl BlockPublisherTrait for ZoneSdkPublisher {
...
                                                     None if <Inscription as AsRef<[u8]>>::as_ref(
                                                         &inscription.payload,
                                                     )
                                                     .is_empty() => {}
-                                                    // An inscription always names
-                                                    // its signer.
+                                                    // Only a config entry has no
+                                                    // signer, and those arrive as
+                                                    // `FinalizedOp::Config`.
                                                     None => undecodable.extend(
```
moudyellaz, line 482: The offence is still `borsh::from_slice::<Block>` failing on the observing build, so during a rolling upgrade the un-upgraded majority sees every upgraded sequencer's block as an offence and 2/3 is exactly the side with the votes: what pins the predicate to something objective before this can fire?

lez/sequencer/actors/slasher/src/actor.rs
```diff
@@ -0,0 +1,817 @@
...
+    async fn handle(&mut self, msg: Approval, _ctx: &mut Context<Self, Self::Reply>) {
+        if msg.signer == self.own_key {
+            return;
+        }
+        // A peer is trusted for its signature, never for the offence itself.
+        if !msg.verify() {
+            return;
+        }
+        // Screened against the last config seen, so junk keys take no slot.
+        if let Some(config) = &self.config
```
moudyellaz: An approval admitted while `self.config` is `None` takes one of the 64 `approvals` slots for a restored offence and persists there with nothing to prune it, so real approvals hit the cap later: should an unknown config reject rather than skip the screen?

lez/programs/sequencer_stake/core/src/lib.rs
```diff
@@ -224,6 +254,18 @@ impl SequencerEntry {
...
 
+/// Approvals a `Slash` must carry: two thirds of the committee, never fewer
+/// than two.
+///
+/// A `Slash` carries no proof the offence happened, so the signatures are the
+/// only thing standing between a key and a peer's stake. One is never enough.
+/// Below three sequencers that puts the threshold above the keys that would
+/// ever sign, which leaves slashing inert rather than unilateral.
+#[must_use]
+pub fn slash_approval_threshold(committee_size: usize) -> usize {
```
moudyellaz, line 276: `accredited_committee_members_count` counts every staked key, so one idle entry at `minimum_sequencer_stake` (149 by default) takes a 3-key committee from threshold 2 with 2 honest signers to threshold 3 with 2, and slashing stops: should the denominator be the keys that can actually sign rather than every entry?

lez/programs/sequencer_stake/core/src/lib.rs
```diff
@@ -7,12 +7,10 @@ pub use lee_core::program::PdaSeed;
 use lee_core::{account::AccountId, program::InstructionData};
 use serde::{Deserialize, Serialize};
 
-/// Approvals a `Slash` must carry. Raising it moves the program id.
-pub const SLASH_APPROVAL_THRESHOLD: usize = 1;
-
 const INVALID_KEY: &str = "invalid Ed25519 public key";
 const SEQUENCER_STAKE_CONFIG_SEED_DOMAIN: [u8; 32] = *b"/LEZ/v0.3/MinSequencerStake/0000";
-const SLASH_APPROVAL_DOMAIN: [u8; 32] = *b"/LEZ/v0.3/SlashApproval/00000000";
+/// Names the fault, so a second fault gets its own domain and its own signatures.
```
moudyellaz, line 12: Since the domain is being widened anyway, should it carry the channel id too, so an approval signed here cannot be replayed on another zone that accredits the same key?

lez/sequencer/core/src/gossip/validation.rs
```diff
@@ -49,6 +57,22 @@ pub fn evaluate_transaction(data: &[u8], max_block_size: u64) -> TxEvaluation {
     TxEvaluation::Accept(authenticated)
 }
 
+/// Decodes a gossiped slash approval and checks its signature; the slasher
+/// judges whether the signer is accredited and the offence real.
+#[must_use]
+pub fn evaluate_approval(data: &[u8]) -> ApprovalEvaluation {
```
erhant: would it be possible to check the accredited key here?

its otherwise possible to spam the p2p mesh with new keys and they will all be validated here, which may make the CPU unhappy

lez/sequencer/actors/slasher/src/actor.rs
```diff
@@ -0,0 +1,817 @@
...
+impl<S: StorageActorTrait> Message<Propose> for SlasherActor<S> {
+    type Reply = Vec<LeeTransaction>;
+
+    async fn handle(
+        &mut self,
+        Propose { config }: Propose,
+        _ctx: &mut Context<Self, Self::Reply>,
+    ) -> Self::Reply {
+        // Kept so the next inbound approval can be screened on arrival.
+        self.config = Some(config.clone());
```
erhant: does config change at runtime? if not, would it be possible to avoid clone here?

it could be some nasty borrow handling & config is not that large anyways so i'm letting you decide

lez/sequencer/core/src/gossip/validation.rs
erhant: it's also what i had intended with gossip at the start too btw, having `accredited.rs` module to that extent (also asked in https://github.com/logos-blockchain/logos-execution-zone/pull/818#pullrequestreview-5154625890)

lez/sequencer/actors/slasher/src/actor.rs
```diff
@@ -0,0 +1,817 @@
...
+    error::Error,
+    protocol::{Approval, Offence, Propose, Report, ReportedOffence, SetApprovalPublisher},
+};
+
+/// Cap on signatures kept per offence.
+const MAX_APPROVERS_PER_OFFENCE: usize = 64;
+/// Cap on distinct offences held in `early`, which are still only a peer's word.
+const MAX_EARLY_OFFENCES: usize = 64;
+
+/// On-disk form, tagged so a later build can migrate instead of failing to decode.
```
erhant, line 35: will the migration happen following this PR?

without it, this PR is more of `feat(lez)!` than `refactor(lez)` I think

erhant (approved):
LGTM, some comments made

also I want to ask, I was thinking that we would use `accredited/mod.rs` module in gossip for these parts, but it appears we do not need it

shall I remove them in my other gossip-related PR? (https://github.com/logos-blockchain/logos-execution-zone/pull/833#discussion_r3968449998)

lez/sequencer/core/src/block_publisher.rs
schouhy: ah I think this is out of the scope of this PR? We'll need to handle this once we have a mechanism for version migrations. Thus far most of our code assumes all nodes are running the same version, not only this bit.

lez/programs/sequencer_stake/core/src/lib.rs
schouhy: what do you mean by an idle entry? if an entry is at or above the minimum, it can sign

lez/programs/sequencer_stake/core/src/lib.rs
schouhy: sounds good!

lez/sequencer/actors/slasher/src/actor.rs
schouhy: ah the first commit is a refactor but the second one is actually what implements the feature and is a `feat`. What do you mean about the migration?

lez/programs/sequencer_stake/core/src/lib.rs
schouhy: done. a bit more complex than it looked like, because the channel id had to go all the way to the stake program for it to verify the signature

lez/sequencer/actors/slasher/src/actor.rs
schouhy: yes. the Option was a leftover from a dumped approach. Removed the Option from it. Also, this surfaced another issue: configs were updated only on production turns. It's now updated also on the follow path so that nodes that are waiting for their turn are in sync

lez/programs/sequencer_stake/core/src/lib.rs
moudyellaz: Fair, "idle" was the wrong word. I meant can-sign is not does-sign: an offline sequencer keeps its full stake so it counts in the denominator while signing nothing. Four keys with one node down and one offender gives threshold 3 and only 2 signers left. Intended failure mode, or should the denominator track live keys?

moudyellaz (approved):
lgtm, thanks! Left one question again.

lez/programs/sequencer_stake/core/src/lib.rs
schouhy: ah no, for a sequencer to go offline it needs to unstake. otherwise it will get slashed because of inactivity. not a slashing reason now, but will be in the future. So the idea is that all keys with stake are active sequencers

lez/sequencer/core/src/gossip/validation.rs
schouhy: ah I'm unfamiliar with this. how should the accredited key check be?

lez/sequencer/actors/slasher/src/actor.rs
erhant: oh I was referring to the comment at the line 33

> On-disk form, tagged so a later build can migrate instead of failing to decode.

indeed otherwise older state will fail to decode, but we can think of restarting the entire chain regardless so there may not be a need of migration per se

lez/sequencer/core/src/gossip/validation.rs
erhant: so the idea was that `AccreditedKeysProvider` trait gives the ability to gossip module to read latest accredited keys set

w.r.t the code here, `Approval`'s `signature` would be checked to actually be part of that set at this point

schouhy: done!

lez/sequencer/actors/slasher/src/actor.rs
schouhy: done! It wasn't that nasty after all

PR #819, opened 2026-09-02
feat(sequencer): multi-node local testing tooling, plus liveness fixes

schouhy:
## 🎯 Purpose

Make it practical to run a local setup with several sequencers: many nodes, one channel, keys joining and leaving the committee. Doing that showed three ways a node stops producing blocks, fixed here too.

## ⚙️ Approach

Fixes so a node keeps producing blocks:

- Fix: an inscription that doesn't decode as a `Block` no longer stops replay. It used to crash the node at startup, so once a channel had one, which is the whole reason slashing v1 exists, no new node could join. `sequencer/core/src/lib.rs:566`
- Fix: a failed production turn no longer kills the process. The error stopped the executor actor, and `main` exits when that happens. Now it is logged and the turn is skipped. `actors/executor/src/actor.rs:230`
- Fix: raise channel posting defaults from 20/10 to 300/25 slots. With a 10 slot timeout and 15s blocks, a healthy sequencer lost its turn between its own blocks. `system_accounts/src/lib.rs:18`
- Add: a `publish_blocked_turns` metric, and a warning after 4 turns skipped behind a tip that never moves. A stuck node looked exactly like a node catching up. This only reports the problem, it does not fix it.
- Change: gossip listens on `0.0.0.0` so nodes on different machines can find each other.

Tools. Each node gets a home folder with its db, keys, wallet, ports and logs, so many nodes on one machine do not conflict:

- `just setup-sequencer <home>`: prepare a node, print the account to fund
- `just run-sequencer <home>`: run it. Creates the channel or stakes itself in when needed
- `just sequencer-leave <home>`: unstake
- `just committee-watch`: stake config next to the live committee
- `just channel-health`: check the channel has every block once, in order, from genesis
- `just inscribe-garbage <home>`: cause a slash. Needs a feature flag that is off by default

## 🧪 How to Test

```
cargo test -p sequencer_core reconstruction
cargo test -p executor_actor
```

Two nodes end to end:

```
just run-bedrock
just run-sequencer                       # creates the channel
just setup-sequencer ~/lez-nodes/seq-1   # prints an account to fund
wallet vault transfer --from Public/<funded> --to Public/<printed> --amount <n>
just run-sequencer ~/lez-nodes/seq-1     # stakes itself in, then produces
just committee-watch --watch 5
just channel-health
```

For the replay fix: stop a node, run `just inscribe-garbage <its home>`, then start a fresh home on the same channel. It catches up. On `dev` it crashes.

## 🔗 Dependencies

None.

## 🔜 Future Work

None

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


lez/system_accounts/src/lib.rs
```diff
@@ -9,14 +9,14 @@ use lee_core::account::{Account, AccountId, Nonce};
...
-/// reclaimed after 10 idle slots if a sequencer stops posting — non-zero so
-/// round robin can move on when a committee has more than one accredited key.
-/// A lone-signature threshold still suffices for config changes.
-pub const DEFAULT_SEQUENCER_POSTING_TIMEFRAME: Slots = 20;
-pub const DEFAULT_SEQUENCER_POSTING_TIMEOUT: Slots = 10;
+/// A 300-slot turn is about twenty blocks at the 15s `block_create_timeout`,
+/// and a turn nobody posts in passes on after 25. The timeout must stay above
+/// that block interval, or a healthy sequencer loses its turn between its own
+/// blocks.
+pub const DEFAULT_SEQUENCER_POSTING_TIMEFRAME: Slots = 300;
```
moudyellaz, line 18: The 25 timeout alone fixes the lost-turn symptom; the 300 is what pushes `sequencer_stake_demo` past its 90s removal wait (30 x 3s, and a posting sequencer now holds its full turn). Could the timeframe stay nearer 20?

lez/sequencer/core/src/lib.rs
```diff
@@ -563,12 +563,18 @@ impl<BP: BlockPublisherTrait, S: StorageActorTrait> SequencerCore<BP, S> {
...
+            // An offence the channel already carries, so replaying it must not
+            // be fatal: skip the payload and let the pin follow the entry, the
+            // same way the follow path treats one.
+            let Ok(block) = borsh::from_slice::<Block>(&zone_block.data) else {
+                warn!(
+                    "Skipping an undecodable inscription {:?} at slot {}",
+                    zone_block.id,
+                    slot.into_inner(),
+                );
+                chain.lock().await.skip_channel_entry(zone_block.id);
```
moudyellaz, line 640: The follow path hands these to `SlashRecord::report`, so a node that was offline for the offence and only replays it never records one. Is slashing on what you followed only the intended split?

lez/sequencer/actors/executor/src/actor.rs
```diff
@@ -40,6 +40,9 @@ use crate::{
 /// How many block lookups a single [`GetBlockRange`] keeps in flight.
 const BLOCK_RANGE_CONCURRENCY: usize = 16;
 
+/// Skips behind an unchanging tip past which this is a stuck pin, not catch-up.
+const BLOCKED_TURNS_BEFORE_WEDGED: u32 = 4;
```
moudyellaz: `ProduceBlock` fires once per `block_create_timeout`, not once per round-robin turn, so this is 4 ticks (~60s at 15s blocks) rather than 4 turns. Rename so the log line and the gauge description read right?

lez/sequencer/actors/executor/src/actor.rs
```diff
@@ -169,24 +202,38 @@ impl<BP: BlockPublisherTrait + Send + Sync + 'static, S: StorageActorTrait> Mess
...
+                     to trigger recovery, or reset the store.",
+                );
+            } else {
+                info!(
+                    "Skipping turn: channel tip {tip} moved past our pin {pin}; catching up first"
+                );
+            }
             return Ok(());
         }
+        if self.blocked_turns.clear() {
```
moudyellaz: The two returns above never reach this, so a node that wedges and is then dropped from the committee holds the gauge non-zero forever and keeps alerting. Clear it on the not-our-turn path too?

lez/sequencer/actors/executor/src/actor.rs
```diff
@@ -169,24 +202,38 @@ impl<BP: BlockPublisherTrait + Send + Sync + 'static, S: StorageActorTrait> Mess
...
-            Ed25519Key::from_bytes(&self.sequencer.sequencer_config().signing_key)
-                .public_key()
-                .as_bytes(),
-        );
-        log::info!("Block with id {id} created by {author_identity:?}");
+        // A failed turn costs this block only. Returning the error would stop
+        // this actor, and the scheduler's interval task gives up for good the
+        // first time it finds us not running — the node then looks healthy and
+        // never produces again.
+        match self.sequencer.run_production_turn().await {
```
moudyellaz, line 249: This was the only construction of `Error::BlockProductionFailed`, so the variant is now dead. Remove it?

lez/sequencer/actors/executor/src/actor.rs
```diff
@@ -169,24 +202,38 @@ impl<BP: BlockPublisherTrait + Send + Sync + 'static, S: StorageActorTrait> Mess
...
+            Ok(id) => log::info!(
+                "Block with id {id} created by {}",
+                self.sequencer.bedrock_public_key_hex()
+            ),
+            Err(err) => warn!("Skipping turn: block production failed: {err:#}"),
```
moudyellaz: A turn that fails every time now looks exactly like a healthy node with no traffic, which is the mode `publish_blocked_turns` was added to catch. Worth a counter here too?

Justfile
```diff
@@ -154,8 +203,7 @@ cross-zone-chat:
 # Clean runtime data
 clean:
     @echo "🧹 Cleaning run artifacts"
-    rm -rf lez/sequencer/service/bedrock_signing_key
-    rm -rf lez/sequencer/service/rocksdb*
+    rm -rf lez/sequencer/service/sequencer_home
```
moudyellaz, line 203: Every existing checkout still has `lez/sequencer/service/rocksdb*` and `bedrock_signing_key` at the old paths, and wiping the store is the remedy the divergence error prints. Keep those two lines alongside this one?

tools/channel_health.py
```diff
@@ -0,0 +1,166 @@
...
+            for tx in block.get("transactions", []):
+                for op in tx.get("mantle_tx", {}).get("ops", []):
+                    if op.get("opcode") != CHANNEL_INSCRIBE_OPCODE:
+                        continue
+                    payload = op["payload"]
+                    if payload.get("channel_id") != channel:
+                        continue
+                    raw = bytes.fromhex(payload["inscription"])
+                    signer = (payload.get("signer") or "?")[:8]
+                    if len(raw) < MIN_BLOCK_LEN:
```
moudyellaz: A garbage payload of 144 bytes or more still lands in `blocks` with its first 8 bytes read as a height, so `just inscribe-garbage --payload <long>` makes this report an enormous gap on the flow the PR documents. Sanity-check the height before counting it?

tools/channel_health.py
```diff
@@ -0,0 +1,166 @@
...
+def report(blocks, others, scanned) -> bool:
+    """Prints the run and returns whether it holds."""
+    slot_from, slot_to, tip = scanned
+    print(f"scanned slots {slot_from}..{slot_to} (tip {tip})")
+    print(f"inscriptions carrying a block: {len(blocks)}   other payloads: {len(others)}")
+    for slot, signer, size, prefix in others:
+        print(f"  non-block at slot {slot} by {signer} ({size} bytes, {prefix})")
+
+    if not blocks:
+        print(f"\n{paint('\u2718  no inscriptions in that range', '1;31')}")
```
moudyellaz: Backslash inside an f-string expression only parses on 3.12+, so this is a SyntaxError on the 3.9 macOS ships. Pull the `\u2718` out into a variable?

tools/committee_watch.py
```diff
@@ -0,0 +1,283 @@
...
+            "staked": r.u128(),
+            "pending": r.u128(),
+        }
+    for e in entries.values():
+        e["net"] = max(e["staked"] - e["pending"], 0)
+    return {"minimum": minimum, "entries": entries}
+
+
+def seat_label(
+    key: str, entry: dict | None, accredited: list[str], turn: int, minimum: int
```
moudyellaz, line 135: `dict | None` needs 3.10 and is evaluated at def time, so on 3.9 this import fails and takes `join_sequencer` and `sequencer_leave` with it (same at `join_sequencer.py:56`). Worth a `from __future__ import annotations`?

tools/setup_sequencer.py
```diff
@@ -0,0 +1,145 @@
...
+    account = re.search(r"account_id Public/(\S+)", out)
+    mnemonic = re.search(r"Recovery phrase:\s*\n\s*(.+)", out)
+    if not account:
+        sys.stderr.write(out)
+        raise SystemExit("could not read the new account id")
+    account = account.group(1)
+
+    seed_path = os.path.join(home, SEED_FILE)
+    with open(seed_path, "w") as handle:
+        handle.write(f"password: {password}\n")
```
moudyellaz, line 103: Worth saying in the printed summary that this file holds the wallet password and mnemonic in cleartext, so nobody copies a home around?

moudyellaz (approved):
lgtm, thanks. Left some comments/questions!

lez/system_accounts/src/lib.rs
schouhy: ah yes, will do for tests

lez/sequencer/core/src/lib.rs
schouhy: This `verify_and_reconstruct` runs after the startup backfill of the zone sdk (inside the block publisher constructor). The offence is reported in the backfill.

lez/sequencer/actors/executor/src/actor.rs
schouhy: done

lez/system_accounts/src/lib.rs
schouhy: I actually ended up implementing something more sophisticated that will be handy for production deployments: I made the timeframe and timeout configurable from sequencer config, and then land on lez on an account (the stake config account that we should later rename to lez config account probably, for another PR). I also used this refactor to make the minimum stake configurable. Also handy for testnet. Tests fallback to using the old timeframes.

lez/sequencer/actors/executor/src/actor.rs
schouhy: done

lez/sequencer/actors/executor/src/actor.rs
schouhy: done

lez/sequencer/actors/executor/src/actor.rs
schouhy: done

Justfile
schouhy: done

tools/channel_health.py
schouhy: done

tools/committee_watch.py
schouhy: done

tools/channel_health.py
schouhy: went with a heuristic approach. I wouldnt worry much about this since it's just for some local testing

tools/setup_sequencer.py
schouhy: done

lez/programs/sequencer_stake/src/main.rs
```diff
@@ -387,6 +395,49 @@ fn verify_approvals(
...
+    // A zero timeframe would leave round robin unable to move off index 0, and
+    // a zero minimum would accredit every key that ever staked a nonzero amount.
+    assert!(
+        channel_params.posting_timeframe > 0,
+        "posting_timeframe must be non-zero"
+    );
+    assert!(
+        channel_params.minimum_sequencer_stake > 0,
+        "minimum_sequencer_stake must be non-zero"
+    );
```
moudyellaz, line 448: `posting_timeout` gets no assert, and a non-zero value under the block interval is the bug this PR fixed, now unfixable once written. Check it against `block_create_timeout` at genesis build?

lez/programs/sequencer_stake/src/main.rs
```diff
@@ -387,6 +395,49 @@ fn verify_approvals(
...
+    self_program_id: ProgramId,
+    pre_states: Vec<AccountWithMetadata>,
+    channel_params: ChannelParams,
+) -> Vec<AccountPostState> {
+    let [config_account] = <[AccountWithMetadata; 1]>::try_from(pre_states)
+        .expect("InitChannelParams requires the config account");
+
+    let mut config = decode_config(&config_account, self_program_id);
+    assert!(
+        config.channel_params.is_none(),
```
moudyellaz, line 430: This assert plus genesis running first is the whole protection on params fixed for the chain's life, on a tx that is unsigned by design. Is that the intended guard?

lez/sequencer/core/src/block_publisher.rs
```diff
@@ -386,12 +400,8 @@ impl BlockPublisherTrait for ZoneSdkPublisher {
                                         .handle()
                                         .channel_config(
                                             new_keys,
-                                            SlotTimeframe::from(
-                                                system_accounts::DEFAULT_SEQUENCER_POSTING_TIMEFRAME,
-                                            ),
-                                            SlotTimeout::from(
-                                                system_accounts::DEFAULT_SEQUENCER_POSTING_TIMEOUT,
-                                            ),
+                                            SlotTimeframe::from(channel_params.posting_timeframe),
```
moudyellaz, line 403: During the upgrade window a node still on dev posts the old `system_accounts` 20/10 here instead, so the first committee update it wins undoes the channel's params. Worth a note on the rollout order?

lez/sequencer/core/src/config.rs
```diff
@@ -105,6 +106,10 @@ pub struct BedrockConfig {
     pub funding_key: ZkPublicKey,
     #[serde(default = "default_priority_fee_percent")]
     pub priority_fee_percent: u64,
+    /// What it takes to write on this channel, fixed at genesis: the stake a
+    /// key needs to be accredited, and the turn timing round robin runs on.
+    /// Every node on the channel must agree on them.
+    pub channel_params: ChannelParams,
```
moudyellaz, line 126: Only the founder's copy ever reaches the channel; a joiner's is silently replaced by the channel's genesis on reconstruction. Required from every operator with no serde default and no mismatch check, deliberate?

moudyellaz (approved):
thanks for addressing my comments! left more on the new commit! Thanks again

lez/sequencer/core/src/block_publisher.rs
schouhy: what do you mean? deployment will stop and wipe the old testnet and start fresh.

lez/programs/sequencer_stake/src/main.rs
schouhy: This is a simplification for RC. later we'll have these be changed through gossiping and signature thresholds. For this version these are set at genesis and fixed.

lez/sequencer/core/src/config.rs
schouhy: yes this is because those channel params are fixed at genesis. The rest of the sequencers will read them from lez state. I didn't want to write default values to have more control on what's useed by just looking at the config. We'll probably need to put other params in this flow, like the lez posting pace and so on.

lez/programs/sequencer_stake/src/main.rs
schouhy: ah yes will add it

lez/programs/sequencer_stake/core/src/lib.rs
```diff
@@ -143,11 +147,38 @@ pub struct PendingUnstake {
...
+    serde::Serialize,
+    serde::Deserialize,
+)]
+pub struct ChannelParams {
+    /// Minimum summed stake for a key to be a committee candidate.
+    pub minimum_sequencer_stake: u128,
+    /// How long one sequencer's posting turn lasts, in slots.
+    pub posting_timeframe: u32,
+    /// Idle slots after which a turn nobody posted in passes on. Must stay
+    /// above `block_create_timeout`, or a healthy sequencer loses its turn
```
agureev, line 168: Do we have something for making sure this property is checked on startup?

lez/sequencer/actors/executor/src/actor.rs
```diff
@@ -166,28 +216,54 @@ impl<BP: BlockPublisherTrait + Send + Sync + 'static, S: StorageActorTrait> Mess
                  waiting for the channel to restore it",
                 self.sequencer.next_block_height().await.saturating_sub(1),
             );
+            // A rewind is its own stall, not a trailing pin.
```
agureev: This is a very loaded comment haha

lez/sequencer/actors/executor/src/actor.rs
```diff
@@ -166,28 +216,54 @@ impl<BP: BlockPublisherTrait + Send + Sync + 'static, S: StorageActorTrait> Mess
...
-        );
-        log::info!("Block with id {id} created by {author_identity:?}");
+        // A failed turn costs this block only. Returning the error would stop
+        // this actor, and the scheduler's interval task gives up for good the
+        // first time it finds us not running — the node then looks healthy and
+        // never produces again.
+        match self.sequencer.run_production_turn().await {
+            Ok(id) => {
+                if std::mem::take(&mut self.failed_attempts) != 0 {
+                    sequencer_core_metrics::record_production_failed_attempts(0);
```
agureev, line 255: I am a bit confused, if the failed attempts is not 0, why do we record 0 production failed attempts? Or am I reading this wrong?

Is this like saying that if we finally get a success we do not care about errors?

agureev (approved):
seems good

lez/programs/sequencer_stake/core/src/lib.rs
schouhy: the config constructur checks it!

lez/sequencer/actors/executor/src/actor.rs
schouhy: this is resetting the prometheus gauge to zero on success, unless it was already zero. does that make sense to you?

lez/sequencer/actors/executor/src/actor.rs
schouhy: lol, reworded

PR #821, opened 2026-09-02
refactor(sequencer): thread the finalized L1 slot through FollowUpdate

erhant:
## 🎯 Purpose

Resolves the FIXME in `sequencer_core::apply_follow_update`: the invalid-finalized stall was recorded with a placeholder `Slot::from(0)` because the zone-sdk did not surface the L1 slot of a finalized inscription. I had added it on https://github.com/logos-blockchain/logos-blockchain/pull/3147 via `FinalizedTx.l1_slot`, and our latest pin now includes it. This PR simply wires it through, so `StallReason` carries the real L1 slot of the block that failed to apply.

## ⚙️ Approach

The slot is per-`FinalizedTx`, not per-update (one `BlocksProcessed` event can carry finalized txs from different L1 blocks), so it rides alongside each block instead of being a single `apply_follow` argument.

- [x] Change `FollowUpdate::finalized` from `Vec<Block>` to `Vec<(Block, Slot)>`, stamping each decoded block with its `FinalizedTx.l1_slot` in the publisher event loop
- [x] Change `ChainState::apply_follow` to take `&[(Block, Slot)]` and drop the blanket `finalized_slot` parameter; each block feeds its own slot into `apply_finalized`
- [x] Change `resolve_own_publishes` to take an `impl IntoIterator<Item = &Block>` (the finalized report is no longer a plain block slice)
- [x] Fix the follow-path call site: FIXME comment and `Slot::from(0)` deleted

## 🧪 How to Test

No behavior change outside the stall's recorded slot; existing coverage exercises the threading (`invalid_finalized_block_sets_final_stall` and the follow-update suite):

```sh
RISC0_DEV_MODE=1 cargo nextest run -p chain_state -p sequencer_core --all-features
```

## 🔗 Dependencies

None

## 🔜 Future Work

None

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] ~Add/update tests~ existing tests were wired

### DIFF
--- a/lez/chain_state/src/chain.rs
+++ b/lez/chain_state/src/chain.rs
@@ -164,8 +164,8 @@ impl ChainState {
     /// Drops our record of every block this update reports on. The sdk omits our
     /// own landed block from `adopted` except on a branch change, so `finalized`
     /// carries the ordinary case.
-    fn resolve_own_publishes(&mut self, reports: [&[Block]; 3]) {
-        for block in reports.into_iter().flatten() {
+    fn resolve_own_publishes<'block>(&mut self, reports: impl IntoIterator<Item = &'block Block>) {
+        for block in reports {
             self.own_publishes.remove(&block.header.hash);
         }
     }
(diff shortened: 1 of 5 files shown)


lez/chain_state/src/chain.rs
```diff
@@ -323,17 +323,21 @@ impl ChainState {
...
+                .iter()
+                .chain(adopted)
+                .chain(finalized.iter().map(|(block, _)| block)),
+        );
 
         let adopted_outcomes = self.apply_channel_update(orphaned, adopted);
         let finalized_outcomes = finalized
             .iter()
-            .map(|block| self.apply_finalized(block, finalized_slot))
+            .map(|(block, l1_slot)| self.apply_finalized(block, *l1_slot))
```
moudyellaz, line 340: nothing asserts the slot that reaches the stall: the stall test calls `apply_finalized` directly and checks only `block_id`, and every `apply_follow` test passes `Slot::from(0)`, the placeholder this PR removes. assert `stall.l1_slot` through `apply_follow` with a non-zero slot?

moudyellaz (approved):
lgtm, thanks. Left one minor question!

lez/chain_state/src/chain.rs
erhant: good point, now tested in https://github.com/logos-blockchain/logos-execution-zone/pull/821/commits/95a8417103cbea38ec8892b1152c59ae71b2af8f

Pravdyvy (approved):
lgtm

PR #823, opened 2026-09-02
feat(sequencer_ffi): Adding Sequencer FFI

Pravdyvy:
## 🎯 Purpose

Added a sequencer FFI for future use in indexer and lez core modules. Also added submodule in `integration_tests` for sequencer FFI setup. 

Currently supports startup and simple queries(one block by id, transaction by hash, account state). More complicated ones require DB updates for sequencer.

## ⚙️ Approach

- [x] Sequencer FFI lifecycle.
- [x] Sequencer FFI queries. 
- [x] Integration tests.

## 🧪 How to Test

Added integration tests in file `sequencer_ffi.rs`.

## 🔗 Dependencies

None

## 🔜 Future Work

- Add optional RPC server setup into FFI. Requires making it optional for usual sequencer as well. 
- Add missing queries. Requires updating DB up to indexer fucntionality. See #878 .
- Add events. Requires updating DB up to indexer fucntionality and event handling.
- Update gossip part when #833 is merged.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


lee/state_machine/src/privacy_preserving_transaction/circuit/mod.rs
```diff
@@ -21,9 +21,19 @@ use crate::{
...
 /// Proof of the privacy preserving execution circuit.
-#[derive(Debug, Clone, PartialEq, Eq, BorshSerialize, BorshDeserialize)]
+#[derive(Clone, PartialEq, Eq, BorshSerialize, BorshDeserialize)]
 pub struct Proof(pub(crate) Vec<u8>);
 
+impl std::fmt::Debug for Proof {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "Too long, cut for readability (altogether {} bytes)",
```
erhant: maybe actually cut if its more than some `k` bytes, or we can just say "[proof redacted for brevity ({} bytes)]"

lez/sequencer/ffi/build.rs
```diff
@@ -0,0 +1,14 @@
+use std::env;
+
+fn main() {
+    let crate_dir = env::var("CARGO_MANIFEST_DIR").unwrap();
+    println!("cargo:rerun-if-changed=src/");
+    cbindgen::Builder::new()
+        .with_crate(crate_dir)
+        .with_language(cbindgen::Language::C)
```
moudyellaz, line 8: Every export here collides with lez/indexer/ffi (init_logger, free_cstring, is_ok, is_error, query_status, all seven query_* and all seven free_ffi_*), and integration_tests now links both into one binary. Add a `sequencer_ffi_` export prefix like wallet-ffi does?

lez/sequencer/ffi/src/sequencer.rs
```diff
@@ -0,0 +1,124 @@
...
+                unsafe { Box::from_raw(self.scheduler_ref.cast::<ActorRef<Scheduler>>()) };
+            // stop the sheduler actor first.
+            let send_res = self.runtime.block_on(scheduler_ref.stop_gracefully());
+            if let Err(err) = send_res {
+                log::error!("Failed to send shutdown signal: {err}");
+            }
+            drop(scheduler_ref);
+        }
+
+        if self.gossip.is_null() {
```
moudyellaz: Inverted? `new()` always boxes the Option, so this never runs and gossip is leaked instead of stopped before the executor.

lez/sequencer/ffi/src/api/query.rs
```diff
@@ -0,0 +1,603 @@
...
+    }
+
+    let sequencer = unsafe { &*sequencer };
+
+    let block_range_resp = sequencer.runtime().block_on(
+        sequencer
+            .executor_ref()
+            .ask(GetBlockRange {
+                range: BoundedRangeInclusive::try_from(before..=(before.saturating_add(limit)))
+                    .unwrap(),
```
moudyellaz: `MAX_BLOCK_RANGE_LEN` is 1024, so any `limit >= 1024` panics out of an `extern "C"` fn and aborts the host. Return `ClientError` instead? Also the range is built forward (`before..=before+limit`) while the doc says `limit` blocks before `before`.

lez/sequencer/ffi/src/api/types/transaction.rs
```diff
@@ -0,0 +1,665 @@
...
+    }
+}
+
+impl TryFrom<FfiTransaction> for LeeTransaction {
+    type Error = OperationStatus;
+
+    fn try_from(value: FfiTransaction) -> Result<Self, Self::Error> {
+        match value.kind {
+            FfiTransactionKind::Public => {
+                let body = unsafe { Box::from_raw(value.body.public_body) };
```
moudyellaz, line 488: `send_transaction` reaches this off `kind` alone, so a caller passing `kind = Public` with a null `public_body` gets UB. Return `NullPointer` on a null body?

lez/sequencer/ffi/src/api/types/transaction.rs
```diff
@@ -0,0 +1,665 @@
...
+            witness_set: lee::public_transaction::WitnessSet::from_raw_parts({
+                let std_vec: Vec<_> = value.witness_set.into();
+                std_vec
+                    .into_iter()
+                    .map(|ffi_val| {
+                        (
+                            Signature {
+                                value: ffi_val.signature.data,
+                            },
+                            PublicKey::try_new(ffi_val.public_key.data).unwrap(),
```
moudyellaz: Reachable from `send_transaction` with any caller-supplied key that is not a valid x-only point, and a panic out of `extern "C"` aborts the host. Make this `TryFrom` and return `CastError`? Same at :297 and at `cast_ffi_validity_window` :630 (start > end).

lez/sequencer/ffi/src/api/query.rs
```diff
@@ -0,0 +1,603 @@
...
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn query_status(sequencer: *const SequencerServiceFFI) -> *mut c_char {
+    if sequencer.is_null() {
+        log::error!(
+            "Attempted to query status on a null sequencer pointer. This is a bug. Aborting."
+        );
+        return std::ptr::null_mut();
+    }
+
+    let json = match serde_json::to_string("Not yet supported") {
```
moudyellaz, line 145: The doc above documents the full `SequencerStatus` schema (state, indexed_block_id, cross_zone_peers) but this returns the bare string `"Not yet supported"`, and that doc ships into sequencer_ffi.h. Trim the doc to match, or return null with an error?

lez/sequencer/ffi/src/api/query.rs
```diff
@@ -0,0 +1,603 @@
...
+    _hash: FfiHashType,
+) -> PointerResult<FfiBlockOpt, OperationStatus> {
+    if sequencer.is_null() {
+        log::error!("Attempted to query a null sequencer pointer. This is a bug. Aborting.");
+        return PointerResult::from_error(OperationStatus::NullPointer);
+    }
+
+    log::error!("Not supported yet");
+
+    PointerResult::from_value(FfiBlockOpt::from_none())
```
moudyellaz: Returning Ok-plus-none makes "not implemented" indistinguishable from "no such block" for a consumer. Return an error status instead? Same for `query_transactions_by_account` at :472.

lez/sequencer/ffi/src/api/query.rs
```diff
@@ -0,0 +1,603 @@
...
+    let last_block_id_resp = sequencer
+        .runtime()
+        .block_on(sequencer.executor_ref().ask(GetLastBlockId).send());
+
+    last_block_id_resp.map_or_else(
+        |e| {
+            log::error!("Failed to query last block id: {e:#}");
+            LastBlockIdResult::error(OperationStatus::ClientError)
+        },
+        LastBlockIdResult::some,
```
moudyellaz: `GetLastBlockId` replies `Result<BlockId>`, not `Option`, so `is_some` is always true and the empty chain the doc describes is indistinguishable from block 0. Drop the field or make the reply optional?

lez/sequencer/ffi/src/api/lifecycle.rs
```diff
@@ -0,0 +1,239 @@
...
+/// # Safety
+/// The caller must ensure that:
+/// - `runtime` is either null or a valid pointer to a [`Runtime`] that outlives the sequencer.
+/// - `config_path` is a valid pointer to a null-terminated C string.
+unsafe fn setup_sequencer(
+    runtime: *const Runtime,
+    config_path: *const c_char,
+) -> Result<SequencerServiceFFI, OperationStatus> {
+    let user_config_path = PathBuf::from(
+        unsafe { std::ffi::CStr::from_ptr(config_path) }
```
moudyellaz, line 160: `stop_sequencer` null-checks its pointer but this one doesn't, and `CStr::from_ptr(null)` is UB. Return `NullPointer` here too?

integration_tests/tests/sequencer_ffi.rs
```diff
@@ -0,0 +1,155 @@
...
+        anyhow::ensure!(
+            leader_block.header.hash.0 == joiner_block.header.hash.data,
+            "Chain divergence at block {id}: leader {:?} vs joiner {:?}",
+            leader_block.header.hash.0,
+            joiner_block.header.hash.data
+        );
+
+        // SAFETY: FFI ensures validity of value.
+        unsafe {
+            sequencer_ffi_helpers::free_ffi_block(joiner_block);
```
moudyellaz: The two `ptr::read`s above leak the outer boxes, so `free_ffi_block_opt` is the call that belongs here. As written no test covers free_ffi_block_opt, free_ffi_block_vec, free_ffi_account, the three free_ffi_transaction*, free_cstring or stop_sequencer, and this file is the only worked example of the C usage. Route the frees through the public entry points?

integration_tests/tests/sequencer_ffi_helpers/mod.rs
```diff
@@ -0,0 +1,309 @@
...
+    // so there is no external runtime whose address we must keep stable.
+    unsafe { start_sequencer(std::ptr::null(), raw_config_path) };
+
+    if res.error.is_error() {
+        anyhow::bail!("Sequencer FFI error {:?}", res.error);
+    }
+
+    let sequencer_ffi =
+    // SAFETY: FFI ensures validity of value.
+    unsafe { std::ptr::read(res.value) };
```
moudyellaz: This bit-copies out of the box and leaks it, and the test then never calls `stop_sequencer`, so the documented lifecycle exit is untested. Keep the pointer and stop through the FFI?

lez/sequencer/ffi/src/api/types/block.rs
```diff
@@ -0,0 +1,209 @@
...
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn free_ffi_block(val: FfiBlock) {
+    // We don't really need all the casts, but just in case
+    // All except `ffi_tx_ffi_vec` is Copy types, so no need for Drop
+    let _ = BlockHeader {
+        block_id: val.header.block_id,
+        prev_block_hash: HashType(val.header.prev_block_hash.data),
+        hash: HashType(val.header.hash.data),
+        timestamp: val.header.timestamp,
+        producer: PublicKey::try_new(val.header.producer.data).unwrap(),
```
moudyellaz: The whole `BlockHeader` rebuild is discarded and every field is Copy, so this only adds a validating unwrap inside a deallocator. Drop the rebuild and just free the body?

lez/sequencer/ffi/src/api/result.rs
```diff
@@ -0,0 +1,29 @@
+/// Simple wrapper around a pointer to a value or an error.
```
moudyellaz, line 1: This file and memory.rs are byte-identical to lez/indexer/ffi, and vectors.rs/block.rs differ by 2 and 12 lines, so roughly 1300 lines are duplicated. Worth lifting the shared type layer into one crate both FFIs depend on?

Cargo.toml
```diff
@@ -17,6 +17,7 @@ members = [
   "lez",
   "lez/system_accounts",
   "lez/chain_state",
+
```
moudyellaz: Stray blank lines here and at :32, unrelated to the change?

integration_tests/tests/sequencer_ffi.rs
```diff
@@ -0,0 +1,155 @@
+//! End-to-end demo of the sequencer self-join flow.
```
moudyellaz: Copied from sequencer_stake_demo.rs; this file is the FFI test.

lee/state_machine/src/privacy_preserving_transaction/circuit/mod.rs
```diff
@@ -21,9 +21,19 @@ use crate::{
 };
 
 /// Proof of the privacy preserving execution circuit.
-#[derive(Debug, Clone, PartialEq, Eq, BorshSerialize, BorshDeserialize)]
+#[derive(Clone, PartialEq, Eq, BorshSerialize, BorshDeserialize)]
 pub struct Proof(pub(crate) Vec<u8>);
 
+impl std::fmt::Debug for Proof {
```
moudyellaz, line 27: What in the sequencer FFI needs this? Nothing in the new crate derives Debug on Proof, so it looks like it belongs in its own PR.

lee/state_machine/src/privacy_preserving_transaction/circuit/mod.rs
Pravdyvy: Redacted proof.

lez/sequencer/ffi/src/sequencer.rs
Pravdyvy: Fixed

lez/sequencer/ffi/src/api/query.rs
Pravdyvy: Fixed.

lez/sequencer/ffi/src/api/types/transaction.rs
Pravdyvy: It is a part of `FfiTransaction` contract, that it holds a valid pointer to a body, which enum `FfiTransactionKind` indicates.

lez/sequencer/ffi/src/api/types/transaction.rs
Pravdyvy: Fixed all.

lez/sequencer/ffi/src/api/query.rs
Pravdyvy: It is better to return valid pointer into useless string that to return null here. Safety of this function guarantees non-null pointer.

lez/sequencer/ffi/src/api/query.rs
Pravdyvy: Fixed, returning `OperationStatus::NotSupported`.

lez/sequencer/ffi/src/api/query.rs
Pravdyvy: Updated reply, now can return none(is_some == false).

lez/sequencer/ffi/src/api/lifecycle.rs
Pravdyvy: Fixed.

integration_tests/tests/sequencer_ffi.rs
Pravdyvy: Fixed now it holds pointers and frees them through ffi function.

integration_tests/tests/sequencer_ffi_helpers/mod.rs
Pravdyvy: Fixed, now it holds pointer and frees it through ffi function.

lez/sequencer/ffi/src/api/types/block.rs
Pravdyvy: Dropped

lez/sequencer/ffi/src/api/result.rs
Pravdyvy: Not worth it for this two FFIs, as we want to drop indexer. Will be worth it in the future to unify with `wallet_ffi`, will be done in future PR.

Cargo.toml
Pravdyvy: Yes, reverted.

integration_tests/tests/sequencer_ffi.rs
Pravdyvy: Fixed.

lee/state_machine/src/privacy_preserving_transaction/circuit/mod.rs
Pravdyvy: Nothing uses it exactly, but it is very useful for debugging purposes. Proof prints too big log.

lez/sequencer/ffi/build.rs
Pravdyvy: Extended all ffi names for sequencer.

moudyellaz (approved):

lez/sequencer/ffi/src/api/types/mod.rs
```diff
@@ -0,0 +1,197 @@
...
+        let ptr = unsafe { self.entries.add(index) };
+        unsafe { &*ptr }
+    }
+}
+
+#[repr(C)]
+pub struct FfiOption<T> {
+    pub value: *mut T,
+    pub is_some: bool,
+}
```
schouhy, line 180: can we drop the `is_some` field and have a `is_some()` method that returns true when the value pointer is null?

schouhy (approved):
LGTm thanks!

lez/sequencer/ffi/src/api/types/mod.rs
Pravdyvy: Well, technically this field is irrelevant, because checking `is_some == true` or `value != NULL` should be the same in cycles, in case if you are not emulating other architecture. 
However, I don't want to give users freedom to mess with FFI structs too much. I think it is better to have kind of similar structure to Rust enum with guarantee, that on `is_some == true` pointer is valid. 
Also, if I do it without bool, it will mean, that every pointer which we return is some kind of FfiOption, and the cost of one byte is fine to not consider this possibility.

PR #825, opened 2026-09-03
feat: fees ffi fixes

erhant:
## 🎯 Purpose

FFI follow-up to #801: make the fee subsystem visible and usable across both C FFI surfaces.
- wallet FFI could not distinguish a fee-admission refusal from a generic network error, and still exposed the public-account `Initialize` path that is removed now
- the indexer FFI silently dropped the fee declaration, so the explorer had no way to show what a transaction declared.

## ⚙️ Approach

**Wallet FFI**
- [x] Add `WalletFfiError::PayerCannotFund` (18) and map the sequencer's fee-admission rejection to it in `map_execution_error`
- [x] Add `ExecutionFailureKind::fee_admission_rejection()` in `wallet`, decoding the structured `AdmissionRejection` from the JSON-RPC error's `data` field (`sequencer_service_rpc` now re-exports `AdmissionRejection` / `ErrorObjectOwned` for this)
- [x] Remove `wallet_ffi_register_public_account`: a bare `Initialize`'s only signer is the unfunded fresh account, which cannot pay, so it is refused at admission by design. A public account is claimed by its first funded transfer instead — `wallet_ffi_transfer_public` already does this (it signs with the recipient's key when the wallet holds it) and its docs now say so. Private registration is unaffected.

**Indexer FFI**
- [x] Carry the fee declaration over the FFI: new inline `FfiFeeDeclaration { payer, gas_limit, tip, max_fee }` plus `has_fee` on `FfiPublicMessage` (no heap allocation, so the free path is unchanged); resolves both `FIXME`s

## 🧪 How to Test

```sh
# Rejection decode + fee roundtrip unit tests
RISC0_DEV_MODE=1 cargo test -p wallet fee_admission_rejection
RISC0_DEV_MODE=1 cargo test -p indexer_ffi fee_roundtrips

# End-to-end claim-by-funded-transfer over the wallet FFI
RISC0_DEV_MODE=1 cargo test -p integration_tests --test wallet_ffi -- wallet_ffi_public_account_is_claimed_by_funded_transfer
```

## 🔗 Dependencies

None.

## 🔜 Future Work

- Bump the logos-execution-zone flake input in lez-indexer-module / logos-execution-zone-module and rebuild against the new headers: wallet_ffi_register_public_account is gone (migrate callers to a funded wallet_ffi_transfer_public), and FfiPublicMessage's layout changed.
- Surface the fee in the explorer UI once the module is bumped.
- Other admission rejections (`MaxFeeBelowReserve`, …) still map to the generic error codes; dedicated codes can be added the same way if consumers need them.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests

### DIFF
--- a/integration_tests/tests/wallet_ffi.rs
+++ b/integration_tests/tests/wallet_ffi.rs
@@ -200,12 +200,6 @@ unsafe extern "C" {
         out_result: *mut FfiTransferResult,
     ) -> error::WalletFfiError;
 
-    fn wallet_ffi_register_public_account(
-        handle: *mut WalletHandle,
-        account_id: *const FfiBytes32,
-        out_result: *mut FfiTransferResult,
-    ) -> error::WalletFfiError;
-
     fn wallet_ffi_register_private_account(
         handle: *mut WalletHandle,
         account_id: *const FfiBytes32,
@@ -839,7 +833,7 @@ fn wallet_ffi_base58_to_account_id() -> Result<()> {
 }
 
 #[test]
-fn wallet_ffi_registering_an_unfunded_public_account_is_refused() -> Result<()> {
+fn wallet_ffi_public_account_is_claimed_by_funded_transfer() -> Result<()> {
     let ctx = BlockingTestContext::new_default()?;
     let home = tempfile::tempdir()?;
     let FfiCreateWalletOutput {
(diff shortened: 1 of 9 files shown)


Pravdyvy (approved):
lgtm

moudyellaz (approved):
lgtm, thanks

schouhy (approved):

PR #827, opened 2026-09-03
fix(test): ensure finalization in test fixture

erhant:
## 🎯 Purpose

Fixes the `valid-proof-test` CI failure on `dev` (`private::private_transfer_to_owned_account`): the sequencer panicked at startup with `Stored block 8 does not replay while restoring chain state: ... Invalid privacy preserving execution circuit proof`. ([see action](https://github.com/logos-blockchain/logos-execution-zone/actions/runs/33752282665/job/100675952162))

**Root cause**: the prebuilt test fixture is generated under `RISC0_DEV_MODE=1`, so the privacy transactions it contains (the private vault claims) carry **fake receipts**, which only verify when dev mode is set at verification time. `restore_chain_state` replays only blocks *above* the dump's final snapshot.

Whether the privacy blocks land above or below that boundary was a race against Bedrock finality during fixture generation (where the problem stems from). Every green run so far was timing luck: older dumps happened to finalize past the privacy blocks. The dump from my #801 snapshot sits at block 6 while the privacy transactions are in blocks 8 and 11, so the one CI job that runs without `RISC0_DEV_MODE` re-verifies fake proofs at restore and dies.

## ⚙️ Approach

- [x] `regenerate_test_fixture`: after stripping the checkpoint and resetting blocks to `Pending`, stamp the **final snapshot at the tip** (`GetLeeState` + `GetLatestBlockMeta` → `ApplyStoreUpdate`). Restore then replays no fixture blocks, so their dev-mode proofs are never re-verified — in any proof mode. Blocks stay `Pending` (`finalized_up_to: None`) so a fresh Bedrock still gets them re-inscribed; their finality events come back as idempotent at-or-below-tip re-deliveries (`apply_finalized_direct`).
- [x] Regenerate and commit the fixture dump.

This removes the race permanently rather than re-rolling the dice with a fresh regen. The valid-proof job still tests exactly what it is for: the fresh privacy transfer the wallet proves *during* the test is verified for real.

## 🧪 How to Test

```sh
# The previously failing test, in real-proof mode (what valid-proof-test runs):
cargo test -p integration_tests -- --exact private::private_transfer_to_owned_account

# Dump round-trip / genesis-follows-config checks still pass in dev mode:
RISC0_DEV_MODE=1 cargo test -p test_fixtures
```

## 🔗 Dependencies

None

## 🔜 Future Work

None

> [!TIP]
>
> If the fixture flow ever changes, note the invariant this PR establishes: a committed dump must never leave privacy transactions above its final-snapshot boundary.

## 📋 PR Completion Checklist
- [x] Complete PR description
- [x] Add/update tests


Arjentix (approved):
LGTM. But ideally I'd leave a TODO with a message that in future when Bedrock communication in moved into a separate actor which can be mocked we can mock it in fixture generation forcing mock to immediately finalize. That way we won't need this extra direct storage manipulations

Pravdyvy (approved):
lgtm

PR #830, opened 2026-09-04
refactor(test): add gossip screening for fees and fix breakpoint test

### LINKED ISSUE #797: fees: screen gossip-ingested transactions through fee admission

erhant:
## 🎯 Purpose

The #801 (fees) merge tripled `nextest` wall time (525s → 1530s; the unit-tests job went 24.5 min → ~53 min and started timing out): every charged transaction now runs up to three zkVM guest emulations (**reserve → action → refund**) and dev mode does not help (as it skips proving, not execution).

This PR takes the free wins from that investigation (kudos to @moudyellaz ):

1. A test `put_block_stores_breakpoint_in_same_batch` that never needed settleable blocks was paying for ~795 guest executions.

2. The block builder ran a full scratch-clone settlement (all three guest runs) on every mempool candidate before discovering it could not pay, and gossiped transactions entered the mempool with no fee screening at all.

## ⚙️ Approach

- [x] `put_block_stores_breakpoint_in_same_batch` (storage): chain `produce_dummy_block`s instead of signed, settled transfer blocks — `put_block` never settles, so the breakpoint-batch assertions are unchanged and the test goes from ~795 guest executions to zero (327s → seconds in CI)
- [x] Builder: run the cheap `fees::screen` (static checks + one balance read, no clone, no execution) on every User/Gossip candidate *before* the scratch clone and settlement, so junk is dropped for a balance read instead of three guest runs
- [x] Gossip ingest: screen a peer's transaction through the same `fees::screen` before it takes a mempool slot, via a new screen-only `ScreenTransaction` executor message — one admission door for RPC and gossip. Deliberately **advisory**: the verdict never affects gossipsub mesh acceptance (admission is priced off the local head state, which drifts, so peers legitimately disagree), and a screened-out tx stays un-`seen` so a rebroadcast can retry once e.g. its payer is funded. Closes #797 
- [x] Document that `MAX_GAS_EXEC` caps action-phase gas only; the reserve/refund invocations and the fee/clock tail run unmetered as a bounded per-transaction constant

## 🧪 How to Test

```sh
# The reverted test, now sub-second:
RISC0_DEV_MODE=1 cargo test -p storage put_block_stores_breakpoint_in_same_batch

# Gossip + builder paths (gossip tests use the new admit-all screen):
RISC0_DEV_MODE=1 cargo test -p sequencer_core --all-features gossip
RISC0_DEV_MODE=1 cargo test -p sequencer_core --all-features tests::
```

## 🔗 Dependencies

None

## 🔜 Future Work

None

## 📋 PR Completion Checklist
- [x] Complete PR description
- [x] Add/update tests


lez/sequencer/core/src/gossip/network.rs
```diff
@@ -522,6 +550,12 @@ pub(crate) fn peer_id_from_ed25519(
     clippy::wildcard_enum_match_arm,
     reason = "SwarmEvent is non_exhaustive; only startup listener events are handled here"
 )]
+/// An [`IngestScreen`] that admits everything; for tests and standalone runs.
+#[must_use]
+pub fn admit_all_screen() -> IngestScreen {
+    Arc::new(|_| Box::pin(async { Ok(()) }))
+}
```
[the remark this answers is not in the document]
erhant: yeah my bad

lez/sequencer/core/src/gossip/network.rs
```diff
@@ -59,6 +60,14 @@ pub struct GossipNetwork {
     tx_tx: mpsc::Sender<LeeTransaction>,
 }
 
+/// Advisory fee-admission screen run at gossip ingest, before a peer's
+/// transaction is pushed to the mempool.
+///
+/// Its verdict never affects mesh acceptance: admission is priced off the
+/// local head state, which drifts, so peers legitimately disagree.
+pub type IngestScreen =
+    Arc<dyn Fn(LeeTransaction) -> BoxFuture<'static, Result<(), String>> + Send + Sync>;
```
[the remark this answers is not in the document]
erhant: it does compile

lez/sequencer/core/src/gossip/network.rs
```diff
@@ -401,15 +416,28 @@ impl DriveTask {
...
-                            log::debug!("Mempool full; forwarding tx {hash:?} without admitting");
+                    // Fee-screen before the mempool sees it, so peers cannot
+                    // fill it with transactions the builder would only drop
+                    // after a full settlement pass. Advisory: the tx is still
+                    // forwarded, and stays unseen so a rebroadcast can retry
+                    // once e.g. its payer is funded.
+                    match (self.screen)(tx.clone()).await {
+                        Err(reason) => {
+                            log::debug!("Not admitting gossiped tx {hash:?}: {reason}");
                         }
```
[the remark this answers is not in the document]
erhant: Considered and kept inline deliberately. 

The screen is static checks plus one balance read under the executor's state lock, the ask round-trip is the same cost RPC submission pays

Decoupling (spawn screen+admit, report acceptance immediately) breaks two semantics that live on `&mut self`: seen is only marked on a successful mempool push so a rebroadcast can retry after a full mempool or an unfunded payer, and that ordering doesn't survive a detached task without adding shared state.

Note that a flood of parseable messages is already rate-bounded by gossipsub itself before it reaches us.

Pravdyvy (approved):
lgtm

lez/storage/src/indexer/tests.rs
erhant, line 1: this part is about the test refactor alone; all others are gossip screening stuff

lez/sequencer/core/src/lib.rs
```diff
@@ -1030,6 +1002,14 @@ impl<BP: BlockPublisherTrait, S: StorageActorTrait> SequencerCore<BP, S> {
...
             // user-submitted ones, so they get the same full state validation.
             TransactionOrigin::User | TransactionOrigin::Gossip => {
+                // The cheap admission screen first: an unfundable or fee-invalid
+                // candidate is dropped before paying for the scratch clone and
+                // the settlement's guest executions.
+                if let Err(rejection) = fees::screen(tx, state) {
+                    log::debug!("Dropping candidate {tx_hash:?} at the fee screen: {rejection}");
+                    return false;
+                }
+
```
Arjentix, line 1040: I think you don't need this since it's checked in Executor Actor

lez/sequencer/core/src/gossip/network.rs
```diff
@@ -514,6 +528,22 @@ pub(crate) fn peer_id_from_ed25519(
         .map(|key| libp2p::identity::PublicKey::from(key).to_peer_id())
 }
 
+/// An [`IngestSubmit`] that pushes straight into `mempool` unscreened; for
+/// tests.
+#[must_use]
+pub fn unscreened_mempool_submit(
```
Arjentix, line 538: Let's mark it with `#[cfg(test)]`

Arjentix (approved):
LGTM, looking forward to make gossiper a separate actor

moudyellaz (approved):
lgtm, thanks!

lez/sequencer/core/src/gossip/network.rs
erhant: i was thinking it could be used for `standalone` mode as well

erhant: but i agree hahah, will do

lez/sequencer/core/src/lib.rs
erhant: the actor screening is done against the head state at **submission** time,
this one screens against the **evolving mid-block state**, without this we can have an attack like:
- fund one account with exactly one fee reserve
- submit N transactions from it
- all N pass the actor's screen (the balance covers the reserve at the door, each checked independently)
- normally the remaining N-1 ones should fail, but now they pass

lez/sequencer/core/src/gossip/network.rs
erhant: done in https://github.com/logos-blockchain/logos-execution-zone/pull/830/commits/9dcbe249a94a1c3ace2a40fdd6c809c5ff174023

PR #831, opened 2026-09-04
perf(lee): memoize guest memory images, keyed by ELF bytes

moudyellaz:
## 🎯 Purpose

Guest execution rebuilds the MemoryImage from the ELF on every call. That build
is ~10.4ms of an ~11.7ms in-process execution; the emulation itself is ~0.5ms.
The same few programs run over and over, so the image is identical every time
and thrown away every time.

This speeds up CI and `wallet-ffi`; the node still uses r0vm until `prove` is
enabled there.

## ⚙️ Approach

- Cache built `MemoryImage`s, byte-bounded with LRU eviction, recorded only
  after a clean run
- Keyed on ELF bytes, since `new_unchecked` lets `Program::id` diverge from
  them. A collision skips the cache, it never returns the wrong image
- Runs the ABI check and the claim the replaced path performs
- Under `prove` only; without it the executor is unchanged

Per call under `--all-features`: 11.74ms → 1.325ms. `-p storage`
`state_replay_falls_back_over_missing_breakpoints` 51s → 3.5s,
`put_block_stores_breakpoint_in_same_batch` 33s → 2.0s.

## 🧪 How to Test

```sh
RISC0_DEV_MODE=1 cargo test --release -p lee --features prove
RISC0_DEV_MODE=1 cargo test --release -p storage --features prove
```

## 🔗 Dependencies

None

## 🔜 Future Work

None

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests

### DIFF
--- a/lee/state_machine/src/program/image_cache/mod.rs
+++ b/lee/state_machine/src/program/image_cache/mod.rs
@@ -1,0 +1,171 @@
+//! Process-global cache of built `MemoryImage`s, keyed by the ELF bytes themselves.
+//!
+//! Keyed on bytes rather than `Program::id`, which `Program::new_unchecked` lets diverge: serving
+//! an image built from bytes other than the ones being charged for is a consensus fault. A
+//! per-process seeded hash picks the slot and a full byte comparison confirms it, so a collision
+//! skips the cache rather than returning the wrong image.
+
+use std::{
+    collections::{HashMap, hash_map::RandomState},
+    hash::{BuildHasher as _, Hasher as _},
+    sync::{Arc, Mutex, MutexGuard, OnceLock},
+};
+
+use anyhow::{Result, bail};
+use risc0_binfmt::{AbiKind, MemoryImage, ProgramBinary, ProgramBinaryHeader};
+use risc0_zkvm::{ExecutorEnv, ExecutorImpl, NullSegmentRef};
+
+use super::SessionOutcome;
+
+#[cfg(test)]
+mod tests;
+
+/// Bytes held at once. Programs are user-deployable, so the budget is bytes rather than a count
+/// of entries, and the least recently used entry is evicted first.
+const MAX_CACHED_BYTES: usize = 32 * 1024 * 1024;
+
+/// Mirrors risc0's private `check_program_version`: a different value would accept guests the
+/// proving path rejects.
+const SUPPORTED_ABI: &str = "^1.0.0";
+
+struct Entry {
+    elf: Box<[u8]>,
+    image: Arc<MemoryImage>,
+    bytes: usize,
+    used: u64,
+}
+
+#[derive(Default)]
+struct Cache {
+    entries: HashMap<u64, Entry>,
+    bytes: usize,
+    tick: u64,
+}
+
+fn cache() -> MutexGuard<'static, Cache> {
+    static CACHE: OnceLock<Mutex<Cache>> = OnceLock::new();
+    CACHE
+        .get_or_init(|| Mutex::new(Cache::default()))
+        .lock()
+        .unwrap_or_else(std::sync::PoisonError::into_inner)
+}
+
+/// Slot selector over the whole ELF, seeded once per process so colliding bytes cannot be
+/// constructed by whoever supplies the program.
+pub fn slot(elf: &[u8]) -> u64 {
+    static SEED: OnceLock<RandomState> = OnceLock::new();
+    let mut hasher = SEED.get_or_init(RandomState::new).build_hasher();
+    hasher.write(elf);
+    hasher.finish()
+}
+
+/// What an entry costs to hold: the ELF copy plus its image. An upper bound, not a measurement,
+/// since `MemoryImage` does not expose its size and measured images come in under their ELF. Only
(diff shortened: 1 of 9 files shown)


lee/state_machine/src/program/image_cache/mod.rs
```diff
@@ -0,0 +1,165 @@
...
+/// `ExecutorImpl::from_elf` runs this before building the image; `ExecutorImpl::new` does not, so
+/// the cached path has to do it itself or an incompatible guest would silently start executing.
+fn check_program_version(header: &ProgramBinaryHeader) -> Result<()> {
+    if header.abi_kind != AbiKind::V1Compat {
+        bail!(
+            "ProgramBinary abi_kind mismatch {:?} != AbiKind::V1Compat",
+            header.abi_kind
+        );
+    }
+    if !semver::VersionReq::parse("^1.0.0")
```
erhant: why is 1.0.0 fixed?

maybe we can put this as a const string somewhere and explain the reason in docstring

lee/state_machine/src/program/mod.rs
```diff
@@ -94,12 +111,22 @@ impl Program {
...
-        default_executor().execute(env, elf).map_err(|e| {
+    ) -> Result<SessionOutcome, LeeError> {
+        #[cfg(feature = "prove")]
+        let raw = image_cache::execute(env, elf);
+        #[cfg(not(feature = "prove"))]
+        let raw = default_executor()
+            .execute(env, elf)
+            .map(|info| SessionOutcome {
+                journal: info.journal.bytes.clone(),
+                cycles: info.cycles(),
```
erhant: if `info.cycles()` come first, we can perhaps get rid of the journal `.clone()` above

lee/state_machine/src/program/image_cache/mod.rs
moudyellaz: It mirrors risc0's private `check_program_version`. Hoisted to a const.

erhant (approved):
LGTM thanks!

Pravdyvy (approved):
lgtm

lez/storage/Cargo.toml
```diff
@@ -24,3 +24,7 @@ zstd.workspace = true
 lee_core = { workspace = true, features = ["host"] }
 programs.workspace = true
 system_accounts.workspace = true
+
+[features]
+# Runs guest execution in-process against the memoized image cache.
+prove = ["lee/prove"]
```
Arjentix, line 30: So where do we enable this feature?

lee/state_machine/src/program/image_cache/mod.rs
```diff
@@ -0,0 +1,169 @@
...
+    tick: u64,
+}
+
+fn cache() -> MutexGuard<'static, Cache> {
+    static CACHE: OnceLock<Mutex<Cache>> = OnceLock::new();
+    CACHE
+        .get_or_init(|| Mutex::new(Cache::default()))
+        .lock()
+        .unwrap_or_else(std::sync::PoisonError::into_inner)
+}
```
Arjentix, line 51: Not a fan of global state like that... But looks like we have no option for now

lee/state_machine/src/program/image_cache/mod.rs
```diff
@@ -0,0 +1,169 @@
...
+pub fn slot(elf: &[u8]) -> u64 {
+    static SEED: OnceLock<RandomState> = OnceLock::new();
+    let mut hasher = SEED.get_or_init(RandomState::new).build_hasher();
+    hasher.write(elf);
+    hasher.finish()
+}
+
+/// What an entry costs to hold: the ELF copy plus its image, which runs somewhat under the ELF.
+const fn footprint(elf: &[u8]) -> usize {
+    elf.len().saturating_mul(2)
```
Arjentix, line 66: Why multiply by 2? How can we know that?

Arjentix (approved):
LGTM

lee/state_machine/src/program/mod.rs
moudyellaz: Good catch, done.

lez/storage/Cargo.toml
moudyellaz: Nothing in the node. CI runs tests with --all-features so it gets the cache, and wallet-ffi already has prove on by default. The node keeps using r0vm until we enable prove there, which is a separate PR since it also pulls in the prover.

lee/state_machine/src/program/image_cache/mod.rs
moudyellaz: Agreed. The image has to outlive the per-call `ExecutorImpl`, and nothing on the execution path owns state across calls, so process-global is the only place it can sit.

lee/state_machine/src/program/image_cache/mod.rs
moudyellaz: It is an upper bound, not a measurement. `MemoryImage` does not expose its size, and the images we measured come in under their ELF, so ELF plus image stays under 2x. Only affects when eviction starts, never correctness. Added a comment saying so.

PR #832, opened 2026-09-04
feat(docker): introduce 4-node setup

Arjentix:
## 🎯 Purpose

Create multi sequencer node docker setup for local testing and as example for infra team.

## ⚙️ Approach

- Sequencer key can now be passed as cli argument
- Multi node configs generation tool
- Configs for these nodes
- Adjusted docker setup

## 🧪 How to Test

```
docker compose up --build
```

Then try in another terminal:
```
just run-wallet check-health
```

Or check grafana at `http://localhost:3000`. You'll see 4 stats

## 🔗 Dependencies

None

## 🔜 Future Work

None

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


tools/devnet_configs/src/main.rs
```diff
@@ -0,0 +1,174 @@
...
+    config.gossip = Some(GossipConfig {
+        listen_addr: GOSSIP_LISTEN_ADDR
+            .parse()
+            .expect("hardcoded gossip listen multiaddr is valid"),
+        bootstrap_peers: vec![],
+    });
+    // Each node is given its own with --signing-key instead.
+    config.signing_key = None;
+
+    let signing_keys: Vec<[u8; 32]> = (0..NODES).map(signing_key).collect();
```
schouhy: this is using a determinstic key derived from the index of the sequencer no? for devnet I think we should sample a random one

Arjentix: In DM we settled that it's okay, because infra team is more interested in configs, not this algorithm which generates. While determinism is good here for local reproducibility

schouhy (approved):
LGTM! only thing I'd change is replace the usage of determinstic keys for stake accounts by random ones

moudyellaz (approved):
lgtm, thanks

erhant (approved):
LGTM!

lez/sequencer/core/src/config.rs
```diff
@@ -89,8 +89,13 @@ pub struct SequencerConfig {
...
     pub retry_pending_blocks_timeout: Duration,
-    /// Sequencer own signing key.
-    pub signing_key: [u8; 32],
+    /// The key this sequencer signs its blocks with.
+    ///
+    /// Absent when the node is given one separately (`--signing-key`), which is
+    /// what lets a set of sequencers share a single config file — it is all
+    /// that otherwise differs between them.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub signing_key: Option<[u8; 32]>,
```
erhant, line 98: I wonder if we should use `PrivateKey` here directly with a custom `serialize_with` etc.
would save us from parsing the key on each call, but negligible really

Arjentix: Created https://github.com/logos-blockchain/logos-execution-zone/issues/836 for that. For now let's merge it

PR #833, opened 2026-09-04
refactor(gossip): make gossip into an actor

erhant:
## 🎯 Purpose

Turns the sequencer's p2p gossip module into a `kameo` actor of its own, resolving the `// TODO: Should be a separate actor` seam in the service's `run()` (suggested by @Arjentix alongside the actor-encapsulation work). 
It now lives as `sequencer_gossip_actor` under `lez/sequencer/actors/`, following the same crate layout and `actor` feature convention as the executor, storage, and RPC server actors (#798).

## ⚙️ Approach

The former `GossipNetwork` drive task *is* the actor now: `GossipActor` owns the libp2p swarm and drives it from an `Actor::next()` override (mirroring `RpcServerActor::next`), selecting over the mailbox, swarm events, and the bootstrap-retry tick. The mailbox replaces the hand-rolled plumbing wholesale:

- [x] Move the whole `sequencer_core::gossip` module to a new `sequencer_gossip_actor` crate at `lez/sequencer/actors/gossip`; rewrite `network.rs` as `actor.rs`.
- [x] Replace the spawned drive task + `CancellationToken` + publish `mpsc` channel + `watch` channel with the actor mailbox: `PublishTransaction` (tell), `GetConnectedPeers` (ask), and `RetryBootstrap` (tell, sent by the service scheduler every 30s, same shape as `ProduceBlock`). The rpc_server publishes through a type-erased `Recipient<PublishTransaction>` with `tell(..).try_send()`, keeping the old non-blocking, drop-on-overflow publish semantics (bounded mailbox, same 1024 depth).
- [x] Implement `on_stop`: the swarm is `Option`-held and dropped there, so the QUIC sockets are freed the moment the actor reports stopped rather than when the struct is dropped — a shutdown observer can rebind immediately.
- [x] Strong types per review: peer keys carried as `Ed25519PublicKey` (map + `GetConnectedPeers` reply, sorted deterministically via `to_bytes`), and the pending-publish queue is a new `common::bounded_vec_deque::BoundedVecDeque` whose eviction is an invariant of the type.
- [x] Preserve the degrade-to-L1 invariant: the gossip `ActorHandle` is deliberately excluded from `SequencerHandle::failed()`/`is_healthy()` (documented on the field). The old driver watchdog becomes `spawn_gossip_outage_watchdog`: it warns every 5 minutes only on an *abnormal* actor stop (panic), stays silent on graceful stop/kill, and its guard aborts the warner on node shutdown.
- [x] Keep startup fail-fast: `GossipActor::new` binds and waits for the listen address before the actor is spawned; bootstrap addrs are captured pre-spawn so `SequencerHandle::gossip_bootstrap_addrs()` stays sync for the multi-node fixtures. Graceful gossip shutdown added to `SequencerHandle::shutdown()`.
- [x] Adopt the #798 `actor` feature convention: without it the crate is protocol-only (`protocol` messages + `accreditation`; `validation` and `seen_cache` are private actor implementation details); `actor` pulls in kameo/libp2p/tokio/borsh and the implementation, with tests declared inside the `actor` module like the other actor crates. Only the service enables it — rpc_server consumes the crate featureless, exactly as it consumes the executor. The `mdns` feature moves here from `sequencer_core` (service forwards it; it implies `actor`).
- [x] Delete the dead `gossip/message.rs` wire-format file — it was never declared as a module (uncompiled), and `sequencer_stake_core::SlashApproval` is the real slash-approval type.

`GossipConfig` stays in `sequencer_core::config` (it is embedded in `SequencerConfig`), which is the one reason core keeps its libp2p dependency. The ingest path is unchanged: gossiped transactions still enter through the executor's admission door via the `IngestSubmit` hook wired in the service.

## 🧪 How to Test

```sh
# check
cargo check -p sequencer_gossip_actor

# test
RISC0_DEV_MODE=1 cargo test -p sequencer_gossip_actor --all-features

# run a sequencer
just run-sequencer
```

## 🔗 Dependencies

None

## 🔜 Future Work

- A mockable publish path if the rpc_server's publish path ever needs testing without a swarm — it holds a type-erased `Recipient<PublishTransaction>`, so a test can hand it a recipient of any lightweight actor.
- Kameo `link`/`supervision` instead of the warn-only outage watchdog — needs defined restart semantics (rebind + publisher re-wiring); `on_stop` already frees the sockets so a supervised restart is possible when we get there.
- Actor supervision/respawn instead of warn-and-restart on a gossip panic; needs re-bind handling and publisher re-wiring, deferred until an actual gossip crash is observed in the wild.
- A gossip-health surface (metric or getter): `is_healthy()` intentionally stays `true` when gossip is down, so the 5-minute log is currently the only operator signal.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


lez/sequencer/core/src/gossip/message.rs
```diff
@@ -1,20 +0,0 @@
...
-/// One gossip payload; the Borsh discriminant tags the variant.
-#[derive(Debug, Clone, BorshSerialize, BorshDeserialize)]
-pub enum GossipMessage {
-    Transaction(LeeTransaction),
-    SlashApproval(SlashApprovalMessage),
-}
-
-/// One sequencer's signature over an offence, for peers to collect.
-#[derive(Debug, Clone, PartialEq, Eq, BorshSerialize, BorshDeserialize)]
-pub struct SlashApprovalMessage {
```
erhant: @schouhy I'm guessing these were going to be used for slashing k/n right? i removed these so that it is more direct for your slashing PR to add later, if that is ok

erhant: reason for removal: they were dead code at the moment

lez/sequencer/service/src/lib.rs
```diff
@@ -29,30 +29,39 @@ pub struct SequencerHandle {
     // NOTE: Order of fields matters as it affects drop order.
     scheduler: ActorHandle<Scheduler>,
     rpc_server: ActorHandle<RpcServerActor>,
+    /// Deliberately NOT part of [`Self::failed`]/[`Self::is_healthy`]: gossip
+    /// going down degrades the node to L1-only, it never halts it (the
+    /// watchdog warns operators instead). `None` when gossip is unconfigured.
```
Arjentix:
```suggestion
```

I'd like to clean the repo from unnecessary Claude's comments

lez/sequencer/service/src/lib.rs
```diff
@@ -29,30 +29,39 @@ pub struct SequencerHandle {
...
     addr: SocketAddr,
-    /// Held for its lifetime: dropping it stops the gossip drive task.
-    /// `None` when gossip is unconfigured.
-    gossip: Option<sequencer_core::gossip::GossipNetwork>,
+}
+
+/// The gossip actor and its companions.
+struct Gossip {
+    actor: ActorHandle<sequencer_gossip_actor::GossipActor>,
+    bootstrap_addrs: Vec<sequencer_gossip_actor::Multiaddr>,
```
Arjentix: Haven't explored gossiper architecture yet, but shouldn't this be inside `GossipActor`?

lez/sequencer/service/src/lib.rs
```diff
@@ -87,6 +99,7 @@ impl SequencerHandle {
             scheduler,
             storage,
             addr: _,
+            // A gossip failure never halts the node; see the field docs.
```
Arjentix: Let's better move that doc here

lez/sequencer/service/src/lib.rs
```diff
@@ -118,6 +131,7 @@ impl SequencerHandle {
...
```
Arjentix: And somewhat here

lez/sequencer/service/src/lib.rs
```diff
@@ -167,20 +181,19 @@ pub fn run(
         let executor_ref = ExecutorActor::spawn(executor);
         info!("Executor Actor spawned");
 
-        // TODO: Should be a separate actor
-        let gossip_network = match gossip_config {
-            None => None,
+        let (gossip, tx_publisher) = match gossip_config {
```
Arjentix: I think we should remove `tx_publisher` and instead pass `ActorRef<GossipActor>` to `RpcServerActor`. Or maybe abstracted version like [`Recipient`](https://docs.rs/kameo/latest/kameo/actor/struct.Recipient.html). So `RpcServerActor` will publish to both `ExecutorActor` And `GossipActor`.

Maybe in ideal future there should be some kind of queue between server and executor / gossiper. But I'd keep it simple for now and later will see

lez/sequencer/service/src/lib.rs
```diff
@@ -29,30 +29,39 @@ pub struct SequencerHandle {
...
-    /// `None` when gossip is unconfigured.
-    gossip: Option<sequencer_core::gossip::GossipNetwork>,
+}
+
+/// The gossip actor and its companions.
+struct Gossip {
+    actor: ActorHandle<sequencer_gossip_actor::GossipActor>,
+    bootstrap_addrs: Vec<sequencer_gossip_actor::Multiaddr>,
+    /// Aborts the gossip outage warner when the handle is dropped.
+    _watchdog: sequencer_gossip_actor::WatchdogGuard,
```
Arjentix, line 50: I'd try to explore ways of removing that in this PR. Probably exploring these things will be useful:
- [`link`](https://docs.rs/kameo/latest/kameo/actor/struct.ActorRef.html#method.link)
- [`supervision`](https://docs.rs/kameo/latest/kameo/supervision/index.html)

lez/sequencer/service/src/lib.rs
Arjentix: Consider moving to `setup_gossip_actor()` or something like that

lez/sequencer/actors/gossip/Cargo.toml
```diff
@@ -0,0 +1,42 @@
+[package]
+name = "sequencer_gossip_actor"
```
Arjentix, line 2: Btw, what do you think about `GossiperActor` and according `gossiper` naming? IMO entity-like naming fits actors well. I just could not came up with a better name for `StorageActor`. But everyone else: `RpcServerActor`, `ExecutorActor` sounds like entities

lez/sequencer/actors/gossip/Cargo.toml
```diff
@@ -0,0 +1,42 @@
...
+[dependencies]
+common.workspace = true
+sequencer_core.workspace = true
+
+logos-blockchain-core.workspace = true
+logos-blockchain-key-management-system-service.workspace = true
+logos-blockchain-zone-sdk.workspace = true
+
+anyhow.workspace = true
+borsh.workspace = true
```
Arjentix: I wonder why this is not enabled only by `actor` feature

lez/sequencer/actors/gossip/Cargo.toml
```diff
@@ -0,0 +1,42 @@
...
+# All mDNS code lives in the actor, so this implies `actor`.
+mdns = ["actor"]
+
+[dependencies]
+common.workspace = true
+sequencer_core.workspace = true
+
+logos-blockchain-core.workspace = true
+logos-blockchain-key-management-system-service.workspace = true
+logos-blockchain-zone-sdk.workspace = true
```
Arjentix: Interesting, I did not expect that dependencies from gossiper

lez/sequencer/service/src/lib.rs
Arjentix: Oh yeah, I can imagine a [topic-based message queue](https://docs.rs/kameo_actors/latest/kameo_actors/message_queue/index.html) which we'll setup in the following way:

```mermaid
sequenceDiagram
    actor RpcServerActor
    participant MessageQueue@{ "type" : "queue" }
    actor GossipActor
    actor ExecutorActor

    RpcServerActor->>+MessageQueue: "transactions/user": Transaction
    MessageQueue->>ExecutorActor: "transactions/user": Transaction
    MessageQueue->>-GossipActor: "transactions/user": Transaction
    GossipActor->>+MessageQueue: "transactions/gossip": Transaction
    MessageQueue->>-ExecutorActor: "transactions/gossip": Transaction
```

(Lol, didn't know mermaid is that good)

lez/sequencer/service/src/lib.rs
```diff
@@ -194,7 +207,7 @@ pub fn run(
                             executor_submit_ref.ask(message).await.map_err(Into::into)
                         })
                     });
```
Arjentix, line 218: Let's implement the idea form [this comment](https://github.com/logos-blockchain/logos-execution-zone/pull/833/changes#r3952179989) and remove this

lez/sequencer/service/src/lib.rs
Arjentix: Ah, probably [`topic-based broker`](https://docs.rs/kameo_actors/latest/kameo_actors/broker/index.html) will be simpler here as we use just one `Transaction` type. But the idea is the same.

lez/sequencer/actors/gossip/src/lib.rs
```diff
@@ -0,0 +1,25 @@
...
+#[cfg(feature = "actor")]
+pub use libp2p::Multiaddr;
+pub use protocol::{GetConnectedPeers, GossipTxPublisher, PublishTransaction};
+
+pub mod accreditation;
+#[cfg(feature = "actor")]
+pub mod actor;
+pub mod protocol;
+pub mod seen_cache;
+pub mod validation;
```
Arjentix: I have a feeling that these modules should be actor implementation detail, no?

lez/sequencer/actors/gossip/src/lib.rs
```diff
@@ -0,0 +1,25 @@
...
+
+#[cfg(feature = "actor")]
+#[cfg(test)]
+mod tests;
```
Arjentix: Let's move into `actor` module as it's done now with other actors

lez/sequencer/actors/gossip/src/lib.rs
```diff
@@ -0,0 +1,25 @@
...
+//! down degrades to L1-only behavior, and a gossip failure after startup
+//! never halts the node.
+
+#[cfg(all(test, feature = "actor"))]
+pub use actor::unscreened_mempool_submit;
+#[cfg(feature = "actor")]
+pub use actor::{GossipActor, IngestSubmit, WatchdogGuard, spawn_gossip_outage_watchdog};
+#[cfg(feature = "actor")]
+pub use libp2p::Multiaddr;
+pub use protocol::{GetConnectedPeers, GossipTxPublisher, PublishTransaction};
```
Arjentix: Hm, other actor don't expose protocol types here, not sure you really need this

lez/sequencer/actors/gossip/src/protocol.rs
```diff
@@ -0,0 +1,32 @@
+//! Messages the gossip actor accepts, consumable without the `actor` feature.
```
Arjentix:
```suggestion
//! Messages the gossip actor accepts.
```

lez/sequencer/actors/gossip/src/protocol.rs
```diff
@@ -0,0 +1,32 @@
...
+
+use common::transaction::LeeTransaction;
+use kameo::actor::Recipient;
+
+/// Ask: Ed25519 public keys of currently connected peers.
+pub struct GetConnectedPeers;
+
+/// Tell: publish a locally-submitted transaction to the mesh.
+pub struct PublishTransaction(pub LeeTransaction);
```
Arjentix: Let's not expose how user should call it `ask` or `tell`. I hate when Claude writes docs how user should use stuff while breaking abstractions

lez/sequencer/actors/gossip/src/actor.rs
```diff
@@ -19,21 +27,24 @@ use libp2p::{
 use logos_blockchain_key_management_system_service::keys::Ed25519Key;
 #[cfg(test)]
 use mempool::MemPoolHandle;
```
Arjentix: I can see a lot of `#[cfg(test)]` in this file which as bit sus to me

Arjentix: Btw mempool hopefully won't be needed here after finishing the refactor

lez/sequencer/actors/gossip/src/actor.rs
```diff
@@ -54,13 +65,31 @@ struct GossipBehaviour {
...
-    connected_rx: watch::Receiver<Vec<[u8; 32]>>,
-    shutdown: CancellationToken,
+/// The gossip network as an actor: owns the libp2p swarm and drives it from
+/// [`Actor::next`], interleaving swarm events with mailbox messages
+/// ([`PublishTransaction`], [`GetConnectedPeers`]).
+///
+/// Stopping the actor (or killing it via its handle) shuts the swarm down.
+/// A gossip failure never halts the node: the service keeps this actor out
+/// of its health/failure aggregation and a [`spawn_gossip_outage_watchdog`] warns
+/// operators instead.
```
Arjentix:
```suggestion
/// The gossip actor. Manages the libp2p swarm.
```

lez/sequencer/actors/gossip/src/actor.rs
```diff
@@ -54,13 +65,31 @@ struct GossipBehaviour {
...
+pub struct GossipActor {
+    swarm: Swarm<GossipBehaviour>,
+    connected: HashSet<PeerId>,
+    /// Ed25519 public keys of peers seen via Identify, keyed by `PeerId`.
+    pubkeys: HashMap<PeerId, [u8; 32]>,
```
Arjentix: Can we have a strong type for Ed25519 key here as well?

lez/sequencer/actors/gossip/src/actor.rs
```diff
@@ -54,13 +65,31 @@ struct GossipBehaviour {
...
+    topic: gossipsub::IdentTopic,
+    seen: SeenCache,
+    max_block_size: u64,
+    submit: IngestSubmit,
+    /// Configured bootstrap peers, re-dialed while the node is isolated.
+    bootstrap: Vec<Multiaddr>,
+    bootstrap_retry: tokio::time::Interval,
```
Arjentix: Hmmmmmm, when I see a retry in actor I immediately think about our current approach with scheduler and executor. So scheduler sends `ConstructBlock` message periodically and `Executor` performs its actions. Maybe this can be applied here as well and we can move time interval from this actor?

Not insisting, just an idea.

lez/sequencer/actors/gossip/src/actor.rs
```diff
@@ -54,13 +65,31 @@ struct GossipBehaviour {
...
+    /// Local transactions whose publish failed, retried when a peer
+    /// subscribes to the topic. Bounded; the oldest is dropped on overflow.
+    pending_publish: VecDeque<LeeTransaction>,
```
Arjentix: I suggest having a strong type like `BoundedVecDequeue` here. To have behavior as an incapsulated invariant. Either researching existing crates or creating our own type is fine to me. I think I've already saw some places where this type could be useful.

lez/sequencer/actors/gossip/src/actor.rs
```diff
@@ -71,24 +100,15 @@ pub struct GossipNetwork {
...
+/// node shutdown.
+pub struct WatchdogGuard(tokio::task::JoinHandle<()>);
 
-impl GossipNetwork {
+impl GossipActor {
     /// Builds the swarm, binds `listen_addr`, seeds Kademlia and dials
-    /// bootstrap peers, and spawns the drive task.
-    pub async fn start(
+    /// bootstrap peers. Call [`Self::spawn`] on the result to start driving
+    /// it; the pre-spawn getters below expose the bound identity.
```
Arjentix:
```suggestion
    /// bootstrap peers.
```

We're using kameo framework, we already know we need to call spawn on all actors

lez/sequencer/actors/gossip/src/actor.rs
```diff
@@ -206,35 +226,39 @@ impl GossipNetwork {
...
         })
     }
 
+    /// Spawns the actor with a mailbox sized for local-publish bursts.
+    pub fn spawn(actor: Self) -> ActorRef<Self> {
+        <Self as kameo::actor::Spawn>::spawn_with_mailbox(
+            actor,
+            kameo::mailbox::bounded(MAILBOX_CAPACITY),
+        )
+    }
```
Arjentix: Oh, wait, wth, let's not mess up with kameo's builtin stuff by creating function with the same name, let's spawn it like that in sequencer service instead

lez/sequencer/actors/gossip/src/actor.rs
```diff
@@ -501,6 +469,65 @@ impl DriveTask {
     }
 }
 
+impl Actor for GossipActor {
```
Arjentix, line 562: Btw, consider implementing `on_stop()`. I can see there are a lot of stuff going on in this Actor. And I can see there's a drop impl doing something important. But unfortunately and quite unexpectedly in kameo actor can be considered stopped without being dropped yet. This means a caller might call [`wait_for_shutdown_with_result()`](https://docs.rs/kameo/latest/kameo/actor/struct.ActorRef.html#method.wait_for_startup_with_result) and expect that after that call all resources are freed. But if you free them only when you drop the actor this won't be true. And things like immediately restarting your actor might fail because of that.

I encountered that issue while working on `StorageActor`, take a look at its `on_stop()` implementation.

I've also created https://github.com/tqwewe/kameo/issues/397.

Arjentix (commented):
Glad to see it's being moved to a separate actor, nice work!

lez/sequencer/actors/gossip/Cargo.toml
erhant: I think it should stay as `gossip`, it's also a noun on it's own (i.e. a gossip is someone who gossips)

> I just could not came up with a better name for StorageActor

on that note, `StorerActor` would look VERY ugly hahah

lez/sequencer/service/src/lib.rs
erhant: yeah yeah, this is partly why this PR was draft hahah

lez/sequencer/actors/gossip/src/actor.rs
erhant: Good point! Added in https://github.com/logos-blockchain/logos-execution-zone/pull/833/commits/874ab272894fad30abb16479516d954dfa55b804

lez/sequencer/actors/gossip/src/actor.rs
erhant: Yes indeed, done in https://github.com/logos-blockchain/logos-execution-zone/pull/833/commits/874ab272894fad30abb16479516d954dfa55b804

Now spawned like `GossipActor::spawn_with_mailbox(actor, kameo::mailbox::bounded(MAILBOX_CAPACITY))`

lez/sequencer/actors/gossip/src/actor.rs
erhant: Fixed in https://github.com/logos-blockchain/logos-execution-zone/commit/874ab272894fad30abb16479516d954dfa55b804

lez/sequencer/actors/gossip/src/actor.rs
erhant: Rolled our own in https://github.com/logos-blockchain/logos-execution-zone/commit/874ab272894fad30abb16479516d954dfa55b804

I do see a crate but its VERY old with extra features, so I think we can go with self-rolled ones here; im doing FIFO so no complicated stuff as well

>  I think I've already saw some places where this type could be useful.

I've put it in `common` so other parts can use it too if needed

lez/sequencer/actors/gossip/src/actor.rs
erhant: I think it makes sense, added the scheduler registration in https://github.com/logos-blockchain/logos-execution-zone/commit/874ab272894fad30abb16479516d954dfa55b804

```rs
scheduler_ref
        .tell(
            SetInterval::new(
                gossip_ref.downgrade(),
                sequencer_gossip_actor::BOOTSTRAP_RETRY_INTERVAL,
                sequencer_gossip_actor::protocol::RetryBootstrap,
            )
            .start_delay(sequencer_gossip_actor::BOOTSTRAP_RETRY_INTERVAL),
        )
        .await?;
```

lez/sequencer/actors/gossip/src/actor.rs
erhant: Done in https://github.com/logos-blockchain/logos-execution-zone/commit/874ab272894fad30abb16479516d954dfa55b804

lez/sequencer/actors/gossip/src/actor.rs
erhant: Done in https://github.com/logos-blockchain/logos-execution-zone/commit/874ab272894fad30abb16479516d954dfa55b804

lez/sequencer/actors/gossip/src/actor.rs
erhant: Moved them to `tests.rs` helper in https://github.com/logos-blockchain/logos-execution-zone/commit/874ab272894fad30abb16479516d954dfa55b804

> Btw mempool hopefully won't be needed here after finishing the refactor

Indeed, note that mempool is a dev dependency for gossip!

lez/sequencer/actors/gossip/src/protocol.rs
erhant: Claude told me that he is sorry that you think like that 😢 

(fixed in https://github.com/logos-blockchain/logos-execution-zone/commit/874ab272894fad30abb16479516d954dfa55b804)

lez/sequencer/actors/gossip/src/protocol.rs
erhant: Done in https://github.com/logos-blockchain/logos-execution-zone/commit/874ab272894fad30abb16479516d954dfa55b804

lez/sequencer/actors/gossip/src/lib.rs
erhant: Yeah, now we import like `sequencer_gossip_actor::protocol::` like the others, done in https://github.com/logos-blockchain/logos-execution-zone/commit/874ab272894fad30abb16479516d954dfa55b804

lez/sequencer/actors/gossip/src/lib.rs
erhant: Done in https://github.com/logos-blockchain/logos-execution-zone/commit/874ab272894fad30abb16479516d954dfa55b804

lez/sequencer/actors/gossip/src/lib.rs
erhant: Yep, done in https://github.com/logos-blockchain/logos-execution-zone/commit/874ab272894fad30abb16479516d954dfa55b804 and moved under `/actor/...` as also noted in your comment below https://github.com/logos-blockchain/logos-execution-zone/pull/833#discussion_r3952205471

lez/sequencer/actors/gossip/Cargo.toml
erhant: These belong to `accreditation` thing, about the L1 `accredited_keys` logic

We could move it elsewhere perhaps? In some other PR maybe, we do need to take a look at this after slashing etc. too

lez/sequencer/actors/gossip/Cargo.toml
erhant: Fixed, `kameo + log + borsh` are now actor-optional in https://github.com/logos-blockchain/logos-execution-zone/commit/874ab272894fad30abb16479516d954dfa55b804

lez/sequencer/service/src/lib.rs
erhant: Done in https://github.com/logos-blockchain/logos-execution-zone/commit/874ab272894fad30abb16479516d954dfa55b804

I could perhaps move it once more though, don't consider this resolved yet!

lez/sequencer/service/src/lib.rs
erhant: Removed a bit in https://github.com/logos-blockchain/logos-execution-zone/commit/874ab272894fad30abb16479516d954dfa55b804

lez/sequencer/service/src/lib.rs
erhant: Done in https://github.com/logos-blockchain/logos-execution-zone/commit/874ab272894fad30abb16479516d954dfa55b804

lez/sequencer/service/src/lib.rs
erhant: Done in https://github.com/logos-blockchain/logos-execution-zone/commit/874ab272894fad30abb16479516d954dfa55b804 akin to https://github.com/logos-blockchain/logos-execution-zone/pull/833#discussion_r3952032512

lez/sequencer/service/src/lib.rs
erhant: `GossipTxPublisher` is gone in https://github.com/logos-blockchain/logos-execution-zone/commit/874ab272894fad30abb16479516d954dfa55b804 and `rpc_server` can now use `Option<Recipient<PublishTransaction>>` directly

I would like to revisit the topic based mq stuff later perhaps!

erhant: also yeah mermaid is love <3

lez/sequencer/service/src/lib.rs
erhant: nice ideas, but I think we can revisit this later in a more contained PR about such a refactor

two things:
- `link` requires a second actor, but _which one?_ is a question one may ask. The answer to that is currently "the watchdog"
- `supervision` requires a well defined restart logic, which we do not have for gossip; it could require a bit more care than this PR I think

> also as per https://github.com/logos-blockchain/logos-execution-zone/pull/833#discussion_r3952319969 I've added `on_stop` which does pave the way for a more graceful restart, but still!

lez/sequencer/service/src/lib.rs
erhant: Done in https://github.com/logos-blockchain/logos-execution-zone/commit/874ab272894fad30abb16479516d954dfa55b804

lez/sequencer/service/src/lib.rs
erhant: Keeping it inside `sequencer/service/src/lib.rs` it seems to be the correct place.

Did a small refactor in https://github.com/logos-blockchain/logos-execution-zone/pull/833/commits/87e30897d500e538bd2e571aa66d5b30a067602d

lez/common/src/bounded_vec_deque.rs
```diff
@@ -0,0 +1,66 @@
+use std::collections::VecDeque;
+
+/// A FIFO queue with a fixed capacity: pushing onto a full queue evicts and
+/// returns the oldest element, so the bound is an invariant rather than a
+/// check every caller must remember.
```
Arjentix:
```suggestion
/// returns the oldest element.
```

lez/sequencer/actors/rpc_server/src/actor/service.rs
```diff
@@ -117,8 +117,12 @@ impl<E: ExecutorActorTrait> sequencer_service_rpc::RpcServer for Service<E> {
...
         // refused. (A full local mempool has already errored above, so a
         // publish cannot be lost to it either.)
-        if let Some(publisher) = &self.gossip_tx_publisher {
-            publisher.publish(authenticated_tx);
+        // Non-blocking: a full gossip mailbox sheds the publish instead of
+        // stalling the RPC response.
+        if let Some(gossip) = &self.gossip
+            && let Err(err) = gossip.tell(PublishTransaction(authenticated_tx)).try_send()
+        {
+            debug!("Dropping local tx publish: gossip mailbox full or closed: {err}");
```
Arjentix:
```suggestion
            warn!("Dropping local tx publish: gossip mailbox full or closed: {err}");
```

lez/sequencer/actors/gossip/Cargo.toml
```diff
@@ -0,0 +1,50 @@
...
```
Arjentix: Shouldn't it be gated by actor feature as well?

lez/sequencer/actors/gossip/src/lib.rs
```diff
@@ -0,0 +1,19 @@
...
+
+#[cfg(feature = "actor")]
+pub use actor::{
+    BOOTSTRAP_RETRY_INTERVAL, GossipActor, IngestSubmit, MAILBOX_CAPACITY, WatchdogGuard,
+    spawn_gossip_outage_watchdog,
+};
+#[cfg(feature = "actor")]
+pub use libp2p::Multiaddr;
+
+pub mod accreditation;
```
Arjentix: So this is not just actor implementation detail? Where it's used then?

lez/sequencer/service/src/lib.rs
Arjentix: Ok, let's do it separately, created https://github.com/logos-blockchain/logos-execution-zone/issues/870

lez/sequencer/service/src/lib.rs
Arjentix: Created https://github.com/logos-blockchain/logos-execution-zone/issues/871

lez/common/src/bounded_vec_deque.rs
erhant: done in https://github.com/logos-blockchain/logos-execution-zone/pull/833/commits/e282307e7ec6978dfaa5fcc1624a7f04771394e1

lez/sequencer/actors/rpc_server/src/actor/service.rs
erhant: done in https://github.com/logos-blockchain/logos-execution-zone/pull/833/commits/e282307e7ec6978dfaa5fcc1624a7f04771394e1

lez/sequencer/actors/gossip/src/lib.rs
erhant: It was to be kept for slashing & gathering signatures for it, but looking at https://github.com/logos-blockchain/logos-execution-zone/pull/818/changes#r3968326121 I can see that this is not needed anymore

let me check in with @schouhy on this to be sure

lez/sequencer/actors/gossip/Cargo.toml
erhant: It is used by accreditation, https://github.com/logos-blockchain/logos-execution-zone/pull/833#discussion_r3968449998

Depending on the resolution of that comment we may remove accreditation module all together

lez/sequencer/actors/gossip/src/lib.rs
erhant: asked in https://github.com/logos-blockchain/logos-execution-zone/pull/818#pullrequestreview-5154625890

lez/sequencer/actors/gossip/src/actor/tests.rs
```diff
@@ -42,59 +101,123 @@ fn invalidly_signed_transaction() -> LeeTransaction {
...
+        [1; 32],
+        Ed25519Key::from_bytes(&[9; 32]),
+        TEST_MAX_BLOCK_SIZE,
+        unscreened_mempool_submit(test_mempool_handle()),
+    )
+    .await
+    .unwrap();
+    let actor_ref = spawn(actor);
+    actor_ref.kill();
+    tokio::time::timeout(Duration::from_secs(5), actor_ref.wait_for_shutdown())
```
moudyellaz: `wait_for_shutdown()` resolves when kameo drops the mailbox in `notify_links`, which runs before `on_stop`, so this returns with the swarm possibly still bound; `wait_for_shutdown_result()` plus a rebind of `listen_addrs[0]` would actually pin the new `on_stop`.

lez/sequencer/actors/gossip/src/actor.rs
```diff
@@ -493,11 +442,93 @@ impl DriveTask {
...
                 log::debug!("Failed to dial gossip bootstrap peer {addr}: {err}");
             }
         }
-        _ = self.swarm.behaviour_mut().kademlia.bootstrap();
+        _ = self.swarm_mut().behaviour_mut().kademlia.bootstrap();
+    }
+}
+
+impl Actor for GossipActor {
+    type Args = Self;
```
moudyellaz, line 563: With `Args = Self` and `on_stop` taking the swarm, a supervised restart hands `on_start` a value whose `swarm` is `None` and the first `next()` panics on `swarm_mut()`; should `Args` carry the config so `on_start` can rebuild and rebind?

lez/sequencer/actors/rpc_server/src/actor/service.rs
```diff
@@ -117,8 +117,12 @@ impl<E: ExecutorActorTrait> sequencer_service_rpc::RpcServer for Service<E> {
...
         // refused. (A full local mempool has already errored above, so a
         // publish cannot be lost to it either.)
-        if let Some(publisher) = &self.gossip_tx_publisher {
-            publisher.publish(authenticated_tx);
+        // Non-blocking: a full gossip mailbox sheds the publish instead of
+        // stalling the RPC response.
+        if let Some(gossip) = &self.gossip
+            && let Err(err) = gossip.tell(PublishTransaction(authenticated_tx)).try_send()
+        {
+            log::warn!("Dropping local tx publish: gossip mailbox full or closed: {err}");
```
moudyellaz, line 125: Gossip being down is a supported steady state and the 5-minute watchdog is the operator signal, so should this stay at `debug` rather than warn on every `sendTransaction` for as long as the node runs L1-only?

moudyellaz (approved):

lez/sequencer/actors/gossip/Cargo.toml
moudyellaz: `accreditation` has no consumers anywhere in the tree, so it is the only thing keeping a featureless build from being protocol-only; removing it drops these three deps rather than gating them.

lez/sequencer/actors/rpc_server/src/actor/service.rs
erhant: I had turned it from `debug` to `warn` on https://github.com/logos-blockchain/logos-execution-zone/pull/833#discussion_r3967811311 actually, but rethinking it, it makes sense to keep it `debug`

we do have a plan to supervise gossip later (https://github.com/logos-blockchain/logos-execution-zone/issues/871) which would give another chance for this kind of issue to be surfaced at a non-debug level at a much less frequency

lez/sequencer/actors/gossip/src/actor/tests.rs
erhant: good point, improved test in https://github.com/logos-blockchain/logos-execution-zone/pull/833/commits/9603a7cdd73c511a41654e78028cfe566c8521d6

lez/sequencer/actors/gossip/src/actor.rs
erhant: that is true, but currently we do not restart the gossip, that will be handled in a more kameo-native way via supervision probably, https://github.com/logos-blockchain/logos-execution-zone/issues/871 tracks it

moudyellaz (approved):

Arjentix (approved):
LGTM, waiting for resolving older comments related to accreditation

lez/sequencer/actors/gossip/src/lib.rs
erhant: We have decided that we do not need accreditation, #818 was merged & I have rebased this PR and removed `accreditation` in https://github.com/logos-blockchain/logos-execution-zone/pull/833/commits/34485dc7051bad12b963cf84cd3818c477f85661

lez/sequencer/actors/gossip/Cargo.toml
erhant: We have decided that we do not need accreditation, #818 was merged & I have rebased this PR and removed `accreditation` in https://github.com/logos-blockchain/logos-execution-zone/pull/833/commits/34485dc7051bad12b963cf84cd3818c477f85661

lez/sequencer/service/src/lib.rs
```diff
@@ -279,11 +251,112 @@ pub fn run(
...
+        kameo::mailbox::bounded(sequencer_gossip_actor::MAILBOX_CAPACITY),
+    );
+    info!("Gossip Actor spawned");
+    let watchdog = sequencer_gossip_actor::spawn_gossip_outage_watchdog(gossip_ref.clone());
+
+    // The slasher's own approvals flow into the mesh through the gossip
+    // mailbox; the channel bridges its `SetApprovalPublisher` API.
+    let (approval_tx, mut approval_rx) =
+        tokio::sync::mpsc::channel(OUTBOUND_APPROVAL_CHANNEL_CAPACITY);
+    slasher_ref.tell(SetApprovalPublisher(approval_tx)).await?;
```
Arjentix, line 330: Ahh, we're using tokio's channel together with actor system, that's weird, hope we refactor this in future

Arjentix (approved):

lez/sequencer/service/src/lib.rs
erhant: This comes from #818 via `SetApprovalPublisher(pub tokio::sync::mpsc::Sender<Approval>)`

erhant: I think we can refactor it later once the smoke settles a bit (with RCs and all that)

PR #834, opened 2026-09-05
feat(lez): move to 2/3 threshold signatures for Bedrock channel config

schouhy:
## 🎯 Purpose

Currently a channel config update only needs one signature, so any accredited node can rotate the committee on its own. This PR adds the actor that collects the signatures a `ChannelConfigOp` needs, so a channel can ask for more than one.

## ⚙️ Approach

The turn holder funds a draft and gossips it. Every other accredited node derives the target from finalized state on its own, checks the draft matches that, and gossips back its signature.

- [x] Add a `ChannelConfigActor` that holds our draft and collects the signatures
- [x] Add a dedicated gossip topic for drafts and signatures, screened by stake on record
- [x] Compute the threshold as two thirds of the committee, capped at size minus one

## 🧪 How to Test

cargo test -p sequencer_channel_config_actor
cargo test -p sequencer_gossip_actor --features actor
cargo test -p sequencer_core --features mock

and the end to end one

RISC0_DEV_MODE=1 cargo test -p integration_tests --test channel_config_threshold

I have not run the integration one yet, only the unit tests.

## 🔗 Dependencies

none. this is rebased on dev after the gossip as actor refactor landed.

## 🔜 Future Work

The slasher publishes its approvals as a bare `Approval` while transactions and configs go through a Publish wrapper. Would be nice to make the three consistent.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


integration_tests/Cargo.toml
erhant, line 1: https://github.com/logos-blockchain/logos-execution-zone/actions/runs/34857314441/job/104020168781?pr=834 fails due to here

lez/sequencer/actors/channel_config/src/actor.rs
```diff
@@ -0,0 +1,856 @@
...
+    type Reply = ();
+
+    async fn handle(&mut self, msg: ChannelView, _ctx: &mut Context<Self, Self::Reply>) {
+        // A draft is bound to one target on one tip; either moving kills it.
+        if self
+            .draft
+            .as_ref()
+            .is_some_and(|draft| msg.target.as_ref() != Some(&draft.target))
+        {
+            self.draft = None;
```
erhant, line 249: draft is only dropped (here), which is when the target changes. what about other cases?

an example: the wallet's fee-note reservation expires after `pending_note_expiry_blocks` (10) immutable blocks if the draft is still waiting for signatures, a block inscription spends the note, and the draft is a double spend from then on. the channel is stuck until an unrelated target change.

note that the old `channel_config()` path **funded** fresh on every attempt so this issue kind of self-healed.

lez/sequencer/actors/channel_config/src/actor.rs
```diff
@@ -0,0 +1,856 @@
...
+        if msg.tx_hash != draft.tx_hash {
+            return;
+        }
+        let index = msg.signature.channel_key_index;
+        let Some(key) = view.live_keys.get(usize::from(index)) else {
+            return;
+        };
+        if key
+            .verify(
+                draft.tx.hash().as_signing_bytes().as_ref(),
```
erhant: could be `draft.tx_hash` instead of `draft.tx.hash()` here?

lez/sequencer/core/src/block_publisher.rs
```diff
@@ -620,14 +634,21 @@ impl BlockPublisherTrait for ZoneSdkPublisher {
...
             channel: self.channel_id,
             // The channel does not exist yet, so the config lineage starts here.
             parent: MsgId::root(),
             keys,
             posting_timeframe: SlotTimeframe::from(channel_params.posting_timeframe),
             posting_timeout: SlotTimeout::from(channel_params.posting_timeout),
-            configuration_threshold: system_accounts::DEFAULT_SEQUENCER_CONFIGURATION_THRESHOLD,
+            // Derived from the founding committee, not a constant: a channel
+            // written with a threshold above its key count could never be
+            // reconfigured, and a one-key channel has to stay single-signer.
```
erhant: the same comment block appears above (637-639) as well

lez/sequencer/actors/channel_config/src/actor.rs
```diff
@@ -0,0 +1,856 @@
...
+fn matches(target: &ConfigTarget, op: &ChannelConfigOp) -> bool {
+    op.parent == target.parent
+        && op.keys.iter().eq(target.keys.iter())
+        && u32::from(op.posting_timeframe.clone()) == target.posting_timeframe
+        && u32::from(op.posting_timeout.clone()) == target.posting_timeout
+        && op.configuration_threshold == target.configuration_threshold
+        && op.transfer_threshold == target.transfer_threshold
+}
+
+#[cfg(test)]
```
erhant, line 404: since tests take quite a lot of lines, maybe we can have them under `tests.rs` like we have on some other actors?

lez/sequencer/actors/gossip/src/actor.rs
```diff
@@ -83,6 +84,14 @@ pub struct GossipActor {
     approval_sink: Recipient<Approval>,
     /// The committee the follow path last read; `None` filters nothing.
     accredited_keys_rx: AccreditedKeysReceiver,
+    /// Keys with stake on record at head; `None` filters nothing.
```
erhant: the `None` here in comment seems a bit misleading since the types are not Option

lez/sequencer/actors/gossip/src/actor.rs
```diff
@@ -692,6 +824,24 @@ pub(crate) fn peer_id_from_ed25519(
...
 }
 
+/// The Ed25519 key an Ed25519 peer id carries inline, so an originator several
+/// hops away needs no Identify exchange.
+fn inlined_ed25519_key(peer_id: &PeerId) -> Option<[u8; 32]> {
+    /// The identity multihash, which keeps the encoded key verbatim.
+    const IDENTITY: u64 = 0;
+
+    let multihash: &libp2p::multihash::Multihash<64> = peer_id.as_ref();
+    if multihash.code() != IDENTITY {
```
erhant, line 837: I think it's worth noting in function docstring that this trick here only works for IDENTITY multihash

I was confused at first like "whoah how can we obtain key from peer id??? its like reverse hash function or something"

test_fixtures/src/config.rs
```diff
@@ -428,14 +428,22 @@ pub fn bedrock_channel_id_from_seed(seed: u32) -> ChannelId {
...
 #[must_use]
 pub fn bedrock_funding_key() -> ZkPublicKey {
-    const PUBLIC_KEY_HEX: &str = "2e03b2eff5a45478e7e79668d2a146cf2c5c7925bce927f2b1c67f2ab4fc0d26";
-
-    let bytes = hex::decode(PUBLIC_KEY_HEX).expect("Fixed funding key must be valid hex");
+    let bytes = hex::decode(BEDROCK_FUNDING_KEY_HEX).expect("Fixed funding key must be valid hex");
     ZkPublicKey::from(BigUint::from_bytes_le(&bytes))
 }
 
+/// The funding key as the node's wallet routes spell it: little endian hex,
```
erhant: maybe no need for the docstring here

lez/sequencer/actors/channel_config/src/actor.rs
```diff
@@ -0,0 +1,856 @@
...
+    }
+}
+
+impl Message<Signature> for ChannelConfigActor {
+    type Reply = ();
+
+    async fn handle(&mut self, msg: Signature, _ctx: &mut Context<Self, Self::Reply>) {
+        self.on_peer_signature(&msg);
+    }
+}
```
erhant: are these used outside of tests? looks like `Wire` does the job for them right below, could be removed?

erhant (commented):
LGTM overall, several comments made, will approve after resolution 🙏🏻

integration_tests/Cargo.toml
schouhy: thanks. done in a3be7fa87

lez/sequencer/actors/channel_config/src/actor.rs
schouhy: good catch thanks. I changed things so that submission of the channel config happens as soon as the threshold is met. Also, a draft that doesn't land in time gets dropped and funded again, so it can't stay stuck on a note that was already spent

done in aabb4f3bd

lez/sequencer/actors/channel_config/src/actor.rs
schouhy: done in a3be7fa87

lez/sequencer/core/src/block_publisher.rs
schouhy: thanks. fixed in a3be7fa87

lez/sequencer/actors/channel_config/src/actor.rs
schouhy: sounds good. Done in 049648274

lez/sequencer/actors/gossip/src/actor.rs
schouhy: thanks, fixed in a3be7fa87

lez/sequencer/actors/gossip/src/actor.rs
schouhy: lol, added the comment on a3be7fa87

lez/sequencer/actors/channel_config/src/actor.rs
schouhy: thanks. Fixed in 049648274

test_fixtures/src/config.rs
schouhy: done in a3be7fa87

lez/sequencer/actors/channel_config/src/actor.rs
```diff
@@ -0,0 +1,389 @@
...
+        // On crossing the threshold, not on standing above it: a further
+        // signature changes nothing.
+        let was_short = draft.signatures.len() < required;
+        draft.signatures.insert(index, msg.signature.signature);
+        if was_short
+            && draft.signatures.len() >= required
+            && let Action::Submit(submission) = self.next_action()
+        {
+            self.send_submission(submission);
+        }
```
erhant: _(i wrote a comment in the commit but it got lost, rewriting here now)_

one question here:

the comment says "on crossing the threshold, not on standing above it"
but when we reach here with `draft.signatures.len() == required - 1` we can only submit if `self.next_action()` is also `Action::Submit`

if not, when we come here later `was_short` will be false and we won't be able to submit anymore

lez/sequencer/actors/channel_config/src/actor.rs
erhant: a follow up comment for commit `aabb4f3` https://github.com/logos-blockchain/logos-execution-zone/pull/834/changes#r4034080709

lez/sequencer/actors/channel_config/src/actor.rs
schouhy: I think next_action should be Submit unless the config is not good. But anyway I refactored and removed the `was_short` in favor of a `submission_sent` flag. Let me know if this looks better!

moudyellaz (commented):
lgtm, thanks! Left some questions/comments

lez/sequencer/core/src/lib.rs
```diff
@@ -1896,12 +2043,78 @@ async fn record_dead_letter_gauge<S: StorageActorTrait>(storage_ref: &ActorRef<S
     }
 }
 
+/// The config the channel should move to, or `None` if it already matches or
+/// a config is still in flight.
+///
+/// Committee membership follows finalized state only. A config is in flight
+/// while the live tip is not the finalized one, and zone-sdk would shed a
+/// draft chained on a tip it has not finalized yet.
+fn config_target(
```
moudyellaz: `committee_update` only fires on a key-set change, so does a channel already at `configuration_threshold = 1` ever raise its bar without a stake event?

lez/sequencer/core/src/lib.rs
```diff
@@ -817,12 +859,22 @@ impl<S: StorageActorTrait, BP: BlockPublisherTrait> SequencerCore<S, BP> {
...
     /// committee-config update the new state calls for.
     pub async fn run_production_turn(&mut self) -> Result<u64> {
+        // A draft still short of signatures a full rotation after funding has
+        // outlived its fee-note reservation, which the block below may spend.
+        if self
+            .config_draft
+            .as_ref()
+            .is_some_and(|draft| !draft.submitted)
+        {
+            warn!("A channel-config draft went a full turn unsigned; funding another");
```
moudyellaz: `run_production_turn` runs once per `block_create_timeout` (10s shipped), not once per turn, so the draft is refunded after one block rather than one rotation; is that the intent?

lez/sequencer/core/src/block_publisher.rs
```diff
@@ -896,22 +952,31 @@ pub async fn post_channel_config(
...
     let tx_hash = mantle_tx.hash();
-    // The admin key is `keys[0]`, hence signature index 0.
-    let signature = IndexedSignature::new(
-        0,
-        signing_key.sign_payload(tx_hash.as_signing_bytes().as_ref()),
-    );
-    let proof = ChannelMultiSigProof::try_new(signature.into())
+    // Creating the channel takes no signature; otherwise the admin key is
+    // `keys[0]`, hence signature index 0.
+    let signatures = if channel.is_some() {
```
moudyellaz: With one signature at index 0, does `configure_channel` still work once the live threshold is above one?

lez/sequencer/actors/channel_config/src/actor.rs
```diff
@@ -0,0 +1,405 @@
...
+            .collect();
+
+        Action::Submit(Box::new(Submission {
+            tx_hash: draft.tx_hash,
+            signatures,
+        }))
+    }
+
+    /// A peer's draft, from gossip or from a direct send.
+    fn on_peer_draft(&self, msg: &Draft) {
```
moudyellaz: Should a draft this node already signed be skipped, so a replayed draft cannot make every peer publish another signature?

lez/sequencer/core/src/lib.rs
moudyellaz, line 2052: `committee_update` only fires on a key-set change, so does a channel already at `configuration_threshold = 1` ever raise its bar without a stake event?

lez/sequencer/core/src/lib.rs
moudyellaz, line 869: `run_production_turn` runs once per `block_create_timeout` (10s shipped), not once per turn, so the draft is refunded after one block rather than one rotation; is that the intent?

lez/sequencer/core/src/block_publisher.rs
moudyellaz, line 957: With one signature at index 0, does `configure_channel` still work once the live threshold is above one?

lez/sequencer/actors/channel_config/src/actor.rs
moudyellaz, line 159: Should a draft this node already signed be skipped, so a replayed draft cannot make every peer publish another signature?

moudyellaz (approved):
lgtm, thanks! Left some questions/comments

lez/sequencer/core/src/lib.rs
schouhy: fixes. committee updates are also triggered by a threshold mismatch

lez/sequencer/core/src/lib.rs
schouhy: fixed the comments. This is intended because the l1 utxos with which the config update is created could also be spent by paying fees for submitting a block. So it's safer to just refund the draft.

lez/sequencer/core/src/block_publisher.rs
schouhy: true. But just realized this code is dead. So I removed it instead of fixing.

lez/sequencer/actors/channel_config/src/actor.rs
schouhy: since signatures go over gossip, they can be lost. I think it's cheap to rebroadcast instead of skipping

Pravdyvy (approved):
lgtm

PR #835, opened 2026-09-05
feat!: sharded program state

agureev:
## 🎯 Purpose

This PR finalizes PR #817 by introducing sharded program state in LEE.

In particular, we now can think of a `program` as a smart contract carrying around the `Data` connected to each `AccountID` that interacts with it. However, for the purposes of proving, we shard it, so that such states are not carried in a smart contract but instead is carried in shards of each acount's state.

Hence, an account becomes - in a flattened presentation -

`Account {nonce, balance, shards: BTreeMap<ProgramAccountID, Data>}`

This narrows "squatting" rather than removing it. A program can only ever occupy its own `Data`, so one program can no longer block an address another program needs. Within a single program's own shard it is still possible: the token program's `InitializeAccount` and `Transfer` gate on the shard being empty, not on authorization, so a predictable address can still be written first. For ATAs that is now repairable rather than sticky, since `create` re-initializes a wrong-definition holding. AMM vaults still stick, tracked in #882. Authorization is also still per account rather than per shard, tracked in #898.

## ⚙️ Approach

Core commits:

In 38437f30f4f4b586e6e1c14d26b5b102eb29f8d1:
- [x] Introduce shards in the `Account` struct
- [x] Remove `AccountWithMatadata` in favor of `AccountInput`

In f8c7a6139c44a69af4dd3ea54d07107f9c2fd44f:
- [x] Change PPC to account for sharded program state

In 0f42f9e545296f4bf2c414db5ed4e1c875a991a3:
- [x] Use `execute_and_prove_with` fetching the entire state of the public account (even that not yet fetched as when a malicious actor updates a random shard of yours)

In 68df4d3645002b615716414e786f86bb4b166035:
- [x] Update transfer, bridge, faucet, fees, stake to use shards

In 5873df47ce84d3ee7cc5e7c268b3f270e6b07984:
- [x] SInce accounts don't have owners, we need to bind the token account ids for AMM and ATA in the seed.

In a4f1058984455d02984431b8546a8ab372979703:
- [x] Utilize shards in cross-zone calls
- [x] Utilize shards in chained calls

In c661c989fb305747b431c4ba0b222bca4f31e418:
- [x] Change genesis accounts given shards

In 7725dd4666102f3d5c0407712ea80aae91d326f4:
- [x] Implement wallet account-view getting, where we fetch only those shards in which we are interested in

## 🔜 Future Work

Account diff should remove the public-state refetch needed to sync-up the limited view of an account's state locally.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


lez/programs/token/src/initialize.rs
```diff
@@ -1,24 +1,21 @@
...
 pub fn initialize_account(
-    definition_account: &AccountWithMetadata,
-    account_to_initialize: &AccountWithMetadata,
+    definition_account: &AccountInput,
+    account_to_initialize: &AccountInput,
+    self_account_id: AccountId,
 ) -> Vec<AccountStateDiff> {
     assert!(
-        account_to_initialize.account.data.is_empty(),
+        account_to_initialize.shard_of(self_account_id).is_empty(),
```
moudyellaz, line 14: Sharding closes the cross-program case nicely, but this still only checks that the token shard is empty: a 0-amount transfer to a predictable ATA or AMM vault address writes a holding under someone else's `definition_id`, and from there create no-ops, transfer panics on the mismatch and burn's ATA re-derivation fails, with no way back. Would a PDA seed grant on the target work here, and should the PR body scope the squatting claim to cross-program so #862 stays visible?

integration_tests/tests/scoped_reads.rs
```diff
@@ -0,0 +1,367 @@
...
+#[test]
+async fn a_bloated_account_defeats_the_whole_account_read_but_not_the_scoped_one() -> Result<()> {
+    let mut ctx = TestContext::new().await?;
+    let victim = ctx.existing_public_accounts()[0];
+
+    let writers = bloat_account(&mut ctx, victim).await?;
+
+    let error = get_account(&ctx, victim)
+        .await
+        .expect_err("the whole-account read must fail once the account is bloated");
```
moudyellaz, line 165: Nice that the scoped read holds up here, though this also shows anyone can make a chosen account permanently unreadable over `getAccount` for the price of four writes, since nothing caps shard count and only the owning program can clear its shard. The wallet's `get_account_public` and the account CLI still take the whole-account path, so would it be worth moving those to `getAccountView` too, and is the indexer's own `get_account` under the same limit?

moudyellaz (approved):

lez/programs/token/src/initialize.rs
agureev: Good catch! And yes, the PDA seed abstraction I think should exactly be the solutuion.

Basically, we should make token holders only PDAs whose IDs are derived from:
1. The holder authority (e.g. Alice's regulard account ID)
2. The ATA
3. The definition ID

on it!

And yes, will work on reworking the PR body as well

integration_tests/tests/scoped_reads.rs
agureev: > anyone can make a chosen account permanently unreadable over getAccount for the price of four writes

Indeed, this is true. The question is: when would you ever be interested in getting your enitre account to begin with? You should usually only be interested in the state of the programs that interest you which would usually fit the limit imposed for fetching.

The only reason you would need to fetch you entire public account is for using a public account in a PPTX. However, this hopefully can be side-stepped when we get incremental updates as I believe instead of accounts we should simply feed in IDs accompanied by the predicates they satisfy. This should make feeding the actual account state of the public account irrelevant, completely. I can note it in a comment or we can make an issue on it, however

> so would it be worth moving those to getAccountView too, and is the indexer's own get_account under the same limit

Indeed, good catch, on it

integration_tests/src/utils.rs
```diff
@@ -187,6 +187,16 @@ pub async fn get_account(ctx: &TestContext, account_id: AccountId) -> anyhow::Re
...
 }
 
+pub async fn get_account_view(
+    ctx: &TestContext,
+    shard_selector: ProgramShardSelector,
+) -> anyhow::Result<lee::Account> {
+    Ok(ctx
+        .sequencer_client()
+        .get_account_view(shard_selector)
+        .await?)
```
erhant, line 197: could be like below maybe?

```rs
ctx
        .sequencer_client()
        .get_account_view(shard_selector)
        .await
```

lez/programs/token/src/initialize.rs
agureev: Ok, I have encountered a problem: this has to be computed in-guest yet the formular for public and private PDA addresses differ, so we cannot really use this...

lee/state_machine/core/src/account.rs
```diff
@@ -112,50 +112,213 @@ pub enum BalanceDiffError {
...
 }
 
-/// Account to be used both in public and private contexts.
+/// An account's balance and program shards.
 #[derive(
     Debug, Default, Clone, Eq, PartialEq, Serialize, Deserialize, BorshSerialize, BorshDeserialize,
 )]
-pub struct Account {
-    pub program_owner: AccountId,
+pub struct AccountData {
```
schouhy, line 150: I think it's confusing that `account::Data` and `AccountData` are different things. Maybe move `account::Data` to `program::Shard`?

lee/state_machine/core/src/account.rs
```diff
@@ -112,50 +112,213 @@ pub enum BalanceDiffError {
...
+    #[must_use]
+    pub const fn new(account_id: AccountId, program_account_id: AccountId) -> Self {
+        Self {
+            account_id,
+            program_account_id: Some(program_account_id),
+        }
+    }
+
+    #[must_use]
+    pub const fn balance_only(account_id: AccountId) -> Self {
```
schouhy, line 236:
```suggestion
    pub const fn balance(account_id: AccountId) -> Self {
```

lee/state_machine/core/src/account.rs
```diff
@@ -112,50 +112,213 @@ pub enum BalanceDiffError {
...
+        }
+    }
 }
 
+/// An account seen as an input to an LEE program.
 #[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize, BorshSerialize, BorshDeserialize)]
-pub struct AccountWithMetadata {
-    pub account: Account,
-    pub is_authorized: bool,
+pub struct AccountInput {
```
schouhy, line 246: wouldn't this be a `ProgramInput`? I think this and the `ProgramShardSelector` belong to the program module

lee/state_machine/core/src/account.rs
```diff
@@ -112,50 +112,213 @@ pub enum BalanceDiffError {
...
     pub account_id: AccountId,
+    pub is_authorized: bool,
+    pub balance: Balance,
+    pub shard: Option<(AccountId, Data)>,
```
schouhy, line 250: there's something iffy imo about the case where the `balance` and `shard` are both given. I think this should provide only one of them, like:
```rust
enum Input {
    BalanceShard(Balance),
    ProgramShard((AccountId, Data)),
}
```
any program needing both balance and a shard as input can have two inputs. And users can choose to use the balance and shard of the same or different accounts. Would this approach hit a wall?

I think conceptually the balance is a shard of a builtin program for native tokens. Its rules checked by the protocol itself instead of a guest program. And so an `AccountInput` should have only one "shard" in any case.

schouhy (commented):
Still reviewing. Leaving some comments to discuss

lee/state_machine/core/src/account.rs
agureev: That's a fair approach to consider, let me check out if we can do it this way.

Another approach is to just attach Balance as a shard of an authorized-transfer program with a predefined ID such as all-zero id.

agureev: ^ That way we would not even need to have this extra field around at all and I think this plays with what @jonesmarvin8 wanted re getting rid of special-treatment for the native tokens

lez/programs/token/src/initialize.rs
agureev: After long discussions, we are to make squatting non-sticky by allowing ATA holders to clean data if authorized by the owner. 

Long-term solution will be ATA migration, see (882)[https://github.com/logos-blockchain/logos-execution-zone/issues/882]

lee/state_machine/core/src/account.rs
agureev: > any program needing both balance and a shard as input can have two inputs.

This is actually impossible right now, since accountIDs used still need to be unique per input, that has not changed.

integration_tests/src/utils.rs
agureev: oh wait, but the return type of `await` is `Result<lee::Account, ClientError>`, so I think we're good here

lee/state_machine/core/src/account.rs
agureev: I think `AccountData` is right, and then for the data of the shard, let's use `ShardData`

a shard - at least of how I thought about it - is more a pair of (program, shard_data`

lee/state_machine/core/src/account.rs
agureev: `ProgramInput` is actually reserved for the entire input of the program, so I am unsure

agureev: i do agree that this should move to the `program`, but I would keep `ProgramShardSelector` here

lez/programs/token/src/initialize.rs
agureev: done: [d2ca401](https://github.com/logos-blockchain/logos-execution-zone/pull/835/commits/d2ca40167ba5b7fd438dbe285209286161dbe435)

integration_tests/tests/scoped_reads.rs
agureev: [d5005a1](https://github.com/logos-blockchain/logos-execution-zone/pull/835/commits/d5005a13547f860a98303a496eb108b1ca2132ed)

[6d5eae4](https://github.com/logos-blockchain/logos-execution-zone/pull/835/commits/6d5eae4e9db09053061619eda22f70c4e9200626)

lee/state_machine/core/src/account.rs
agureev: [e91d062](https://github.com/logos-blockchain/logos-execution-zone/pull/835/commits/e91d062dae111cc8c3d3671c74b0c7c49b73366e)

lee/state_machine/core/src/account.rs
agureev: [cb21c1a](https://github.com/logos-blockchain/logos-execution-zone/pull/835/commits/cb21c1aa81dfc05b6108b3daa0da5e927fb71224)

lee/state_machine/core/src/account.rs
agureev: [e27b291](https://github.com/logos-blockchain/logos-execution-zone/pull/835/commits/e27b29173892009f96b1add6bde771345c755b5e)

lee/state_machine/core/src/account.rs
agureev: This is postponed, see [thread](https://discord.com/channels/973324189794697286/1547912678603890738)

lez/programs/amm/src/new_definition.rs
```diff
@@ -112,16 +111,17 @@ pub fn new_definition(
...
         BalanceDiff::Add(0),
-        Data::from(&pool_post_definition),
+        ShardData::from(&pool_post_definition),
     );
 
-    let token_program_id: AccountId = user_holding_a.account.program_owner;
-
     // Chain call for Token A (user_holding_a -> Vault_A)
     let vault_a_seed = compute_vault_pda_seed(pool.account_id, definition_token_a_id);
     let call_token_a = ChainedCall::new(
```
moudyellaz, line 119: The vaults have the same hole one file over: `vault_a` is derivable before the pool exists, so a transfer of any token to it leaves a foreign `definition_id` and this `Transfer` panics on the mismatch, with no way back since AMM never chains an `InitializeAccount` for a vault. You already delegate `vault_a_seed` here and token now accepts an authorized overwrite, so same treatment as `create.rs`, or fold it into 882?

lez/explorer_service/src/pages/account_page.rs
```diff
@@ -116,21 +112,44 @@ pub fn AccountPage() -> impl IntoView {
...
                                                 </div>
-                                                <div class="info-row">
-                                                    <span class="info-label">"Data:"</span>
-                                                    <span class="info-value">{format!("{data_len} bytes")}</span>
-                                                </div>
                                             </div>
                                         </div>
 
+                                        <div class="account-shards">
+                                            <h2>"Shards"</h2>
```
moudyellaz, line 123: The scoped sweep caught the wallet, the ffi and `cross_zone_chat`, but this page still renders from `api::get_account`. Same four writes as `scoped_reads` push it past jsonrpsee's 10 MiB default once `ShardData` is JSON number arrays, and the page stops loading for that address for good. Point it at the `get_account_view` you just added?

moudyellaz (approved):
lgtm, thanks for addressing my comments! Added other comments on the new commits

lez/programs/amm/src/new_definition.rs
moudyellaz: Moving this one to #882 rather than growing this PR again.

lez/explorer_service/src/pages/account_page.rs
moudyellaz: Fixed in 06752ed0a.

lez/programs/program_loader/core/src/lib.rs
```diff
@@ -121,17 +102,16 @@ pub fn write_segment(
...
 ) -> Vec<AccountStateDiff> {
     assert!(
         !pre_states.is_empty(),
         "CreateHeader requires at least the header target account"
     );
-    assert_eq!(
-        pre_states[0].account,
-        Account::default(),
+    assert!(
+        pre_states[0].shard_of(PROGRAM_LOADER_ACCOUNT_ID).is_empty(),
```
schouhy, line 124: is there something protecting against a user deploying a program at the program loader id? Seems like the account at that id starts empty and someone could populate it. Probably the program loader can then be impersonated to change any other program after?

lee/state_machine/src/validated_state_diff/mod.rs
```diff
@@ -304,31 +314,33 @@ impl ValidatedStateDiff {
...
                             ));
                         }
                     };
-                    Ok(AccountWithMetadata::new(
-                        account,
-                        is_authorized(account_id),
-                        *account_id,
+                    Ok(AccountInput::at(
+                        *shard_selector,
+                        is_authorized(&account_id),
```
schouhy, line 342: a signature provides authorization for all shards. I'd consider either making authorization targeted to a specific program, or state that using the same account for different programs is bad practice. Shards are cool to save us from the claiminig sheningans, but using the same account to store the data of two different programs, then either both go as authorized or none in all executions. That doesnt look good.

schouhy (approved):
This is great thanks! Left two comments I think are important. Otherwise lgtm

lee/state_machine/src/validated_state_diff/mod.rs
agureev: This is a fair concern, but to be explicit, this is not something that is really new for the Solana authorization model if you project it onto the previously-supported one.

In Solana, it is possible to e.g. transfer balance and then make a new program owner in the same transaction. This is allowed and that would propagate authorization through different program owners (read as "different shards" in this model). Similarly, you can even clear-and-claim in the same tx in Solana (with supporting programs) as well, so you can indeed do this for non-default (system) program as well.

In other words, I still do not see it as an underlying problem and would recommend a separate PR/research in this direction. Authorization actors are still the same. So regular accounts are still authorized via secret key possession proof and PDAs can only be authorized by their programs (recall that that authorization mechanism does not go away anywhere).

So the authorizing can never be "blind". The authorization agent always sees what they are authorizing. Programs can control to which shards they make calls and prevent free-assignment of these while users need - well - only sign transactions which do not do weird things (this is true regardless of the model).

So, to reiterate, I still do not see it as a problem and hence would suggest postponing this proposal to another PR as a separate concern.

examples/program_deployment/README.md
erhant, line 1: the commands here use `wallet account get --raw --account-id ...` but the `--scope` defaults to "balance" on the Wallet CLI

perhaps we should indicate `--scope all` as well in the commands here? (or we can make default `--scope` to be  "all")

lez/wallet/src/lib.rs
erhant: here we still read the whole account just for the `nonce`.

with reasoning similar to https://github.com/logos-blockchain/logos-execution-zone/pull/835#discussion_r3974675014 I would suggest we use `get_account_view(ProgramShardSelector::balance(payer))` which gives the nonce without that exposure

lez/wallet-ffi/src/generic_transaction.rs
```diff
@@ -134,7 +134,7 @@ impl Default for FfiTransactionResult {
 ///
 /// # Parameters
 /// - `handle`: Valid pointer to wallet handle
-/// - `account_identities`: Valid pointer to list of `FfiAccountIdentity`
+/// - `account_mentions`: Valid pointer to list of `FfiAccountMention`
 /// - `instruction_data`: Valid pointer to instruction data bytes
 /// - `payer`: Fee payer, or null to self-pay from the first funded signing account in
 ///   `account_identities` (the first signing account if none is funded). May be one of those
```
erhant:
```diff
- account_identities
+ account_mentions
```

lez/wallet-ffi/wallet_ffi.h
```diff
@@ -600,7 +642,7 @@ enum WalletFfiError wallet_ffi_bridge_withdraw(struct WalletHandle *handle,
  *
  * # Parameters
  * - `handle`: Valid pointer to wallet handle
- * - `account_identities`: Valid pointer to list of `FfiAccountIdentity`
+ * - `account_mentions`: Valid pointer to list of `FfiAccountMention`
  * - `instruction_data`: Valid pointer to instruction data bytes
  * - `payer`: Fee payer, or null to self-pay from the first funded signing account in
  *   `account_identities` (the first signing account if none is funded). May be one of those
```
erhant:
```diff
- account_identities
+ account_mentions
```

test_programs/guest/src/bin/time_locked_transfer.rs
```diff
@@ -44,7 +44,11 @@ fn main() {
     assert_eq!(clock_pre.account_id, CLOCK_01_PROGRAM_ACCOUNT_ID);
 
     // Read the current timestamp from the clock account.
-    let clock_data = ClockAccountData::from_bytes(&clock_pre.account.data);
+    let (_, clock_bytes) = clock_pre
+        .shard
+        .as_ref()
+        .expect("the clock shard selector must name a record");
```
erhant, line 50: wouldn't `clock_pre.shard_of(CLOCK_01_PROGRAM_ACCOUNT_ID)` be more suitable here?

lez/programs/program_loader/core/src/lib.rs
moudyellaz: ok public dispatch is safe (it branches on the address first), but the privacy path has no such branch: after a squat it resolves the attacker's image and runs it as the loader, which can then write the loader shard on any account, i.e. every program's header and segments. it was pre-existing on dev so i fixed by refusing that address in write_segment and create_header, segments too since a minimal one decodes as a header. 7ee2c8d89

lee/state_machine/src/validated_state_diff/mod.rs
moudyellaz: fixed -> the authorized set is now built from the signed selectors, and the message already names one shard per account, so nothing widens. Balance stays account-level, it carries no program. The private path stays account-level: its message carries no selectors, only a signed post state, so narrowing it needs a design for private accounts whose ask covers the whole account. Filing that separately. [45f65faba](https://github.com/logos-blockchain/logos-execution-zone/pull/835/changes/45f65fabac8b775ca414b93341ecdc80e015cd67)

lez/programs/token/src/initialize.rs
```diff
@@ -1,32 +1,30 @@
...
-    account_to_initialize: &AccountWithMetadata,
+    definition_account: &AccountInput,
+    account_to_initialize: &AccountInput,
+    self_account_id: AccountId,
 ) -> Vec<AccountStateDiff> {
     assert!(
-        account_to_initialize.account.data.is_empty(),
-        "Only Uninitialized accounts can be initialized"
+        account_to_initialize.shard_of(self_account_id).is_empty()
+            || account_to_initialize.is_authorized,
```
erhant, line 15: so if a token is already initialized, it can be initialized again? that sounds like a bug or am i missing something

erhant (commented):
Some comments made, will re-review after resolved 🙏🏻

lee/state_machine/src/validated_state_diff/mod.rs
moudyellaz: reverted this one as agreed.

lez/wallet/src/lib.rs
moudyellaz: good catch, now reads `get_account_view(balance(payer))`. 1aad85a25

examples/program_deployment/README.md
moudyellaz: Right, the expected outputs show shards that the default scope would not print. Added `--scope all` to all ten commands rather than change the default, since balance is the safe default on a bloatable account. 1aad85a25

lez/wallet-ffi/src/generic_transaction.rs
moudyellaz: Done. 1aad85a25

lez/wallet-ffi/wallet_ffi.h
moudyellaz: Done. 1aad85a25

test_programs/guest/src/bin/time_locked_transfer.rs
moudyellaz: I tried this and it breaks the test: `CLOCK_01_PROGRAM_ACCOUNT_ID` is the clock data account's address, while shards are keyed by the writing program, so `shard_of` on it asserts a shard nothing writes. The guest has no constant for the clock program's own id, since it derives from the image id, so it pins the account by address and takes the selected shard. Leaving as is.

lez/programs/token/src/initialize.rs
moudyellaz: Not missing anything, that is real: an authorized target gets overwritten with a zeroized holding, so re-initialization is destructive rather than idempotent. It came in with the squat repair. It is also the smaller half, the same authorization inheritance lets any program you call drain the holding outright via `token::Transfer`, so fixing just the wipe would not buy much. Tracked in #898.

lee/state_machine/src/validated_state_diff/mod.rs
moudyellaz: Filed as #898, with the drain via `token::Transfer` and `ATA::Transfer` and the `UpdateHeader` reach written up alongside your point, plus why the public-only narrowing was reverted so nobody re-derives it.

erhant (approved):
LGTM, thanks for addressing the comments!

#898 should be kept in mind for future!

PR #837, opened 2026-09-07
feat(state_machine): handle non-zero program exit codes

erhant:
## 🎯 Purpose

A guest that fails via `panic!` bails out of the zkVM executor and drops the session, so the host has no cycle count and charges the full declared budget. A guest that fails via `env::exit(n)` halts normally: the executor keeps the session, the journal, and the metered cycles — but `execute_session` never looked at `exit_code`, so:

- a non-zero halt with a committed journal was **accepted as success** (as was `Paused(_)`);
- a non-zero halt without a journal surfaced as a `malformed program journal frame` failure, with the metered cycles discarded.
- a pause (as in `env::pause` instead of `env::exit`) is treated as a failure and charges the whole budget, because we do not expect multiple tx'es to chain with checkpoints in between, there is no use case for `pause` in LEZ at the time of writing

This PR treats a non-zero exit as a typed, metered revert. It is the host-side prerequisite for rewriting programs to report errors through exit codes instead of panics.

## ⚙️ Approach

- [x] Add `LeeError::ProgramExitedWithCode { code, cycles }` (chargeable, like every post-execution failure).
- [x] `SessionOutcome` carries `exit_code` on both executor paths (r0vm IPC and the in-process `image_cache`).
- [x] `execute_session` gates on it: `Halted(0)` → ok; `Halted(n)` → `ProgramExitedWithCode`; `Paused`/`SystemSplit` → `ProgramExecutionFailed`.
- [x] `execute_authorized` folds a callee's exit-code cycles into `cycles_used` before propagating; `from_public_transaction_metered` charges `cycles_used` for an exit-code revert and still the full budget for panics / session-limit bails.
- [x] Anchor the session-limit match to `root_cause().starts_with("Session limit exceeded:")`. A guest panic always roots at `Guest panicked: …`, so the separate anti-spoof `contains` goes. A typed error is not available: risc0 bails with anyhow and the r0vm IPC boundary flattens errors to a string.

## 🧪 How to Test

```sh
RISC0_DEV_MODE=1 cargo test -p lee --all-features
```

New test guest `exits_nonzero` backs two tests:

- `program::tests::nonzero_exit_is_rejected_with_its_cycles` program exits with the correct error
- `validated_state_diff::tests::metered_nonzero_exit_is_charged_its_metered_cycles` the metered path reports 0 < cycles < budget and a nonce-only diff.

> [!NOTE]
>
> When we later re-write our programs to use non-zero exit codes, we can perhaps get rid of this artificial test.

## 🔗 Dependencies

None

## 🔜 Future Work

- Rewrite programs to fail via `env::exit(code)`; perhaps add a `lee_core` guest-side `abort(code)` helper so programs don't import `risc0_zkvm::guest::env` for this alone.
- Tracker issue opened for `lez-programs`: https://github.com/logos-blockchain/lez-programs/issues/356

## 📋 PR Completion Checklist
- [x] Complete PR description
- [x] Add/update tests

### DIFF
--- a/lee/state_machine/src/program/mod.rs
+++ b/lee/state_machine/src/program/mod.rs
@@ -9,7 +9,7 @@ use lee_core::{
 };
 #[cfg(not(feature = "prove"))]
 use risc0_zkvm::default_executor;
-use risc0_zkvm::{ExecutorEnv, ExecutorEnvBuilder};
+use risc0_zkvm::{ExecutorEnv, ExecutorEnvBuilder, ExitCode};
 
 use crate::error::LeeError;
 
@@ -31,6 +31,7 @@ pub const DEFAULT_PUBLIC_CYCLE_BUDGET: Cycles = 1024 * 1024 * 32; // 32M cycles
 pub(crate) struct SessionOutcome {
     pub journal: Vec<u8>,
     pub cycles: Cycles,
+    pub exit_code: ExitCode,
 }
 
 #[derive(Clone, Debug, PartialEq, Eq, BorshSerialize, BorshDeserialize)]
@@ -109,8 +110,6 @@ impl Program {
     /// Runs the session, translating the executor's session-limit bail into the
     /// typed [`LeeError::OutOfGas`]. The only place that error string is
     /// recognized.
-    ///
-    /// FIXME: This is a brittle string match; the executor should provide a typed error.
     pub(crate) fn execute_session(
         env: ExecutorEnv<'_>,
         elf: &[u8],
@@ -125,10 +124,13 @@ impl Program {
             SessionOutcome {
                 journal: info.journal.bytes,
                 cycles,
+                exit_code: info.exit_code,
             }
         });
 
-        raw.map_err(|e| {
+        // map the r0 error to LeeError
+        // FIXME: This is a brittle string match; the executor should provide a typed error.
+        let outcome = raw.map_err(|e| {
             // check for "Guest panicked" to prevent spoofing
             // via `panic!("Session limit exceeded")` cases
             let message = format!("{e:#}");
@@ -139,7 +141,24 @@ impl Program {
             } else {
                 LeeError::ProgramExecutionFailed(e.to_string())
             }
-        })
+        })?;
+
+        // check the exit code for non-zero terminations
+        #[expect(
+            clippy::wildcard_enum_match_arm,
+            reason = "we only care about ExitCode::Halted"
+        )]
+        match outcome.exit_code {
+            ExitCode::Halted(0) => Ok(outcome),
+            ExitCode::Halted(code) => Err(LeeError::ProgramExitedWithCode {
+                code,
+                cycles: outcome.cycles,
+            }),
+            // anything other than Halted is an actual error (e.g. r0vm termination)
+            other => Err(LeeError::ProgramExecutionFailed(format!(
+                "unexpected exit {other:?}"
+            ))),
+        }
     }
 
     /// Writes a `CallKind::Execute` frame followed by the guest's `ProgramInput` as a single
--- a/lee/state_machine/src/program/mod.rs
+++ b/lee/state_machine/src/program/mod.rs
@@ -145,25 +145,12 @@ impl Program {
             }
         })?;
 
-        // check the exit code for non-zero terminations
-        #[expect(
-            clippy::wildcard_enum_match_arm,
-            reason = "we only care about ExitCode::Halted"
-        )]
-        match outcome.exit_code {
-            ExitCode::Halted(0) => Ok(outcome),
-            ExitCode::Halted(code) => Err(LeeError::ProgramExitedWithCode {
-                code,
-                cycles: outcome.cycles,
-            }),
-            // anything other than Halted is an actual error (e.g. r0vm termination)
-            //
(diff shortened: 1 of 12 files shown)


lee/state_machine/test_methods/guest/src/bin/exits_nonzero.rs
```diff
@@ -0,0 +1,11 @@
+use risc0_zkvm::guest::env;
+
+// FIXME: we can remove this program when other guests are re-written
+// andalso handle non-zero codes
```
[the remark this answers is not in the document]
erhant: Fixed in https://github.com/logos-blockchain/logos-execution-zone/pull/837/commits/ae04990e679ef797049a9cde536e9b45d98a048f

lee/state_machine/src/validated_state_diff/mod.rs
```diff
@@ -359,13 +359,22 @@ impl ValidatedStateDiff {
...
                     cycle_budget.saturating_sub(*cycles_used),
-                )?;
+                ) {
+                    Ok(ok) => ok,
+                    Err(err) => {
+                        // if there is a non-zero halt, we should charge up to the cycle
+                        if let LeeError::ProgramExitedWithCode { cycles, .. } = err {
+                            *cycles_used = cycles_used.saturating_add(cycles);
+                        }
+                        return Err(err);
```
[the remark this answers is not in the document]
erhant: ??? it does compile bro

Pravdyvy (approved):
lgtm

lee/state_machine/src/validated_state_diff/mod.rs
```diff
@@ -151,11 +151,11 @@ impl ValidatedStateDiff {
...
         );
-        // any failure pays the full declared budget
-        let cycles = if result.is_err() {
-            cycle_budget
-        } else {
-            cycles_used
+        // A panic or session-limit bail loses the count and pays the full budget; a non-zero
+        // exit code is metered like a success.
+        let cycles = match &result {
+            Ok(_) | Err(LeeError::ProgramExitedWithCode { .. }) => cycles_used,
```
moudyellaz: `program/mod.rs:97` binds `let cycles = session.cycles` and then drops it on a malformed journal frame, so a clean session pays the full budget while a non-zero exit with the same count is metered. Allowlist the variants that actually lose the count?

lee/state_machine/src/program/mod.rs
```diff
@@ -125,21 +124,46 @@ impl Program {
...
+            ExitCode::Halted(0) => Ok(outcome),
+            ExitCode::Halted(code) => Err(LeeError::ProgramExitedWithCode {
+                code,
+                cycles: outcome.cycles,
+            }),
+            // anything other than Halted is an actual error (e.g. r0vm termination)
+            //
+            // NOTE: ExitCode::SessionLimit actually exists, but r0vm does not use it
+            // see their own code for the comment
+            other => Err(LeeError::ProgramExecutionFailed(format!(
```
moudyellaz: The private path never reaches this gate (`privacy_preserving_transaction/circuit/mod.rs:377` proves without checking the exit code) and `env::verify` resolves assumptions on `Halted(0)` or `Paused(0)`, so a program ending in `env::pause(0)` now reverts publicly and succeeds privately. Gate the exit code in `execute_and_prove_program` too?

lee/state_machine/src/validated_state_diff/mod.rs
```diff
@@ -359,13 +359,22 @@ impl ValidatedStateDiff {
...
                     caller_data.account_id,
                     &real_pre_states,
                     &chained_call.instruction_data,
                     cycle_budget.saturating_sub(*cycles_used),
-                )?;
+                ) {
+                    Ok(ok) => ok,
+                    Err(err) => {
+                        // if there is a non-zero halt, we should charge up to the cycle
+                        if let LeeError::ProgramExitedWithCode { cycles, .. } = err {
```
moudyellaz: Both new tests run `exits_nonzero` at top level, where `cycles_used` is still 0, so the accumulation this branch exists for is never exercised. Add a case where a chained callee exits non-zero after the caller has burned cycles?

moudyellaz (approved):
lgtm, thanks. Left some comments!

lee/state_machine/src/validated_state_diff/mod.rs
erhant: Added `chained_nonzero_exit_adds_callee_cycles_to_callers` test in https://github.com/logos-blockchain/logos-execution-zone/pull/837/commits/7300f9da1cb73939342bf49479dfac7a68517bc7

lee/state_machine/src/program/mod.rs
erhant: ah yes good catch...
added handling in https://github.com/logos-blockchain/logos-execution-zone/commit/7300f9da1cb73939342bf49479dfac7a68517bc7

lee/state_machine/src/validated_state_diff/mod.rs
erhant: I think a malformed journal counts as a bad program, akin to a panicking one, so we can charge the full budget.

moudyellaz (approved):

PR #843, opened 2026-09-08
chore(lez): add logging

schouhy:
## 🎯 Purpose

Add more logging on the sequencer for devnet debugging.

One unrelated fix found on the way.

## ⚙️ Approach

Logging, all extracted into free functions so `apply_follow_update` only calls them:

Unrelated fix: the orphan path dropped `ChannelUpdateTx::Custom`, while the adopt path decoded it. A block could be adopted into the head and never reverted out. Both directions now use one extractor.

## 🧪 How to Test

```
cargo test -p sequencer_core --features mock channel_update_extraction
cargo test -p sequencer_core --features mock
```

The new lines on a running node:

```
just run-bedrock
just run-sequencer
```

## 🔗 Dependencies

None.

## 🔜 Future Work

None

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests

### DIFF
--- a/lez/sequencer/core/src/lib.rs
+++ b/lez/sequencer/core/src/lib.rs
@@ -47,6 +47,7 @@ use tokio_retry::{Retry, strategy::FixedInterval};
 use crate::{
     block_publisher::{BlockPublisherTrait, MsgId, NoteId, ZoneSdkPublisher},
     block_store::SequencerStore,
+    logging::{log_high_water_lowered, log_parked, log_rewind, log_update, pin_str},
     task_group::TaskGroup,
 };
 
@@ -57,6 +58,7 @@ pub mod config;
 pub mod cross_zone_watcher;
 pub mod fees;
 pub mod gossip;
+pub mod logging;
 #[cfg(feature = "mock")]
 pub mod mock;
 pub mod slashing;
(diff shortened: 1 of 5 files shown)


lez/sequencer/core/src/lib.rs
```diff
@@ -1843,6 +1852,95 @@ async fn record_dead_letter_gauge<S: StorageActorTrait>(storage_ref: &ActorRef<S
...
+        }
+        _ => "none".to_owned(),
+    }
+}
+
+fn block_ids(blocks: &[Block]) -> Vec<u64> {
+    blocks.iter().map(|block| block.header.block_id).collect()
+}
+
+fn pin_str(pin: Option<MsgId>) -> String {
```
erhant: was the term `pin` used anywhere else? if not, may be worth a docstring here to indicate that it is about pinning

lez/sequencer/core/src/lib.rs
```diff
@@ -1843,6 +1852,95 @@ async fn record_dead_letter_gauge<S: StorageActorTrait>(storage_ref: &ActorRef<S
...
+            &finalized
+                .iter()
+                .map(|(b, _)| b.header.block_id)
+                .collect::<Vec<_>>()
+        ),
+    );
+}
+
+/// Adoptions that did not apply, with the pin they were left behind by.
+fn log_parked(
```
erhant: perhaps these can go under `utils.rs`? or if we are even more restrictive, a `logging.rs` or such

erhant (approved):
LGTM, few nits

lez/sequencer/core/src/lib.rs
schouhy: ah yes, its used as the name for what will the parent be in next inscription

lez/sequencer/core/src/lib.rs
schouhy: yes that's much better. done.

Pravdyvy (approved):
lgtm

moudyellaz (approved):

PR #844, opened 2026-09-08
ci: repair host cache saves, move valid-proof-test to nightly, cancel superseded runs

moudyellaz:
## 🎯 Purpose

Two structural CI problems, both cheap to fix.

The host-runner Rust cache has never worked: `run-in-ci-image` runs docker as root, so `target/` comes back root-owned and the cache save dies with `tar: Permission denied`. `gh cache list` shows zero host-family entries. `integration-tests-prebuild`, which gates all 35 shards, compiles cold on every run (13m41s).

`valid-proof-test` is 78m and is the entire critical path of a merge run; everything else finishes by ~55m. Its 60m is one real proof, irreducible on a standard runner. The `auth_transfer` shard already covers that test under dev mode, so this job only establishes that real proving works, which is a daily property.

Expected: PR runs ~53m to ~40m, push runs ~83m to ~42m.

## ⚙️ Approach

- [x] Restore workspace ownership after `run-in-ci-image` so the cache can save
- [x] Move `valid-proof-test` to a nightly workflow with `workflow_dispatch`
- [x] Split `ci-rust-cache` into `ci-container` and `ci-host`, which were racing to save
- [x] Add top-level `concurrency:` so superseded PR runs cancel
- [x] Retries 5 to 2 (1 for the proof test), and no `cargo clean` between attempts
- [x] Buildkit `cache-to: mode=min`; the repo is at 11.28 GB against a 10 GB limit
- [x] Add a `ci` nextest profile so slow tests show up in the log

## 🧪 How to Test

The cache fix cannot show on this PR: `save-if` is limited to `main`/`dev`, so a PR branch never writes a cache. It populates on the first merge to `dev` and pays off from the run after.

After merge:

1. `gh cache list` shows `v0-rust-ci-host-*` entries, where there were none.
2. `integration-tests-prebuild` logs a cache restore and compiles well under 13m41s.
3. `gh workflow run valid-proof-test.yml` still passes in ~78m.
4. Push twice quickly to a PR branch; the first run cancels within seconds.

## 🔜 Future Work

- Route dev-mode proving through the existing `image_cache`, bypassed by ~76 `default_prover()` call sites, to cut the 25-35m `unit-tests` job.
- `image_cache` is behind `#[cfg(feature = "prove")]`, which the sequencer and indexer Dockerfiles do not enable, so the node still spawns r0vm per guest execution while CI does not.
- Rebalance the integration matrix: 35 shards, mean 10.6m, max 31m. Sharding is one per file, so it needs the file split.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [ ] Add/update tests

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,0 +1,13 @@
+# Timing visibility for CI. There is deliberately no `terminate-after`: some
+# tests legitimately run for many minutes, and the point here is to see which
+# ones, not to kill them.
+[profile.ci]
+# Print a SLOW line every period a test is still running, so the log shows the
+# expensive tests without needing the JUnit file.
+slow-timeout = { period = "120s" }
+# One failing test should not hide the timings of the rest.
+fail-fast = false
+status-level = "slow"
+
+[profile.ci.junit]
+path = "junit.xml"
--- a/.github/actions/run-in-ci-image/action.yml
+++ b/.github/actions/run-in-ci-image/action.yml
@@ -61,3 +61,19 @@ runs:
           "${env_args[@]}" \
           "$INPUT_IMAGE" \
           bash -e -o pipefail -c "$INPUT_RUN"
+
+    # The container runs as root, so everything it writes through a bind mount
+    # comes back root-owned and the unprivileged runner cannot tar it for the
+    # cache. Covers every mount above: rust-cache saves `.cargo/git` as well as
+    # `.cargo/registry`, so missing one still breaks the save. Runs on failure
+    # too, since a failed build leaves the same root-owned files behind.
+    - name: Restore workspace ownership
+      if: always()
+      shell: bash
+      run: |
+        set -euo pipefail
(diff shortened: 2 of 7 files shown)


.config/nextest.toml
```diff
@@ -0,0 +1,13 @@
...
+[profile.ci]
+# Print a SLOW line every period a test is still running, so the log shows the
+# expensive tests without needing the JUnit file.
+slow-timeout = { period = "120s" }
+# One failing test should not hide the timings of the rest.
+fail-fast = false
+status-level = "slow"
+
+[profile.ci.junit]
+path = "junit.xml"
```
erhant, line 13: are we using this?

erhant (approved):
LGTM, small comment

.config/nextest.toml
moudyellaz: Good catch, nothing reads it.

schouhy (approved):
lgtm

PR #845, opened 2026-09-08
feat(program_loader)!: efficient elf deployment by omitting kernel

jonesmarvin8:
## 🎯 Purpose

Each LEZ program's elf is stored in LEZ's public state. These elfs consist of two components: `kernel_elf` and `user_elf`. The `kernel_elf` is repeated for each program (and is constant in Risc0). This PR optimizes program deployment to exclude `kernel_elf`. 

BREAKING CHANGE!: programs deployed before this PR will no longer work due to their inclusion of `kernel_elf`; does not affect future deployment of testnet.

## ⚙️ Approach

- [X] Attach `kernel_elf` to `user_elf` before public executions.
- [X] Update wallet logic for privacy executions to expect `kernel_elf` to be given.

## 🧪 How to Test

All previous tests should work.

## 🔗 Dependencies

N/A

## 🔜 Future Work

N/A

## 📋 PR Completion Checklist

- [X] Complete PR description
- [X] Add/update tests


lee/state_machine/src/state/mod.rs
```diff
@@ -309,9 +309,10 @@ impl V03State {
     /// through [`get_program_via`] with the real address instead.
     #[must_use]
     pub fn get_program(&self, program_id: ProgramId) -> Option<(ProgramId, Vec<u8>)> {
-        get_program_via(AccountId::from(program_id), |account_id| {
+        let (image_id, user_elf) = get_program_via(AccountId::from(program_id), |account_id| {
             self.get_account_by_id(account_id)
-        })
+        })?;
+        Some((image_id, crate::program::attach_kernel(&user_elf)))
     }
```
Arjentix, line 316: So here we're basically trading runtime lookup speed for storage efficiency. Not saying it's bad, just and observation

lez/programs/program_loader/core/Cargo.toml
```diff
@@ -11,3 +11,4 @@ workspace = true
 lee_core.workspace = true
 borsh.workspace = true
 risc0-binfmt = "3.0.2"
+risc0-zkos-v1compat = "2.2.0"
```
Arjentix: Second time I see this dependency, let's move it to workspace

lez/wallet/src/lib.rs
```diff
@@ -104,6 +104,13 @@ pub enum ExecutionFailureKind {
     AccountDataError(AccountId),
     #[error("Program bytecode splits into {expected} segment(s) but {actual} were supplied")]
     SegmentCountMismatch { expected: usize, actual: usize },
+    #[error("Program bytecode is not a valid RISC0 program binary: {0}")]
+    InvalidProgramBinary(#[source] anyhow::Error),
```
Arjentix:
```suggestion
    #[error("Program bytecode is not a valid RISC0 program binary")]
    InvalidProgramBinary(#[source] anyhow::Error),
```

Common mistake. No need to include error string there. Information about error is already included in the error chain and can be pretty-printed if on the top call you wrap it with `anyhow` and then format with `{:#}`. But that only works if it's marked as `#[source]` or `#[from]`. For other data like `String` including it into error message is still good.

Maybe need to fix other variants in this type.

lez/wallet/Cargo.toml
```diff
@@ -23,6 +23,8 @@ programs.workspace = true
 system_accounts.workspace = true
 associated_token_account_core.workspace = true
 program_loader_core.workspace = true
+risc0-binfmt = "3.0.2"
```
Arjentix: Also consider workspace dependency

lee/state_machine/src/state/mod.rs
jonesmarvin8: Absolutely.

From your point of view; do you think this trade off is worth it?

lez/wallet/src/lib.rs
jonesmarvin8: Thanks for the clarification.

lee/state_machine/src/program/mod.rs
```diff
@@ -178,6 +185,12 @@ impl Program {
     }
 }
 
+/// Re-attaches the protocol's fixed kernel ELF to `user_elf`, producing a full `ProgramBinary`
+/// blob ready to decode and execute.
+pub(crate) fn attach_kernel(user_elf: &[u8]) -> Vec<u8> {
```
moudyellaz, line 190: The round-trip pin runs on `test_methods::noop()`, built in-tree against the same v1compat, so nothing catches a committed artifact whose kernel drifts; can it cover a builtin too?

lez/wallet/src/program_facades/program_loader.rs
```diff
@@ -177,7 +177,14 @@ impl ProgramLoader<'_> {
...
         }
-        let chunks: Vec<&[u8]> = bytecode.chunks(MAX_SEGMENT_DATA_LEN).collect();
+        // Every LEZ program is built with the protocol's default kernel, so only `user_elf` is
+        // segmented and uploaded.
+        let binary = risc0_binfmt::ProgramBinary::decode(&bytecode)
+            .map_err(ExecutionFailureKind::InvalidProgramBinary)?;
+        if binary.kernel_elf != risc0_zkos_v1compat::V1COMPAT_ELF {
+            return Err(ExecutionFailureKind::UnsupportedKernelElf.into());
+        }
+        let chunks: Vec<&[u8]> = binary.user_elf.chunks(MAX_SEGMENT_DATA_LEN).collect();
```
moudyellaz, line 187: `wallet_ffi.h:1025` and `:1052` still say `segments_len` matches the chunks `elf_data` splits into, which is now `user_elf`; worth correcting since downstream computes that count.

moudyellaz (approved):

lez/wallet/src/program_facades/program_loader.rs
jonesmarvin8: Fixed in [5c5d1ed](https://github.com/logos-blockchain/logos-execution-zone/pull/845/commits/5c5d1ed4efc719bc75dac290584f5db879afcbb0)

lee/state_machine/src/program/mod.rs
jonesmarvin8: Added a test for the privacy preserving circuit in [5c5d1ed](https://github.com/logos-blockchain/logos-execution-zone/pull/845/commits/5c5d1ed4efc719bc75dac290584f5db879afcbb0).

Cargo.toml
```diff
@@ -147,6 +147,8 @@ async-trait = "0.1"
 trait-variant = "0.1.2"
 risc0-zkvm = { version = "3.0.5", default-features = false, features = ['std'] }
 risc0-build = "3.0.5"
+risc0-binfmt = "3.0.2"
+risc0-zkos-v1compat = "2.2.0"
```
erhant: perhaps we should do `=2.2.0` here? a `cargo update` to move this could affect `V1COMPAT_ELF`

not that risc0 is alive and updated but worth being more safe

lez/wallet/src/program_facades/program_loader.rs
erhant: https://github.com/logos-blockchain/logos-execution-zone/blob/marvin/kernel-elf/lez/wallet/src/cli/programs/program_loader.rs#L77 and https://github.com/logos-blockchain/logos-execution-zone/blob/marvin/kernel-elf/lez/wallet/src/cli/programs/program_loader.rs#L99 still has outdated comment, worth fixing those as well!

erhant (approved):
LGTM, two comments!

Cargo.toml
jonesmarvin8: Pinned in [e62aa3a](https://github.com/logos-blockchain/logos-execution-zone/pull/845/commits/e62aa3a7c61130c55656682f8cf090e84c3ab502).

lez/wallet/src/program_facades/program_loader.rs
jonesmarvin8: Thanks. Fixed in [e62aa3a](https://github.com/logos-blockchain/logos-execution-zone/pull/845/commits/e62aa3a7c61130c55656682f8cf090e84c3ab502).

lez/programs/program_loader/core/Cargo.toml
jonesmarvin8: Thanks.

lez/wallet/Cargo.toml
jonesmarvin8: Thanks.

lee/state_machine/src/state/mod.rs
Arjentix: Yeah, looks fine to me

Arjentix (approved):
LGTM

PR #846, opened 2026-09-08
refactor(lee/lez): clean up associated to `ProgramId` and `Program`.

jonesmarvin8:
## 🎯 Purpose

This PR handles clean up from program as accounts PRs leading into RC.

## ⚙️ Approach

- [X] lowers data cap back to 100 KiB
- [X] cleans up `ProgramId`/`AccountId` usage.
- [X] Introduces a `DEFAULT_PROGRAM_OWNER` in place of `DEFAULT_PROGRAM_ID`
- [X] Reworks program deployment in genesis to be appropriately segmented using correct data cap.
- [X] Remove dead code.

## 🧪 How to Test

All previous tests should continue to work.

## 🔗 Dependencies

N/A

## 🔜 Future Work

N/A

## 📋 PR Completion Checklist

- [X] Complete PR description
- [X] Add/update tests


lee/state_machine/core/src/program/mod.rs
```diff
@@ -571,28 +568,6 @@ impl ProgramOutput {
...
-
-    pub fn valid_from_timestamp(mut self, ts: Option<Timestamp>) -> Result<Self, InvalidWindow> {
-        self.timestamp_validity_window = (ts, self.timestamp_validity_window.end()).try_into()?;
-        Ok(self)
-    }
-
-    pub fn valid_until_timestamp(mut self, ts: Option<Timestamp>) -> Result<Self, InvalidWindow> {
-        self.timestamp_validity_window = (self.timestamp_validity_window.start(), ts).try_into()?;
-        Ok(self)
-    }
```
jonesmarvin8, line 595: This code was identified as dead code. Let me know if it should stay.

examples/program_deployment/methods/guest/src/bin/simple_tail_call.rs
```diff
@@ -1,38 +1,31 @@
...
+// deploy time, so it can't be a compile-time constant.
 
-fn hello_world_program_id() -> ProgramId {
-    let hello_world_program_id_bytes: [u8; 32] = hex::decode(HELLO_WORLD_PROGRAM_ID_HEX)
-        .unwrap()
-        .try_into()
-        .unwrap();
-    bytemuck::cast(hello_world_program_id_bytes)
-}
+type Instruction = AccountId;
```
agureev: I think we can get rid of this alias, wdyt?

examples/program_deployment/methods/guest/src/bin/tail_call_with_pda.rs
```diff
@@ -1,43 +1,39 @@
...
 
-fn hello_world_program_id() -> ProgramId {
-    let hello_world_program_id_bytes: [u8; 32] =
-        hex::decode(HELLO_WORLD_WITH_AUTHORIZATION_PROGRAM_ID_HEX)
-            .unwrap()
-            .try_into()
-            .unwrap();
-    bytemuck::cast(hello_world_program_id_bytes)
-}
+type Instruction = AccountId;
```
agureev: Similarly here

lee/state_machine/core/src/program/mod.rs
```diff
@@ -33,14 +27,17 @@ pub const MAX_PROGRAM_SEGMENTS: usize = 20;
...
+    /// The default `AccountId` a builtin program is deployed at: a genesis-seeded program lives
+    /// here until (if ever) redeployed elsewhere via `program_loader`, at which point resolution
+    /// goes through the real segment chain instead — never trust this as a live address for a
+    /// program that could have moved.
+    ///
+    /// A 1:1, information-preserving mapping (both types are exactly 32 bytes) rather than a
+    /// hash — `ProgramId` is already content-derived (RISC0's `image_id`), so no extra domain
+    /// separation is needed just to reinterpret its 8 little-endian `u32` words as 32 raw bytes.
+    #[must_use]
+    pub fn builtin_default_address(program_id: ProgramId) -> Self {
```
Arjentix: Can you explain more why this is not okay for non-default programs and why it was ok before? I don't get it

Arjentix:
```suggestion
    pub fn from_builtin_program(program_id: ProgramId) -> Self {
```

Maybe like that? Just an idea

Arjentix: Btw what if we indeed redeploy such default programs? Their id might change. Or do we forbid that and that requires recompiling the chain code and redeploying?

lee/state_machine/core/src/encryption/mod.rs
```diff
@@ -234,7 +234,10 @@ mod tests {
         let receiver_ss = SharedSecretKey::decapsulate(&epk, &d, &z).unwrap();
 
         let account = Account {
-            program_owner: [12_u32; 8].into(),
+            program_owner: AccountId::new([
+                12, 0, 0, 0, 12, 0, 0, 0, 12, 0, 0, 0, 12, 0, 0, 0, 12, 0, 0, 0, 12, 0, 0, 0, 12,
+                0, 0, 0, 12, 0, 0, 0,
+            ]),
```
Arjentix, line 240: So maybe


```suggestion
            program_owner: AccountId::new([12; 32]),
```

Will work as well but looks much shorter? Same idea for other places like that

lez/sequencer/core/src/tests.rs
```diff
@@ -1982,38 +1993,61 @@ async fn user_tx_that_chain_calls_clock_is_dropped() {
...
+        let mut segment_account_ids = vec![segment_ids[i]];
+        segment_account_ids.extend(segment_ids.get(i + 1).copied());
+        let segment_message = lee::public_transaction::Message::try_new_with_fees(
+            lee_core::program::PROGRAM_LOADER_ACCOUNT_ID,
+            segment_account_ids,
+            vec![
+                lee_core::account::Nonce(0),
+                lee_core::account::Nonce(payer_nonce),
+            ],
+            program_loader_core::Instruction::WriteSegment {
```
Arjentix, line 2021: Looks like I missed that thing called program_loader. Can you please explain what it does?

lee/state_machine/src/state/tests/mod.rs
```diff
@@ -271,42 +274,54 @@ pub fn init_pda_witness(
...
+                        bytecode: chunk.to_vec(),
+                        next_segment: segment_account_ids.get(i + 1).copied(),
+                    }
+                    .to_bytes(),
+                )
+                .unwrap(),
+                ..Account::default()
+            },
+        );
+    }
```
Arjentix: Looks like it's used often in tests, no? Maybe we can move it into a special function for state? Maybe guarded by `#[cfg(test)]`

lez/sequencer/core/src/tests.rs
jonesmarvin8: `program_loader` is program used to deploy LEZ programs; it runs in native Rust (at state level) rather than in Risc0. Dealing with the elf code directly resulted in excessive Risc0 cycles, so it was decided to just run it in native Rust.

`program_loader` handles the account types for a program: `ProgramHeader` and `ProgramSegment`. The `ProgramHeader` holds (`image_id`, `program_first_segment` and `immutable`). Elfs are too large to store in one `Account`'s `data` entry. So, programs are stored in a linked-list of `ProgramSegment` accounts.

`program_loader` can create header, update header (change `immutable = false` to `true`, or change `program_first_segment`). program segments can only be created (currently, cannot be updated).

lee/state_machine/core/src/program/mod.rs
jonesmarvin8: Yes. A non-default program is deployed using any `AccountId` independent of the elf. Built in program's `AccountId`s are generated based on the elf's `image_id`.

Prior, all programs used a `ProgramId` that was the elf's `image_id`. However, this prevented programs from being (nicely) updated. So, in the transition of all programs deployed as accounts, we needed to fix this.

So, user deployed accounts use `AccountId`s that the user provides. And built in accounts use `AccountId`s generated from elfs. Honestly computed `AccountId`s won't collide since these use a different domain separator in their construction (built in vs programs deployed as regular/PDA accounts)

lee/state_machine/src/state/tests/mod.rs
jonesmarvin8: Moved it to `V03State::register_program`, alongside `force_insert_account` in the existing test/test-utils-gated impl block, and updated all 9 call sites to `state.register_program(...)`.

lee/state_machine/core/src/program/mod.rs

lee/state_machine/core/src/encryption/mod.rs
jonesmarvin8: Fixed; there were two spots that had to remain unchanged due to its usage.

lee/state_machine/core/src/program/mod.rs
```diff
@@ -33,14 +27,17 @@ pub const MAX_PROGRAM_SEGMENTS: usize = 20;
...
+    /// The default `AccountId` a builtin program is deployed at: a genesis-seeded program lives
+    /// here until (if ever) redeployed elsewhere via `program_loader`, at which point resolution
+    /// goes through the real segment chain instead — never trust this as a live address for a
+    /// program that could have moved.
+    ///
+    /// A 1:1, information-preserving mapping (both types are exactly 32 bytes) rather than a
+    /// hash — `ProgramId` is already content-derived (RISC0's `image_id`), so no extra domain
+    /// separation is needed just to reinterpret its 8 little-endian `u32` words as 32 raw bytes.
+    #[must_use]
+    pub fn from_builtin_program(program_id: ProgramId) -> Self {
```
moudyellaz, line 40: `From<ProgramId> for AccountId` is gone but its inverse at line 49 stays, and that is now the misleading direction: `convert.rs:284` turns an arbitrary header address into a "ProgramId" that is no image id at all; should it go too?

lee/state_machine/src/state/mod.rs
```diff
@@ -409,20 +404,83 @@ impl V03State {
...
+    /// `ProgramWithDependencies::from(program)` assumes.
+    /// `check_privacy_preserving_circuit_proof_is_valid` claims every top-level/dependency
+    /// program against real chain state, so any test that runs a privacy-preserving transaction
+    /// through `Self::transition_from_privacy_preserving_transaction` needs the program
+    /// registered here first, not just known to the local prover.
+    pub fn register_program(&mut self, program: &Program) {
+        let self_account_id = AccountId::from_builtin_program(program.id());
+        let elf = program.elf();
+        let chunks: Vec<&[u8]> = elf
+            .chunks(program_loader_core::MAX_SEGMENT_DATA_LEN)
```
moudyellaz, line 419: Segment addresses here are the header address with `i + 1` added to byte 0, while `insert_program` twenty lines up uses the domain-separated `genesis_segment_account_id(header, i)`; why two conventions?

moudyellaz (approved):

`examples/program_deployment/README.md:586` still names `DEFAULT_PROGRAM_ID`, which this PR removes.

lee/state_machine/core/src/program/mod.rs
moudyellaz: Confirmed dead: `try_with_timestamp_validity_window`, `valid_from_timestamp` and `valid_until_timestamp` have no callers at the merge base and none in the other org repos, so the removal is safe.

lee/state_machine/src/state/mod.rs
jonesmarvin8: `register_program` now uses `genesis_segment_account_id`'s approach. [6ca5ed9](https://github.com/logos-blockchain/logos-execution-zone/pull/846/commits/6ca5ed9eb40972c80063373989fae16a6bc2dd58)

lee/state_machine/core/src/program/mod.rs
jonesmarvin8: Taken care of in  [6ca5ed9](https://github.com/logos-blockchain/logos-execution-zone/pull/846/commits/6ca5ed9eb40972c80063373989fae16a6bc2dd58).

lee/state_machine/src/state/mod.rs
```diff
@@ -409,20 +404,77 @@ impl V03State {
...
     }
+
+    /// Registers `program` at its own bijection address, as a `program_loader` deploy chunked
+    /// into segments the same way a live deploy would be — the shape
+    /// `ProgramWithDependencies::from(program)` assumes.
+    /// `check_privacy_preserving_circuit_proof_is_valid` claims every top-level/dependency
+    /// program against real chain state, so any test that runs a privacy-preserving transaction
+    /// through `Self::transition_from_privacy_preserving_transaction` needs the program
+    /// registered here first, not just known to the local prover.
+    pub fn register_program(&mut self, program: &Program) {
```
erhant: following on top of https://github.com/logos-blockchain/logos-execution-zone/pull/846/changes#r3984747181, I want to say that this function looks very much like `insert_program`, maybe they can be made into one?

jonesmarvin8: Addressed in [c143fdb](https://github.com/logos-blockchain/logos-execution-zone/pull/846/commits/c143fdb9819c71e766a3ae6180dc650a850bbff8)

examples/program_deployment/README.md
erhant, line 1: the README does not mention `payer_account_id` and it's related commands may panic as described here

lee/state_machine/src/lib.rs
```diff
@@ -34,6 +34,12 @@ mod state;
 pub mod test_utils;
 mod validated_state_diff;
 
+#[cfg(all(feature = "test-utils", not(debug_assertions)))]
```
erhant, line 39: hmm while I like the extra caution here, `debug_assertions` is not always "if-and-only-if" for a test build

e.g. one can do `cargo test ... --release` or have `[profile.release] debug-assertions = true`

perhaps worth noting if there in case some test unexpectedly fails here

examples/program_deployment/src/lib.rs
```diff
@@ -0,0 +1,27 @@
...
+///
+/// `payer` must be an existing, funded account. A freshly-claimed account cannot pay for its own
+/// claim (funding it first would claim it via the transfer guest instead), so deployment is
+/// always paid for by a separate, already-funded account.
+pub async fn deploy_program(
+    wallet_core: &mut WalletCore,
+    bytecode: Vec<u8>,
+    payer: AccountId,
+) -> anyhow::Result<AccountId> {
+    let segment_count = bytecode.len().div_ceil(MAX_SEGMENT_DATA_LEN).max(1);
```
erhant: we likely don't need `.max(1)` here; div-ceil will already return something at least 1; and the only case it would return 0 is when `bytecode` is empty. that case will be handled by `ProgramLoader::deploy` below with a `bail`

erhant: or we can simply add an assert at the top to check bytecode is non-empty

lez/indexer/service/protocol/src/lib.rs
erhant, line 1: we can also remove ProgramId structure in this file?

https://github.com/logos-blockchain/logos-execution-zone/blob/marvin/program-removal/lez/indexer/service/protocol/src/lib.rs#L60

lez/indexer/service/Cargo.toml
```diff
@@ -11,6 +11,7 @@ workspace = true
 indexer_service_protocol = { workspace = true, features = ["convert"] }
 indexer_service_rpc = { workspace = true, features = ["server"] }
 indexer_core.workspace = true
+lee_core.workspace = true
```
erhant: from what I can see `lee_core` is used only in tests, if true, let's make it a dev dependency instead

erhant (approved):
LGTM overall, some comments made

examples/program_deployment/README.md
jonesmarvin8: Created [Issue 895](https://github.com/logos-blockchain/logos-execution-zone/issues/895). I'm pretty sure tutorials are also stale with all the recent updates. Will address these in a future PR.

lez/indexer/service/Cargo.toml
jonesmarvin8: Fixed in [743fcbb](https://github.com/logos-blockchain/logos-execution-zone/pull/846/commits/743fcbbefe7fbd5b4f248a902d632695a2272c0e).

lez/indexer/service/protocol/src/lib.rs
jonesmarvin8: Fixed in [743fcbb](https://github.com/logos-blockchain/logos-execution-zone/pull/846/commits/743fcbbefe7fbd5b4f248a902d632695a2272c0e).

examples/program_deployment/src/lib.rs
jonesmarvin8: Fixed in [743fcbb](https://github.com/logos-blockchain/logos-execution-zone/pull/846/commits/743fcbbefe7fbd5b4f248a902d632695a2272c0e).

lee/state_machine/src/lib.rs
jonesmarvin8: Fixed in [743fcbb](https://github.com/logos-blockchain/logos-execution-zone/pull/846/commits/743fcbbefe7fbd5b4f248a902d632695a2272c0e).

lee/state_machine/core/src/program/mod.rs
Arjentix: > So, user deployed accounts use AccountIds that the user provides.

Oh, that's something new, cool. However why not to just assign some beautiful `AcctountId`s for built-in programs as well? Won't it reduce the pain?

lez/sequencer/core/src/tests.rs
Arjentix: Wow, sounds cool, thanks

lee/state_machine/core/src/program/mod.rs
Arjentix: Because I had a discussion with @agureev about dependencies from auth transfer. And having a fixed AccountId for it would help us if I remember correctly

Arjentix (approved):
LGTM, one question [here](https://github.com/logos-blockchain/logos-execution-zone/pull/846#discussion_r4030399785)

lee/state_machine/core/src/program/mod.rs
jonesmarvin8: I'm not opposed to doing that for built-in programs;
`auth_transfer_account_id` being given by a hash based. Though, if we do it, we should avoid using `LEE/V0.3/` formatting. This is to avoid any (future) breaking changes in case the LEE protocol has a major version update.

For example, the key protocol has a separate version-ing approach as well.

This is the only caution I'd recommend for the "beautiful AccountIds".

jonesmarvin8:
> LGTM, one question [here](https://github.com/logos-blockchain/logos-execution-zone/pull/846#discussion_r4030399785)

Replied to the comment.

lez/wallet-ffi/wallet_ffi.h
```diff
@@ -624,7 +629,7 @@ enum WalletFfiError wallet_ffi_send_generic_public_transaction(struct WalletHand
                                                                uintptr_t account_identities_size,
                                                                const uint8_t *instruction_data,
                                                                uintptr_t instruction_data_size,
-                                                               struct FfiProgramId program_id,
+                                                               struct FfiBytes32 program_account_id,
```
moudyellaz, line 632: Deleting `FfiProgramId` breaks `logos-execution-zone-module`: `src/lez_core_module.cpp:836` declares one and passes it to this function, and `tests/stubs/wallet_ffi.h` redeclares the type. That repo builds against this header, so it stops compiling once this lands, and it has no PR open for the fix.

moudyellaz (approved):
thanks for addressing my comments. Left a new one on the new commits!

lez/wallet-ffi/wallet_ffi.h
jonesmarvin8: Added companion [PR](https://github.com/logos-blockchain/logos-execution-zone-module/pull/59)

lee/state_machine/core/src/program/mod.rs
jonesmarvin8: Added [Issue 906](https://github.com/logos-blockchain/logos-execution-zone/issues/906)

PR #874, opened 2026-09-10
ci: make integration-tests-prebuild the sole writer of the host rust cache

moudyellaz:
## 🎯 Purpose

#844 fixed the *save* half of the host Rust cache. The *restore* half never paid off: `integration-tests-prebuild` reported a cache hit and still compiled for ~14m.

Cause, in two parts. All four host jobs shared `shared-key: ci-host` while building different things, so a faster job wrote the cache and prebuild restored a `target/` cargo could not use. Worse, prebuild's own environment hash is `30985b06` while the entry under that key was `2f32086d`, so the "hit" was a restore-key prefix fallback from a different environment. rust-cache then reported `Cache up-to-date` and never replaced it, so the wrong cache persisted indefinitely.

Renaming the key breaks that stalemate: nothing exists under the new name, so the sole writer creates it and restores its own tree thereafter.

Measured on this branch, prebuild:

| | compile | job |
|---|---|---|
| cold (three runs: 13m41s, 13m56s, 14m20s) | ~14m | 20m8s |
| own cache restored | **4m18s** | **14m22s** |

About 10 minutes off the compile, ~6 minutes off the job after paying 73s to restore 2.37GB. prebuild gates all 35 shards, so this comes off the PR critical path, which is prebuild then the longest shard.

The key was also unstable run to run. `rust-cache` hashes every toolchain on the host into its key, and the runner image ships a rolling Rust (1.98.0 vs 1.98.1 today), so runs landed in two pools that can never restore each other:

    Restore Key:  v0-rust-ci-host-prebuild-Linux-x64-2f32086d

The environment hash is part of the restore key, not just the exact key. Nothing builds on the host here (all four jobs compile inside the CI image on its pinned 1.94.0), so the new `pin-host-toolchain` action drops the preinstalled toolchain and leaves only the one from `rust-toolchain.toml`. The key then derives solely from a repo-pinned version.

## ⚙️ Approach

- [x] Rename the host key to `ci-host-prebuild`, abandoning the unusable shared entry
- [x] `integration-tests-prebuild` is the sole writer; `test-fixtures-tests`, `artifacts` and the nightly `valid-proof-test` are read-only (`save-if: false`)
- [x] Drop the runner's preinstalled Rust before `rust-cache` so the key is deterministic

Budget-neutral: still one host cache, and the repo is at 9.16GB of the 10GB limit. Also removes the `Unable to reserve cache` race.

## 🧪 How to Test

Measured pre-merge by temporarily letting the writer fire on this branch, then removing that clause. Sequence on the branch: cold run wrote the cache (`No cache found` then `Saving cache`), next run restored it (`Cache hit` then `Finished in 4m 18s`).

After merge, on the second dev run: prebuild should log a hit on `v0-rust-ci-host-prebuild-*` and compile in ~4m rather than ~14m. The first dev run writes the cache under the new key and will still be cold.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [ ] Add/update tests


erhant (approved):
LGTM, perhaps we can tune down the comment verbosity a bit

Pravdyvy (approved):
lgtm

PR #875, opened 2026-09-10
refactor(lez)!: remove the faucet program and replace its genesis rol…

schouhy:
## 🎯 Purpose

The bridge escrow balance is defined by genesis config. Devnet shipped `1_000_000`. This is problematic for deposits over that amount which is the normal devnet case. Also, the faucet program is, apart from genesis supply, pretty much unused. This PR removes the faucet program and relies on the bridge program for genesis funding. The escrow now holds the whole supply from genesis.

## ⚙️ Approach

- [x] Change: the bridge escrow starts at `u128::MAX` and the faucet at nothing. What the escrow still holds is the unissued supply. `system_accounts/src/lib.rs:36`
- [x] Change: genesis allocations and founding stakes are bridge `Deposit`s, not faucet transfers. `sequencer/core/src/lib.rs:2491`
- [x] Add: `genesis_deposit_op_id`, the op id of the nth genesis allocation. Literal ASCII `/LEZ/v0.3/GenesisDeposit` plus the index, which no L1 op id can be because those are hashes. Unique per allocation is the part that matters: a shared op id means a shared receipt PDA, and every allocation after the first is a silent no-op. `sequencer/core/src/lib.rs:2476`
- [x] Remove: `GenesisAction::SupplyBridgeAccount`, and the entry from the four shipped configs
- [x] Remove: the faucet program, crate, artifact, system account, its restricted-list entry, and its four guard tests. The bridge equivalents already exist in `integration_tests/tests/bridge.rs` and the `bridge_guard_*` unit tests.

Nothing in the bridge program changed, so `BRIDGE_ID`, the escrow address and every receipt PDA are the same.

## 🧪 How to Test

```
cargo test -p sequencer_core --lib --features mock,test-utils
cargo test -p common -p testnet_initial_state -p test_fixtures --lib
```

Genesis follows config, both paths:

```
cargo test -p test_fixtures --test prebuilt_fixture
```

## 🔗 Dependencies

None.

## 🔜 Future Work

None

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


lez/sequencer/core/src/lib.rs
```diff
@@ -2481,19 +2467,50 @@ pub fn is_sequencer_only_program(program_account_id: AccountId) -> bool {
...
+/// Genesis allocations are `Deposit`s with no L1 event behind them, so their op
+/// ids must be unmistakable: an L1 op id is a hash, and this is a literal ASCII
+/// domain followed by the index, which no hash realistically produces. The
+/// receipt PDA each one claims is what stops a later block replaying it.
+fn genesis_deposit_op_id(index: usize) -> [u8; 32] {
+    const DOMAIN: &[u8; 24] = b"/LEZ/v0.3/GenesisDeposit";
+
+    let mut op_id = [0_u8; 32];
+    op_id[..DOMAIN.len()].copy_from_slice(DOMAIN);
+    op_id[DOMAIN.len()..].copy_from_slice(&u64::try_from(index).unwrap_or(u64::MAX).to_le_bytes());
```
erhant: we could `.expect` instead of `.unwrap_or(u64::MAX)` I think, index should fit in u64

lez/sequencer/core/src/config.rs
```diff
@@ -36,19 +36,12 @@ pub const MAX_PUBLISHABLE_BLOCK_SIZE: u64 =
...
+    /// `u64`, not [`Balance`]: the bridge's `Deposit` amount is `u64`, and this
+    /// is funded through it.
+    SupplyAccount { account_id: AccountId, balance: u64 },
+    /// Funds a holder's holding PDA at genesis with one replayable genesis
     /// credit; the balance-only PDA needs no claim.
-    SupplyBridgeLockHolding {
-        holder: AccountId,
-        amount: Balance,
-    },
+    SupplyBridgeLockHolding { holder: AccountId, amount: Balance },
```
erhant: the `balance` here is `Balance` but for `SupplyAccount` it's u64 (i see the explanation but still), could we make this one u64 too maybe or vice-versa?

erhant (approved):
LGTM, two comments

lez/sequencer/core/src/lib.rs
moudyellaz: Seconding, and there is a concrete reason: `genesis_stake_deposit_op_id` passes `usize::MAX`, which is `u32::MAX` on a 32-bit host, so the op id and its receipt PDA differ from a 64-bit node's. Take `index: u64` and use `u64::MAX` as the stake sentinel; the conversion disappears.

lez/sequencer/core/src/config.rs
moudyellaz: u64 is the right answer for this one: an `amount` above `u64::MAX` deserializes fine and then panics at `lib.rs:2201`, so the cap belongs in serde at load, not in an expect at genesis.

lez/sequencer/core/src/lib.rs
```diff
@@ -2415,22 +2414,11 @@ fn build_stake_genesis_transactions(
...
-    .expect("Failed to build genesis funding message");
-    // The funding account signs even though it is only receiving: the stake
-    // transactions below count their nonces from 1 on the strength of it.
-    let fund_witness_set =
-        lee::public_transaction::WitnessSet::for_message(&fund_message, &[&funding_key]);
-
-    let mut txs = vec![PublicTransaction::new(fund_message, fund_witness_set)];
+    let mut txs = vec![build_supply_account_genesis_transaction(
+        &genesis_stake_funding_account(),
+        u64::try_from(total).expect("genesis stake total exceeds u64"),
```
moudyellaz: One deposit funds every founding stake, so `minimum_sequencer_stake * founding_count` must fit u64 or genesis panics here, while the param itself is u128. Cap it in a doc comment, or emit one deposit per stake with its own op id?

integration_tests/tests/sequencer_stake_demo.rs
```diff
@@ -60,7 +60,7 @@ async fn stake_transaction_joins_the_bedrock_committee() -> Result<()> {
                 .with_sequencer_partial_config(fast_blocks())
                 .with_genesis(vec![GenesisAction::SupplyAccount {
                     account_id: funding_id,
-                    balance: FUNDING_BALANCE,
+                    balance: u64::try_from(FUNDING_BALANCE).expect("funding balance fits u64"),
```
moudyellaz: `tps.rs` changed its const to u64; make `FUNDING_BALANCE` u64 too and drop the conversion.

moudyellaz (commented):
lgtm

lez/sequencer/core/src/lib.rs
```diff
@@ -2481,19 +2467,50 @@ pub fn is_sequencer_only_program(program_account_id: AccountId) -> bool {
         || program_account_id == programs::fee().id().into()
 }
 
+/// Op id of the `index`-th genesis allocation.
+///
+/// Genesis allocations are `Deposit`s with no L1 event behind them, so their op
+/// ids must be unmistakable: an L1 op id is a hash, and this is a literal ASCII
+/// domain followed by the index, which no hash realistically produces. The
+/// receipt PDA each one claims is what stops a later block replaying it.
+fn genesis_deposit_op_id(index: usize) -> [u8; 32] {
```
Arjentix: What do you think about introducing enum instead of fake hashes? Maybe that's more code but at least this invariant is more readable in the whole system

Arjentix (approved):
LGTM, one suggestion

agureev (approved):

lez/sequencer/core/src/lib.rs
schouhy: one thing to note is that, at any point, the channel's net value (what has been deposited into it and not yet withdrawn) is exactly what can be withdrawn back to L1. I think a problem with our current flow is that it creates the channel and writes the genesis funding txs with native tokens that are note backed by any deposit from  L1. That smells bad, because it means the circulating supply in LEZ exceeds the net value (forever).
I think the right approach would be: 
1. The channel is created by submitting the initial channel config with all the initial channel writers. At this point the channel is just controlled by those keys. Still has nothing specific to LEZ.
2. Each of the writers deposit funds into the channel. This provides real `l1_deposit_op_id` for each.
3. The genesis block is constructed using lez bridge txs that point to real l1 deposits chained with stake txs to put those at stake for each of the channel writers.
4. At this point each of the channel writers becomes a sequencer and LEZ is live. And all the native tokens staked are backed by some real deposit.

That's why I think we should keep the fake op ids because they are just temporary.

lez/sequencer/core/src/lib.rs
schouhy: done

lez/sequencer/core/src/config.rs
schouhy: done

lez/sequencer/core/src/lib.rs
schouhy: done. One deposit per stake

integration_tests/tests/sequencer_stake_demo.rs
schouhy: done

PR #876, opened 2026-09-10
fix(nix): follow logos-nix and apply its crates.io overlays

igor-sirotin:
## 🎯 Purpose

#838 fixed the one hand-rolled fetch, which unblocked *evaluation*. It did not touch the fetchers `nixpkgs` itself uses, so the build is still broken cold. It passes on dev machines and cached runners because these are content-addressed FODs that `cache.nixos.org` serves without contacting crates.io.

161 fixed-output derivations in the build closure download from `https://crates.io/api/v1/crates/<crate>/<version>/download`, which answers **403** to the `curl/...` User-Agent `fetchurl` sends ([rust-lang/crates.io#13482](https://github.com/rust-lang/crates.io/issues/13482)).

## ⚙️ Approach

- [x] Follow `logos-nix` rather than `logos-liblogos`.
- [x] Apply `logos-nix.lib.nativeOverlays` in `mkPkgs`.

logos-nix recently got fixes to fix Rust builds:
- https://github.com/logos-co/logos-nix/pull/9
- https://github.com/logos-co/logos-nix/pull/10
- https://github.com/logos-co/logos-nix/pull/11
- https://github.com/logos-co/logos-nix/pull/12

`logos-nix` is used centrally across most Logos repositories.
Future versions will bump `nixpkgs` to a newer version and gain proper Rust fix, but this implies a Qt upgrade which we're trying to avoid.

`logos-liblogos` was an input purely to carry that pin — it is referenced nowhere in `outputs` — so `flake.lock` goes from 450 nodes to 10. The deleted nodes are the Logos Core Qt graph, including 147 duplicated `nixpkgs` entries.

## 🧪 How to Test

```
nix-store -qR --include-outputs $(nix eval --raw .#packages.<system>.default.drvPath) \
  | grep '\.drv$' | xargs grep -l 'crates.io/api/v1' | wc -l
```

161 on `dev`, 1 here. That one is the unpatched `fetch-cargo-vendor-util` script, still a `nativeBuildInput` of the input-addressed `-vendor` step, which does no network I/O; all five network-facing `-vendor-staging` FODs invoke the patched helper.

## 🔗 Dependencies

None.

## 🔜 Future Work

- The zkr-hash IFD is built for the target platform, so `packages.x86_64-linux` cannot be evaluated from an aarch64 machine — follow-up PR, stacked on this one.
- `packages.x86_64-windows` is declared in `systems` but does not evaluate at all: `import nixpkgs { system = "x86_64-windows"; }` is not a valid package set. Left alone; `logos-nix.lib.mkWindowsPkgs` is the supported cross path if Windows is actually wanted.

## 📋 PR Completion Checklist

- [x] Complete PR description

<details>
<summary>AI-made decisions</summary>

- Re-locked with `nix flake lock`, not `nix flake update`: the latter also bumps `rust-overlay` (`c06d86d` → `577bb1e`) and the Rust toolchain with it.
- Passed the whole `lib.nativeOverlays` list rather than naming individual overlays, per the logos-nix README, so a future overlay is not silently missed.
- Ordered `nativeOverlays ++ [ rust-overlay ]`; checked that the toolchain still resolves to `rust-default-1.96.0`.
- Crane is not implicated — its pinned rev already downloads from `static.crates.io`. Nor are `logos-blockchain-circuits` / `rust-rapidsnark`, which only fetch GitHub release assets.

</details>


erhant (approved):
LGTM

Arjentix (approved):

PR #878, opened 2026-09-10
feat(sequencer_ffi): Fullfillment of missing queries 1

Pravdyvy:
## 🎯 Purpose

Updating sequencer query interfaces, setting it to a state of current indexer.

## ⚙️ Approach

- [x] Implemented block by hash query.
- [x] Implemented transaction by account query.

## 🧪 How to Test

Added corresponding integration test.

## 🔗 Dependencies

#823 

## 🔜 Future Work

- Implement event query. Needs DB modifications. See #891 .
- Implement status query. Needs substantial DB modifications.
- Paginate queries.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


lez/sequencer/actors/storage/src/actor.rs
```diff
@@ -384,6 +434,71 @@ impl StorageActor {
...
+            log::info!(
+                "############################### Meta acc data len is {}",
+                meta_acc_data.length
+            );
+
+            let mut affecting_txs = vec![];
+            let mut local_offset = 0;
+            let mut local_limit = 0;
+
+            'outer: for i in 0..meta_acc_data.length {
```
schouhy, line 589: shouldnt this be paginated? for clock accounts this will be a very long loop eventually

lez/sequencer/actors/storage/src/actor/entities.rs
```diff
@@ -40,14 +41,20 @@ pub enum ColumnFamily {
     Meta,
     /// Small records deleted as the work they track settles.
     Pending,
+    /// Many small records, maps block hases to block ids.
+    BlockHashToBlockIdMap,
```
schouhy, line 45: are these pruned when blocks are deleted on reorgs?

lez/sequencer/actors/storage/src/actor.rs
```diff
@@ -209,6 +210,55 @@ impl StorageActor {
...
+
+            let affected_accounts = block
+                .block
+                .body
+                .transactions
+                .iter()
+                .flat_map(LeeTransaction::affected_public_account_ids);
+
+            for acc in affected_accounts {
+                let meta_acc_opt = self.db().get::<entities::AccountIdToBlockIdMetaLen>(&acc)?;
```
schouhy: this is inside a bigger loop that goes through different blocks. If two or more of those blocks affect the same account, I think only the first one will have the right `meta_acc_opt` value. Because writes are to the batch and dont apply immediately. So the next get still reads the same value?

schouhy (approved):
lgtm! left some comments

lez/sequencer/actors/storage/src/actor.rs
```diff
@@ -384,6 +434,71 @@ impl StorageActor {
...
+        offset: u64,
+        limit: u64,
+    ) -> Result<Option<Vec<LeeTransaction>>> {
+        let meta_acc_opt: Option<entities::AccountIdToBlockIdMetaLen> =
+            self.db()
+                .get::<entities::AccountIdToBlockIdMetaLen>(&account_id)?;
+
+        if let Some(meta_acc_data) = meta_acc_opt {
+            log::info!(
+                "############################### Meta acc data len is {}",
```
erhant: i guess this is a debug-log from before that should be removed?

lez/sequencer/actors/storage/src/actor/entities.rs
```diff
@@ -40,14 +41,20 @@ pub enum ColumnFamily {
     Meta,
     /// Small records deleted as the work they track settles.
     Pending,
+    /// Many small records, maps block hases to block ids.
```
erhant: typo here `hashes`

lez/sequencer/actors/storage/src/actor.rs
```diff
@@ -384,6 +434,71 @@ impl StorageActor {
...
+                    )))?
+                    .block
+                    .body
+                    .transactions
+                    .into_iter()
+                    .filter(|tx| tx.affected_public_account_ids().contains(&account_id));
+
+                for aff_tx in affecting_transactions {
+                    if local_offset >= offset {
+                        if local_limit < limit {
```
erhant, line 614: can be a single `if` with &&

lez/sequencer/ffi/src/api/types/vectors.rs
erhant, line 1: i wonder if `lists.rs` is a better name for this guy

lez/sequencer/ffi/src/api/lifecycle.rs
```diff
@@ -0,0 +1,229 @@
...
+    // SAFETY: The caller must ensure the validness of the pointer arguments.
+    unsafe { setup_sequencer(runtime, config_path) }.map_or_else(
+        InitializedSequencerServiceFFIResult::from_error,
+        InitializedSequencerServiceFFIResult::from_value,
+    )
+}
+
+/// Creates all components of sequencer service except RPC.
+///
+/// `ToDo`: Make RPC construction optional. Probably needs modification of configs to be alighned
```
erhant:
```diff
- `ToDo`: Make ...
+ TODO: Make ...
```

we should write todo comments as `TODO: ...`, this convention is picked up by many IDE tools & is generally the standard

(same goes for `FIXME: ...`)

lez/sequencer/ffi/src/runtime.rs
erhant, line 1: good stuff, i wonder if there is a library for this though

lez/sequencer/ffi/src/sequencer.rs
```diff
@@ -0,0 +1,141 @@
...
+use sequencer_executor_actor::ExecutorActor;
+use sequencer_service::Gossip;
+use sequencer_slasher_actor::SlasherActor;
+use sequencer_storage_actor::StorageActor;
+
+use crate::Runtime;
+
+/// FFI-owned sequencer.
+///
+/// - A [`ActorRef<StorageActor>`] used to get acess to db.
```
erhant: Instead of:

```
- A [`ActorRef<StorageActor>`] used to get acess to db.
```

we can maybe say:

```
- `storage_ref`: an [`ActorRef<StorageActor>`] used to get acess to db.
```

and such; it's better to pair the docstring to arg when by name rather than by position, in case this changes later as well (or some renames take place)

lez/sequencer/ffi/src/api/query.rs
```diff
@@ -0,0 +1,692 @@
...
+    limit: u64,
+) -> PointerResult<FfiVec<FfiBlock>, OperationStatus> {
+    if sequencer.is_null() {
+        log::error!("Attempted to query a null sequencer pointer. This is a bug. Aborting.");
+        return PointerResult::from_error(OperationStatus::NullPointer);
+    }
+
+    let sequencer = unsafe { &*sequencer };
+
+    let before_opt = before.is_some.then_some(unsafe { before.value.read() });
```
erhant: should do lazy eval here

```diff
- let before_opt = before.is_some.then_some(unsafe { before.value.read() });
+ let before_opt = before.is_some.then(|| unsafe { before.value.read() });
```

erhant (approved):
LGTM, some comments made

lez/sequencer/actors/storage/src/actor.rs
Pravdyvy: `offset` and `limit` parameters kind of do that. Anyhow, this request on indexer was not paginated, so can be done later.

lez/sequencer/actors/storage/src/actor/entities.rs
Pravdyvy: Added deletion on reorgs for all mapping columns.

lez/sequencer/actors/storage/src/actor.rs
Pravdyvy: True. Refactored this part, now accumulate all changes for this column before pushing.

lez/sequencer/actors/storage/src/actor.rs
Pravdyvy: Removed.

lez/sequencer/actors/storage/src/actor/entities.rs
Pravdyvy: Fixed.

lez/sequencer/actors/storage/src/actor.rs
Pravdyvy: Yes, but it is much more obvious this way, what do you think?

lez/sequencer/ffi/src/api/types/vectors.rs
Pravdyvy: Maybe, for now let's stay that way in line with indexer ffi.

lez/sequencer/actors/storage/src/actor.rs
erhant: up to you 🙏🏻 , although I recall clippy usually noticed these

lez/sequencer/ffi/src/runtime.rs
Pravdyvy: I am not sure, it was taken from indexer ffi which in turn was taken from bedrock ffi, how it entered there I have no clue.

lez/sequencer/ffi/src/api/lifecycle.rs
Pravdyvy: Fixed.

lez/sequencer/ffi/src/sequencer.rs
Pravdyvy: Fixed.

lez/sequencer/ffi/src/api/query.rs
Pravdyvy: Updated.

lez/sequencer/actors/executor/src/protocol.rs
```diff
@@ -146,3 +146,13 @@ pub struct RequeueCrossZoneDeadLetter {
...
+
+pub struct GetBlockHashToBlockIdMapItem {
+    pub block_hash: HashType,
+}
+
+pub struct GetAccountIdToAffectingTxMapItemUptoLimit {
+    pub account_id: AccountId,
+    pub offset: u64,
+    pub limit: u64,
+}
```
Arjentix, line 158:
```suggestion
pub struct GetBlockByHash {
    pub block_hash: HashType,
}

pub struct GetAccountTransactions {
    pub account_id: AccountId,
    pub offset: u64,
    pub limit: u64,
}
```

Let's rename to that so that it's more readable and doesn't expose internal implementation with map

lez/sequencer/actors/storage/src/protocol.rs
```diff
@@ -126,6 +126,16 @@ pub struct SetCrossZonePeerTip {
...
 
+pub struct GetBlockHashToBlockIdMapItem {
+    pub block_hash: HashType,
+}
+
+pub struct GetAccountIdToAffectingTxMapItemUptoLimit {
+    pub account_id: AccountId,
+    pub offset: u64,
+    pub limit: u64,
+}
```
Arjentix:
```suggestion
pub struct GetBlockByHash {
    pub block_hash: HashType,
}

pub struct GetAccountTransactions {
    pub account_id: AccountId,
    pub offset: u64,
    pub limit: u64,
}
```

Same

lez/sequencer/actors/storage/src/mock.rs
```diff
@@ -245,6 +246,18 @@ mockall::mock! {
...
             ctx: &mut Context<Self, Result<DbDump>>
         ) -> Result<DbDump>;
+
+        pub fn handle_get_block_hash_to_block_id_map_item(
+            &mut self,
+            msg: GetBlockHashToBlockIdMapItem,
+            ctx: &mut Context<Self, Result<Option<u64>>>
+        ) -> Result<Option<u64>>;
+
+        pub fn handle_get_account_id_to_block_id_map_item_uo_to_a_limit(
```
Arjentix:
```suggestion
        pub fn handle_get_block_by_hash(
            &mut self,
            msg: GetBlockHashToBlockIdMapItem,
            ctx: &mut Context<Self, Result<Option<u64>>>
        ) -> Result<Option<u64>>;

        pub fn handle_get_account_transactions(
```

Same

lez/sequencer/actors/storage/src/actor.rs
```diff
@@ -206,30 +229,193 @@ impl StorageActor {
...
+                        accounts_affected_transactions
+                            .entry(acc)
+                            .and_modify(|meta_update| meta_update.removed_block_ids.push(stale_id))
+                            .or_insert_with(|| AffectedAccountsMetaUpdate {
+                                old_meta_length: starting_meta_acc_opt.map_or(0, |val| val.length),
+                                updated_block_ids: vec![],
+                                removed_block_ids: vec![stale_id],
+                            });
+                    }
+                }
```
Arjentix: I can see this is also repeated in `stage_new_blocks()`, maybe we can move it into a function?

lez/sequencer/actors/storage/src/actor.rs
```diff
@@ -206,30 +229,193 @@ impl StorageActor {
...
+            } else {
+                break;
+            }
+
+            cycle_counter = cycle_counter.saturating_add(1);
+            iteration_key = entities::AccountIdToBlockIdKey::from((
+                account_id,
+                meta_pre_state.saturating_sub(cycle_counter),
+            ));
+        }
```
Arjentix: Can you use `self.db().iter()` instead? I think this should work if key ordering is right

lez/sequencer/actors/storage/src/actor.rs
```diff
@@ -384,6 +570,66 @@ impl StorageActor {
...
+        &self,
+        account_id: AccountId,
+        offset: u64,
+        limit: u64,
+    ) -> Result<Option<Vec<LeeTransaction>>> {
+        let meta_acc_opt: Option<entities::AccountIdToBlockIdMetaLen> =
+            self.db()
+                .get::<entities::AccountIdToBlockIdMetaLen>(&account_id)?;
+
+        if let Some(meta_acc_data) = meta_acc_opt {
```
Arjentix:
```suggestion
        let Some(meta_acc_data) = meta_acc_opt else {
            return Ok(None);
        }
```

Let's do like that to remove big nesting

lez/sequencer/actors/storage/src/actor.rs
```diff
@@ -206,30 +229,193 @@ impl StorageActor {
...
         batch: &mut db::WriteBatch,
         highest_block_id: BlockId,
         head_tip: Option<BlockMeta>,
-    ) -> Vec<BlockId> {
+        accounts_affected_transactions: &mut BTreeMap<AccountId, AffectedAccountsMetaUpdate>,
+    ) -> Result<Vec<BlockId>> {
         let mut removed_block_ids = Vec::new();
         if let Some(head_tip) = head_tip {
             for stale_id in head_tip.id.saturating_add(1)..=highest_block_id {
+                // TODO: Handle absence of stale block
```
Arjentix, line 325: Is this need to be addressed in this PR?

lez/sequencer/actors/storage/src/protocol.rs
Arjentix: I understand that my idea will contradict to your already done implementation, but what do you think about in-memory index for these queries instead of really storing them in database? Here is my view:

Pros:
1. It removes db consistency errors which are easy to do and hard to track
2. Consistent with existing `TransactionIndex`, I think you could expand `TransactionIndex` with affected accounts and create separate `BlockIndex`.
3. Less space for dumps
4. Faster lookup

Cons:
1. Startup requires filling index (but that's okay IMO)
2. Need to track index updates and invalidation (but at least an error here doesn't mean we need to clean db in production)
3. May take a lot of RAM (in future we can counter that with lazy-indexing for requested accounts/blocks and LRU-based clean-up)

Arjentix: So the philosophy here becomes:

1. Every non-derivable information is stored on disk
2. Everything else is stored in memory

integration_tests/tests/sequencer_stake_demo.rs
```diff
@@ -5,6 +5,8 @@
     reason = "Integration tests live at crate root and don't care about these lints"
 )]
 
+//
+
```
Arjentix: Some accidental change?

lez/sequencer/actors/storage/src/actor/entities.rs
```diff
@@ -40,14 +41,20 @@ pub enum ColumnFamily {
     Meta,
     /// Small records deleted as the work they track settles.
     Pending,
+    /// Many small records, maps block hashes to block ids.
+    BlockHashToBlockIdMap,
+    /// Many small records, maps account ids to block ids, which affect them.
+    ///
+    /// Contains chains of records.
+    AccountIdToBlockIdMap,
```
Arjentix: Ok, so this is not very like `Meta` because it's written on every block, and it's not `Block` because these are not big chunks of data. So I propose to just use `BlockMeta` for both  of them. Distinction between concrete entities types is achieved by `TYPE_NAME`

lez/sequencer/actors/storage/src/actor/entities.rs
```diff
@@ -357,3 +364,66 @@ impl db::Storable<ColumnFamily> for CrossZonePeerTip {
     const COLUMN_FAMILY: ColumnFamily = ColumnFamily::Meta;
     const TYPE_NAME: &'static str = db::type_name!(CrossZonePeerTip);
 }
+
+/// The map entry between block hashes and block ids.
+#[derive(BorshSerialize, BorshDeserialize)]
+pub struct BlockHashToBlockIdMappingDestination {
+    pub id: u64,
```
Arjentix:
```suggestion
    pub block_id: BlockId,
```

Renamed `id` to `block_id` for consistency with `AccountIdToBlockIdDestination`

lez/sequencer/actors/storage/src/actor/entities.rs
```diff
@@ -357,3 +364,66 @@ impl db::Storable<ColumnFamily> for CrossZonePeerTip {
...
+impl AsRef<[u8]> for AccountIdToBlockIdKey {
+    fn as_ref(&self) -> &[u8] {
+        self.key_bytes.as_ref()
+    }
+}
+
+/// The map entry between account id and block id, which affect this account.
+#[derive(BorshSerialize, BorshDeserialize)]
+pub struct AccountIdToBlockIdDestination {
+    pub block_id: u64,
```
Arjentix:
```suggestion
    pub block_id: BlockId,
```

lez/sequencer/actors/storage/src/actor/entities.rs
```diff
@@ -357,3 +364,66 @@ impl db::Storable<ColumnFamily> for CrossZonePeerTip {
...
+    fn from(value: (AccountId, u64)) -> Self {
+        let mut key_bytes = [0; 32 + 8];
+
+        key_bytes[..32].copy_from_slice(value.0.as_ref());
+        #[expect(clippy::big_endian_bytes, reason = "We use big endian for `u64` in DB")]
+        key_bytes[32..].copy_from_slice(value.1.to_be_bytes().as_ref());
+
+        Self { key_bytes }
+    }
+}
```
Arjentix: Maybe let it be a `AccountIdToBlockIdKey::new()` so that `u64` parameter will be more readable? Because currently for me it's hard to tell what it means

lez/sequencer/actors/storage/src/actor/entities.rs
```diff
@@ -357,3 +364,66 @@ impl db::Storable<ColumnFamily> for CrossZonePeerTip {
...
+
+    const COLUMN_FAMILY: ColumnFamily = ColumnFamily::AccountIdToBlockIdMap;
+    const TYPE_NAME: &'static str = db::type_name!(AccountIdToBlockIdDestination);
+}
+
+/// The metadata entry to store length of map for some account id.
+#[derive(BorshSerialize, BorshDeserialize)]
+pub struct AccountIdToBlockIdMetaLen {
+    pub length: u64,
+}
```
Arjentix, line 419: I don't quite get why we need this

lez/sequencer/ffi/Cargo.toml
```diff
@@ -0,0 +1,38 @@
...
+env_logger.workspace = true
+log = { workspace = true }
+tokio = { features = ["rt-multi-thread"], workspace = true }
+serde_json.workspace = true
+kameo.workspace = true
+kameo_actors.workspace = true
+anyhow.workspace = true
+
+[build-dependencies]
+cbindgen = "0.29"
```
Arjentix: I think this is not the first time we use this dependency, right? Let's move to workspace dependency then

lez/sequencer/ffi/src/runtime.rs
Arjentix: Idk, this was a part of my small refactoring because I needed borrowed runtime in integration tests but was scared of using raw pointers so I did some abstraction around it

lez/sequencer/ffi/src/sequencer.rs
```diff
@@ -0,0 +1,144 @@
...
+/// - `gossip`: an [`Option<Gossip>`] right now is unused and exists only to pin gossip.
+/// - `runtime`: the [`Runtime`] used to run async queries against the store (either owned or
+///   borrowed), already FFI-safe.
+#[repr(C)]
+pub struct SequencerServiceFFI {
+    storage_ref: *mut c_void,
+    slasher_ref: *mut c_void,
+    executor_ref: *mut c_void,
+    scheduler_ref: *mut c_void,
+    gossip: *mut c_void,
```
Arjentix, line 31: Looks strange to me. I think we should only store pointer to `SequencerHandle` which encapsulates the usage of actors and their setup

lez/sequencer/ffi/src/api/types/account.rs
Arjentix, line 1: Maybe time to move common types to something like `ffi_common` or `ffi_core`?

lez/sequencer/ffi/src/api/types/event.rs
```diff
@@ -0,0 +1,66 @@
...
+    pub block_id: FfiBlockId,
+    pub tx_index: u32,
+    pub tx_hash: FfiHashType,
+    pub program_id: FfiProgramId,
+    pub selector: FfiSelector,
+    pub data: FfiVecU8,
+}
+
+impl From<EventRecord> for FfiEventRecord {
+    fn from(value: EventRecord) -> Self {
```
Arjentix: I don't understand, I can't see `EventRecord` defined or imported in this file... How does this compile?

lez/sequencer/ffi/src/api/lifecycle.rs
```diff
@@ -0,0 +1,229 @@
...
+        InitializedSequencerServiceFFIResult::from_error,
+        InitializedSequencerServiceFFIResult::from_value,
+    )
+}
+
+/// Creates all components of sequencer service except RPC.
+///
+/// TODO: Make RPC construction optional. Probably needs modification of configs to be alighned
+/// with mainline sequencer.
+async fn make_sequencer_compoments(
```
Arjentix, line 50: As I said in a comment above, let's not do it here as breaks the abstraction around `sequencer_service` and also requires support. E.g. adding new actor would also require to change code here which is not obvious. Can we just reuse `SequencerHandle` returned by `sequencer_service::run()`?

Arjentix: Or maybe implement some functionality needed here in `sequencer_service`

lez/sequencer/actors/executor/src/protocol.rs
Pravdyvy: Renamed

lez/sequencer/actors/storage/src/protocol.rs
Pravdyvy: Renamed

lez/sequencer/actors/storage/src/mock.rs
Pravdyvy: Renamed

lez/sequencer/actors/storage/src/actor.rs
Pravdyvy: Moved into a function

lez/sequencer/actors/storage/src/actor.rs
Pravdyvy: `self.db().iter()` is doing kind of a different thing. If we imagine rocksdb as a HashMap, then current implementation of `self.db().iter()` is `HashMap::values()` and for deletion here I need full `HashMap::iter()`.

lez/sequencer/actors/storage/src/actor.rs
Pravdyvy: Removed.

lez/sequencer/actors/storage/src/actor.rs
Pravdyvy: Probably not, the main concern is that `delete_cf` does nothing in case if key is absent, and we get no notification if something potentially bad happened in DB.

lez/sequencer/actors/storage/src/protocol.rs
Pravdyvy: The main problem with that is that we do not know how much data exactly we need to store in memory. One additional Gb on disc is bearable, one additional Gb in memory is a tragedy. 
So the only sustainable and scalable variant is to store all the data in DB and some data in memory (much like you mentioned in Cons 3).
DB is the harder part, so it is better to be done first, so we can observe all bugs earlier.

integration_tests/tests/sequencer_stake_demo.rs
Pravdyvy: Yes, removed.

lez/sequencer/actors/storage/src/actor/entities.rs
Pravdyvy: How iteration will work if they in the same column? For `BlockHashToBlockIdMap` iteration is probably useless, but for `AccountIdToBlockIdMap` could be a possibility.

lez/sequencer/actors/storage/src/actor/entities.rs
Pravdyvy: Renamed

lez/sequencer/actors/storage/src/actor/entities.rs
Pravdyvy: Renamed

lez/sequencer/actors/storage/src/actor/entities.rs
Pravdyvy: Rewrote with `new`.

lez/sequencer/ffi/Cargo.toml
Pravdyvy: Moved

lez/sequencer/actors/storage/src/actor/entities.rs
Pravdyvy: How it works here and in indexer:
We store such map: (Account Id, Idx) -> (Block Id in which account id was affected). So each account id have a list in DB.
To know, where we should stop iteration, in `Meta` column we store length of a list for every account.
This construction is better than (Account Id) -> (List of blocks) due to reading concerns.

Ultimately, the best construction is actually 2 maps:
(Account ID) -> (Unique account id identifier) and (UAId, Idx) -> (Block Id in which account id was affected).
 This works if total number of accounts is lesser that `u64::MAX`(which is probably true), but this will be traumatic to read.

lez/sequencer/ffi/src/api/types/event.rs
Pravdyvy: It is from the next PR, removed for now.

lez/sequencer/ffi/src/api/types/account.rs
Pravdyvy: In future we plan to remove indexer ffi and add sequencer initialization into wallet ffi, so, probably everyone should just use sequencer ffi primitives as a common.

lez/sequencer/ffi/src/sequencer.rs
Pravdyvy: The main issue is that when we start sequencer from FFI, we don't need RPC. So, basically we need just some parts for sequencer for an FFI.
The correct way to solve it is tomake RPC optional for the main sequecner and do as you suggest, the problem with that is that this will inflate this PR by 1000 more lines, because it will afeect a lot of places.
I already noted this in #823 ToDo's. This will be in queue.

lez/sequencer/ffi/src/api/lifecycle.rs
Pravdyvy: Same as for `SequencerServiceFFI` signature.

lez/sequencer/actors/storage/src/actor.rs
Arjentix: Ah I see what you mean. I thought about that during initial implementation and here are my thoughts:

1. At a time no code needed keys
2. If someone will need keys during iteration they can just store key in value it-self
3. This let's us not to decode keys when noone needs them
4. Although I agree that this might not be very convenient, so if we have a use-case for keys iteration right now, I don't mind reimplementing `db.iter()` to produce `impl Iterator<Item = (T::Key, T)>`

So to me either storing key inside value or reworking `db.iter()` are good

lez/sequencer/actors/storage/src/protocol.rs
Arjentix: Hmm, yeah, potential user hardware should better have as little requirements as possible... And RAM costs much more than disk. Makes sense

lez/sequencer/actors/storage/src/actor/entities.rs
Arjentix: Iteration takes care of it, it uses `TYPE_NAME` to determine start and end range of type stored in column. So it should work

lez/sequencer/actors/storage/src/actor/entities.rs
Arjentix: > To know, where we should stop iteration, in Meta column we store length of a list for every account.
This construction is better than (Account Id) -> (List of blocks) due to reading concerns.

Hmm, why can't we just kind of use `BTreeMap` thinking and just store `AccountId -> Vec<BlockId>`?

lez/sequencer/ffi/src/sequencer.rs
Arjentix: Ah, yes, I see. Yeah, I think Rpc can be easily disabled. Probably even with a compile feature. In fact this will be just not constructing `RpcServerActor` and that's it. But then someone else should send requests and transactions to executor. Probably we can create some ffi-related actor or something like that.

This conversation can be resolved.

Arjentix (approved):
LGTM, but would like to continue discussing unresolved threads

lez/sequencer/actors/storage/src/actor/entities.rs
Pravdyvy: Ok. Merged this columns into `BlockMeta`.

lez/sequencer/actors/storage/src/actor.rs
Pravdyvy: Key inside a value in not an option. Reworking `db.iter()` is fine, but I don't believe I can do it fast. 
Would it be fine to leave as TODO?

lez/sequencer/actors/storage/src/actor/entities.rs
Pravdyvy: It is quite inefficient. The main problem is that this list grows. It would be fine if it would be constant-size.
We would need to rewrite big blob of data, potentially every block. The account which can provide issues is, for example, clock. But user accounts can be affected as much.

lez/sequencer/actors/storage/src/actor.rs
Arjentix: Yeah, sure

lez/sequencer/actors/storage/src/actor/entities.rs
Arjentix: I see now, thanks

lez/sequencer/ffi/src/runtime.rs
erhant: feel free to resolve this one!

PR #879, opened 2026-09-10
fix(nix): resolve the Metal toolchain from its cryptex mount

igor-sirotin:
# Description

1. Rewrote `metalStub` to resolve `metal` / `metallib` from the cryptex mount directly, instead of clearing `DEVELOPER_DIR`/`SDKROOT` and delegating the lookup to `xcrun`.

On recent macOS the Metal Toolchain is a **per-user cryptex**. `xcrun` resolves it only for the user that ran `xcodebuild -downloadComponent MetalToolchain`; a nix build runs as `_nixbld`, gets Xcode's stub instead, and `risc0-build-kernel` panics with `Could not build metal kernels`. This reproduces on a correctly configured host — Xcode selected, toolchain downloaded, sandbox already off — so no local setup fixes it.

```
$ xcrun --find metal                    # the installing user
/var/run/.../Metal.xctoolchain/usr/bin/metal
$ sudo -u nobody xcrun --find metal     # any other user, i.e. the builder
/Applications/Xcode.app/.../XcodeDefault.xctoolchain/usr/bin/metal
```

Only the *lookup* is per-user — the mount is world-readable, which is what makes this fixable.

<details>
<summary>AI-made decisions</summary>

- `--sdk` is parsed off and passed to the compiler as `-isysroot`, resolved via `xcrun --sdk <name> --show-sdk-path` (that call works for any user; it is the SDK, not the Metal toolchain). risc0 invokes `xcrun --sdk <name> metal …`, so dropping the flag silently would have changed what the compiler sees. `metallib` is a linker and gets no sysroot.
- Falls back to the previous behaviour when no cryptex is mounted, so hosts whose Metal tools live inside `Xcode.app` are unaffected. It degrades to today's error rather than a new one.
- The tool name is matched positionally and only the first occurrence is treated as the tool, so a later `metal` appearing as a filename argument is passed through.
- Based on `dev`, not on #877, so the two are independently mergeable. #877 is what lets a non-darwin machine evaluate this flake at all; without it I could only evaluate the native system here.
- **Not verified on macOS.** I have no darwin host. What is checked: the flake evaluates with the new stub, and the generated script passes `bash -n`. The cryptex path resolution and the `-isysroot` handoff need a real build on a Mac.

</details>


flake.nix
```diff
@@ -112,15 +112,46 @@
...
             tool=
-            for a in "$@"; do
-              case "$a" in metal|metallib) tool=1 ;; esac
+            args=()
+            while [ $# -gt 0 ]; do
+              case "$1" in
+                --sdk) sdk=$2; shift 2 ;;
+                metal|metallib)
+                  if [ -z "$tool" ]; then tool=$1; else args+=("$1"); fi
+                  shift
```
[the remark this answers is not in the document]
igor-sirotin: fixed in ed861b4. `--find`/`-f` now resolves and prints the cryptex path rather than running the compiler with the flag — passing through to `/usr/bin/xcrun` would have returned Xcode's stub, which is the path this PR exists to avoid.

flake.nix
```diff
@@ -112,15 +112,46 @@
...
+              for cand in /var/run/com.apple.security.cryptexd/mnt/*/Metal.xctoolchain/usr/bin/"$tool"; do
+                [ -x "$cand" ] || continue
+                if [ "$tool" = metal ] && [ -n "$sdk" ]; then
+                  # --sdk was xcrun's job; hand the compiler the sysroot itself.
+                  sysroot=$(/usr/bin/xcrun --sdk "$sdk" --show-sdk-path 2>/dev/null || true)
+                  if [ -n "$sysroot" ]; then
+                    exec "$cand" -isysroot "$sysroot" "''${args[@]}"
+                  fi
+                fi
+                exec "$cand" "''${args[@]}"
```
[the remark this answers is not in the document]
igor-sirotin: fixed in ed861b4 — a failed `--show-sdk-path` now exits non-zero with a message instead of falling through and compiling without `-isysroot`.

flake.nix
erhant, line 109: the "clears those two vars" (i guess `unset DEVELOPER_DIR SDKROOT`) belongs to the old code, now we do that in the fallback case, worth rewording here

flake.nix
```diff
@@ -112,15 +112,60 @@
...
+                fi
+                if [ "$tool" = metal ] && [ -n "$sdk" ]; then
+                  # --sdk was xcrun's job; hand the compiler the sysroot itself.
+                  sysroot=$(/usr/bin/xcrun --sdk "$sdk" --show-sdk-path 2>/dev/null) || sysroot=
+                  if [ -z "$sysroot" ]; then
+                    echo "xcrun: cannot resolve SDK '$sdk'" >&2
+                    exit 1
+                  fi
+                  exec "$cand" -isysroot "$sysroot" "''${args[@]}"
+                fi
```
erhant, line 158: the `-isysroot` block can probably go, as my Claude says (but would love it to be double-checked):

- The cryptex `metal` already reads SDKROOT from the env. Exec'd with no flag it emits `-isysroot <nix apple-sdk>/MacOSX.sdk` on its own.
- DEVELOPER_DIR is still set when `xcrun --show-sdk-path` runs, so it returns nix's apple-sdk 14.4, not Xcode's SDK as the old flow did. The two paths produce byte-identical `.air` output.

So `exec "$cand" "${args[@]}"` alone is equivalent, and the --sdk case just needs to **consume** its value.

That also removes the "cannot resolve SDK" branch; it was checking a value that never reached the compiler, and risc0 only ever passes macosx/iphoneos.

We can keep this as well, but then the L151 comment should say it resolves nix's SDK.

erhant (approved):
LGTM, few comments!

flake.nix
igor-sirotin: ~~Reworded in 5b858ac5 — the header now says the wrapper resolves metal/metallib from the cryptex mount, and only clears those two vars in the no-cryptex fallback.~~

flake.nix
igor-sirotin: ~~You're right that it resolves nix's SDK, not Xcode's — DEVELOPER_DIR is still set at that point — so I fixed that comment in 5b858ac5. I kept the `-isysroot` block for now: the "byte-identical .air" part needs a real Mac build to confirm, and without it an explicit `--sdk` that disagrees with SDKROOT (e.g. iphoneos) would silently get the wrong sysroot. Verifying on a Mac and will follow up here.~~

igor-sirotin:
Thanks for the comments. Claude confirms both are valid, but it seems that it breaks the Mac build again. I'm looking into it.

flake.nix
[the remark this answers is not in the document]
igor-sirotin: Reverted in 72f373b6. Making this fatal broke the build on macOS: `xcrun --sdk ... --show-sdk-path` can fail for the nix build user, and the original tolerated that and let the compiler fall back to SDKROOT — a real Mac build succeeds at the original stub and fails at every commit after it.

flake.nix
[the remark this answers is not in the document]
igor-sirotin: Also reverted in 72f373b6 — verified on a Mac, the build succeeds with the original stub and fails with these changes, so I restored the working version rather than keep a `--find` branch no real build exercises. Worth revisiting only with a Mac build to confirm it against.

flake.nix
igor-sirotin: Your read was right, and a Mac build now backs it up: the `--show-sdk-path` lookup can fail for the nix build user, and the build only works because that failure is tolerated and the compiler falls back to SDKROOT. I kept the block (72f373b6 restores the verified-working stub) but the comment now says it resolves nix's SDK and that the lookup may fail. Dropping `-isysroot` entirely looks sound on this evidence — happy to do it as a follow-up with a Mac build to confirm.

igor-sirotin:
@erhant the applied fixes for your comments don't actually compile on Mac, so I've left it as before — current state works perfectly for me

erhant:
> @erhant the applied fixes for your comments don't actually compile on Mac, so I've left it as before — current state works perfectly for me

alright, thanks for the effort!

flake.nix
erhant: resolving per https://github.com/logos-blockchain/logos-execution-zone/pull/879#issuecomment-5698151216

Arjentix:
`nix build --extra-experimental-features 'nix-command flakes'` fails for me on Mac:

```
error: Cannot build '/nix/store/16p2pnybb9l93mq2c6kh5f39gbjd66mv-logos-execution-zone-wallet-ffi-deps-0.1.0.drv'.
       Reason: builder failed with exit code 101.
       Output paths:
         /nix/store/zw0d3xnk0khvl4g8y3x84vlisvbp7zr5-logos-execution-zone-wallet-ffi-deps-0.1.0
       Last 25 log lines:
       >   note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
       >
       >   thread '<unnamed>' (28993885) panicked at /nix/store/jb4f3sqd4a73zd41arsjbza9sbvvyx2y-vendor-cargo-deps/c19b7c6f923b580ac259164a89f2577984ad5ab09ee9d583b888f934adbbe8d0/risc0-build-kernel-2.0.1/src/lib.rs:253:29:
       >   Could not build metal kernels
       >   error: error: cannot execute tool 'metal' due to missing Metal Toolchain; use: xcodebuild -downloadComponent MetalToolchain
       >
       >   thread '<unnamed>' (28993883) panicked at /nix/store/jb4f3sqd4a73zd41arsjbza9sbvvyx2y-vendor-cargo-deps/c19b7c6f923b580ac259164a89f2577984ad5ab09ee9d583b888f934adbbe8d0/risc0-build-kernel-2.0.1/src/lib.rs:253:29:
       >   Could not build metal kernels
       >   error: error: cannot execute tool 'metal' due to missing Metal Toolchain; use: xcodebuild -downloadComponent MetalToolchain
       >
       >   thread '<unnamed>' (28993881) panicked at /nix/store/jb4f3sqd4a73zd41arsjbza9sbvvyx2y-vendor-cargo-deps/c19b7c6f923b580ac259164a89f2577984ad5ab09ee9d583b888f934adbbe8d0/risc0-build-kernel-2.0.1/src/lib.rs:253:29:
       >   Could not build metal kernels
       >   error: error: cannot execute tool 'metal' due to missing Metal Toolchain; use: xcodebuild -downloadComponent MetalToolchain
       >
       >   thread '<unnamed>' (28993887) panicked at /nix/store/jb4f3sqd4a73zd41arsjbza9sbvvyx2y-vendor-cargo-deps/c19b7c6f923b580ac259164a89f2577984ad5ab09ee9d583b888f934adbbe8d0/risc0-build-kernel-2.0.1/src/lib.rs:253:29:
       >   Could not build metal kernels
       >   error: error: cannot execute tool 'metal' due to missing Metal Toolchain; use: xcodebuild -downloadComponent MetalToolchain
       >
       >   thread '<unnamed>' (28993886) panicked at /nix/store/jb4f3sqd4a73zd41arsjbza9sbvvyx2y-vendor-cargo-deps/c19b7c6f923b580ac259164a89f2577984ad5ab09ee9d583b888f934adbbe8d0/risc0-build-kernel-2.0.1/src/lib.rs:253:29:
       >   Could not build metal kernels
       >   error: error: cannot execute tool 'metal' due to missing Metal Toolchain; use: xcodebuild -downloadComponent MetalToolchain
       >
       >   thread '<unnamed>' (28993884) panicked at /nix/store/jb4f3sqd4a73zd41arsjbza9sbvvyx2y-vendor-cargo-deps/c19b7c6f923b580ac259164a89f2577984ad5ab09ee9d583b888f934adbbe8d0/risc0-build-kernel-2.0.1/src/lib.rs:253:29:
       >   Could not build metal kernels
       > warning: build failed, waiting for other jobs to finish...
       For full logs, run:
         nix-store -l /nix/store/16p2pnybb9l93mq2c6kh5f39gbjd66mv-logos-execution-zone-wallet-ffi-deps-0.1.0.drv
error: Cannot build '/nix/store/lp6a5gijzp20qdznn8j79ppb7lyqx7dx-logos-execution-zone-wallet-ffi-0.1.0.drv'.
       Reason: 1 dependency failed.
       Output paths:
         /nix/store/d1kqvjw0ffcnvmhyb3mqhwmg1z9rk66z-logos-execution-zone-wallet-ffi-0.1.0
error: Build failed due to failed dependency
```

Or am I doing something wrong?

igor-sirotin:
@Arjentix can you please confirm that you're on the right commit?
```
> git rev-parse HEAD --short
72f373b61d2687aed6b1848dedd638ca44670554

> zsh ✗ nix build
```

Passes for me with nor errors 🤔

erhant:
I've tried again now at the latest head, `nix build` passes (in ~5 minutes)

erhant:
> error: error: cannot execute tool 'metal' due to missing Metal Toolchain; use: xcodebuild -downloadComponent MetalToolchain

worth noting that this error is nasty as in if it works it works, and it can only ever break again after some specific events like MacOS update and such (if i remember correctly)

that being said, the point of this PR was to fix this anyways so I'm curious why it doesnt work on Daniil

Arjentix:
> @Arjentix can you please confirm that you're on the right commit?

Yeah it's the same...

image

I'm trying my Claude with it right now

Arjentix:
Ok, actually it worked after `xcodebuild -downloadComponent MetalToolchain`

Arjentix (approved):

igor-sirotin:
Nice! merging then. Thanks everyone!

PR #883, opened 2026-09-11
ci: stop gating integration shards on test-fixtures-tests

moudyellaz:
## 🎯 Purpose

The integration shards wait on `test-fixtures-tests` before they start, but it doesn't produce anything they use. The archive comes from `integration-tests-prebuild`.

On the last dev run the archive was ready at +11m and the shards still didn't start until +30m, because `test-fixtures-tests` was ahead of them in the queue (it waits on `bedrock-node`, which takes 19m). Dropping it from `needs:` lets them start at +19m instead, so the run goes from ~53m to ~42m.

This wasn't worth doing before #874. Prebuild used to take 29m, so both finished around the same time and removing one changed nothing.

The gate was there to catch an outdated db dump before it breaks most of the integration tests. To keep that, `abort-on-stale-fixtures` cancels the run if `test-fixtures-tests` fails, so a stale dump still shows up as one clear failure instead of 35 confusing ones. The shards just don't wait for it any more.

## ⚙️ Approach

- [x] Drop `test-fixtures-tests` from the `integration-tests` `needs:` list
- [x] Cancel the run if `test-fixtures-tests` fails, so the check is preserved without the wait

## 🧪 How to Test

Next dev run, check when the shards start. Should be ~+19m rather than ~+30m, with the run landing around 42m.

The cancel path isn't exercised unless the dump actually goes stale. Worth knowing that it makes the run read as "cancelled" with `test-fixtures-tests` the single red job, rather than "failed".

## 🔜 Future Work

None.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [ ] Add/update tests

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -256,6 +256,17 @@ jobs:
             RUST_LOG=info,kameo=warn
           run: cargo nextest run -p test_fixtures
 
+  abort-on-stale-fixtures:
+    needs: test-fixtures-tests
+    if: failure()
+    runs-on: ubuntu-latest
+    permissions:
+      actions: write
+    steps:
+      - run: gh run cancel ${{ github.run_id }}
+        env:
+          GH_TOKEN: ${{ github.token }}
+
   # Not a `container:` job: the test binaries bake CARGO_MANIFEST_DIR in at
   # compile time, so the archive has to be built on the same path the matrix
   # jobs later run it from.
@@ -328,7 +339,6 @@ jobs:
           echo "Discovered integration targets: $targets_json"
 
   integration-tests:
-    # Not gated on test-fixtures-tests: prebuild produces the archive.
     needs: [ci-image, bedrock-node, integration-tests-prebuild]
     runs-on: ubuntu-latest
     timeout-minutes: 90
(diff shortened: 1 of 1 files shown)


erhant (approved):
LGTM good catch!

.github/workflows/ci.yml
```diff
@@ -328,7 +328,8 @@ jobs:
           echo "Discovered integration targets: $targets_json"
 
   integration-tests:
-    needs: [ci-image, bedrock-node, test-fixtures-tests, integration-tests-prebuild]
+    # Not gated on test-fixtures-tests: prebuild produces the archive.
```
Arjentix:
```suggestion
```

Arjentix (approved):
Overall the idea of running `test-firtures-tests` earlier than integration tests is to catch outdated db dump which will cause most of the integration tests fail.
Indeed before this was taking pretty much the same time as another job blocking integration tests, so it was not the problem.
I like the speed up, but would like to somehow preserve the check. Maybe we can abort integration tests runs if `test-fixtures-tests` failed? So that dev can clearly see where to dig instead of trying to debug integration tests. Because even if some integration tests will pass it makes no sense with outdated db.

Pravdyvy (approved):
lgtm

erhant (approved):

Arjentix (approved):
LGTM

Pravdyvy (approved):
lgtm

PR #884, opened 2026-09-11
feat(wallet): makes gas limit configurable per wallet

adklempner:
## 🎯 Purpose

The wallet had a hardcoded gas limit that was below the maximum allowed for a transaction. If a program required more cycles than 2M, there was no way to use the wallet to interact with that program.

## ⚙️ Approach

This PR replaces the hardcoded gas limit with a new field in the wallet config, with the previous cap set as the default.

## 🧪 How to Test

Against a local chain with a funded account, with `LEE_WALLET_HOME_DIR` pointing at that wallet:

```bash
wallet config get gas_limit
# 2000000

wallet auth-transfer send --from Public/<A> --to Public/<B> --amount 100
# Transaction is included in block 4
# fee: Some(FeeDeclaration { gas_limit: 2000000, tip: 0, max_fee: 134400000 })

wallet config set gas_limit 10000001                  # one past MAX_GAS_EXEC
wallet auth-transfer send --from Public/<A> --to Public/<B> --amount 100
# Error: Sequencer client error: ... message: "Incorrect fee"

wallet config set gas_limit 10000000                  # the cap
wallet auth-transfer send --from Public/<A> --to Public/<B> --amount 100
# Transaction is included in block 9
# fee: Some(FeeDeclaration { gas_limit: 10000000, tip: 0, max_fee: 134400000 })
```

A send echoes its `FeeDeclaration`, so the declared gas is directly visible. On `dev` it reads
2000000 whatever the config says.

The same sequence as a test:

```bash
just resolve-bedrock-node   # once
cargo nextest run -p integration_tests --test config \
    transactions_declare_the_configured_gas_limit
```

## 🔗 Dependencies

TO COMPLETE IF APPLICABLE

## 🔜 Future Work

- Configuring gas limit per call; wider change that the FFI and every caller
- `lez/wallet/src/config.rs` already declares a `GasConfig` with `gas_limit_runtime` and `gas_limit_deploy`, but nothing in the workspace seems to construct or read it and it is not a field of `WalletConfig`

## 📋 PR Completion Checklist

- [ ] Complete PR description
- [ ] Implement the core functionality
- [ ] Add/update tests


erhant (changes requested):
ty for the PR!

one important note: you should also derive `DEFAULT_MAX_FEE` from the given `self.config.gas_limit`, otherwise you will not be able to pay for the fee that your gas requires; we derive the default via the following:

```rs
(DEFAULT_GAS_LIMIT as u128 + ASSUMED_DATA_BYTES) * ASSUMED_BASE_FEE;
```

these were just placeholders though, indeed they will be more sequencer-aware and configurable later (at FFI level), as also noted by your future work

erhant (approved):
LGTM, ty for addressing the comment!

moudyellaz (approved):
lgtm, thanks!

PR #885, opened 2026-09-12
refactor!(sequencer): move bedrock communication into BedrockActor

Arjentix:
## 🎯 Purpose

Bedrock communication was the next good candidate for moving into a separate actor from sequencer_core and this PR implements it.

## ⚠️  Breaking

Bedrock sequencer checkpoint now uses different (de-)encoding and starting from an old database would fail

## ⚙️ Approach

- Added `BedrockActor` in a separate crate
- Added `BedrockActorTrait` to replace `BlockPublisherTrait`
- Removed `block_publisher.rs`
- Removed `block_store.rs` as not meaningful anymore
- `reconstruction.rs` tests moved into `sequencer_executor_actor` as now they can be properly mocked with actors
- Used `kameo_actors::broker::Broker` for topic-based subscription to `BedrockActor` notifications, topics are:
  - `channel/<channel_id>/update`
  - `channel/<channel_id>/turn`

## 🧪 How to Test

CI must pass

## 🔗 Dependencies

None

## 🔜 Future Work

I still have some ideas for improvements in further PRs:

- Cross zones are still using manual `Node` from zone sdk. This is breaks incapsulation and I want to address this in future.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


lez/sequencer/actors/bedrock/src/actor.rs
```diff
@@ -0,0 +1,739 @@
...
+            node.clone(),
+            zone_sdk_config,
+            initial_checkpoint,
+        );
+
+        // Wait for cold-start backfill to complete before returning so callers
+        // can publish immediately without racing readiness.
+        while !sequencer.is_ready() {
+            // Zone SDK sequencer will process ready event internally.
+            sequencer.next_event().await;
```
schouhy, line 132: what about BlocksProcessed events with deposits that happened while offline? aren't they discarded here?

lez/sequencer/service/src/lib.rs
```diff
@@ -192,15 +196,37 @@ pub fn run(
...
+                topic,
+                recipient: executor_prepared.actor_ref().clone().recipient(),
+            })
+            .await?;
+        info!("Bedrock Broker Actor spawned");
+
+        let bedrock = setup_bedrock_actor(&config, storage_ref.clone(), bedrock_broker_ref.clone())
+            .await
+            .context("Failed to set up Bedrock Actor")?;
+        let bedrock_ref = BedrockActor::spawn(bedrock);
```
schouhy, line 220: this will start sending messages to the executor that hasn't yet started no? can't this fill the executor mailbox and start dropping messages?

lez/sequencer/actors/executor/src/actor.rs
```diff
@@ -550,3 +522,24 @@ impl<S: StorageActorTrait, BP: BlockPublisherTrait + Send + Sync + 'static>
...
+    type Reply = ();
+
+    async fn handle(
+        &mut self,
+        msg: sequencer_bedrock_actor::protocol::ChannelEvent,
+        _ctx: &mut Context<Self, Self::Reply>,
+    ) -> Self::Reply {
+        match msg {
+            sequencer_bedrock_actor::protocol::ChannelEvent::Update(channel_update) => {
+                self.sequencer.on_channel_update(*channel_update).await;
```
schouhy, line 538: this will write the checkpoint to disk. the executor also writes it after publishing on record_produced_block. A channel event message queued while the a block is built can end up overwriting a checkpoint?

schouhy (approved):
lgtm! thanks!

lez/sequencer/service/src/lib.rs
```diff
@@ -360,3 +388,84 @@ async fn setup_gossip(
...
+
+    BedrockActor::new(bedrock_actor_config, storage_ref, bedrock_broker_ref).await
+}
+
+#[cfg(feature = "standalone")]
+#[expect(
+    clippy::unused_async,
+    reason = "Must be async to match the signature of the non-standalone version"
+)]
+async fn setup_bedrock_actor(
```
schouhy, line 423: this is missing the expect_handle_get_channel_id call no?

lez/sequencer/service/src/lib.rs
moudyellaz: Good catch, and I think it is worse than dropping: with `Guaranteed` the broker blocks, so once the undrained executor mailbox fills and then the broker does too, the `CheckIsOurTurn` ask in `ExecutorActor::new` never comes back. Would subscribing after `executor_prepared.spawn(executor)` work?

lez/sequencer/service/Cargo.toml
```diff
@@ -14,7 +14,8 @@ sequencer_gossip_actor = { workspace = true, features = ["actor"] }
 sequencer_rpc_server_actor.workspace = true
 sequencer_executor_actor = { workspace = true, features = ["actor"] }
 sequencer_slasher_actor.workspace = true
-sequencer_storage_actor = { workspace = true, features = ["actor"] }
+sequencer_storage_actor = { workspace = true, features = ["actor", "mock"] }
```
moudyellaz, line 17: This puts mockall in the release image, since `configure_channel` has no `required-features` and always builds. Could the `mock` feature hang off a bin feature the way `inscribe_garbage` does?

lez/sequencer/actors/bedrock/src/protocol.rs
```diff
@@ -0,0 +1,151 @@
...
+};
+pub use sequencer_stake_core::ChannelParams;
+
+/// A boxed, pinned, Send stream.
+pub type BoxStream<T> = std::pin::Pin<Box<dyn futures::Stream<Item = T> + Send>>;
+
+/// Event happened on configured chain.
+#[derive(Debug, Clone)]
+pub enum ChannelEvent {
+    Update(Box<ChannelUpdate>),
```
moudyellaz, line 17: The broker clones the message per delivery, so every update deep-copies its blocks. Would `Arc<ChannelUpdate>` work here instead of `Box`?

lez/sequencer/actors/bedrock/src/actor.rs
```diff
@@ -0,0 +1,739 @@
...
+
+impl Message<ReadChannel> for BedrockActor {
+    type Reply = Result<BoxStream<(ZoneMessage, Slot)>>;
+
+    async fn handle(
+        &mut self,
+        ReadChannel { after }: ReadChannel,
+        _ctx: &mut Context<Self, Self::Reply>,
+    ) -> Self::Reply {
+        const BATCH_SIZE: Slot = Slot::new(100);
```
moudyellaz, line 591: This is the same batching loop as `ZoneIndexer::next_messages`, which is still used by the indexer and the cross-zone watcher. Worth reusing it, maybe with an owned variant so the stream can be boxed?

lez/sequencer/actors/rpc_server/src/actor/service.rs
```diff
@@ -334,8 +334,10 @@ fn map_executor_error<M>(
             handle_err @ (sequencer_executor_actor::error::Error::BackgroundTaskFinishedUnexpectedly
             | sequencer_executor_actor::error::Error::BlockPublisherFinishedUnexpectedly
```
moudyellaz, line 335: This arm is unreachable now, nothing constructs `BlockPublisherFinishedUnexpectedly` since the `driver_cancellation` arm went away. Worth dropping the variant while you are here?

lez/sequencer/core/src/lib.rs
```diff
@@ -788,29 +788,263 @@ impl<S: StorageActorTrait, BP: BlockPublisherTrait> SequencerCore<S, BP> {
...
+            // block finalized at its height.
+            let mut finalized_dispatch_keys = HashSet::new();
+            for block in &irreversible {
+                finalized_dispatch_keys
+                    .extend(settled_dispatch_keys(&self.storage_ref, block).await);
+            }
+
+            // A persist failure is fatal: the in-memory chain has already advanced,
+            // and continuing would leave a permanent gap in the store. The `panic!`
+            // ends the drive task, whose cancellation halts the node.
```
moudyellaz, line 985: The follow path runs in the executor's handler now, so there is no drive task to cancel (the halt still works, via kameo's `on_panic`). Same wording at 1033 and 1545 if you want to catch them together.

lez/sequencer/service/src/lib.rs
```diff
@@ -144,19 +143,24 @@ impl SequencerHandle {
...
             storage,
             addr: _,
             gossip: _,
         } = self;
 
-        executor.is_healthy()
+        scheduler.is_healthy()
+            && bedrock_broker.is_healthy()
             && rpc_server.is_healthy()
             && scheduler.is_healthy()
```
moudyellaz, line 160: Small one: `scheduler.is_healthy()` is already on line 157.

moudyellaz (approved):
lgtm, thanks. Left some comments!

lez/sequencer/service/src/lib.rs
Arjentix: Hmm, good catch, this is only triggered if such a request comes from RPC, that's why I haven't catched that during testing

lez/sequencer/actors/bedrock/src/protocol.rs
Arjentix: Good idea, sequencer never needed that thing by value so Arc landed without any troubles

lez/sequencer/actors/rpc_server/src/actor/service.rs
Arjentix: Thanks!

lez/sequencer/service/Cargo.toml
Arjentix: Good catch!

lez/sequencer/core/src/lib.rs
Arjentix: Ah, tracking Claude's outdated comments is an issue if you don't code everything with Claude. Thanks

lez/sequencer/actors/bedrock/src/actor.rs
Arjentix: Yes, that's true, but I did this intentionally. One day we'll remove `ZoneIndexer` I already have a branch where I removed it from cross zones. And when Oleksandr will remove Indexer as a whole we can safely remove `ZoneIndexer`. So let's keep this duplication for now

lez/sequencer/service/src/lib.rs
Arjentix: Ah, yes, this is complicated.

1. I can't subscribe `ExecutorActor` after spawning it, because it needs a running `BedrockActor` in it's constructor. That means `BedrockActor` could start producing events before `ExecutorActor` is subscribed leading to missed events
2. It will indeed deadlock as Moudy described
3. I can address this with unbounded mailbox for executor, I think this is the way

lez/sequencer/actors/bedrock/src/actor.rs
Arjentix: Ah, right, thanks. Fixed with applying `on_event()`

lez/sequencer/service/src/lib.rs
Arjentix: I think there are other solutions with custom broker implementations. Like TTL parameter for messages. But I don't think this idea will be suitable here because we indeed want guaranteed delivery

lez/sequencer/actors/executor/src/actor.rs
Arjentix: Hmm, no, because mailbox is drained sequentially. So first our block production will be fully done. And only after that `ExecutorActor` will take `ChannelEvent` from it's queue and process it

schouhy: sure it's sequential. But once the block is produced, the channel event message is processed and writes again the checkpoint no? will any checkpoint coming from a channel event always be more recent than any checkpoint at publishing?

PR #887, opened 2026-09-14
test: resolve the Bedrock node in an image that builds on both architectures

moudyellaz:
## 🎯 Purpose

`just resolve-bedrock-node` cannot run on an arm64 host, so every Docker-backed test fails with "Resolved Bedrock node is missing". It builds the CI image, which installs risc0, and rzup publishes no linux-aarch64 artifacts.

## ⚙️ Approach

- Resolve in a dedicated `rust:1.94.0-trixie` image rather than the CI image. It carries no risc0, so it builds on both architectures.
- `BEDROCK_RESOLVER_IMAGE` now skips the image build instead of only renaming the tag.
- Set `CARGO_HOME=/cargo` so the mounted cargo caches live in a directory the image defines, rather than depending on the base image leaving `/usr/local/cargo` world writable.
- Reject a resolved node whose architecture does not match the Docker daemon, which was five silent `exec format error` retries.

CI resolves inside the CI image and does not use this script.

## 🧪 How to Test

```bash
just resolve-bedrock-node
RISC0_DEV_MODE=1 cargo nextest run -p integration_tests -E 'binary(config)'
```

## 🔗 Dependencies

None.

## 🔜 Future Work

The locked Logos revision matches no published release, so every resolution is a source build, in CI too. Pinning a revision that carries published assets would make it a download.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


bedrock/tools/resolve_bedrock_node_in_docker.sh
```diff
@@ -21,25 +21,29 @@ case "$docker_arch" in
...
 docker run --rm \
     --platform "$docker_platform" \
     --user "$(id -u):$(id -g)" \
     --volume "$repo_root:$repo_root" \
-    --volume "$cargo_home/registry:/usr/local/cargo/registry" \
-    --volume "$cargo_home/git:/usr/local/cargo/git" \
+    --volume "$cargo_home/registry:/cargo/registry" \
+    --volume "$cargo_home/git:/cargo/git" \
     --volume "$HOME/.cache/logos/blockchain:$resolver_home/.cache/logos/blockchain" \
+    --env CARGO_HOME=/cargo \
```
erhant, line 46: should we change `RUSTUP_HOME` as well i wonder

erhant (approved):
LGTM!

tested locally & it works 👌🏻

bedrock/tools/resolve_bedrock_node_in_docker.sh
moudyellaz: No. The base image leaves `/usr/local/rustup` at 0777, so rustup installs the pinned 1.98.0 as the invoking uid. A host-mounted `RUSTUP_HOME` would only cache toolchains between runs, a minute or two against a 19 minute node build.

Pravdyvy (approved):
lgtm. Can not test, but looks fine.

PR #892, opened 2026-09-16
fix(lez): chain consistency fixes

schouhy:
## 🎯 Purpose

Three fixes on the sequencer follow and produce paths.

The head used to rewind onto any adopted block at a height it already held, without the channel reporting anything orphaned. Now it only rewinds on an orphan.

An adopted block that didn't apply was dropped. Now we keep it and apply it when the block in front of it arrives.

A turn used to produce whenever it was ours, even with the head behind the channel, which is how we write an id the channel already carries. Now it produces only if the pin still walks back to the head.

## ⚙️ Approach

- [x] `apply_adopted` only applies a block that extends the head. Anything else parks, and the head shortens through `orphaned` only. `chain_state/src/chain.rs:206`
- [x] Parked blocks are kept (up to 256) and applied when the block in front of them arrives. `chain_state/src/chain.rs:397`
- [x] `ChannelLineage`, the unfinalized entries from `checkpoint.pending_txs`, each with its parent and the block it carries. `chain_state/src/lineage.rs`
- [x] `ChainState::publish_verdict`. Walks from the pin back to the head's entry; a block we could apply in between, or missing entries, holds the turn. Gated in the executor and in `build_block_from_mempool`. `chain_state/src/chain.rs:228`

Entries with no block are walked through, so a pin on garbage still produces. Same for a duplicate below our head, and for a block we already tried and parked: neither will ever be our head, and waiting on them would stall every honest node.

## 🧪 How to Test

```bash
cargo test -p chain_state --lib
cargo test -p sequencer_core --lib --all-features
cargo test -p sequencer_executor_actor --all-features
```

New tests in `sequencer_core`:

- `a_duplicate_height_on_the_channel_does_not_rewind_the_head`
- `a_whole_duplicate_branch_parks`
- `a_lost_race_still_rebases_the_head`
- `a_turn_is_skipped_while_the_head_trails_the_channel`
- `a_turn_is_skipped_when_the_entries_below_the_pin_are_missing`
- `parked_blocks_apply_and_persist_once_the_hole_is_filled`
- `publishing_is_clean_past_a_successor_we_could_not_apply` (`chain_state`)

`chain_state` covers each verdict in the `publishing_*` tests.

## 🔗 Dependencies

None.

## 🔜 Future Work

- Produce at the start of a turn instead of on the fixed tick. A turn is 25 slots, the tick is 10s, so we waste part of every turn.
- Carry entry ids with adopted blocks, so the produce path can compare the pin with the head's entry directly.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


lez/chain_state/src/chain.rs
```diff
@@ -297,10 +380,74 @@ impl ChainState {
...
+        let mut drained = Vec::new();
+        while let Some(idx) = self.head_tip().and_then(|tip| {
+            self.parked.iter().position(|held| {
+                held.header.block_id == tip.block_id.saturating_add(1)
+                    && held.header.prev_block_hash == tip.hash
+            })
+        }) {
+            let block = self.parked.remove(idx);
+            if !matches!(self.apply_adopted(&block), AcceptOutcome::Applied) {
+                break;
```
erhant: should we perhaps re-push the non-applied block here? it could be applied later?

erhant (approved):
LGTM, one important comment

I wonder if not pushing a parked-but-failed-to-apply block again to be parked causes an issue

lez/chain_state/src/chain.rs
schouhy: yes, added in [d4bf2bb](https://github.com/logos-blockchain/logos-execution-zone/pull/892/commits/d4bf2bba75fbe6608d5f89b56d33cd06fa055a8d). But this whole block id bookkeeping is nonsense. I think we can get rid of most of it and these annoying patches. For another PR though.

Pravdyvy (approved):
lgtm

lez/chain_state/src/chain.rs
erhant: agreed!

moudyellaz (approved):

PR #894, opened 2026-09-16
fix(lee)!: reject duplicate witness keys

### LINKED ISSUE #893: fix(lee): reject duplicate witness keys in public and private transactions
A transaction may list the same public key more than once in its witness set. Nothing rejects it, and the nonce is then advanced once per entry.

- `authenticate_public_transaction_signers` (`lee/state_machine/src/validated_state_diff/mod.rs:755-785`) checks only that `message.nonces.len()` equals the witness count, verifies each signature independently, and zips signers against nonces comparing each pair to the *unmodified* pre-state. So a duplicate signer `[A, A]` with nonces `[n, n]` validates because they both become `[n+1, n+1]`.
- The private path (`from_privacy_preserving_transaction`, `mod.rs:618-640`) does the same. Its `n_unique` guards cover `public_account_ids`, nullifiers and commitments, not the witness set.
- `apply_state_diff` (`lee/state_machine/src/state/mod.rs:253-257`) iterates `signer_account_ids` as a `Vec` and increments A's nonce twice.
- Stateless admission (`lez/common/src/transaction.rs`, `transaction_stateless_check`) verifies signatures only, so the tx reaches the mempool.

## Impact

Not exploitable against others: only the key holder can produce the duplicate signature, program authorization is set-based, fee authorization scans for any valid payer signature, and nothing derives value or ordering from the nonce. Deterministic across nodes, so no fork risk.

The effect is a self-inflicted nonce burn: the signer's next tx (expecting `n+1`, chain at `n+2`) fails strict equality and is dropped. With #880 that dropped tx also occupies its signer's mempool lane head for a turn.

## Fix

Reject duplicate witness keys in both authentication paths, reusing `n_unique` over `signer_account_ids`, with a test per path mirroring `account_id_list_cant_have_duplicates` (`lee/state_machine/src/public_transaction/transaction.rs:166`). Optionally reject at `transaction_stateless_check` too, so the tx never takes a mempool slot.

> [!WARNING]
>
> The fix would be a block-validity rule change: a block containing such a tx is valid today and becomes invalid after.

erhant:
## 🎯 Purpose

A witness set could list the same key twice; validation passed and the nonce advanced once per entry. This PR adds checks for that to reject such a case.

Closes #893 which was created as a future work from #880 .

## ⚙️ Approach

- [x] Reject repeated signers in both `authenticate_public_transaction_signers` and `from_privacy_preserving_transaction`, before signature verification.
- [x] Reject at `transaction_stateless_check` too, with a new `DuplicateSigner` malformation error, so the tx doesnt enter the mempool
- [x] One test per path, asserting on the duplicate error specifically.

> [!WARNING]
>
> BREAKING: Block-validity rule change: a block carrying such a tx was valid before and is invalid now.

## 🧪 How to Test

```sh
RISC0_DEV_MODE=1 cargo test -p lee duplicate_signers
cargo test -p common
```

## 🔗 Dependencies

None

## 🔜 Future Work

None

## 📋 PR Completion Checklist
- [x] Complete PR description
- [x] Add/update tests


lee/state_machine/src/public_transaction/transaction.rs
```diff
@@ -182,6 +182,30 @@ pub mod tests {
         assert!(matches!(result, Err(LeeError::InvalidInput(_))));
     }
 
+    #[test]
+    fn witness_set_cant_have_duplicate_signers() {
```
jonesmarvin8, line 186: `witness_set_cannot_have_dulicate_signers`

jonesmarvin8 (approved):
Lgtm. One minor comment.

lee/state_machine/src/public_transaction/transaction.rs
erhant: done in https://github.com/logos-blockchain/logos-execution-zone/pull/894/commits/0abed5c0a7dd3833ec44cc1c8ae12a935518966e

Pravdyvy (approved):
lgtm

PR #897, opened 2026-09-17
feat(indexer): select the blockchain backend with a cargo feature

bacv:
`NodeClient` is now initialized according to the selected feature: `backend-http` (default) uses `NodeHttpClient`, `backend-module` uses `NodeModuleClient` from `logos-blockchain-module`, which reads the blockchain through the in-process `blockchain_module`. Enabling neither is a compile error.

The backend is a path dependency until logos-blockchain/logos-blockchain-module#82 lands.


bacv:
Converted to draft, but still good to review, I just need to merge the stack on blockchain-module and change crate paths to point to commits in master.

pradovic (approved):
LGTM!

PR #423, opened 2026-04-01
fix: correct standalone sequencer command to use config file path

vpavlin:
## 🎯 Purpose

The standalone mode command was missing the '--' separator and the full config file path. Changed from directory path to the complete config file path: sequencer/service/configs/debug/sequencer_config.json

## ⚙️ Approach

*Describe the core changes introduced by this PR.*

- [ ] Changed README command for standalon sequencer

## 🧪 How to Test

*How to verify that this PR works as intended.*

Run the command

## 🔗 Dependencies

*Link PRs that must be merged before this one.*

None

## 🔜 Future Work

*List any work intentionally left out of this PR, whether due to scope, prioritization, or pending decisions.*

None

## 📋 PR Completion Checklist

*Mark only completed items. A complete PR should have all boxes ticked.*

- [x] Complete PR description
- [ ] Implement the core functionality
- [ ] Add/update tests
- [ ] Add/update documentation and inline comments

PR #410, opened 2026-03-26
fix: reject transfer when --from and --to are the same account

ygd58:
## Summary

Fixes #264 (partially).

When `--from` and `--to` point to the same account, wallet previously
submitted the tx and user saw unhelpful error:
`error: Transaction not found in preconfigured amount of blocks`

## Change

Added upfront validation: if `from == to`, wallet now returns:
`Invalid transfer: --from and --to cannot be the same account`

## Note

Other invalid tx scenarios (#265, #268) require deeper changes at
the sequencer level and are out of scope for this PR.

Fixes #264

PR #408, opened 2026-03-26
fix: improve security advisory descriptions in deny.toml

ygd58:
## Summary

Refs #321.

Improves the tracking comments for ignored security advisories in `.deny.toml`:

- **RUSTSEC-2025-0055** (`tracing-subscriber`): Clarified that v0.2.25 is pulled transitively by `ark-relations v0.4.0`. Fix requires upstream to upgrade to `tracing-subscriber >=0.3.20`.
- **RUSTSEC-2025-0141** (`bincode`): Documented recommended alternatives (postcard, bitcode, rkyv) for future migration.

These changes make it easier to track the status of each advisory and what action is needed.

Refs #321

PR #899, opened 2026-09-17
chore: reapply small fixes from closed external PRs

moudyellaz:
## 🎯 Purpose

Four external contributions carried fixes that are still needed on `dev`, but their branches predate the `lee/`+`lez/` restructure and target `main`, so they could not be rebased. The PRs were closed and the content is reapplied here.

Original work by @ygd58 (#407, #408, #410), @vpavlin (#423) and @paschal533 (#441).

## ⚙️ Approach

- [x] Fix the integration test command in the README. `integration_tests` has no binary target, so the documented `cargo run` cannot work.
- [x] Fix the standalone sequencer command. It takes a config file path, not a directory.
- [x] Expand two advisory ignore reasons in `.deny.toml` to name the upstream blocker and reference #321.
- [x] Reject a native token transfer when `--from` and `--to` resolve to the same privacy-qualified identity.
- [x] Render stateless-check RPC errors with `Display` instead of `Debug`.

The transfer guard compares the privacy-qualified identity rather than the bare account id, so shielding and deshielding your own account (`Public(A)` to `Private(A)`) stay allowed.

## 🧪 How to Test

```
cargo test -p wallet native_token_transfer::tests
cargo deny --config .deny.toml check advisories
```

## 🔜 Future Work

Two gaps found during the same triage, not addressed here: the native balance is not displayed for token-owned accounts (`lez/wallet/src/cli/account.rs`), and `checkHealth` returns no report.

## 📋 PR Completion Checklist

- [x] Complete PR description
- [x] Add/update tests


erhant (approved):
LGTM

Pravdyvy (approved):
lgtm

